PR #215, opened 2018-09-28
Kv/psr recv

kvand:
This seems to be working now, and I've added rudimentary documentation for it. No rush on merging it in, but I think it's a straightforward merge.


andrerenard:
This shouldn't be too hard to pull forward, just take a look at some of the other `stages` to see how we implemented the name changes/name spaces, and remember to change `kotekan_process` to `kotekan_stage` in the config files.    Let me know if you'd like me to do that.

andrerenard:
@kvand Do you think you'll be able to look into updating this to match the new frequency mapping  in the VDIF packets (once the encoding is worked out)?  Or should we try to assign someone else to tackle this PR.   It would also be good to at least get this upto date with `develop` and ideally into a state we could merge it.

andrerenard:
I'm going to close this, since it will be superseded by the new TCP based VDIF assembler #804

PR #263, opened 2018-11-08
Coding standards

nritsche:
This is WIP until more complete, please propose additions in the comments or commit to this branch.

Please disagree with the code guidelines in the comments or as a review.

The next step, once it is in `develop`, is enforcing it.


.clang-format
```diff
@@ -0,0 +1,95 @@
...
+  AfterObjCDeclaration: false
+  AfterStruct:     false
+  AfterUnion:      false
+  BeforeCatch:     false
+  BeforeElse:      false
+  IndentBraces:    false
+BreakBeforeBinaryOperators: None
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializers: AfterColon
+ColumnLimit:     80
```
andrerenard: I'm always kind of torn on a column limit of `80`...   I know it's the standard from the days of 80x24 terminal windows, but I find it to be a little overly restrictive in the days of large high resolution monitors.  I'd prefer a column limit of 100, which is google's Java standard, especially in doxygen style indented comments.  Although I know there are always lots of opinions on this.

.clang-format
```diff
@@ -0,0 +1,95 @@
...
+ObjCBlockIndentWidth: 2
+ObjCSpaceAfterProperty: false
+ObjCSpaceBeforeProtocolList: true
+PenaltyBreakBeforeFirstCallParameter: 19
+PenaltyBreakComment: 300
+PenaltyBreakFirstLessLess: 120
+PenaltyBreakString: 1000
+PenaltyExcessCharacter: 1000000
+PenaltyReturnTypeOnItsOwnLine: 60
+PointerAlignment: Left
```
andrerenard: I will have to get used to this.... I'm a pointer right person, but I know there is a good reason to move it to the left, since it's really part of the type not the variable name.

kotekan/kotekan.cpp
```diff
@@ -178,13 +189,13 @@ bool set_gps_time(Config &config) {
  * @brief Starts a new kotekan mode (config instance)
  *
  * @param config The config to generate the instance from
- * @param requires_gps_time If set to true, then the config must provide a valid time
- *                          otherwise an error is thrown.
+ * @param requires_gps_time If set to true, then the config must provide a valid
```
andrerenard: From a readably point of view I prefer having the second line indented to the start of the first line.   Although I'm guessing this might be hard to enforce with `clang-format`.

lib/core/basebandApiManager.cpp
```diff
@@ -1,39 +1,45 @@
...
     std::lock_guard<std::mutex> lock(map_lock);
     return readout_map.end();
 }
 
-void to_json(json &j, const basebandDumpStatus& d) {
-    j = json{{"file_name", d.request.file_name},
-             {"total", d.bytes_total},
-             {"remaining", d.bytes_remaining}};
+void to_json(json& j, const basebandDumpStatus& d) {
+    j = json {{"file_name", d.request.file_name},
```
andrerenard: This is an interesting effect with initializer lists...  Seems to go against the standard of function calls not having a space, but also super confusing for an autoformater.

lib/core/core.cmake
```diff
@@ -0,0 +1,17 @@
+set (KOTEKAN_CORE_SOURCES
```
andrerenard: I see why you are breaking these files up `CMakeLists.txt` plus `core.cmake`, etc.  for the auto-formatter, but I don't like the idea of having to remember or lookup which file in the folder actually contains the sources when I need to add too it.    Maybe name this file `sources.cmake` in every folder to make it standard.  Or just apply the auto-formatter to every .cpp file you find recursively?

CMakeLists.txt
```diff
@@ -253,3 +253,81 @@ if (${BOOST_TESTS})
...
+
+    foreach(f ${ARGN})
+        list(APPEND listVar "${prefix}/${f}")
+    endforeach()
+
+    set(${var} "${listVar}" PARENT_SCOPE)
+endfunction()
+
+if(CLANG_FORMAT_PATH)
+    # Gather absolute path to all .cpp files.
```
andrerenard: Is there a reason to only apply this to `.cpp` files, should we include `.hpp` files as well?  I think you'd be safe to save some of the complexity here and just recursively scan `kotekan/` and `lib/` for `.cpp|.hpp` files.

andrerenard:
Looks like there is also a CMake issue that's blocking this from building as is.

andrerenard:
Also I'd say let's try to get the auto formatter stuff in fairly quickly, even if that means some of the other documentation gets moved to another PR.  This is going to cause a huge amount of conflicts for outstanding branches, so let's get this in while the release cycle is just starting.

docs/sphinx/dev/dev_style_guide_code.rst
```diff
@@ -1,28 +1,169 @@
...
+    #endif
+
+at the end of the file.
+
+
+#include Style
+^^^^^^^^^^^^^^^
+The list of #includes should be in the order:
+
+ - Header file implementing the interface of this `.cpp` file.
```
jrs65: This order doesn't quite make sense to me. I'd suggest:
- Interface file
- Local/private
- Project
- System

as it's most to least specific. The suggestion doesn't seem to have a consistent order. Personally I'd usually do the other way (i.e. least to most), but that's Python influence I think.

docs/sphinx/dev/dev_style_guide.rst
```diff
@@ -26,9 +31,12 @@ Detailed Rules
...
-* ``kotekanProcess`` files should be placed in the **lib/processes/** folder, and named in *camelCase* to match the process class they contain.
+* ``kotekanProcess`` files should be placed in the **lib/processes/** folder,
+and named in *CamelCase* to match the process class they contain.
 
-* Accelerator interfaces (such as e.g. OpenCL or CUDA) should create their own folder within **lib/**, for storage of kernel handling objects and the kernels themselves.
+* Accelerator interfaces (such as e.g. OpenCL or CUDA) should create their own
+folder within **lib/**, for storage of kernel handling objects and the kernels
+themselves.
 
 * Support scripts and components can be placed in the **script/** folder.
```
jrs65: I'd add that Python components/scripts should be going to the `python/` dir if reasonable.

docs/sphinx/dev/dev_style_guide.rst
```diff
@@ -26,9 +31,12 @@ Detailed Rules
 File Organization
 ------------------
 
-* ``kotekanProcess`` files should be placed in the **lib/processes/** folder, and named in *camelCase* to match the process class they contain.
+* ``kotekanProcess`` files should be placed in the **lib/processes/** folder,
+and named in *CamelCase* to match the process class they contain.
```
jrs65: Yes!

docs/sphinx/dev/dev_style_guide_code.rst
```diff
@@ -1,28 +1,169 @@
...
+      int f(int bbbbbb,
+            int cccccc,
+            int dddddd,
+            int eeeeee) {}
+
+- Don't add a newline before an opening curly bracket, e.g.:
+
+  .. code-block:: c++
+
+      void f(bool a) {
```
jrs65: I've had mixed feelings about this one. I quite like Strousop style placing the `{` on the next line (for functions/methods only).

nritsche:
Thanks for commenting on this! I am bringing my car to the shop now and then I will change everything here to the popular choice. So please comment on all options you are not agreeing with. @jrs65On Dec 14, 2018 6:30 AM, Andre Renard <notifications@github.com> wrote:@andrerenard requested changes on this pull request.

This looks great!   Thanks for taking the time to do this.    I have some comments inline.  I'm impressed by how well that auto-formatter did on its own, I'm used to these things often going a little sideways.

In docs/sphinx/dev/dev_style_guide_code.rst:
>  
-Variables
+Private member variables should start with an underscore, e.g.
+``_my_private_variable``.

I think we had decided (with poor follow through) that we'd use underscores for variables which match config parameters. The idea being it's not as important to tell if a parameter is a private member variable or not, but useful to know if variable is config derived.



In docs/sphinx/dev/dev_style_guide_code.rst:
>  ^^^^^^^^^^
-Variables in the code should use underscore naming, e.g. ``my_favourite_variable``.
+Avoid `using namespace X;`. Instead specify where you are using classes or
+functions from a namespace, e.g. `std::vector<std::string> my variable;`.
+Never do `using namespace std;`.

I'm ok with things like using std::string; but only in the context of .cpp files where the using scope won't escape the object.   i.e. no using statements in headers.



In docs/sphinx/dev/dev_style_guide_code.rst:
> +- Don't indent namespaces.
+
+- Pointer and reference symbols should be left aligned, e.g.
+
+  .. code-block:: c++
+
+      int* a;
+      int& b;
+
+- Operands of binary and ternary expressions, trailing comments should be
+  horizontally aligned if
+  they are split over multiple lines, e.g.:
+
+  .. code-block:: c++
+
+      int a = abc *     // My comment

Should we pick a standard for where the operator goes in this case?   I like it starting on the next line (second example) since it's easier to then see that line is related to the previous one.



In docs/sphinx/dev/dev_style_guide_code.rst:
> +  .. code-block:: c++
+
+      int a = f(bbbbbb,
+                cccccc,
+                dddddd,
+                eeeeee);
+      int f(int bbbbbb,
+            int cccccc,
+            int dddddd,
+            int eeeeee);
+      int f(int bbbbbb,
+            int cccccc,
+            int dddddd,
+            int eeeeee) {}
+
+- Don't add a newline before an opening curly bracket, e.g.:

I approve of this style and the next rule on spaces as well.



In .clang-format:
> +BraceWrapping:   
+  AfterClass:      true
+  AfterControlStatement: false
+  AfterEnum:       false
+  AfterFunction:   false
+  AfterNamespace:  false
+  AfterObjCDeclaration: false
+  AfterStruct:     false
+  AfterUnion:      false
+  BeforeCatch:     false
+  BeforeElse:      false
+  IndentBraces:    false
+BreakBeforeBinaryOperators: None
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializers: AfterColon
+ColumnLimit:     80

I'm always kind of torn on a column limit of 80...   I know it's the standard from the days of 80x24 terminal windows, but I find it to be a little overly restrictive in the days of large high resolution monitors.  I'd prefer a column limit of 100, which is google's Java standard, especially in doxygen style indented comments.  Although I know there are always lots of opinions on this.



In .clang-format:
> +IndentWrappedFunctionNames: false
+KeepEmptyLinesAtTheStartOfBlocks: true
+MacroBlockBegin: ''
+MacroBlockEnd:   ''
+MaxEmptyLinesToKeep: 2
+NamespaceIndentation: None
+ObjCBlockIndentWidth: 2
+ObjCSpaceAfterProperty: false
+ObjCSpaceBeforeProtocolList: true
+PenaltyBreakBeforeFirstCallParameter: 19
+PenaltyBreakComment: 300
+PenaltyBreakFirstLessLess: 120
+PenaltyBreakString: 1000
+PenaltyExcessCharacter: 1000000
+PenaltyReturnTypeOnItsOwnLine: 60
+PointerAlignment: Left

I will have to get used to this.... I'm a pointer right person, but I know there is a good reason to move it to the left, since it's really part of the type not the variable name.



In kotekan/kotekan.cpp:
> @@ -178,13 +189,13 @@ bool set_gps_time(Config &config) {
  * @brief Starts a new kotekan mode (config instance)
  *
  * @param config The config to generate the instance from
- * @param requires_gps_time If set to true, then the config must provide a valid time
- *                          otherwise an error is thrown.
+ * @param requires_gps_time If set to true, then the config must provide a valid

From a readably point of view I prefer having the second line indented to the start of the first line.   Although I'm guessing this might be hard to enforce with clang-format.



In lib/core/basebandApiManager.cpp:
>      std::lock_guard<std::mutex> lock(map_lock);
     return readout_map.end();
 }
 
-void to_json(json &j, const basebandDumpStatus& d) {
-    j = json{{"file_name", d.request.file_name},
-             {"total", d.bytes_total},
-             {"remaining", d.bytes_remaining}};
+void to_json(json& j, const basebandDumpStatus& d) {
+    j = json {{"file_name", d.request.file_name},

This is an interesting effect with initializer lists...  Seems to go against the standard of function calls not having a space, but also super confusing for an autoformater.



In lib/core/core.cmake:
> @@ -0,0 +1,17 @@
+set (KOTEKAN_CORE_SOURCES

I see why you are breaking these files up CMakeLists.txt plus core.cmake, etc.  for the auto-formatter, but I don't like the idea of having to remember or lookup which file in the folder actually contains the sources when I need to add too it.    Maybe name this file sources.cmake in every folder to make it standard.  Or just apply the auto-formatter to every .cpp file you find recursively?



In CMakeLists.txt:
> +    set(DO_CLANG_FORMAT "${CLANG_FORMAT_PATH}" "-i -style=file")
+endif()
+
+# Helper function to make absolute from relative paths.
+function(prepend var prefix)
+    set(listVar "")
+
+    foreach(f ${ARGN})
+        list(APPEND listVar "${prefix}/${f}")
+    endforeach()
+
+    set(${var} "${listVar}" PARENT_SCOPE)
+endfunction()
+
+if(CLANG_FORMAT_PATH)
+    # Gather absolute path to all .cpp files.

Is there a reason to only apply this to .cpp files, should we include .hpp files as well?  I think you'd be safe to save some of the complexity here and just recursively scan kotekan/ and lib/ for .cpp|.hpp files.

.clang-format
jrs65: Tangentially related, but this CppCon talk (~5 mins) is a lot of fun. https://www.youtube.com/watch?v=Dw0UBuTKHHg

nritsche:
I will force push into this branch.

docs/sphinx/dev/dev_style_guide_code.rst
```diff
@@ -1,28 +1,169 @@
...
+      int* a;
+      int& b;
+
+- Operands of binary and ternary expressions, trailing comments should be
+  horizontally aligned if
+  they are split over multiple lines, e.g.:
+
+  .. code-block:: c++
+
+      int a = abc *     // My comment
```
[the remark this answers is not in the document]
nritsche: I agree. I am setting it to always break before the operator (also assignments), e.g.:
```
LooooooooooongType loooooooooooooooooooooongVariable
    = someLooooooooooooooooongFunction();

bool value = aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
                     + aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
                     == aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
                     && aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
                     > ccccccccccccccccccccccccccccccccccccccccc;
```

.clang-format
nritsche: I will set it to 100, but it makes me sad and I think this trend is a conspiracy by the monitor manufacturing industry.

kotekan/kotekan.cpp
nritsche: We could tell clang-format not to touch `@param`. That would only result in some still exceeding the 100 columns.

nritsche:
Some more changes I just did: 
- No space after `template` keyword.
- No one-line functions, also not in header files. (except when they are empty: `{ }`).

kotekan/kotekan.cpp
nritsche: In the end this didn't work, because it would still butcher the line after the pragma. We can still indent by hand after clang-format took care of overlength lines. It should leave it like that afterwards.

lib/hsa/hsaDeviceInterface.cpp
```diff
@@ -194,18 +207,17 @@ void hsaDeviceInterface::sync_copy_gpu_to_host(void *dst, void *src, int length)
...
-                                           src, gpu_agent,
-                                           length, 0, NULL, sig);
+    hsa_status = hsa_amd_memory_async_copy(dst, cpu_agent, src, gpu_agent, length, 0, NULL, sig);
     assert(hsa_status == HSA_STATUS_SUCCESS);
 
-    while (hsa_signal_wait_acquire(sig, HSA_SIGNAL_CONDITION_LT, 1,
-                                   UINT64_MAX, HSA_WAIT_STATE_ACTIVE));
+    while (
+        hsa_signal_wait_acquire(sig, HSA_SIGNAL_CONDITION_LT, 1, UINT64_MAX, HSA_WAIT_STATE_ACTIVE))
+        ;
```
nritsche: `pass`?

nritsche:
@jrs65 I turned off auto-formatting for `factory.hpp`, so we don't force you to check if you agree with what the formatting does to all those fancy macros. You can still later enable the formatting in that file and see if that's okay.

docs/sphinx/dev/dev_style_guide_code.rst
cubranic: Richard's order makes sense to me too, not least because it helps me make a mental model of the module dependency hierarchy. 

This is also the order recommended in "Thinking in C++" (via [SO](https://stackoverflow.com/a/14243688)), although the [Google style guide](http://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) suggests the reverse:

> 1. dir2/foo2.h.
> 2. A blank line
> 3. C system files.
> 4. C++ system files.
> 5. A blank line
> 6. Other libraries' .h files.
> 7. Your project's .h files.

Technically, either is OK, as long as the interface file is first, because that helps catch errors with missing includes in the interface header.

we should emphasize that within each section, includes should be sorted alphabetically.

docs/sphinx/dev/dev_style_guide_code.rst
cubranic: I know it looks like Java, but like Rick, I also like keeping the opening brace on the same line as the function signature.

cubranic: Not *just* Java, Go also keeps the opening function brace on the same line. :-)

nritsche:
@andrerenard Thanks!
I fixed the problems you spotted, this is ready to merge.

Issue #288, opened 2018-11-19
negative timespecs


nritsche:
undefined behaviour in the metrics and warnings generated by `receiveFlags`. Caused by negative `timespec` values. Possible fix by making the timespec subtraction function throw an exception instead of returning a negative result.

andrerenard:
Was this fixed @nritsche ?  If so can we close this.

nritsche:
I don't think so

PR #290, opened 2018-11-19
fix negative timespec bug #288

nritsche:


andrerenard:
Is this still being targeted for this release?    What's still required (if anything) @nritsche, @jrs65 ?   At the very least, we need fix the merge problems.

jrs65:
I think not for this release.

andrerenard:
@jrs65 Is this something we want to target finishing for the next release?

jrs65:
Probably. In theory all the time related code do with a refactor, and wanting to do that is why this never got finished up. I don't think it brings any actual science returns though, so we can put it on the list but with a low priority.

nritsche:
Replaced by #789

PR #348, opened 2018-12-12
Add dM endpoint: force register everything with the broker

nritsche:
This adds an endpoint called `/dataset-manager/force-update`. It can be used to force the dM to register everything it has with the broker. When the receiver node gets restarted, this can be used to force a GPU node to register stuff instead of doing so by restarting that node.
If someone has a moment to review this, I think it could be quite useful to get in (edit: looks like I am too late).

This also unrelatedly changes the endpoint string in `restClient::make_request` to be passed by reference.


andrerenard:
@jrs65 are you still planning to review this, or should I take over the review for this code?

jrs65:
I've vaguely skimmed it and it looks fine.

@nritsche it would be very helpful fresh feature branch for things like this in the future? The commit history seems to be cluttered with stuff that I guess is actually now merged, and it's actually quite useful to have it for reviewing.

PR #349, opened 2018-12-13
fix unused variable warnings in Release build with _GOSSEC

nritsche:


lib/core/buffer.c
```diff
@@ -592,7 +594,10 @@ void print_buffer_status(struct Buffer* buf)
         }
     }
     status_string[buf->num_frames] = '\0';
-//TODO: temporary solution to not print buffer status on gossec
+#else
+    (void)buf;
+#endif
+
 #ifndef _GOSSEC
     INFO("Buffer %s, status: %s", buf->buffer_name, status_string);
```
andrerenard: This statement should be moved into the `#if` block above.

nritsche: I think that would result in the line being printed two times in a non-gossec debug build.

nritsche: In order to prevent the buildup of a macro hell, we could replace `_GOSSEC` with a `print_buffer_status` conf value (default true).

andrerenard: But currently if `_GOSSEC && DEBUGGING` then we generate a status string, but don't do anything with it which seems odd.  So maybe the solution is just to wrap the entire thing in `#indef _GOSSEC` if we never want to print anything in the `_GOSSEC` case.

andrerenard: Moving to a config option seems like it might be good, but I'm hoping to work on the new buffers over the break which will have more dynamic logging, so I'm not sure it's worth doing that now.

nritsche: Sounds good. I'll just wrap the function call in `#ifndef _GOSSEC` then.

Issue #355, opened 2018-12-16
Balancing with multithreaded recv processes


jrs65:
In the two multithreaded receiver processes (`applyGains`, `baselineCompression`) we distribute the work to the threads by round robin assigning frames too them. This has an issue that if one thread runs slower than the other one half of the buffer can fill (alternating frames) without it looking like a problem, this causes the whole system to slow down. At the moment this seems to be the dominant cause of frames being dropped in the recv system, so it needs to be fixed.

One option that Rick suggested is that each thread would just claim the next available frame rather than only the next alternating frame. To make this work you need a single `frameID` for all threads, and at the start of the main loop you would lock the frame ID, make a copy and increment it, unlock and then process the frame using the copied value.

jrs65:
From @nritsche:

> The 2 threads of the baselineCompression stage are not processing frames at the same speed. One of them processes even frames, the other processes uneven ones. This leads to one of them congesting, the other one starving.
> ```
> Buffer visbuf_10s_gain, status: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
> Buffer visbuf_10s_stack, status: _X_X_X_X_X_X_X_X_X_X_X_X_X_X_X_X_X_X_X_X_X_X_X_X_X_X_X_X_X_X_X__
> /buffer_recv/26m/instance: No free buffer frames, dropping data from 10.1.2.16
> ```
> I propose they should share one frame ID instead.

jrs65:
Closing #350 as duplicate (this one was longer!)

andrerenard:
Closed by #379

PR #356, opened 2018-12-17
Balancing with multithreaded recv processes

nritsche:
#355


jrs65:
Rebuilding now.

lib/processes/applyGains.cpp
```diff
@@ -25,14 +25,16 @@ REGISTER_KOTEKAN_PROCESS(applyGains);
 applyGains::applyGains(Config& config, const string& unique_name,
                        bufferContainer& buffer_container) :
     KotekanProcess(config, unique_name, buffer_container,
-                   std::bind(&applyGains::main_thread, this)) {
+                   std::bind(&applyGains::main_thread, this)),
+    in_buf(get_buffer("in_buf")),
+    out_buf(get_buffer("out_buf")),
+    frame_id_in(frameID(in_buf)),
```
jrs65: Does this now work if you just use `frame_id_in(in_buf)` without the intermediate object being constructed?

nritsche: Thanks! Fixed.

andrerenard:
Have we decided on if we want to do this as a hot fix, or just merge it in now?  Also it looks like there are now some conflicts with `develop`.

andrerenard:
This needs to be repointed at `master` for the hotfix, and we'll need to fix any strangeness that comes up as a result of that.   Depending on what the tree looks like, it might be easier to start a new branch off master and `patch` these changes into that.

nritsche:
Replaced by #379

PR #358, opened 2019-01-07
Embed the yaml_to_json scripts in the binary

andrerenard:
This change allows us to run `kotekan` from any folder without using the `-d` option and installing the python scripts to a standard location.

Adds a `yamllint` step to the config reader to help track down yaml config errors.

Speeds up the CI build process by allowing multi-threaded builds. 

Also provides an option for giving a different URL for the GPS time-server (`ch_master`) for non-CHIME installations.


andrerenard:
Note we'll likely want to also `yamllint` the configs as a part of the CI tests, but that's lower priority for now.

PR #360, opened 2019-01-08
Add a missing Python dependency on `requests`

cubranic:


cubranic:
BTW, how did this not trip up the CI server? The virtualenv used for testing is not transient?

jrs65:
Which code needs `requests`? Maybe it's not run through the CI server.

cubranic:
The following files contain "import requests":

- scripts/gps_yaml_to_json.py
- python/kotekan/runner.py
- python/scripts/kernel_utilization.py
- python/scripts/rfi_receiver.py
- python/kotekan/scripts/ctl.py

cubranic:
Specifically, it's used in the test runner to execute REST commands: see [line 93](https://github.com/kotekan/kotekan/blob/develop/python/kotekan/runner.py#L93)

jrs65:
Weird that it didn't get caught then.

Issue #275, opened 2018-11-13
Rename KotekanProcess


jrs65:
I'm going to make the controversial suggestion that we find and replace through the whole codebase to rename `KotekanProcess`. This is because the terminology is extremely awkward and confusing, as we are always having to qualify the difference between a KotekanProcess class and the kotekan OS process that's running.

Having chatted with @nritsche and @tristpinsm about it now, suggestions that we have (in my personal, descending order):
- `kotekanTask`
- `kotekanAction`
- `kotekanWorker`
- `kotekanUnit`
- `kotekanModule`
- `kotekanThread` (though they don't actually map 1-1 to threads so that seems like a bad idea)

@andrerenard thoughts?

cubranic:
I’m working on renaming KotekanProcess to Stage. The rename was entirely mechanical and quick, but now I realize there are all these places talking about “process”. Example:

- in core/kotekan_process.hpp, there is a macro REGISTER_KOTEKAN_PROCESS, which is used by every subclass of Stage. Should I change that to REGISTER_KOTEKAN_STAGE?

- prometheusMetrics: method “add_process_metric”

- processFactory: the class name; method “kotekan_register_process”

- various comments talking about “processes”

- the name of the directory lib/process

Should I update all of those to “stage”?

cubranic:
OK, I think I'm done for now and would like some feedback: see PR #362. There are a few things left unfinished, but I shouldn't necessarily be the one to do them.

- config file parameter is still named "kotekan_process", rather than "kotekan_stage" because I wanted changes to config files to be done by someone more familiar with their syntax and effects

- subclasses of `kotekan::Stage` (formerly `KotekanProcess`) that are not in `lib/core` need to have updated their class, method, member, and variable names, as well as comments (API and in-code) to reflect the change "process" -> "stage". I'd like this to be done by their authors (I'll handle the baseband ones), if we agree that the changes (and choices) I made are OK.

- I think @andrerenard said that stage subclasses can remain outside the "kotekan" namespace for now. I'm not sure where they should go, but probably they should be put *somewhere*.
- directory 'lib/processes' needs to be renamed to 'lib/stages'. I can do this,  but wanted to do it last.

**Important:** Where I renamed classes, I used capitalized `CamelCase`, as we've agreed. One ambiguity in our style standard was file names. I choose lowercased `snake_case`, e.g., `stage.hpp` and `stage_factory.hpp`. Opinions please?

andrerenard:
This was done in #362

PR #362, opened 2019-01-15
Introduce 'kotekan' namespace and rename KotekanProcess to Stage

cubranic:
This is mostly a mechanical rename as discussed in #275, with just some light proofreading of code comments.

There are still some inconsistencies in naming, e.g.:

- config file parameter is still named "kotekan_process", rather than "kotekan_stage" because I wanted changes to config files to be done by someone more familiar with their syntax and effects
- subclasses of `kotekan::Stage` (formerly `KotekanProcess`) that are not in `lib/core` need to have updated their class, method, member, and variable names, as well as comments to reflect the change "process" -> "stage"
- directory 'lib/processes' needs to be renamed to 'lib/stages', but I wanted to do this last

### DIFF
--- a/lib/core/StageFactory.hpp
+++ b/lib/core/StageFactory.hpp
@@ -18,18 +18,13 @@
 #include <map>
 #include <string>
 
-// Name space includes.
-using json = nlohmann::json;
-using std::map;
-using std::string;
-
 namespace kotekan {
 
 class Stage;
 
 class StageMaker {
 public:
-    virtual Stage* create(Config& config, const string& unique_name,
+    virtual Stage* create(Config& config, const std::string& unique_name,
                           bufferContainer& host_buffers) const = 0;
 };
 
@@ -43,15 +38,16 @@ public:
     // Creates all the stages listed in the config file, and returns them
     // as a vector of Stage pointers.
     // This should only be called once.
-    map<string, Stage*> build_stages();
+    std::map<std::string, Stage*> build_stages();
 
 private:
-    void build_from_tree(map<string, Stage*>& stages, json& config_tree, const string& path);
+    void build_from_tree(std::map<std::string, Stage*>& stages, nlohmann::json& config_tree,
+                         const std::string& path);
 
     Config& config;
     bufferContainer& buffer_container;
 
-    Stage* create(const string& name, Config& config, const string& unique_name,
+    Stage* create(const std::string& name, Config& config, const std::string& unique_name,
                   bufferContainer& host_buffers) const;
 };
 
@@ -75,7 +71,7 @@ public:
     StageMakerTemplate(const std::string& key) {
         StageFactoryRegistry::kotekan_register_stage(key, this);
     }
-    virtual Stage* create(Config& config, const string& unique_name,
+    virtual Stage* create(Config& config, const std::string& unique_name,
                           bufferContainer& host_buffers) const override {
         return new T(config, unique_name, host_buffers);
     }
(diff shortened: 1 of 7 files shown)


cubranic:
> Why wouldn't you name each file after the class in it? They would then use capitalized CamelCase, like here.

The files *are* named after the class in them, just cased differently. I got the file naming convention from the Google style guide, because I couldn't find any mention of it in our style guide (I missed the section you linked, thank). That's why I wanted it discussed here. I find naming files CamelCase.hpp gross, TBH, but I don't have to look at it too often, so I'll go along with the consensus here.

cubranic:
> I also propose to move `restServer`, `connectionInstance`, `HTTP_RESPONSE`, etc. to a `kotekan::REST` namespace.

Why bother starting a new namespace for 3-4 classes?

Also, namespace is lowercased, so that would be `kotekan::rest`. (It looks a little weird, I know. But it's still better that than introducing confusion over what's namespace and what's class name.)

And if we do go with this, the classes could drop the "rest" prefix from their name, so `Server` and `Client` in `kotekan::rest` namespace.

nritsche:
> Why bother starting a new namespace for 3-4 classes?
> 
> Also, namespace is lowercased, so that would be `kotekan::rest`. (It looks a little weird, I know. But it's still better that than introducing confusion over what's namespace and what's class name.)
> 
> And if we do go with this, the classes could drop the "rest" prefix from their name, so `Server` and `Client` in `kotekan::rest` namespace.

Just because I think `kotekan::HTTP_RESPONSE` sounds wrong. I agree with the `rest`.

andrerenard:
We could make `HTTP_RESPONSE`  a public member of `RestServer` i.e. `kotekan::RestServer::HTTP_RESPONSE` although if it's also used by `RestClient` that could be an issue.

andrerenard:
Also not to open this can of worms, but it seems like a number of people don't like the look of `CamelCase`, and we are in the process of renaming everything so...   What do people think of using underscores and lowercase for everything: variables, members, functions *and* types?   I've always liked having a distinction for types, but I also dislike having to type capitals, so I would be open to changing that if people feel strongly at this junction.

jrs65:
CamelCase for types.
snake_case for everything else.

I have a strong opinion on this as I do about everything.

andrerenard:
@jrs65 that aligns with my opinion, so we'll close that can of worms, and keep with that. 

The file names could be open for debate, but I'd prefer they match the main class name they contain.  So we'd would have `CamelCase.[c|h]pp` for _most_ files.

cubranic:
@nritsche I notice you removed various "using" statements from header files. I appreciate the help, and it's true that we discussed doing it, but it wasn't really part of this PR and while doing it you introduced some unnecessary "kotekan::" qualifiers to signatures for types that were already in the "kotekan" namespace. E.g., in lib/core/metadataFactory.hpp:

```c++
namespace kotekan {

class metadataFactory {

public:
    // One metadataFactory should be created for each set of config and buffer_container
    metadataFactory(kotekan::Config& config);
```

I'm cleaning the Git history by moving this out into a separate commit. Please don't make any more changes in this branch until I'm done.

cubranic:
I'm done with the Git history rewriting. We might want to consider moving the last one into a separate PR, and do it more widely. (There are still many header files that contain "using".)

cubranic:
That commit history looks a bit different from my local one: 2f97799c should be the last commit, I'm not sure why it's being listed before 1769734 above.

PR #363, opened 2019-01-17
Bug in pulsar date math and PP truncation

cherryng:
Critical bug fixes:
- incomplete date math in the pulsar phase calculation (affecting month of January and February) 
- proper truncation of large negative numbers in pulsar post processing

### DIFF
--- a/lib/hsa/hsaPulsarUpdatePhase.cpp
+++ b/lib/hsa/hsaPulsarUpdatePhase.cpp
@@ -126,6 +126,10 @@ void hsaPulsarUpdatePhase::calculate_phase(struct psrCoord psr_coord, timespec t
     timeinfo = localtime (&time_now.tv_sec);
     uint year = timeinfo->tm_year+1900;
     uint month = timeinfo->tm_mon+1;
+    if (month <3) {
+        month = month +12;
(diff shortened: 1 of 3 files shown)


jrs65:
Just looking at this to see what the changes going in are.

One suggestion (@cherryng @kvand), I think you can simplify the whole date calculation by noting that JD and Unix time are linearly related. Just take:

```C++
double JD = time_now.tv_sec / 86400.0 + 2440587.5;
```
which would replace all of lines 126-134. @tristpinsm also has an implementation (to MJD) in `lib/utils/pulsarTiming.hpp` called `ts2mjd` which directly converts a time spec. Note that neither this nor your original implementation correctly accounts for leap seconds, but it seems like you don't need that level of accuracy.

At the moment you're going to broken down time calendar time, and then essentially reversing that step to get to an epoch time again.

andrerenard:
I agree with @jrs65 that we should just use the `ts2mjd` function or a simplified function here.  However since we want to get this on the sky quickly, and this patch has already been tested. I'm going to go with @kvand's thoughts and approve this as is and deploy, and then open a ticket to update this for the next release.

PR #366, opened 2019-01-23
Read debug frames

tristpinsm:
Added the ability to read back GPU buffer dumps in python.

### DIFF
--- a/lib/processes/rawFileRead.cpp
+++ b/lib/processes/rawFileRead.cpp
@@ -91,6 +91,8 @@ void rawFileRead::main_thread() {
             break;
         }
 
+        fclose(fp);
+
         INFO("rawFileRead: Read frame data from %s into %s[%i]", full_path, buf->buffer_name,
              frame_id);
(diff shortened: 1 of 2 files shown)


jrs65:
@tristpinsm this looks great. However, could you rebase only the relevant commits on top of the current `develop` branch? I think you've started this from a non merged branch and thus it looks like there's an extra 84 commits in addition to the five that look useful. It makes it much easier to review.

jrs65:
You'll need to do the rebase and then force push to replace this branch.

tristpinsm:
Oh, ok. Is this because commits are being squashed into one after the PR?

jrs65:
> Oh, ok. Is this because commits are being squashed into one after the PR?

Partly yes. Though you'd be missing merge commits if you didn't do this and we just used a normal merge. Generally it's way better to start a fresh branch from develop every time you start to develop something new unless you really need something else in a different branch. Thanks!

PR #367, opened 2019-01-29
Generic pattern test stage

nritsche:
This new `visTestPattern` is idle until an update is sent to it's endpoint. The update includes the test pattern FPGA buffer, a test name and number of frames. This stage will then run the given test (compare against the expected vis data calculated from the given FPGA buffer) for the given number of frames. Afterwards it is idle until the next update is received.
The endpoint is to be called by [`capo`](https://github.com/chime-experiment/capo).

**done:**   ~Needs a fix: don't wait in restServer callback~


lib/processes/visTestPattern.cpp
```diff
@@ -38,71 +51,41 @@ visTestPattern::visTestPattern(Config& config, const std::string& unique_name,
...
-    outfile.open(outfile_name);
-    if (!outfile.is_open()) {
-        throw std::ios_base::failure("visCheckTestPattern: Failed to open "
-                                     "out file "
-                                     + outfile_name);
-    }
-    outfile << "fpga_count,time,freq_id,num_bad,avg_err,min_err,max_err" << std::endl;
+    // Subscribe to the dynamic config update: used to start a test for a number of frames.
+    kotekan::restServer::instance().register_post_callback(
+        endpoint_name, std::bind(&visTestPattern::receive_update, this, std::placeholders::_1,
```
andrerenard: There should be a `remove_json_callback` in the destructor for this endpoint.

lib/processes/visTestPattern.cpp
```diff
@@ -124,131 +107,393 @@ void visTestPattern::main_thread() {
 
     uint64_t i_frame = 0;
 
+    // Wait for the first frame
+    if (wait_for_full_frame(in_buf, unique_name.c_str(), frame_id) == nullptr) {
+        if (outfile.is_open())
+            outfile.close();
+        return;
```
andrerenard: Do you also want to `raise(SIGINT)` here, or some other error condition?

lib/processes/visTestPattern.hpp
```diff
@@ -10,17 +11,32 @@
...
+#include "json.hpp"
+
 #include <fstream>
 #include <stddef.h>
 #include <string>
 #include <vector>
 
+using json = nlohmann::json;
+
+#define REPORT_PRECISION 7
```
andrerenard: Should this be made a function of the tolerance?

lib/testing/fakeVis.cpp
```diff
@@ -120,10 +121,35 @@ fakeVis::fakeVis(Config& config, const string& unique_name, bufferContainer& buf
...
+            config.get<std::vector<cfloat>>(unique_name, "input_values");
+        if (input_values.size() != num_elements) {
+            throw std::invalid_argument("fakeVis: lengths of input values ("
+                                        + std::to_string(input_values.size())
+                                        + ") and number of elements ("
+                                        + std::to_string(num_elements) + ") have to be equal.");
+        }
+
+        size_t num_prods = num_elements * (num_elements + 1) / 2;
+        if (num_prods < 3)
```
andrerenard: What case is this catching?   Wouldn't `num_elements == 1` just have one product, i.e. it's autocorrelation?

lib/processes/visTestPattern.cpp
nritsche: Thanks!
Should we rename `register_post_callback` `register_json_callback` or is there a reason in the naming difference between the `register_` and `remove_` methods?

andrerenard: We should rename them, I'm not sure why we ended up with different names.  I think both should be `*_json_callback` since it's not a generic POST callback, it does assume a JSON encoded string.

lib/processes/visTestPattern.cpp
nritsche: Why? This is a normal exit condition, even though as bit unusual on start. I will add an `INFO`.

andrerenard: Oh, sorry, I must have misread that statment.  I agree this is fine as is.

lib/processes/visTestPattern.hpp
nritsche: That's a good idea.

nritsche:
[CHIME wiki on Automatic Pattern Tests](https://bao.chimenet.ca/wiki/index.php/Automatic_Pattern_Test)

PR #370, opened 2019-02-07
Implement metrics for number of received and completed baseband requests

cubranic:
Received requests are counted at the node (kotekan) level by metric
`kotekan_baseband_requests_total`. It has no labels, other than Kotekan's
standard "stage".

Completed requests are tracked by `kotekan_baseband_readout_total`. This is kept
per-stage, and provides two useful labels: "freq_id" provides the stage's
frequency channel, and "status" separates the time series based on the outcome
of the request ("done" if successful, "error" if there was a problem writing the
data, and "no_data" if there was no available baseband data for the requested
time interval.

### DIFF
--- a/lib/core/basebandApiManager.cpp
+++ b/lib/core/basebandApiManager.cpp
@@ -52,6 +52,13 @@ void to_json(json& j, const basebandDumpStatus& d) {
     }
 }
 
+basebandApiManager::basebandApiManager() :
+    // clang-format off
+    metrics(prometheusMetrics::instance()) // clang-format on
+{
+    metrics.add_stage_metric("kotekan_baseband_requests_total", "baseband", request_count);
+}
+
 basebandApiManager& basebandApiManager::instance() {
     static basebandApiManager _instance;
     return _instance;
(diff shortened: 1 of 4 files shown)


cubranic:
OK, this is reformatting of the constructor member initializers is just plain wrong:
```
-basebandApiManager::basebandApiManager() :
-    metrics(prometheusMetrics::instance())
-{
+basebandApiManager::basebandApiManager() : metrics(prometheusMetrics::instance()) {
```

I've never seen it squished on the same line as the signature instead of below, one member initialization per line. It's much more difficult to read it that way.

Can we make it accept my formatting here, @andrerenard and @nritsche?

andrerenard:
@cubranic you can exclude that line from clang with the magic comment described here: http://lwlab.dunlap.utoronto.ca/kotekan/dev/dev_style_guide_code.html  However I agree that's an odd way to format things, and I'd prefer not to have a lot of `clang-format off` lines to deal with that.  @nritsche can you see if that's an easy rule to disable? or is it tied up with another useful rule?

nritsche:
There is currently no option to force a line break here, it seems. So clang-format is just consistent with the column limit (100) and doesn't break the line.

PR #372, opened 2019-02-09
bugfix: missing throw keyword

nritsche:
Without `throw` these Exceptions were only created, but probably never seen by anyone.

In the case of `visRawReader`, `munmap` could also be moved to the end of `main_thread` (destructors shouldn't throw). But in any case, how would throwing an exception there make sense anyways? I think it's not caught and kotekan would be exiting already anyways.


nritsche:
Replaced by #382 for hotfix.

Issue #369, opened 2019-01-31
About half the nodes are rejecting gating config update


tristpinsm:
Whenever I send a config update for the pulsar gating through `KotekanMaster` about half the nodes (~100) respond with 
```
{'UnknownError': 'Response={u\\'message\\': u\"configUpdater: Update to endpoint \\'/updatable_config/gating/psr0_config\\' contained value \\'dm\\' of type float (expected type number).\", u\\'code\\': 400}, Error:HTTP 400: BAD_REQUEST\\n'}
```
 The other nodes are fine and seem to be producing the gated stream as expected.

Here a typical JSON update :
```
curl csBfs:54323/update-pulsar-gating -X POST -H "Content-Type: application/json" -d '{"coeff": [[8.451188749386428e-10, -0.00013213758449419458, -6.627836033917244e-08, -4.732131671123605e-10, 2.331131336850821e-13, 4.552811969358192e-16, -1.401626224241363e-19, -3.6089049190384454e-22, -5.078171175765845e-25, 4.410000424496888e-27, 1.0968063735702019e-29, -4.2400626494554845e-32], [1.4344346305358498e-10, -0.0002592509101170875, -2.711651590138689e-07, 8.52119140931717e-11, 5.599454690706729e-13, -8.343620189603956e-17, -3.6209919173919025e-19, 1.5344960473424132e-22, 3.130489472711657e-25, -3.232114419701468e-27, -2.675647370327176e-30, 3.085769738832778e-32], [4.074491285715008e-10, -0.0003465987739188493, 4.6654664661639314e-08, 5.166329675391741e-10, 5.284976788026692e-14, -4.987223509650824e-16, -4.02888773664995e-20, 5.492535530479008e-22, 3.583726999905542e-25, -1.125232102606971e-26, -6.222557697821949e-30, 1.4221809126102123e-31]], "t_ref": [58514.6041666666, 58514.8124999999, 58515.0208333333], "phase_ref": [2909514952.315827, 2909565128.0972075, 2909615303.838368], "rot_freq": 2.787546496219, "dm": 158.521, "segment": 18000.0, "pulsar_name": "B1933+16", "pulse_width": 0.0177, "enabled": true}'
```
The relevant key is `"dm": 158.521`.

It's worth noting that if I send an empty update to disable gating, then all the nodes seem to respond successfully.
```
[...]

PR #373, opened 2019-02-11
configUpdater: strict about number types (#369)

nritsche:
configUpdater is now strict about number types and does not accept updates that were of a different number type before. The number types are `integer`, `unsigned integer` and `float`.

### DIFF
--- a/lib/core/configUpdater.cpp
+++ b/lib/core/configUpdater.cpp
@@ -3,6 +3,7 @@
 #include "Stage.hpp"
 #include "errors.h"
 #include "restServer.hpp"
+#include "visUtil.hpp"
 
 #include "fmt.hpp"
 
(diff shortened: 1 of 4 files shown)


andrerenard:
I think we should consider `unsigned` and `signed` as `signed`.  I think it's reasonable to expect the default config might have a positive number, but be willing to accept a negative number.  However without getting into having a scheme, this means we have to make default values negative (and initial updates all have to be negative) to support a signed number.   So I think we should just assume integers are all signed, and leave it to subscribers to accept or reject negatives.  This problem doesn't exist for floats, since you can always pass e.g. `val: 1.0` instead of `val: 1` when you want to make sure the number is treated as a float.

andrerenard:
Also I think we should point this at `master` since it's a target for the hotfix release at the end of this week.

jrs65:
Oh yikes. I hadn't thought about signed vs unsigned. Hmmm....

I would make them all signed, but I did vote for the strict validation version.

nritsche:
I just assumed yaml files would have to use something like `123u` for unsigned and unittests would fail if if this is not true. Thanks, Andre. Fix is coming.

nritsche:
replaced by #374

PR #374, opened 2019-02-11
configUpdater: more strict about number types (#369)

nritsche:
configUpdater is now more strict about number types and does not accept `float` updates that were of a number type `integer` or `unsigned` before. `integer` and `unsigned` values are allowed to be used to update each other.

### DIFF
--- a/lib/core/configUpdater.cpp
+++ b/lib/core/configUpdater.cpp
@@ -2,7 +2,9 @@
 #include "errors.h"
 #include "restServer.hpp"
 #include "KotekanProcess.hpp"
+#include "visUtil.hpp"
 
+#include "json.hpp"
 #include "fmt.hpp"
 #include <iostream>
 #include <signal.h>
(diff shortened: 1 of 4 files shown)


lib/core/configUpdater.cpp
```diff
@@ -166,28 +168,19 @@ void configUpdater::rest_callback(connectionInstance &con, nlohmann::json &json)
...
-
-            if (it.value().type() == nlohmann::json::value_t::number_float &&
-                _init_values[uri].at(it.key()).type() != it.value().type()) {
-                std::string msg = fmt::format(
-                            "configUpdater: Update to endpoint '{}' contained" \
-                            " value '{}' of type float (expected type {}).",
-                            uri.c_str(), it.key().c_str(),
-                            _init_values[uri].at(it.key()).type_name());
+            // ...and for correct types:
+            // If type_name differs, they are of different type and we care.
```
andrerenard: It might be worth saying what we are trying to achieve here in more detail.  i.e. something like "Reject changes in general type (string vs. number vs. option), and reject changes between (un)signed integers and floats, but allow both signed and unsigned integers to be interchanged"   As written it's fairly hard to parse from the comment what this does.

nritsche: Done

PR #379, opened 2019-02-12
Balancing with multithreaded recv processes

nritsche:
#355 

replaces #356

### DIFF
--- a/lib/processes/applyGains.hpp
+++ b/lib/processes/applyGains.hpp
@@ -91,10 +91,10 @@ private:
     /// The gains and when to start applying them in a FIFO (len set by config)
     updateQueue<gainUpdate> gains_fifo;
 
-    /// Output buffer with gains applied
-    Buffer * out_buf;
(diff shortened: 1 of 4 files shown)


lib/processes/applyGains.hpp
```diff
@@ -111,14 +111,22 @@ class applyGains : public KotekanProcess {
...
-    void apply_thread(int thread_id);
+    void apply_thread();
 
     ///Vector to hold the thread handles
     std::vector<std::thread> thread_handles;
 
     /// Number of parallel threads accessing the same buffers (default 1)
     uint32_t num_threads;
 
+    /// Input frame ID, shared by compress threads.
```
andrerenard: I think you mean `apply threads`

nritsche:
```suggestion
    /// Input frame ID, shared by apply threads.
```

nritsche: Done

PR #380, opened 2019-02-12
Psr time offset

cherryng:
- align pulsar vdif packet to %1.6ms instead of %625 samples
- remove fpga_seq_num_low32 from vdif header (to avoid confusion)


arunnaidu123:
Code looks fine. I will make required code changes on the pulsar side.

PR #383, opened 2019-02-15
Turned down the number of EV iterations

jrs65:


andrerenard:
@jrs65 have we decided these numbers are good now?  If so I'd like to merge this in, and then pull the hot fixes back into `develop`.

PR #385, opened 2019-03-02
Patch Python2 numpy dtypes to work with the future.unicode_literals import

tristpinsm:
Mark numpy dtype labels as ASCII to work with unicode_literals. Only in VisRaw for now. Also made VisRaw work with stacked data.

If this is deemed correct I can go through and fix the other dtypes in the file. Also, it just occurred to me that this could potentially mess things up in Python 3... Any idea?

### DIFF
--- a/python/kotekan/visbuffer.py
+++ b/python/kotekan/visbuffer.py
@@ -3,6 +3,7 @@
 # Python 2/3 compatibility
 from __future__ import (absolute_import, division,
                         print_function, unicode_literals)
+from future.utils import native_str
 from builtins import (ascii, bytes, chr, dict, filter, hex, input,
                       int, map, next, oct, open, pow, range, round,
                       str, super, zip)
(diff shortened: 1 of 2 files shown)


tristpinsm:
I found a similar issue trying to run kotekan on a machine where python 3 is the default interpreter. Casting to `str` in the yaml parsing script seems to have fixed it.

andrerenard:
Thanks for finding that problem with the yaml script @tristpinsm   I'm good with merging this in, but maybe wait for @jrs65 or @nritsche who know more about python3 to also comment on if this is the right solution.

p.s. the unit test failure is because we changed the version of `clang-format` on the CI server.  There is a fix for it in another branch I'm hoping to merge in soon.

jrs65:
Yes. That does indeed mess things up in Python 3. You need to use the native string type on each platform. You can get access to that as `future.utils.native_str`. I recommend writing a routine which just processes the list of tuples and processes each string (unicode) using `native_str(...)`

tristpinsm:
OK, I think I've ironed out the remaining bugs. I was able to run it in both Python 2 and 3 successfully.

python/kotekan/visbuffer.py
```diff
@@ -285,36 +286,40 @@ def __init__(self, filename, mmap=False):
...
+            ("vis", np.complex64, self.num_stack),
+            ("weight", np.float32, self.num_stack),
+            ("flags", np.float32, self.num_elements),
             ("eval", np.float32,  self.num_ev),
             ("evec", np.complex64, self.num_ev * self.num_elements),
             ("erms", np.float32,  1),
             ("gain", np.complex64, self.num_elements),
-        ], align=True)
+        ]
+        # Process dtype labels to ensure Python 2/3 compatibility
```
jrs65: Could you start this like: `# TODO: Python 3 ...`, I'm trying to make it easy to search for Python 2/3 work arounds when we want to remove them.

tristpinsm: Done.

jrs65:
Changed it slightly just to match the exact form I've used everywhere else (easier to grep!)

andrerenard:
I had to cancel the CI builds for this one because I was trying to resolve issues with the MacOS system at the time they were posted.   Can you please pull the current develop and push, that should trigger a new CI build and resolve build issues that showed up today.

tristpinsm:
OK, looks like it's running now.

PR #386, opened 2019-03-04
[WIP] Support 16/256-element systems

andrerenard:
This is currently the result of merging the working with 16-element branch with the current `develop`, that resulted in hundreds of merge conflicts thanks to all the renaming and other updates, so it's very possible this branch unexpected things in it still.


andrerenard:
I'm going to try another approach to merging this in.  The current 16-element code branch as deviated quite a bit from this merge attempt, so I'm going to close this PR and try another way of getting things into develop.

PR #387, opened 2019-03-04
Merge 2018.12 hotfixes into develop

andrerenard:


nritsche:
To prevent what just happened here, I found it useful to always hit `make clang-format` twice.

The merge of my changes looks good to me.

andrerenard:
@nritsche  I tried running it a few more times.  It still doesn't pick up that change on my system.

nritsche:
Did you update clang-format on both?

andrerenard:
Looks like `clang-format` on the CI server is running `8.0`, and I'm running `9.0`  So the newer one doesn't apply this rule, or does it differently.

andrerenard:
In fact `clang-format 9.0` on my system reverts the change that `clang-format 8.0` wants on the CI server.  Which is super annoying.  I guess we need to force people to use just one clang version.  Let me see about updating clang-format on the CI server.  I assume you are using 9.0 @nritsche ?

nritsche:
Let's update it on the CI-server then. This is about line break weights...

andrerenard:
FWIW I do like the `clang-format-9` formatting of that line better as well.

PR #390, opened 2019-03-05
Frb upchan bp

cherryng:
16-value correction for the upchannelized bandpass. it is derived empirically.


andrerenard:
Also you'll need to pull in the latest develop to fix that `clang-format` issue.

PR #401, opened 2019-03-11
Update receiver config for ringmaps

tristpinsm:
This makes small changes to the receiver config to interface with the ringmap maker on `recv1`.

The list of frequencies I included is just a placeholder that I used for testing, but I will update it shortly with selected frequencies, so please don't merge until then. I just wanted to make sure this is to get this going before the deployment window closes.


tristpinsm:
I've updated the list of 64 frequencies to avoid RFI bands and nodes that were down in the `20190215T225253Z_chimecal_corr` acquisition. I haven't tested this last commit on `recv2` yet.

PR #402, opened 2019-03-11
Patterntest fixes after testing

nritsche:
- Speed up of expected data computation
- Export metrics also for good frames

### DIFF
--- a/lib/processes/visTestPattern.cpp
+++ b/lib/processes/visTestPattern.cpp
@@ -471,15 +471,17 @@ void visTestPattern::compute_expected_data() {
     for (size_t f = 0; f < num_freqs; f++)
         expected[f].resize(num_prods);
 
-    // Sort input feed values by frequency.
+        // Sort input feed values by frequency.
 #ifdef _OPENMP
     INFO("Speeding up calculation of exptected data with OpenMP.");
 #pragma omp parallel for
 #endif
     for (size_t p = 0; p < num_prods; p++) {
         // Generate auto product.
-        std::vector<cfloat> buf_a = fpga_buf_pattern.at(inputs.at(prods.at(p).input_a).correlator_input);
-        std::vector<cfloat> buf_b = fpga_buf_pattern.at(inputs.at(prods.at(p).input_b).correlator_input);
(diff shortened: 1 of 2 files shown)


lib/processes/visTestPattern.cpp
```diff
@@ -463,23 +466,31 @@ void visTestPattern::compute_expected_data() {
...
 
+    INFO("Computing expected data...");
     expected.resize(num_freqs);
+    for (size_t f = 0; f < num_freqs; f++)
+        expected[f].resize(num_prods);
 
-    // Sort input feed values by frequency.
-    for (size_t f = 0; f < num_freqs; f++) {
+        // Sort input feed values by frequency.
+#ifdef _OPENMP
```
andrerenard: One concern I have here is that OPENMP doesn't respects the cpu affinity model that kotekan uses.  On the `recv2` style nodes where having exclusive cores isn't as important, I think this is ok for these kind of testing `stages`.  However we might want to consider implementing some kind of control over the OPENMP thread affinity.  From what I can tell online I'm not sure that can actually be done on a per `stage`/MP call basis, but we might be able to set the global OPENMP affinity via the config file (then having kotekan internally modifying the OPENMP environment variables).   That is likely a change external to this PR, and as long as we don't think this will cause problems for say the networking threads or disk writing threads while the test is running, then I'm ok with change as is.

lib/processes/visTestPattern.cpp
```diff
@@ -463,23 +466,31 @@ void visTestPattern::compute_expected_data() {
     size_t num_freqs = freqs.size();
     size_t num_prods = prods.size();
 
+    INFO("Computing expected data...");
```
andrerenard: To me this seems like more of a debug statement (same with the second one).  Maybe switch this to debug, but then set the config to make this stage run at the `debug` log level if this is useful.

nritsche: I left it there as INFO, because when run without OMP, this takes around 10min and drops all frames in that time.

nritsche:
@andrerenard Can we merge this?

PR #403, opened 2019-03-11
Fix bug where stacked weights can be set to NaN when should be 0

tristpinsm:


tristpinsm:
I've checked that this fixes the issue running on my laptop using the `test_pattern_freq` mode of `fakeVis` to generate data with zero weights.

jrs65:
Great. Can you fix the build failure (formatting)?

tristpinsm:
oh I didn't even realize we had that...

jrs65:
Okay. Thanks Tristan. Looks good to me, but I'm not sure what the current merge/deploy plan is. @andrerenard should decide what to do with this.

nritsche:
@tristpinsm if you install clang-format 9, you can do `make clang-format` in `kotekan/build`

tristpinsm:
ah ok! thanks for the tip, that'll save me some useless commits

PR #415, opened 2019-05-09
RFI receiver script config tracking

nritsche:


nritsche:
Thanks, @andrerenard . I added the registration of the script startup time and git version. This can get merged now, I think.

PR #418, opened 2019-05-16
Fix double counting of RFI and lost packets

andrerenard:
It's possible that `hsaRfiMaskOutput.cpp` will count packet loss zeroed samples from `zeroSamples.cpp` twice.   This is because first `zeroSamples` adds to the lost samples counter for packet loss, and then later the RFI system might zero out the same samples (but in 256 time sample blocks), leading to double counting.   This fix checks if any of the samples zeroed in `zeroSamples` would be counted twice in `hsaRfiMaskOutput` and corrects for them. 

Also adds a new metadata value for number of RFI flagged samples.

### DIFF
--- a/lib/hsa/hsaRfiMaskOutput.cpp
+++ b/lib/hsa/hsaRfiMaskOutput.cpp
@@ -85,7 +85,7 @@ void hsaRfiMaskOutput::finalize_frame(int frame_id) {
         // Total number of samples flaged as RFI
         uint32_t flagged_samples = 0;
 
-        // The number of samples flaged as RFI less the number of
+        // The number of samples flagged as RFI less the number of
         // samples flaged as missing (i.e. packet loss/currupt packet).
         uint32_t net_lost_samples = 0;
 
--- a/lib/hsa/hsaRfiMaskOutput.cpp
+++ b/lib/hsa/hsaRfiMaskOutput.cpp
@@ -98,8 +98,7 @@ void hsaRfiMaskOutput::finalize_frame(int frame_id) {
                 // Remove any samples which were also counted as lost samples
                 // in this block of data.
                 for (uint32_t j = 0; j < _sk_step; ++j) {
-                    uint32_t lost_samples_idx = subframe * _sub_frame_samples +
-                                                i * _sk_step + j;
+                    uint32_t lost_samples_idx = subframe * _sub_frame_samples + i * _sk_step + j;
                     assert(lost_samples_idx < (uint32_t)_lost_samples_buf->frame_size);
                     if (lost_samples_frame[lost_samples_idx] == 1)
                         net_lost_samples--;
--- a/lib/hsa/hsaRfiMaskOutput.cpp
+++ b/lib/hsa/hsaRfiMaskOutput.cpp
@@ -86,7 +86,7 @@ void hsaRfiMaskOutput::finalize_frame(int frame_id) {
         uint32_t flagged_samples = 0;
 
         // The number of samples flagged as RFI less the number of
-        // samples flaged as missing (i.e. packet loss/currupt packet).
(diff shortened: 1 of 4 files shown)


lib/hsa/hsaRfiMaskOutput.cpp
```diff
@@ -66,40 +72,72 @@ hsa_signal_t hsaRfiMaskOutput::execute(int gpu_frame_id, hsa_signal_t precede_si
...
+    uint32_t total_flagged = 0;
+    uint32_t total_net_lost_samples = 0;
+    uint32_t sub_frame_mask_len = _sub_frame_samples / _sk_step;
+
+    for (uint32_t subframe = 0; subframe < _num_sub_frames; ++subframe) {
+
+        // Total number of samples flaged as RFI
+        uint32_t flagged_samples = 0;
+
+        // The number of samples flaged as RFI less the number of
```
nritsche:
```suggestion
        // The number of samples flagged as RFI less the number of
```

lib/hsa/hsaRfiMaskOutput.cpp
```diff
@@ -66,40 +72,72 @@ hsa_signal_t hsaRfiMaskOutput::execute(int gpu_frame_id, hsa_signal_t precede_si
...
+        // samples flaged as missing (i.e. packet loss/currupt packet).
```
nritsche:
```suggestion
        // samples flagged as missing (i.e. packet loss/corrupt packet).
```

lib/hsa/hsaRfiMaskOutput.cpp
```diff
@@ -66,40 +72,72 @@ hsa_signal_t hsaRfiMaskOutput::execute(int gpu_frame_id, hsa_signal_t precede_si
...
+    uint8_t* lost_samples_frame = _lost_samples_buf->frames[_lost_samples_buf_id];
+
+    // Add the number of flagged samples to the metadata
+    uint32_t total_flagged = 0;
+    uint32_t total_net_lost_samples = 0;
+    uint32_t sub_frame_mask_len = _sub_frame_samples / _sk_step;
+
+    for (uint32_t subframe = 0; subframe < _num_sub_frames; ++subframe) {
+
+        // Total number of samples flaged as RFI
```
nritsche:
```suggestion
        // Total number of samples flagged as RFI
```

nritsche:
Closes #418

lib/hsa/hsaRfiMaskOutput.cpp
```diff
@@ -66,40 +72,72 @@ hsa_signal_t hsaRfiMaskOutput::execute(int gpu_frame_id, hsa_signal_t precede_si
...
 
-            total_lost += lost_in_subframe;
+        for (uint32_t i = 0; i < sub_frame_mask_len; ++i) {
+            uint32_t mask_idx = subframe * sub_frame_mask_len + i;
+            assert(mask_idx < (uint32_t)_rfi_mask_output_buf->frame_size);
+            if (mask_frame[mask_idx] == 1) {
+                flagged_samples += _sk_step;
+                net_lost_samples += _sk_step;
+                // Remove any samples which were also counted as lost samples
+                // in this block of data.
```
nritsche: Were did they get double counted?

andrerenard: They get counted in `zeroSamples.cpp`.  However zeroed packets often look like RFI and trigger the RFI zeroing, which means if the RFI system adds to the lost sample count because of zeroing it needs to correct for the zeroing of the packet, but only if the two are coincident in time.

nritsche: Ok, makes sense now!

PR #419, opened 2019-05-17
dataset management: hardcoded state type names

nritsche:
The dataset broker now requires a state type name sent with the states. This is fixed here in 21c1182.

These type names until now were what comes out of `typeid::name()` and looked like `6freqState`, `9prodState`. The type names will in the future be used in the CHIME database to look up states, so they have to look better. I first wanted to just add a hardcoded string to each state that is used for that. In the end I decided against that and demangled the strings. That only works with g++ and is deactivated for all other compilers, resulting in something more like the above. The reason for the decision is that the datasetState factory actually needs the type names at compile time and everything I came up with would have resulted in extra work for adding new states that was hard to enforce.


jrs65:
I've not had more than a skim of it yet, but I'm a little concerned if we're implementing something that will have different external behaviour on different compilers. In particular I use `clang++` a lot for development, is there a similar demangling we can use for clang?

A hardcoded string doesn't seem like a bad option, but I remember it being a pain. I think you can't have a `const static std::string` member be initialised inline until C++17, so you had to define and then annoyingly initialise in the cpp file.

andrerenard:
I might just be missing an edge case, but it seems overly complex to go through `typeid(T).hash_code()` and use a map to link those hash values to hard coded names.   Why not just add a string like `datasetName` to the `datasetState` and then hardcode the value of that string in the `datasetState` subclasses?

nritsche:
I think what you describe would be fine, too. I decided to do it like that because a static string was a bit of a pain. And because I couldn't find a way to enforce child classes to hardcode this variable, I think this is a nice way to reduce the amount of stuff a new state has to add in different places, since it is now needed as an argument to the factory registration macro.

nritsche:
Yeah, Richard proposed the same. But that string will end up in the CHIME database in a table of state types. That's why we really want it to be pretty.

andrerenard:
That makes sense.  I'll go ahead a merge this now.

PR #420, opened 2019-05-21
Track pulsar gating

nritsche:
Add all pulsar gating config to the gateSpec. This is for it to end up getting tracked by the dataset broker.


nritsche:
`coeff` is not saved as a gateSpec class member. Should it be tracked as well?
```
     * @param coeff      Polynomial coefficients in Tempo format.
```

tristpinsm:
that's what I was just wondering... It's not saved as a member because it is only needed to initialize the polyco objects and not afterwards. But this is also the case for a few of the other parameters. I'm not quite sure why I would have treated it differently.

In the spirit of tracking everything I would say we should keep it also, but that would require changing the structure of the class a bit.

jrs65:
Sorry, I didn't realise you'd already implemented this one. I don't really want this to go live until we've changed the model by which the dataset ids cause the writer to start new acquisitions. So let's not merge this into master.

andrerenard:
@jrs65 are we deferring this change again for this cycle?

andrerenard:
@jrs65 what's the status of this PR?  Is there a ticket with the details of what needs to be done to finish it, or is that already done and we can merge?

jrs65:
Yeah, we should get this merged.

andrerenard:
@nritsche could you resolve the conflicts in this?  @jrs65 can we merge as soon as Rick is done fixing the conflicts, or does this require any extra work?

jrs65:
I'd like to re-review it, as we rewrote a bunch of the dM and tracking stuff since then.

nritsche:
@jrs65 This is ready for your re-review

PR #421, opened 2019-05-22
Update ctl script

tristpinsm:
Some fixes to make the kotekan_ctl script more usable and installation work better. The timeouts may need to be tweaked, I wasn't sure how long would be needed.

### DIFF
--- a/python/kotekan/scripts/ctl.py
+++ b/python/kotekan/scripts/ctl.py
@@ -6,17 +6,17 @@ from builtins import (ascii, bytes, chr, dict, filter, hex, input,
                       str, super, zip)
 
 import requests
-from urlparse import urljoin, urlsplit
+from requests.compat import urljoin, urlsplit
 import click
 import yaml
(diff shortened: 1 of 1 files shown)


nritsche:
Great, thanks Tristan!
I'm not sure if the scripts directory needs an __init__ file

tristpinsm:
In python 3 (on my laptop) the script gets properly installed without the `__init__.py`, but when I tried it on a receiver node with python 2 it was unable to find the script module without that file...

nritsche:
Oh I see, ctl.py is actually not a script, but a module. I got confused by the directory name.

tristpinsm:
Yeah, it's a bit confusing how this `load_entry_points` thing works. I'll go ahead and merge then.

andrerenard:
@tristpinsm this looks good!  thanks for adding this.  just please remember to squash merge, and move the description to the squash commit.

PR #426, opened 2019-05-30
Move solar transit toggle into rfi_reciever service

andrerenard:
Also updates the rfi_reciever's configuration system and version tagging.


jrs65:
How serious a review do you want? This implementation could do with a lot of refinement!

Also, how often is this set to run in the crontab?

andrerenard:
This is how it's set to run: `10 11 * * * python /home/chime/kotekan/python/scripts/rfi_toggle_cron.py`

And I agree it could use refinement...  Feel free to review as seriously as you want.

jrs65:
Okay, the basic premise of how this one runs seems suspect. It runs half as cronjob, half as a background service, i.e. once a day (at 11:10) it is started up and then it sleeps until it should turn off the excision, then sleeps some more until it's time to turn it off. 

It seems that it should either: run all the time as a service, appropriately sleeping in time to turn excision on/off; or be entirely stateless, running at regular intervals (e.g. 15 mins) and turning off/on excision depending on the time of the day (probably it should check with KM for the current state before running).

Anyway, that seems like maybe slightly more work than we want to do right now, so I'll comment on the low hanging fruit.

python/scripts/rfi_toggle_cron.py
```diff
@@ -0,0 +1,57 @@
+import datetime
+import requests
+import json
+import time
+
+#Wait until the correct UTC time (deals with daylight savings time)
+while(datetime.datetime.utcnow().time() < datetime.time(19,10)):
```
jrs65: If only there was a way of calculating and sleeping for the exact number of seconds between now and 19:10....

python/scripts/rfi_toggle_cron.py
```diff
@@ -0,0 +1,57 @@
...
+    if(not r.ok):
+        print("RFI CRONJOB: Something went wrong in the request.")
+    else:
+        on = False
+        print("RFI CRONJOB: Request sent")
+except:
+    print("RFI CRONJOB: Failure to Contact Kotekan Master, is it running?")
+
+#If we successfully turned RFI Zeroing off
+if(not on):
```
jrs65: This logic seems wrong if the connection above failed, i.e. if we don't connect and hit the except clause, we then continue with `on = True` which means we never attempt to turn excision back on, even when we are at the correct time to do so.

python/scripts/rfi_toggle_cron.py
```diff
@@ -0,0 +1,57 @@
...
+    time.sleep(5)
+
+downtime = 60 #minutes
+
+#Endpoint parameters
+url = 'http://csBfs:54323/toggle-rfi-zeroing'
+headers = {'content-type': 'application/json', 'Accept-Charset': 'UTF-8'}
+
+#Payload
+payload = dict()
```
jrs65:
```suggestion
payload = {'rfi_zeroing': False}
```

james-s-willis:
I've update this script now

james-s-willis:
I just need a way of testing it

python/scripts/rfi_receiver.py
```diff
@@ -845,6 +845,7 @@ def rfi_zeroing():
                 print("RFI CRONJOB: Request sent")
         except:
             print("RFI CRONJOB: Failure to Contact Kotekan Master, is it running?")
```
nritsche: kotekan_master was replaced by comet

james-s-willis: Ah thanks, I'll change those messages.

andrerenard:
Annoying I cannot review this since I opened it, but there are still some things to fix.  The print statements all talk about it as if it's a `cron` job.  And the solar transit time is still just based on UTC.

andrerenard:
Sorry my page was showing old information.  Looks like the time stuff should be good, but the print statements would still be good to update.

python/scripts/rfi_receiver.py
```diff
@@ -787,6 +789,73 @@ def http_server2():
     httpd.serve_forever()
 
 
+def rfi_zeroing():
+    # Wait until the correct UTC time (deals with daylight savings time)
+    t_now = datetime.utcnow()
+    t_transit = ephemeris.solar_transit(t_now)[0]
+    t_diff = datetime.utcfromtimestamp(t_transit) - t_now
```
andrerenard: Shouldn't we be subtracking half the `downtime` from this number?  It looks to me like we stop zeroing exactly at solar transit, not 30 minutes before.

python/scripts/rfi_receiver.py
```diff
@@ -787,6 +789,73 @@ def http_server2():
...
+
+    # Turn Off
+    on = True
+    try:
+        r = requests.post(url, data=json.dumps(payload), headers=headers)
+        if not r.ok:
+            print("RFI CRONJOB: Something went wrong in the request.")
+        else:
+            on = False
+            print("RFI CRONJOB: Request sent")
```
andrerenard: Maybe `RFI CRONJOB` -> `RFI Solar Transit Toggle:`, or something like that.  This is no longer using `cron`

python/scripts/rfi_receiver.py
```diff
@@ -787,6 +789,73 @@ def http_server2():
     httpd.serve_forever()
 
 
+def rfi_zeroing():
+    # Wait until the correct UTC time (deals with daylight savings time)
```
jrs65: Doesn't this need a loop, otherwise it'll only run once?

python/scripts/rfi_receiver.py
```diff
@@ -787,6 +789,73 @@ def http_server2():
...
+    t_now = datetime.utcnow()
+    t_transit = ephemeris.solar_transit(t_now)[0]
```
andrerenard: How is this getting the site coordinates?

python/scripts/rfi_receiver.py
jrs65: Also, if I restart this script half way through the solar transit (after the transit itself, but before the end of the window), isn't it going to sleep until the next solar transit when it turns it off again.

It should presumably ensure that zeroing is off, and then wait until the end of transit window when it turns it back on.

Similarly if it starts up between transits, it should ensure that zeroing is turned on.

python/scripts/rfi_receiver.py
jrs65: The `ch_util.ephemeris` routines all know where CHIME is and use it in all calculations. As this is a fairly CHIME specific script I suggested that James just use that.

andrerenard: Ah good, thanks!

andrerenard: But I think we still need to subtract `downtime/2` minutes correct?

andrerenard: As well as fix the other issues you pointed out below.

jrs65: Yeah. Indeed, we need to subtract half of downtime. Good spot as I missed that one!

jrs65: I like this review tag teaming!!!

python/scripts/rfi_receiver.py
jrs65: I would recommend just using a Python logger object instead. Then you can give the logger a decent name, and it'll give that plus other context.

python/scripts/rfi_receiver.py
james-s-willis: How do I check RFI zeroing is turned on/off? Can I do something like: `r = requests.get(url, data=json.dumps(payload), headers=headers)`?

james-s-willis: Also, if the script starts up and zeroing is off am I to assume that it is going through a solar transit window?

jrs65: It's pretty cheap to send the update, and this script shouldn't restart often, so I would just *always* set the state to what it should be when things start up. i.e. if you start during the window the first thing you do is send the command to turn zeroing off, and then wait until the end of the period.

james-s-willis: Ah okay thanks, I will do that then.

james-s-willis: I've added that edge case now.

andrerenard:
Now that we've merged all the version stuff, can you update this to use those version scripts?

james-s-willis:
I've updated it now, so it uses the common `_version.py`.

python/scripts/rfi_receiver/rfi_receiver.py
```diff
@@ -162,14 +167,7 @@ def __init__(self):
...
             )
-            # TODO: this was disabled because it ignores the structure of the kotekan
-            # config, looks in the whole kotekan config tree for the keys from the
-            # config hardcoded in this file. It ignores that the kotekan config has
-            # variable names reused in different parts of the config and just applies
-            # all of them, resulting in wrong values, i.e. for samples_per_data_set.
-            # To solve this issue, only the hardcoded config is used, but it should
-            # become a config file or fin d a better way to read the kotekan config.
-            # parse_dict(self, yaml.safe_load(open(argument.config)))
+            parse_dict(self, yaml.safe_load(open(argument.config)))
```
andrerenard: Since we are pushing the conf options to the file, I think we should remove them from the source file, see line 77

andrerenard:
@james-s-willis Can you look at the one comment above, and @jrs65 can you give this another quick check?  Then I think we can merge.

python/scripts/rfi_receiver/rfi_receiver.py
james-s-willis: The `config` struct is needed as part of the `CommandLine` class

andrerenard:
Actually I think we should add the service file: `carillon:/etc/systemd/system/kotekan_rfi_receiver.service` to the repo, and ideally make it a part of the setup.py install process.

andrerenard:
@james-s-willis could you please add the service file stuff above, and @jrs65 could you give this a final review?

andrerenard:
Also pull in `develop` so that the unit tests pass.

james-s-willis:
@andrerenard I've added the service file and merged `develop` in now. Did you want the `setup.py` to start the service?

andrerenard:
@jrs65 or @nritsche can you please review this change?  Since I opened it, and then @james-s-willis took over, I cannot approve it myself.

python/scripts/rfi_receiver/rfi_receiver.py
```diff
@@ -787,6 +790,121 @@ def http_server2():
...
+
+            time.sleep(sleep_time_s)
+
+            logger.info(
+                "RFI Solar Transit Toggle: Waking up to disable RFI zeroing during solar transit"
+            )
+
+        # Create payload
+        payload = {"rfi_zeroing": False}
+        payload["rfi_zeroing"] = False
```
jrs65, line 850: Isn't this line completely redundant with the one above?
```suggestion
```

python/scripts/rfi_receiver/rfi_receiver.py
```diff
@@ -787,6 +790,121 @@ def http_server2():
...
+            # Wait until sun has passed
+            time.sleep(downtime_s)
+
+            logger.info(
+                "RFI Solar Transit Toggle: Solar transit ended. Waking up to turn RFI zeroing back on."
+            )
+
+            # Payload
+            payload = {"rfi_zeroing": True}
+            payload["rfi_zeroing"] = True
```
jrs65, line 887: Same here
```suggestion
```

PR #431, opened 2019-06-10
FRB updates for July deployment

cherryng:
1. added the 16-bandpass correction to the CPU pipeline for FRB verification. use `chime_frb_verification.yaml`
2. Zeroing dropped packets in FRB postprocess
3. change speed of light constant from 3e8 to 299792458.
4. PSR pointings as updatable config with 10 individual endpoints
5. renamed to follow updatable_config convention


lib/testing/gpuBeamformSimulate.cpp
```diff
@@ -449,6 +449,9 @@ void gpuBeamformSimulate::main_thread() {
             }
         }
 
+        //16-bandpass correction
+        float BP[16] {0.52225748,0.58330915,0.6868705,0.80121821,0.89386546,0.95477358,0.98662733,0.99942558,0.99988676,0.98905127,0.95874124,0.90094667,0.81113021,0.6999944,0.59367968,0.52614263};
```
andrerenard: This likely won't pass the clang-format line length rule (100 characters).   Although looks like the CI server isn't working for some reason...  I will look into that.

andrerenard:
Please address the clang format issues, and then we can merge this in.

config/chime_psr_verification.yaml
```diff
@@ -127,9 +178,7 @@ gpu:
     kotekan_stage: hsaProcess
     gpu_id: 0
     updatable_gain_psr: /pulsar_gain
-    source_ra: [300.022454326,  299.772454326, 300.272454326 , 300.022454326, 300.022454326, 299.771454326 , 299.771454326 , 300.273454326,  300.273454326, 300.022454326]
-    source_dec: [40.7839418602, 40.7839418602, 40.7839418602, 41.0339418602, 40.5339418602, 40.5329418602 , 41.0349418602 , 40.5329418602 , 41.0349418602 , 40.7839418602]
-    psr_scaling: [48,48,48,48,48,48,48,48,48,48]
+    updatable_psr_pt: /pulsar_pointing
```
nritsche: I think we once decided to only use a fixed key here: `updatable_config` (https://github.com/kotekan/kotekan/blob/556a62159b0a303f652febb1f851751f64b573b9/lib/core/configUpdater.hpp#L32), but I leave it to @andrerenard if we should apply that strictly.

andrerenard: This is a bit of an odd case because we have a bit of an internal framework with-in this stage, with its own subcomponents.   However in the interest of consistency, yes, let's change that to: 
```
updatable_config:
   gain_psr: /pulsar_gain
   psr_pt: /pulsar_pointing
```
And then update the `hsaCommand` objects that use those.

andrerenard:
Two quick comments:  Could you please update the title and description of this PR to reflect the extra items it addresses?  And also it looks like the clang-format check is still failing.  It might be easiest just to run the `make clang-format` command on your code and push the changes.  Just note I would only accept changes in makes to code you've also changed in this PR, there are some annoying differences between version with clang-format that can sometimes result in other code that passes on the CI server already getting changed locally when you run it.

lib/processes/frbPostProcess.cpp
```diff
@@ -205,24 +222,37 @@ void frbPostProcess::main_thread() {
...
-                        __m256 _ofs = _mm256_broadcast_ss(&off);
                         int f_per_m = sizeof(__m256) / sizeof(float);
                         unsigned char utr[256], tr[256];
                         for (int t = 0; t < _timesamples_per_frb_packet; t++) {
+                            scl = (253.) / (max - min);
+                            off = -ofs * scl;
+                            // Loop through 384 time to check for drop packet, if so, set scl and
+                            // off to be zero
+                            for (int tz = 0; tz < _downsample_time * _factor_upchan; tz++) {
+                                if (lost_samples_frame[T * _factor_upchan * _downsample_time + tz]
```
andrerenard: Precompute large T lost samples array to make this more efficient.

lib/processes/frbPostProcess.cpp
```diff
@@ -205,24 +238,32 @@ void frbPostProcess::main_thread() {
                         frb_header_offset[b * _num_gpus + thread_id] = ofs;
```
andrerenard: In the two lines above it looks like `scl` could be zero in the case all packets are lost, if that happened there will be a divide by zero error.  Also in that case I think the header should just have `scale` and `offset` set to `0,0` or `1,0` if that makes more sense to L1.  Since in this case those values are otherwise undefined.

lib/processes/frbPostProcess.cpp
```diff
@@ -205,24 +238,32 @@ void frbPostProcess::main_thread() {
...
                         // Apply scale and offset
                         float off = -ofs * scl;
```
andrerenard: This line can be removed.  `off` always get reset below before use.

lib/processes/frbPostProcess.hpp
```diff
@@ -112,6 +121,9 @@ class frbPostProcess : public kotekan::Stage {
     int32_t udp_packet_size;
     int32_t udp_header_size;
     int16_t fpga_counts_per_sample;
+
+    // flag for dropped packet
+    int* droppacket;
```
andrerenard:
```suggestion
    uint8_t* droppacket;
```

lib/processes/frbPostProcess.cpp
```diff
@@ -50,12 +49,18 @@ frbPostProcess::frbPostProcess(Config& config_, const string& unique_name,
...
+    register_consumer(lost_samples_buf, unique_name.c_str());
+    lost_samples_buf_id = 0;
+
     // Dynamic header
     frb_header_beam_ids = new uint16_t[_nbeams];
     frb_header_coarse_freq_ids = new uint16_t[_num_gpus];
     frb_header_scale = new float[_nbeams * _num_gpus];
     frb_header_offset = new float[_nbeams * _num_gpus];
 
+    droppacket = (int*)calloc(num_samples, sizeof(int));
```
andrerenard:
```suggestion
    droppacket = (int*)calloc(num_samples, sizeof(uint8_t));
```

andrerenard: I don't think you need this to be a full int, since it's just 0 or 1.

config/chime_science_run_gpu.yaml
```diff
@@ -91,6 +91,58 @@ pulsar_gain:
     - /mnt/gong/cherry/FRB-GainFiles/broker_CygA_10_26_scl/
     - /mnt/gong/cherry/FRB-GainFiles/broker_CygA_10_26_scl/
 
+pulsar_pointing:
```
nritsche: For generating endpoints automatically from the kotekan config in the km-replacement (https://github.com/chime-experiment/coco/issues/31), the I propose that we stick to a common place for the updatable config blocks (like [here](https://github.com/kotekan/kotekan/blob/556a62159b0a303f652febb1f851751f64b573b9/config/chime_science_run_recv.yaml#L103)). That way, the path in the config is the same as the endpoint name. That way, coco can get the endpoint name from the kotekan config.

nritsche:
Is the `/update-east-west-beam` endpoint going to change as well?

cherryng:
@nritsche: No, `/update-east-west-beam` and `/update-north-south-beam`  will not be updatable config. we don't intend to change them on the fly.

andrerenard:
@cherryng do we ever intend to change them in the middle of a run?  If so should we just remove those endpoints entirely? and tell @nritsche that he doesn't need to add them to `coco`.  Otherwise I think if we intend to still have the option to update them in the middle of the run, then they should be updatable config, so we can have auto registration.

config/chime_science_run_gpu.yaml
andrerenard: @nritsche would it not work to just generate the endpoint based on the paths to `kotekan_update_endpoint` keys in the config?   Currently that's how kotekan generates the endpoint paths.   Having said that we did say for standardization reasons we'd put all the updatable config options in an `updatable_config` block.  So I think it's still worth doing that for consistency - but I'm not sure we want that as a hard requirement.

nritsche:
> @cherryng do we ever intend to change them in the middle of a run? If so should we just remove those endpoints entirely? and tell @nritsche that he doesn't need to add them to `coco`. Otherwise I think if we intend to still have the option to update them in the middle of the run, then they should be updatable config, so we can have auto registration.

I added them for now.

config/chime_science_run_gpu.yaml
nritsche: Oh yeah you are right. So if this value is always called `kotekan_update_endpoint`, I can use it (this is also not enforced).

andrerenard:
Can you update this branch with the latest develop to resolve the conflicts?

andrerenard:
This has been addressed with other pull requests already #438 #467

PR #435, opened 2019-06-24
Compilation for ROCM >=2.4

james-s-willis:
Added a CMAKE option to build kotekan with ROCM 2.3 or older. Added flag -mno-code-object-v3 to allow kernel compilation and updated symbol naming convention for ROCM 2.4 or newer.


lib/hsa/hsaPresumKernel.cpp
```diff
@@ -1,5 +1,13 @@
 #include "hsaPresumKernel.hpp"
 
+// Use old symbol naming convention if 
+// compiled with ROCM version 2.3 or older
+#ifdef USE_OLD_ROCM
```
andrerenard: This will need to be done for every file.  Also it should be possible to simplify this to just add `.kd` or ` ` below and include it in the `hsaCommand.hpp` file.

CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ option(USE_OPENCL "Build OpenCL GPU Framework" OFF)
 option(USE_DPDK "Enable DPDK Framework" OFF)
 option(USE_HDF5 "Build HDF5 output stages" OFF)
 option(USE_OMP "Enable OpenMP" OFF)
+option(USE_OLD_ROCM "Build for ROCm versions 2.3 or older" OFF)
```
andrerenard: Please add this new option to the README.md and the build documentation next to the current HSA options.

docs/sphinx/compiling/general.rst
```diff
@@ -85,6 +85,8 @@ Cmake build options
     Includes DPDK support.
 * ``-DUSE_HSA=ON``
     Build with HSA support if available. On by default.
```
andrerenard:
```suggestion
    Build with HSA support if available. Off by default.
```

Issue #353, opened 2018-12-16
prometheusMetric timestamp issues


jrs65:
Prometheus drops metrics where the time samples are much older (or newer) than what it thinks the current time is. This causes metrics that are updated irregularly to dissappear within time series.

I propose that we either: 
- Move to the behaviour that prometheus expects and the timestamp returned on a `/metrics` scrape is the current timestamp.
- We move to a policy where each metric is updated every cycle by the relevant process, but the counts are only increased when they change, i.e. dropped frame counts are posted to prometheusMetrics every cycle, but the counter is only incremented when a frame is dropped.

The former seems like a better option to me.

cubranic:
I *think* we don't care about timestamps anywhere, in which case there is no reason *not* to go with option 1.

If we do care about the timestamps, a variation for the second option is to distinguish between "counter" and "gauge" type metrics. (Prometheus terminology). 

Counters can only by increased (this should be controlled by an API), but don't need to be refreshed by the stages every cycle, since they remain "current" between updates. For them, we always return the "now" timestamp on Prometheus scrape. (Or, more simply, don't include a timestamp and Prometheus server doing the scrape fills it in.)

Gauges are explicitly _set_ to a value that is current at that time, so for them we do keep the timestamp of the last update and return it on Prometheus scrape.

jrs65:
@cubranic @andrerenard do you think this would be a lot of work to fix? Could be a good hot fix candidate actually, as it messes up some of the metric analysis I want to do for dropped frames.

jrs65:
Also, I guess there are a few things that we output which probably are gauges. As Davor says we might want to distinguish between them. To some degree that's what's causing the problem I'm seeing in the first place as my "counters" are getting updated fairly infrequently, and we're using a "gauge" like behaviour.

cubranic:
@jrs65 I started on this -- have a look in branch [_prometheus-metrics-gauge-counter_](https://github.com/kotekan/kotekan/tree/prometheus-metrics-gauge-counter), in particular the latest commit, 475b0ec6b6640fd1e9ac3bc783d1c94b8acd7b74. It gives you the API similar to other Prometheus client libraries, where you create the metric in a constructor or at the beginning of your Stage's main loop:
```c++
request_count_metric(
	        prometheusMetrics::instance()
	        .add_stage_counter("kotekan_baseband_requests_total", "baseband")
	        )
```
and then increment it when necessary with: `request_count_metric.inc()`.

[...]

PR #436, opened 2019-06-27
Rework Prometheus metrics API - counters vs gauges

cubranic:
As discussed in #353

### DIFF
--- a/lib/core/prometheusMetrics.hpp
+++ b/lib/core/prometheusMetrics.hpp
@@ -54,7 +54,7 @@ public:
 
 private:
     /// The actual value to be returned
-    int value = 0;
+    uint64_t value = 0;
 };
 
 /**
--- a/lib/core/prometheusMetrics.hpp
+++ b/lib/core/prometheusMetrics.hpp
@@ -69,8 +69,14 @@ public:
     std::ostringstream& to_string(std::ostringstream& out) override;
 
 private:
+    /// Internal function to get the time in
+    static uint64_t get_time_in_milliseconds();
+
     /// The actual value to be returned
     double value = 0;
+
+    /// Time stamp in milliseconds.
+    uint64_t last_update_time_stamp;
 };
 
 /**
--- a/lib/core/prometheusMetrics.hpp
+++ b/lib/core/prometheusMetrics.hpp
@@ -18,41 +18,12 @@ namespace kotekan {
 namespace prometheus {
 
 /**
- * @class metric
- * @brief An internal base class for storing metric values
- */
-class metric {
-public:
-    /// @brief Returns the stored value as a string.
-    virtual string to_string() = 0;
-    virtual ~metric() = default;
-
-protected:
-    /// The actual value to be returned
-    double value = 0;
-
-    /// Internal function to get the time in
-    static uint64_t get_time_in_milliseconds();
-};
-
-class gauge : public metric {
-public:
-    /// @brief Sets the current metric value
-    void set(double value);
-    string to_string() override;
-
-private:
-    /// Time stamp in milliseconds.
-    uint64_t last_update_time_stamp;
-};
-
-/**
  * @class Metric
- * @brief An internal base class for storing metric values
+ * @brief An internal base class for storing metric value for a given combination of label values
  */
 class Metric {
 public:
-    Metric(const std::vector<string>&);
+    Metric(const std::vector<string>& label_values);
     virtual ~Metric() = default;
 
     /// @brief Returns the stored value as a string.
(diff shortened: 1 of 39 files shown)


lib/core/prometheusMetrics.hpp
```diff
@@ -16,6 +16,116 @@ using std::string;
 
 namespace kotekan {
```
andrerenard: I wonder if this should have it's own namespace.  Just because `kotekan::Family` and `kotekan::Metric` seems overly generic.   I would either subname space this, or add some indication this is for Prometheus to all the names.

lib/core/prometheusMetrics.hpp
```diff
@@ -16,6 +16,116 @@ using std::string;
...
+    /// @brief Returns the stored value as a string.
+    virtual string to_string() = 0;
+    virtual std::ostringstream& to_string(std::ostringstream& out) = 0;
+    const std::vector<string> label_values;
+};
+
+class Counter : public Metric {
+public:
+    Counter(const std::vector<string>&);
+    void inc();
```
andrerenard: Tempted to operator overload `++` here.

cubranic:
I consciously modelled the new API on other Prometheus clients, where declaring and using the metric is done in separate spots. This also catches duplicate declarations, and throws an error. But we could have a shim that follows the current API, where there is single function that both looks-up/declares a metric and set its current value. 

Davor

andrerenard:
I see the point about catching duplicate declarations, we have had that happen in the past.  So let's not bother with the extra function to lookup/declare for now.   However if possible it would be good to address the other comments before we merge, mostly around naming.  And if possible a fix to avoid the need for `std::to_string()` would be nice, but not critical for this release.

cubranic:
I'll work on this today -- it should be pretty mechanical and quick to do.

lib/core/prometheusMetrics.hpp
cubranic: I'm inclined to go with the subname: either `kotekan::prometheus::` or `kotekan::metrics`. It would probably make sense to move the `prometheusMetrics` class into it as well, no?

andrerenard: I would go with `kotekan::prometheus::`, the word "metrics" might well apply to other systems of tracking stats.

cubranic: Sounds good. And now `kotekan::prometheusMetrics` becomes `kotekan::prometheus::Metrics`.

lib/core/prometheusMetrics.hpp
```diff
@@ -16,6 +16,116 @@ using std::string;
 
 namespace kotekan {
 
+/**
+ * @class metric
+ * @brief An internal base class for storing metric values
+ */
+class metric {
```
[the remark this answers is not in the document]
cubranic: The lower case ones were used in the old API. I removed them now.

lib/core/prometheusMetrics.hpp
cubranic: I'm leaving it alone for now, since `.inc()` is used in other Prometheus clients. Let's discuss this and other ways to improve the API from developers' point after this cycle.

lib/dpdk/iceBoardShuffle.hpp
```diff
@@ -169,7 +169,37 @@ class iceBoardShuffle : public iceBoardHandler {
...
+                               unique_name, {"port"})),
+    second_missing_short_error(kotekan::prometheusMetrics::instance()
+                               .AddGauge("kotekan_dpdk_shuffle_fpga_second_stage_missing_short_errors_total",
+                                         unique_name, {"port"})),
+    second_long_error(kotekan::prometheusMetrics::instance()
+                      .AddGauge("kotekan_dpdk_shuffle_fpga_second_stage_long_errors_total",
+                                unique_name), {"port"}),
+    second_fifo_overflow_error(kotekan::prometheusMetrics::instance()
+                               .AddGauge("kotekan_dpdk_shuffle_fpga_second_stage_fifo_overflow_errors_total",
+                                         unique_name, {"port"})) {
```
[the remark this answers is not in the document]
cubranic: Can you clarify the suggested change here? `kotekan_dpdk_shuffle_fpga_second_stage_fifo_overflow_errors_total` only uses the "port" label. Only `...second_stage_shuffle_errors_total` and `...third_stage_shuffle_errors_total` also add the `fpga_lane` label.

cubranic:
Well, it wasn't that mechanical, but I've finalized the API renaming, and hopefully addressed all of your points.

One thing I just realized is that we should probably change the type of the "value" of the counter from `int` to `uint64_t` since they can't be negative (start at 0 and count up), or maybe `double`, since that's the underlying storage type in Prometheus.

andrerenard:
Just need to fix the formatting issue (see `#kotekan_dev` on slack), but otherwise this looks good to me.

PR #438, opened 2019-07-05
Frb july2019 part1

cherryng:
A few changes to be deployed during first week of July:
- updated speed of light
- zeroing dropped packet in FRB postprocess and correctly normalize the scaling
- FRB cpu pipeline missing bandpass correction
- changed FRB & PSR gain paths in config file to move away from `gong` 

updated two relevant config files:
- chime_science_run_gpu.yaml
- chime_frb_verification.yaml


lib/processes/frbPostProcess.cpp
```diff
@@ -118,6 +124,18 @@ void frbPostProcess::main_thread() {
                 return;
         }
 
+        // Information on drop packets
+        uint8_t* lost_samples_frame = lost_samples_buf->frames[lost_samples_buf_id];
+        for (uint t = 0; t < num_samples; t++) {
+            // check if drop packet by reading 384 original times, if so flag that t
```
andrerenard: This array never gets reset, we also need to make sure when there is no loss that `t` gets set to `0`
```suggestion
            // check if drop packet by reading 384 original times, if so flag that t
            droppacket[t] = 0;
```

PR #439, opened 2019-07-05
Clean up metrics created by a stage after the stage exits

cubranic:
- Removes stage metrics after stage exit in `kotekanMode`

### DIFF
--- a/lib/core/basebandApiManager.cpp
+++ b/lib/core/basebandApiManager.cpp
@@ -61,10 +61,6 @@ basebandApiManager::basebandApiManager() :
         ) // clang-format on
 {}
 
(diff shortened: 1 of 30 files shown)


cubranic:
That would work too, and is less error-prone, I agree. But I also saw some stages clean up their rest callbacks in the destructor?

andrerenard:
Yes the rest callbacks never had a system that clearly associated them with stages so it was harder to add this kind of framework level cleanup, but might go back when we have time and do the same with the rest callbacks - because we've had a number of errors associated with forgetting to remove them in the past.

cubranic:
@andrerenard as requested: move stage metric removal from individual destructors to `~kotekanMode`.

PR #442, opened 2019-07-08
Extract 21-cm hydrogen line data beams

james-s-willis:
Extracts 21-cm hydrogen line data (Hyper Fine Beam - HFB) from FRB pipeline:
-------------------------------------------------------------------
- Taps FRB pipeline in the upchannelization stage before downsampling occurs
- Both polarizations are summed over all time sampes for each frequency
- 1024 (beams) x 128 (freq) x 10 (samples) data is output from the `upchannelize_flip.cl` kernel
- `sum_hfb.cl` kernel sums across samples
- `integrateHFBData.cpp` integrates `num_frames_to_integrate` frames of data together
- `gpuBeamformSimulate.cpp` simulates the output of `sum_hfb.cl` on the CPU and cross checks with the GPU data
- `chime_frb_verification.yaml` has been updated to include a HFB test

Other changes:
--------------------
- Removed pre-compiled kernels that are deprecated


andrerenard:
Also it might be worth pulling the renaming of `process` to `stage` to another PR that we can merge in on a shorter time scale.   That will also prevent problems like the merge conflict showing up above.

james-s-willis:
I just ran `chime_frb_verification.yaml` on `csDg8` and compared the performance with the `develop` branch:

`develop`
-----------
```
| -------- GPU[0] Kernel timing --------
| Kernel name        |     time | utilization   |
|--------------------+----------+---------------|
| reorder            | 5.3e-05  | 0.5404%       |
| zero_padded_FFT512 | 0.000841 | 8.5513%       |
| transpose          | 0.000381 | 3.8785%       |
| upchannelize       | 0.000425 | 4.3261%       |
| Total:             | 0.0017   | 17.2962%      |
| -------- Host->GPU DMA timing --------
| Copy in name      |     time | utilization   |
|-------------------+----------+---------------|
| hsaInputData      | 0.000741 | 7.5389%       |
| hsaOutputDataZero | 0.001598 | 16.2595%      |
| Total:            | 0.002339 | 23.7984%      |
| -------- GPU->Host DMA timing --------
| Copy out (GPU->host)   |    time | utilization   |
|------------------------+---------+---------------|
| hsaBeamformOutputData  | 5.7e-05 | 0.5778%       |
| Total:                 | 5.7e-05 | 0.5778%       |
```

`frb-tap`
----------
```
| -------- GPU[0] Kernel timing --------
| Kernel name        |     time | utilization   |
|--------------------+----------+---------------|
| reorder            | 5.2e-05  | 0.5257%       |
| zero_padded_FFT512 | 0.000839 | 8.5317%       |
| transpose          | 0.000379 | 3.8508%       |
| upchannelize       | 0.000448 | 4.5621%       |
| sum_hfb            | 7e-05    | 0.7113%       |
| Total:             | 0.001787 | 18.1816%      |
| -------- Host->GPU DMA timing --------
| Copy in name      |     time | utilization   |
|-------------------+----------+---------------|
| hsaInputData      | 0.000716 | 7.2834%       |
| hsaOutputDataZero | 0.001572 | 15.9877%      |
| Total:            | 0.002288 | 23.2710%      |
| -------- GPU->Host DMA timing --------
| Copy out (GPU->host)     |    time | utilization   |
|--------------------------+---------+---------------|
| hsaBeamformOutputData    | 4.9e-05 | 0.4997%       |
| hsaBeamformHFBOutputData | 3.9e-05 | 0.3955%       |
| Total:                   | 8.8e-05 | 0.8952%       |
```

lib/hsa/kernels/sum_hfb_wrapper.c
```diff
@@ -0,0 +1,765 @@
+//  gcc -std=gnu99 -O3 -Wall sum_hfb_wrapper.c -I$AMDAPPSDKROOT/include -lOpenCL -lm -L$AMDAPPSDKROOT/lib/x86_64/ -o sum_hfb_wrapper
```
andrerenard: I'm not sure it makes sense to commit this wrapper here...  Can we drop this from the PR for now?

lib/hsa/kernels/upchannelization_test.c
```diff
@@ -0,0 +1,812 @@
+//  gcc -std=gnu99 -O3 -Wall fourierTesting_128_upchannelization.c -I$AMDAPPSDKROOT/include -lOpenCL -lm -L$AMDAPPSDKROOT/lib/x86_64/ -o fourierTest128upchan
```
andrerenard: Same for this one, remove from this PR for now.

config/chime_science_run_gpu_21cm.yaml
```diff
@@ -0,0 +1,3596 @@
...
+#
+# Config to support the CHIME project with 2048 elements, and 1024 frequencies.
+#
+# This config contains the following main components:
+#
+# - Packet capture from the ICE boards including FPGA flag handling and normalization
+# - N^2 kernels and output for various data projects to receiver systems
+# - FRB FFT Fan beam and packet Tx
+# - Pulsar 10 coherent beam system abd packet Tx
+# - RFI detection and summary stats system (only reports on RFI, no normalization yet)
```
andrerenard: We should just merge the changes for the absorber config directly into the main `science_run` and `no_output` configs at this point.

andrerenard: Also add a comment in here about the absorber system

config/chime_science_run_recv_21cm.yaml
```diff
@@ -0,0 +1,63 @@
+##########################################
+#
+# chime_science_run_recv_21cm.yaml
+#
+# CHIME 21cm absorber receiver node configuration used in the December 2019 run.
```
andrerenard: I would remove the date reference, and include some details about how it writes out the files (i.e. raw output).

config/chime_science_run_recv_21cm.yaml
```diff
@@ -0,0 +1,63 @@
...
+udp_packet_size: 4928
+num_data_sets: 1
+samples_per_data_set: 32768
+buffer_depth: 64
+baseband_buffer_depth: 282 # 282 = ~34 seconds after accounting for active frames
+num_gpu_frames: 128
+block_size: 32
+cpu_affinity: [1,6,7,9,14,15]
+num_frb_total_beams: 1024
+num_sub_freqs: 128
```
andrerenard: It's not obvious to me that all of these values are needed.  Can we remove the ones that aren't?

config/integrateHFBData_test.yaml
```diff
@@ -0,0 +1,93 @@
+##########################################
+#
+# Test integrateHFBData stage
```
andrerenard: Should this file be in the `config/tests/` folder?

lib/hsa/hsaBeamformHFBOutput.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+hsaBeamformHFBOutputData::~hsaBeamformHFBOutputData() {}
+
+int hsaBeamformHFBOutputData::wait_on_precondition(int gpu_frame_id) {
+    (void)gpu_frame_id;
+    uint8_t* frame =
+        wait_for_empty_frame(output_buffer, unique_name.c_str(), output_buffer_precondition_id);
+    if (frame == NULL)
+        return -1;
+    output_buffer_precondition_id = (output_buffer_precondition_id + 1) % output_buffer->num_frames;
```
andrerenard: You should also do a `wait_for_full_frame()` on the network buffer here.   It's not strictly needed, but it helps avoid potential bugs if this is used in different configs, and improves the internal statistics.

lib/hsa/hsaBeamformHFBOutput.cpp
```diff
@@ -0,0 +1,61 @@
...
+    return 0;
+}
+
+hsa_signal_t hsaBeamformHFBOutputData::execute(int gpu_frame_id, hsa_signal_t precede_signal) {
+
+    void* gpu_output_ptr =
+        device.get_gpu_memory_array("hfb_sum_output", gpu_frame_id, output_buffer->frame_size);
+
```
andrerenard: Minor style point, no need for a space here.  These are both variable declarations.

lib/hsa/hsaBeamformHFBOutput.hpp
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * @class hsaBeamformHFBOutput
+ * @brief hsaCommand for copying HFB output gpu to host.
+ *
+ * This is an hsaCommand that async copy HFB output buffer from GPU
+ * to CPU. It marks the HFB output buffer to be full when done so that
+ * it can be reused.
+ *
+ * @par GPU Memory
+ * @gpu_mem  hfb_output      Output from the FRB pipeline, size 1024x128x1
```
andrerenard: Where are the size numbers coming from?  i.e. maybe add something like beams*freq*....   Also shouldn't it be x4 at the end since they are floats?

lib/hsa/hsaBeamformHFBOutput.hpp
```diff
@@ -0,0 +1,64 @@
...
+    int wait_on_precondition(int gpu_frame_id) override;
+
+    /// Async copy output form gpu to host
+    hsa_signal_t execute(int gpu_frame_id, hsa_signal_t precede_signal) override;
+
+    /// Marks output full when done and passes metadata
+    void finalize_frame(int frame_id) override;
+
+private:
+    /// The very first input data from dpdk
```
andrerenard:
```suggestion
    /// The data from the dpdk network thread (used for metadata)
```

lib/hsa/hsaBeamformHFBSum.hpp
```diff
@@ -0,0 +1,68 @@
...
+ * @brief hsaCommand to upchannelize and downsample FRB data
+ *
+ * This is an hsaCommand that launches the kernel (sum_hfb) for
+ * summing HFB data across all samples. Final output in float of
+ * 1024 beams x 128 freq.
+ *
+ * @requires_kernel   sum_hfb.hasco
+ *
+ * @par GPU Memory
+ * @gpu_mem  hfb_output  Input data of size input_frame_len
```
andrerenard: This is missing the `compressed_lost_samples` memory block.  Also it might be worth adding a note above in the summary about how it excludes samples from the sum where there is 1 or more lots samples.

lib/hsa/hsaBeamformHFBZeroData.cpp
```diff
@@ -0,0 +1,41 @@
...
+                                               hsaDeviceInterface& device) :
+    hsaSubframeCommand(config, unique_name, host_buffers, device, "hsaBeamformHFBZeroData", "") {
+    command_type = gpuCommandType::COPY_IN;
+
+    int block_size = config.get<int>(unique_name, "block_size");
+    int num_elements = config.get<int>(unique_name, "num_elements");
+    _num_blocks = (int32_t)(num_elements / block_size) * (num_elements / block_size + 1) / 2.;
+    output_len = 1024
+                 * 128 // No. of frequencies
+                 //* (_samples_per_data_set / _downsample_time)
```
andrerenard: Remove this line?  Also shouldn't we be getting these values from the config? The 1024 and 128?

lib/hsa/hsaBeamformHFBZeroData.hpp
```diff
@@ -0,0 +1,23 @@
+#ifndef HSA_HFB_ZERO_DATA_H
+#define HSA_HFB_ZERO_DATA_H
+
+#include "hsaSubframeCommand.hpp"
+
+class hsaBeamformHFBZeroData : public hsaSubframeCommand {
```
andrerenard: This file is missing comments.  Although I'm wondering is it possible to avoid having to zero the output by adjusting the kernel in some way?

lib/hsa/hsaBeamformTranspose.cpp
```diff
@@ -8,6 +8,7 @@ REGISTER_HSA_COMMAND(hsaBeamformTranspose);
 hsaBeamformTranspose::hsaBeamformTranspose(Config& config, const string& unique_name,
                                            bufferContainer& host_buffers,
                                            hsaDeviceInterface& device) :
+
```
andrerenard:
```suggestion
```
Remove added line.

lib/hsa/hsaBeamformUpchan21cm.hpp
```diff
@@ -0,0 +1,94 @@
...
+ *     @gpu_mem_format          Array of @c float
+ *
+ * @conf   num_elements         Int (default 2048). Number of elements
+ * @conf   samples_per_data_set Int (default 49152). Number of time samples in a data set
+ * @conf   downsample_time      Int (default 3). Downsample factor in time
+ * @conf   downsample_freq      Int (default 8). Downsample factor in freq
+ * @conf   num_frb_total_beams  Int (default 1024). Number of total FRB formed beams
+ * @conf   num_sub_freqs        Int (default 128). Number of sub frequencies
+ *
+ * @todo   Check that the 16 freq axis has the correct orientation
```
andrerenard: I really hope this is actually correct at this point :)   Should probably remove the comment.

lib/stages/rawFileWrite.hpp
```diff
@@ -37,6 +37,7 @@ class rawFileWrite : public kotekan::Stage {
     std::string base_dir;
     std::string file_name;
     std::string file_ext;
+    uint32_t _num_frames_per_file;
```
andrerenard: This needs to be documented (here and in the list of options for this stage).   Also if we are going to add this functionality, we should also add the ability to read these files back with rawFileRead.  Ideally `rawFileRead` shouldn't need to know how many frames there with another `_num_frames_per_file` but be able to figure that out based on how long the file is.

lib/testing/testDataCheck.hpp
```diff
@@ -102,7 +102,7 @@ void testDataCheck<A_Type>::main_thread() {
             } else { // N2 numbers are int
                 // INFO("Checking non float numbers-----------");
                 if (first_value != second_value) {
-                    if (num_errors++ < 10000)
+                    if (num_errors++ < 20)
```
andrerenard: We've changed this value enough times that it should be a config option with a default of something like `100`

lib/testing/testDataGenMissingFrames.hpp
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * This is CPU only pipeline to verify the FRB beamformer.
+ * This CPU pipeline does the equivalent of:
+ *   - hsaBeamformReorder
+ *   - name: hsaBeamformKernel
+ *   - name: hsaBeamformTranspose
+ *   - name: hsaBeamformUpchan
+ *
+ * @author Cherry Ng
+ **/
```
andrerenard: The comments here don't match what this stage actually does...  Also I think Richard and/or Davor worked on a similar stage for testing their systems with dropped frames.   Can you check if there is any duplication going on here?    Finally the name of this `stage` also doesn't really match what it's doing - it's dropping frames based on a pattern.

lib/testing/testDataGenMissingFrames.cpp
```diff
@@ -0,0 +1,67 @@
...
+            break;
+
+        if (std::find(_missing_frames.begin(), _missing_frames.end(), num_frames)
+            == _missing_frames.end()) {
+            float* output =
+                (float*)wait_for_empty_frame(output_buf, unique_name.c_str(), output_buf_id);
+            if (output == NULL)
+                break;
+
+            memcpy(output, input, output_buf->frame_size);
```
andrerenard: You could likely use `swap_frame()` to make this zero copy.   Just check for the edge cases with that function call.

lib/stages/compressLostSamples.cpp
```diff
@@ -0,0 +1,85 @@
...
+                    compressed_lost_samples_frame[sample / (3 * _num_sub_freqs)] = 1;
+                    total_lost_samples += 3 * _num_sub_freqs;
+                    break;
+                }
+            }
+        }
+
+        // Create new metadata
+        allocate_new_metadata_object(out_buf, out_buffer_ID);
+        zero_lost_samples(out_buf, out_buffer_ID);
```
andrerenard: We should actually do a `copy_metadata()` from the `in_buf` here and then zero the lost samples and set it with this new value.   That way we also carry along all the stream/seq number information.

lib/stages/compressLostSamples.cpp
```diff
@@ -0,0 +1,85 @@
...
+        if (in_frame == NULL)
+            goto end_loop;
+
+        // Information on dropped packets
+        uint8_t* lost_samples_frame = in_buf->frames[in_buffer_ID];
+        uint8_t* compressed_lost_samples_frame = out_buf->frames[out_buffer_ID];
+        uint32_t total_lost_samples = 0;
+
+        // Compress lost samples buffer by checking each sample for a flag
+        for (uint sample = 0; sample < _samples_per_data_set; sample += 3 * _num_sub_freqs) {
```
andrerenard: Can we make 3 at least a constant?  It shows up a lot here.

lib/stages/rawFileWrite.hpp
james-s-willis: Where do you mean for the list of options?

lib/testing/testDataGenMissingFrames.hpp
james-s-willis: I gave Davor this stage and he has modified for his needs. I will remove my version of it and use Davor's instead.

lib/stages/integrateHFBData.cpp
```diff
@@ -0,0 +1,228 @@
...
+}
+
+integrateHFBData::~integrateHFBData() {}
+
+void integrateHFBData::initFirstFrame(float* input_data, float* sum_data,
+                                      const uint32_t in_buffer_ID) {
+
+    int64_t fpga_seq_num_start =
+        fpga_seq_num_end - (_num_frames_to_integrate - 1) * _samples_per_data_set;
+    memcpy(&sum_data[0], &input_data[0], _num_frb_total_beams * _num_sub_freqs * sizeof(float));
```
andrerenard:
```suggestion
    memcpy(sum_data, input_data, _num_frb_total_beams * _num_sub_freqs * sizeof(float));
```
If it complains then cast to `(void *)`, but there is no pointing getting the pointer to the first element of an array, when the array itself is that value.

lib/stages/integrateHFBData.cpp
```diff
@@ -0,0 +1,228 @@
...
+    while (!stop_thread) {
+        // Get an input buffer, This call is blocking!
+        in_frame = wait_for_full_frame(in_buf, unique_name.c_str(), in_buffer_ID);
+        if (in_frame == NULL)
+            goto end_loop;
+
+        compressed_lost_samples_frame = wait_for_full_frame(
+            compressed_lost_samples_buf, unique_name.c_str(), compress_buffer_ID);
+        if (compressed_lost_samples_frame == NULL)
+            goto end_loop;
```
andrerenard: Just to make sure we write it down, currently this could result in a desync between the `in_buf` and `compressed_lost_samples_buf` if some frames go missing in one buffer but not the other.   You might want to see what Richard did for this in `RfiFrameDrop`, it doesn't involve lost samples, but it does cover this case.   I think Davor also did something like this in one of the frb/pulsar post process stages.   Given this pattern has shown up a few times, we might want to abstract this pattern into its own stage.

lib/stages/integrateHFBData.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+    uint in_buffer_ID = 0,
+         compress_buffer_ID = 0; // Process only 1 GPU buffer, cycle through buffer depth
+    uint8_t* in_frame;
+    uint8_t* compressed_lost_samples_frame;
+    int out_buffer_ID = 0, first = 1;
+    int64_t fpga_seq_num_end_old;
+    total_timesamples = _samples_per_data_set * _num_frames_to_integrate;
+    total_lost_timesamples = 0;
+    fpga_seq_num = 0, fpga_seq_num_end = (_num_frames_to_integrate - 1) * _samples_per_data_set;
```
andrerenard: From a style point of view I think this should be two lines.

lib/stages/integrateHFBData.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+                // Get a new output buffer
+                out_buffer_ID = (out_buffer_ID + 1) % out_buf->num_frames;
+                out_frame = wait_for_empty_frame(out_buf, unique_name.c_str(), out_buffer_ID);
+                if (out_frame == NULL)
+                    goto end_loop;
+
+                sum_data = (float*)out_buf->frames[out_buffer_ID];
+            } else
+                DEBUG("Integration discarded. Too many lost samples.");
```
andrerenard: Style point, I'd like this to be a fully `{` open and `}` close.  I find it harder to read these one line if results when one part is not one line.

lib/stages/rawFileRead.cpp
```diff
@@ -59,47 +62,59 @@ void rawFileRead::main_thread() {
...
-        if (metadata_size != 0) {
+          // If metadata exists then lets read it in.
+          if (metadata_size != 0) {
             allocate_new_metadata_object(buf, frame_id);
             struct metadataContainer* mc = get_metadata_container(buf, frame_id);
             assert(metadata_size == mc->metadata_size);
             if (fread(mc->metadata, metadata_size, 1, fp) != 1) {
-                ERROR("rawFileRead: Failed to read file {:s} metadata,", full_path);
-                break;
+              ERROR("rawFileRead: Failed to read file {:s} metadata,", full_path);
```
andrerenard: The formatting of this file is off, I'm not sure why.

lib/hsa/hsaBeamformUpchanHFB.hpp
```diff
@@ -0,0 +1,99 @@
...
+ *     @gpu_mem_type            staging
+ *     @gpu_mem_format          Array of @c float
+ *
+ * @conf   num_elements         Int (default 2048). Number of elements
+ * @conf   samples_per_data_set Int (default 49152). Number of time samples in a data set
+ * @conf   downsample_time      Int (default 3). Downsample factor in time
+ * @conf   downsample_freq      Int (default 8). Downsample factor in freq
+ * @conf   num_frb_total_beams  Int (default 1024). Number of total FRB formed beams
+ * @conf   num_sub_freqs        Int (default 128). Number of sub frequencies
+ *
```
andrerenard: Missing comment for `factor_upchan` which seems to be the same value as `num_sub_freqs`, can we reduce this to one variable maybe?

lib/hsa/kernels/sum_hfb.cl
```diff
@@ -0,0 +1,42 @@
+// Input data is float with sample-beam-freq, sum each frequency across samples
+//LWS = {64,  1  }
```
andrerenard: It would be nice to have more comments here on what this kernel does, and what the global/local dimensions represent.

lib/hsa/kernels/upchannelize_flip_hfb.cl
```diff
@@ -0,0 +1,309 @@
...
+
+__constant float BP[16] = { 0.52225748 , 0.58330915 , 0.6868705 , 0.80121821 , 0.89386546 , 0.95477358 , 0.98662733 , 0.99942558 , 0.99988676 , 0.98905127 , 0.95874124 , 0.90094667 , 0.81113021 , 0.6999944 , 0.59367968 , 0.52614263};
+__constant float HFB_BP[16] = { 1.f, 1.f, 1.f, 1.f, 1.f, 1.f, 1.f, 1.f, 1.f, 1.f, 1.f, 1.f, 1.f, 1.f, 1.f, 1.f};
+
+#define BIT_REVERSE_7_BITS(index) ((( ( (((index) * 0x0802) & 0x22110) | (((index) * 0x8020)&0x88440) ) * 0x10101 ) >> 17) & 0x7F)
+//input data is float2 with beam-pol-time, try to do 3 N=128 at once so that we can sum 3 time samples
+//LWS = {     64 ,  1  } 
+//GWS = {nsamp/6*, 1024} 
+
+__kernel void upchannelize(__global float2 *data, __global float *results_array, __global float *hfb_output_array){
```
andrerenard: I realize the original kernel likely doesn't have comments either, but it would be nice to say at least how this one differs from the original.

lib/stages/bufferCopy.hpp
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @warning The sizes of the frames must be the same in all buffers, and the
+ *       metadata types and underlying pools must also be the same.
+ *
+ *
+ * @par Buffer
+ * @buffer in_buf Buffer to copy frames from
+ *        @buffer_format Matches the output buffers
+ *        @buffer_metadata Matches the output buffers
+ * @buffer out_bufs Array of output buffers to copy frames into.
```
andrerenard: These comments appear to have been copied from `bufferMerge` which has an optional complex selection process.  This `stage` doesn't really need that fancy "internal_name" functionality, and including that in the comments suggests it might do something which it doesn't.  I would simplify this, and also simplify the config reading logic above.  We really just need the version without the internal names here.

lib/stages/rawFileWrite.hpp
```diff
@@ -19,6 +20,7 @@
  * @conf base_dir  String. Directory to write into.
  * @conf file_name String. Base filename to write.
  * @conf file_ext  String. File extension.
+ * @conf _num_frames_per_file Integer. No of frames to write into a single file.
```
andrerenard:
```suggestion
 * @conf num_frames_per_file Integer. No of frames to write into a single file.
```
The config file doesn't include the prefix.

lib/stages/rawFileWrite.cpp
```diff
@@ -36,6 +39,8 @@ rawFileWrite::rawFileWrite(Config& config, const std::string& unique_name,
     base_dir = config.get<std::string>(unique_name, "base_dir");
     file_name = config.get<std::string>(unique_name, "file_name");
     file_ext = config.get<std::string>(unique_name, "file_ext");
+    _num_frames_per_file = config.get_default<uint32_t>(unique_name, "num_frames_per_file", 1);
+    prefix_hostname = config.get_default<bool>(unique_name, "prefix_hostname", true);
```
andrerenard: Can we clean up these names to include the `_` prefix while we are at it?  I don't like seeing mixed standards.

config/chime_science_run_gpu.yaml
```diff
@@ -11,6 +11,7 @@
 # - FRB FFT Fan beam and packet Tx, with Valves protecting input buffers from overflowing
 # - Pulsar 10 coherent beam system and packet Tx, with Valves protecting input buffers from overflowing
 # - RFI detection and summary stats system (only reports on RFI, no normalization yet)
+# - Extracts 21cm absorber data from FRB pipeline
```
andrerenard:
```suggestion
# - Extracts 21cm absorber data from FRB pipeline.  This is referred to as the High Frequency Beam (HFB) below. 
```
Same for the `no_output` file.

config/chime_science_run_gpu.yaml
```diff
@@ -60,6 +61,7 @@ downsample_freq: 8
 factor_upchan: 128
 factor_upchan_out: 16
 num_frb_total_beams: 1024
+num_sub_freqs: 128
```
andrerenard: This variable is the same as `factor_upchan` in the FRB options.   We should merge this down to one variable.

lib/hsa/hsaBeamformHFBSum.hpp
```diff
@@ -0,0 +1,81 @@
...
+ *     @gpu_mem_format       Array of @c uint8_t
+ *     @gpu_mem_metadata     chimeMetadata
+ * @gpu_mem  hfb_sum_output     Output data of size output_frame_len
+ *     @gpu_mem_type            staging
+ *     @gpu_mem_format          Array of @c float
+ *
+ * @conf   num_frb_total_beams  Int (default 1024). Number of total FRB formed beams
+ * @conf   num_sub_freqs  Int (default 128). Number of sub frequencies
+ * @conf   samples_per_data_set Int. Number of time samples in a data set.
+ *
```
andrerenard: Some of the config variable are missing here.

PR #443, opened 2019-07-08
Move RFI metadata processing

andrerenard:
Move RFI metadata processing out of the GPU stage into its own stage called `rfiUpdateMetadata`

This is trying to address the problem that when RFI flagging spikes, it requires checking the `lost_samples_buffer` for lost packets which happened at the same time as the RFI zeroed data.  This is to prevent double counting of RFI zeroed data, and lost packets.  However, this seemed to lead to instability when dealing with large RFI spikes, and caused a backlog in the GPU part of the system.


lib/hsa/hsaRfiMaskOutput.hpp
```diff
@@ -36,8 +36,6 @@ class hsaRfiMaskOutput : public hsaSubframeCommand {
     virtual ~hsaRfiMaskOutput();
     /// Wait for output buffer to be empty, keep track of _rfi_mask_output_buf_precondition_id
     int wait_on_precondition(int gpu_frame_id) override;
-    /// Function to handle updatble config rest server calls for rfi zeroing toggle
-    bool update_rfi_add_lostsamples_flag(nlohmann::json& json);
```
nritsche: was this ever implemented?

lib/hsa/hsaRfiMaskOutput.hpp
```diff
@@ -47,37 +45,15 @@ class hsaRfiMaskOutput : public hsaSubframeCommand {
...
-    uint32_t _num_local_freq;
-    /// Total number of frequencies (1024)
-    uint32_t _num_total_freq;
-    /// Number of time samples per frame (Usually 32768 or 49152)
-    uint32_t _samples_per_data_set;
-    // RFI config parameters
-    /// The kurtosis step (How many timesteps per kurtosis estimate)
-    uint32_t _sk_step;
-    /// A Flag wich sets whether or not we compute and add the lost samples due to RFI flagging
-    bool _rfi_add_lostsamples;
```
nritsche: Was this also never used or am I confused?

lib/hsa/hsaRfiMaskOutput.hpp
```diff
@@ -47,37 +45,15 @@ class hsaRfiMaskOutput : public hsaSubframeCommand {
...
     int32_t _rfi_mask_output_buf_precondition_id;
     /// ID for _rfi_mask_output_buf_execute
     int32_t _rfi_mask_output_buf_execute_id;
-    // General Config Parameters
-    /// Number of elements (2048 for CHIME or 256 for Pathfinder)
-    uint32_t _num_elements;
-    /// Number of frequencies per GPU (1 for CHIME or 8 for Pathfinder)
-    uint32_t _num_local_freq;
-    /// Total number of frequencies (1024)
-    uint32_t _num_total_freq;
```
nritsche: Also never used?

config/chime_science_run_gpu.yaml
```diff
@@ -535,6 +535,29 @@ gpu:
       rfi_output_buf: gpu_rfi_output_buffer_3
       rfi_mask_output_buf: gpu_rfi_mask_output_buffer_3
 
+
+rfi_update_metadata:
+  update_metadata_0:
+    kotekan_stage: rfiUpdateMetadata
+    rfi_mask_buf: gpu_rfi_mask_output_buffer_0
+    lost_samples_buf: lost_samples_buffer
+    gpu_correlation_buf: gpu_n2_output_buffer_0
```
nritsche: Maybe add a comment here in the config to explain why this is used as an output of this as well as `hsaProcess` stage? It had confused me.

lib/processes/rfiUpdateMetadata.cpp
```diff
@@ -0,0 +1,114 @@
...
+        for (uint32_t subframe = 0; subframe < _num_sub_frames; ++subframe) {
+
+            // Total number of samples flagged as RFI
+            uint32_t flagged_samples = 0;
+
+            // The number of samples flagged as RFI less the number of
+            // samples flagged as missing (i.e. packet loss/corrupt packet)
+            // in the same time block as an RFI flag.
+            uint32_t net_lost_samples = 0;
+            uint32_t sub_frame_mask_len = _sub_frame_samples / _sk_step;
```
nritsche: Can be done in the constructor

lib/processes/rfiUpdateMetadata.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+            for (uint32_t i = 0; i < sub_frame_mask_len; ++i) {
+                uint32_t mask_idx = subframe * sub_frame_mask_len + i;
```
nritsche: `subframe * sub_frame_mask_len` can be done in the outer loop...

lib/processes/rfiUpdateMetadata.cpp
```diff
@@ -0,0 +1,114 @@
...
+                assert(mask_idx < (uint32_t)rfi_mask_buf->frame_size);
+                if (rfi_mask_frame[mask_idx] == 1) {
+                    flagged_samples += _sk_step;
+                    net_lost_samples += _sk_step;
+                    // Remove any samples which were also counted as lost samples
+                    // in this block of data.
+                    for (uint32_t j = 0; j < _sk_step; ++j) {
+                        uint32_t lost_samples_idx =
+                            subframe * _sub_frame_samples + i * _sk_step + j;
```
nritsche: ...and used here as well.

lib/hsa/hsaRfiMaskOutput.cpp
```diff
@@ -15,25 +15,11 @@ hsaRfiMaskOutput::hsaRfiMaskOutput(Config& config, const string& unique_name,
...
-    register_consumer(_lost_samples_buf, unique_name.c_str());
-
     _rfi_mask_output_buf = host_buffers.get_buffer("rfi_mask_output_buf");
     register_producer(_rfi_mask_output_buf, unique_name.c_str());
 
-    _output_buf = host_buffers.get_buffer("output_buf");
-    register_producer(_output_buf, unique_name.c_str());
-
-    // Config parameters
-    _num_local_freq = config.get<uint32_t>(unique_name, "num_local_freq");
```
nritsche: Read from config but never used?

lib/hsa/hsaRfiMaskOutput.hpp
andrerenard: At one point yes, but looks like when we changed it, this got left behind.

lib/hsa/hsaRfiMaskOutput.hpp
andrerenard: Yeah this was a part of the above function, which got removed at one point.

lib/hsa/hsaRfiMaskOutput.hpp
andrerenard: Ah this was used at one point when it supported Pathfinder type configurations.  However we decided to remove that support, but it looks like this wasn't cleaned up.

lib/hsa/hsaRfiMaskOutput.cpp
andrerenard: yeah that was part of the removed Pathfinder support that apparently didn't get cleaned up.

andrerenard:
Thanks for the review.  I think I've addressed all the comments now.

PR #444, opened 2019-07-08
Update polyco script

tristpinsm:
We are approaching a time when the 26m will be operational again and we can resume testing/start observations with pulsar gating, so I improved the script that generates and sends gating configs. It should now be able to generate the config update and send it as well as disable gating all within the script.


andrerenard:
Is this being targeted for this release, or can we delay getting it in until after the current deployment.

tristpinsm:
This only touches the polyco python script, so it doesn't need to be included in the release, but it also has no effect on kotekan.

andrerenard:
@tristpinsm any objections to me merging this in now?

tristpinsm:
@andrerenard no objections. I've been using it to enable/disable gating a few times now and no more bugs cropped up.

PR #446, opened 2019-07-10
Initialize `kotekan_baseband_readout_in_progress` to 0 at start of main loop

cubranic:
Otherwise the metric is missing until the node receives a baseband dump request.


cubranic:
Hmm, maybe this PR is not that important because the gauge will be timestamped, and so will almost certainly be considered stale before the next baseband readout comes along. So it's more for tidiness than any real benefit.

Issue #276, opened 2018-11-13
Error handling framework


jrs65:
It would be nice to have a common error handling framework. I see a lot of places in the code with the pattern:

```
ERROR("Describe some error");
std::raise(SIGINT);
return;
```

Maybe wrapping this somehow into a `FATAL_ERROR` macro.

andrerenard:
I'd propose the following two ways to exit:

This function: `exit_kotekan(enum ReturnCode)`.  Which raises `SIGINT` and sets a return value based on a `ReturnCode` enum.  That return value is then returned by the binary after the shutdown process is finished.

Or for common fatal errors this simple macro:

`FATAL_ERROR(message, ...)`

Which then calls `ERROR(message, ...)` and `exit_kotekan(ReturnCode::FATAL_ERROR)`

Some possible error codes we might consider would be: `FATAL_ERROR`, `CLEAN_EXIT`, `TEST_PASSED`, `TEST_FAILED`.  The codes numbers should follow Linux conventions where possible.   The `TEST_` errors would allow us to make testing configurations with an easy to script binary return code for pass/fail for use with the CI server.

PR #447, opened 2019-07-10
Exit kotekan

james-s-willis:
Created a function to exit `kotekan` with a status code. Addresses issue: #276.

@andrerenard @jrs65 do you agree with these changes? If so, I will replace all instances of:

```
ERROR("Describe some error");
std::raise(SIGINT);
return;
```
with:

```
FATAL_ERROR("Describe some error");
```

### DIFF
--- a/kotekan/kotekan.cpp
+++ b/kotekan/kotekan.cpp
@@ -499,7 +499,7 @@ int main(int argc, char** argv) {
         }
     }
 
-    INFO("kotekan shutdown with status: %s", get_exit_code_string(get_exit_code()));
+    INFO("kotekan shutdown with status: %s\n%s", get_exit_code_string(get_exit_code()), get_error_message());
 
     closelog();
 
--- a/kotekan/kotekan.cpp
+++ b/kotekan/kotekan.cpp
@@ -454,7 +454,8 @@ int main(int argc, char** argv) {
 
     rest_server.register_get_callback("/kill", [&](connectionInstance& conn) {
         ERROR("/kill endpoint called, raising SIGINT to shutdown the kotekan system process.");
-        set_error_message("/kill endpoint called, raising SIGINT to shutdown the kotekan system process.");
+        set_error_message(
+            "/kill endpoint called, raising SIGINT to shutdown the kotekan system process.");
         exit_kotekan(ReturnCode::CLEAN_EXIT);
         conn.send_empty_reply(HTTP_RESPONSE::OK);
     });
@@ -501,10 +502,10 @@ int main(int argc, char** argv) {
     }
 
     INFO("kotekan shutdown with status: %s", get_exit_code_string(get_exit_code()));
-    
+
     // Print error message if there is one.
     if (string(get_error_message()) != "not set") {
-      INFO("Fatal error message was: %s", get_error_message());
+        INFO("Fatal error message was: %s", get_error_message());
     }
 
     closelog();
--- a/kotekan/kotekan.cpp
+++ b/kotekan/kotekan.cpp
@@ -124,6 +124,8 @@ kotekanMode* kotekan_mode = nullptr;
 bool running = false;
 std::mutex kotekan_state_lock;
 volatile std::sig_atomic_t sig_value = 0;
+extern enum ReturnCode status_code;
+extern const char *returnCodeNames[RETURN_CODE_COUNT];
 
 void signal_handler(int signal) {
     sig_value = signal;
@@ -499,9 +501,9 @@ int main(int argc, char** argv) {
         }
     }
 
-    INFO("kotekan shutdown successfully.");
+    INFO("kotekan shutdown successfully, with status: %s", returnCodeNames[status_code]);
 
     closelog();
 
-    return 0;
+    return status_code;
 }
--- a/kotekan/kotekan.cpp
+++ b/kotekan/kotekan.cpp
@@ -499,7 +499,12 @@ int main(int argc, char** argv) {
         }
     }
 
-    INFO("kotekan shutdown with status: %s\n%s", get_exit_code_string(get_exit_code()), get_error_message());
+    INFO("kotekan shutdown with status: %s", get_exit_code_string(get_exit_code()));
+    
+    // Print error message if there is one.
+    if (string(get_error_message()) != "not set") {
+      INFO("Fatal error message was: %s", get_error_message());
+    }
 
     closelog();
 
--- a/kotekan/kotekan.cpp
+++ b/kotekan/kotekan.cpp
@@ -124,8 +124,6 @@ kotekanMode* kotekan_mode = nullptr;
 bool running = false;
 std::mutex kotekan_state_lock;
 volatile std::sig_atomic_t sig_value = 0;
-extern enum ReturnCode __status_code;
-extern const char *returnCodeNames[RETURN_CODE_COUNT];
 
 void signal_handler(int signal) {
     sig_value = signal;
@@ -501,9 +499,9 @@ int main(int argc, char** argv) {
         }
     }
 
-    INFO("kotekan shutdown with status: %s", returnCodeNames[__status_code]);
+    INFO("kotekan shutdown with status: %s", get_exit_code_string(get_exit_code()));
 
     closelog();
 
-    return __status_code;
+    return get_exit_code();
 }
--- a/kotekan/kotekan.cpp
+++ b/kotekan/kotekan.cpp
@@ -124,7 +124,7 @@ kotekanMode* kotekan_mode = nullptr;
 bool running = false;
 std::mutex kotekan_state_lock;
 volatile std::sig_atomic_t sig_value = 0;
-extern enum ReturnCode status_code;
+extern enum ReturnCode __status_code;
 extern const char *returnCodeNames[RETURN_CODE_COUNT];
 
 void signal_handler(int signal) {
@@ -501,9 +501,9 @@ int main(int argc, char** argv) {
         }
     }
 
-    INFO("kotekan shutdown with status: %s", returnCodeNames[status_code]);
+    INFO("kotekan shutdown with status: %s", returnCodeNames[__status_code]);
 
     closelog();
 
-    return status_code;
+    return __status_code;
 }
--- a/kotekan/kotekan.cpp
+++ b/kotekan/kotekan.cpp
@@ -501,7 +501,7 @@ int main(int argc, char** argv) {
         }
     }
 
-    INFO("kotekan shutdown successfully, with status: %s", returnCodeNames[status_code]);
+    INFO("kotekan shutdown with status: %s", returnCodeNames[status_code]);
 
     closelog();
 
--- a/kotekan/kotekan.cpp
+++ b/kotekan/kotekan.cpp
@@ -453,7 +453,9 @@ int main(int argc, char** argv) {
     });
 
     rest_server.register_get_callback("/kill", [&](connectionInstance& conn) {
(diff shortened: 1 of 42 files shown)


jrs65:
@james-s-willis I might be better to try and tie the error message to the end of the shutdown procedure for better visibility. By that I mean that once kotekan sends the shutdown signal to itself it tends to output another few hundred lines of log output of closing and processes and buffers. So if you log the message, send the signal, and then exit, you need to scroll up some number of lines to find *why* it exited.

It's not clear exactly how to do that without attaching a lot more state at the global level (I must admit I'm not a fan of the return code global you use at the moment).

As a larger change, potentially you could use a singleton instance class designed for errorHandling, that gives a way for threads to post error states, and then the loop within the `main` function can just poll the state of the error handling singleton. That would allow removing `SIGINT` as the mechanism for communicating an error, which always felt ugly, and any use of explicitly global state. I guess you'd want to put in a handler that calls the error handler if anyone sends SIGINT externally.

Something like:

```
// Error in stage
exitHandler::instance().exit(code, "Error message to report.");

//...

auto& exit_handler = exitHandler::instance();
// In main loop
     for(EVER) {
        if (exit_handler.should_exit()) {
             // Shutdown and join...
             // ...
             exit_handler.report_error()
        }
    }
    return exit_handler.exit_code();
```

jrs65:
One thing I forgot to say, is that I'm very glad we're finally addressing this! Your changes is already way better than what were already doing so I'd basically be happy to merge it as is.

james-s-willis:
Thanks for the comments @jrs65. I agree with you about the global variable but I couldn't see a way of implementing the functionality without it. 

I think the singleton instance class would be a better way of doing it, but would take a bit longer to implement. How about we merge what I have already done and keep the issue open for future changes?

What do you think @andrerenard?

lib/core/errors.h
```diff
@@ -92,6 +96,10 @@ void internal_logging(int log, const char * format, ...);
 // Should be used sparingly, and limited to useful areas.
 #define INFO(m, a...) if (__log_level > 2) { internal_logging(LOG_INFO, m, ## a); }
 
+// Use this for fatal errors that kotekan can't recover from.
+// Prints an error message and raises a SIGINT.
+#define FATAL_ERROR(message, a...) { ERROR(message); exit_kotekan(ReturnCode::FATAL_ERROR);}
```
andrerenard: I don't think this works with variadic calls to `FATAL_ERROR`, i.e. `FATAL_ERROR("Error value %d", myval);`.  Not completely sure my suggestion works either since this is compound macro, but this should be tested regardless. 
```suggestion
#define FATAL_ERROR(message, a...) { ERROR(message, ## a); exit_kotekan(ReturnCode::FATAL_ERROR);}
```

andrerenard:
And yes it would be very nice to get the fatal error printed as the last line as well, that would make it much easier to parse the logs.

kotekan/kotekan.cpp
```diff
@@ -124,6 +124,8 @@ kotekanMode* kotekan_mode = nullptr;
 bool running = false;
 std::mutex kotekan_state_lock;
 volatile std::sig_atomic_t sig_value = 0;
+extern enum ReturnCode status_code;
```
andrerenard: You could have a function like `get_exit_code()` in `errors.h` to avoid having to `extern` the variable holding the value here.  That's a little cleaner than doing it this way.

lib/core/errors.c
```diff
@@ -5,6 +5,9 @@ int __log_level = 3;
 int __enable_syslog = 1;
 const int __max_log_msg_len = 1024;
 
+enum ReturnCode status_code = CLEAN_EXIT;
+const char *returnCodeNames[RETURN_CODE_COUNT] = {"CLEAN_EXIT", "FATAL_ERROR", "TEST_PASSED", "TEST_FAILED"};
```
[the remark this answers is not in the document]
james-s-willis: So you would prefer a case statement instead?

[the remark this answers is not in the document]
andrerenard: Yes

lib/core/errors.c
```diff
@@ -5,6 +5,9 @@ int __log_level = 3;
 int __enable_syslog = 1;
 const int __max_log_msg_len = 1024;
 
+enum ReturnCode __status_code = CLEAN_EXIT;
+char __err_msg[1024] = "Hello there";
```
nritsche: Was the `"Hello there"` forgotten here?

kotekan/kotekan.cpp
```diff
@@ -499,9 +499,9 @@ int main(int argc, char** argv) {
         }
     }
 
-    INFO("kotekan shutdown successfully.");
+    INFO("kotekan shutdown with status: %s\n%s", get_exit_code_string(get_exit_code()), get_error_message());
```
andrerenard: The syslog version of this message won't include the newline. 
```suggestion
    INFO("kotekan shutdown with status: %s", get_exit_code_string(get_exit_code()));
    If (string(get_error_message()) != "not set") {
        INFO("Fatal error message was: %s", get_error_message());
    } 
    
```

lib/core/errors.c
james-s-willis: Yep, I just realised that. I've removed it now.

andrerenard: This message will get printed unless the other change is also applied.
```suggestion
char __err_msg[__max_log_msg_len] = "not set";
```

kotekan/kotekan.cpp
```diff
@@ -453,8 +453,7 @@ int main(int argc, char** argv) {
     });
 
     rest_server.register_get_callback("/kill", [&](connectionInstance& conn) {
-        INFO("/kill endpoint called, raising SIGINT to shutdown the kotekan system process.");
-        raise(SIGINT);
+        FATAL_ERROR("/kill endpoint called, raising SIGINT to shutdown the kotekan system process.");
```
andrerenard: This one is a little odd.  The message at the end makes sense, but this isn't an error, so the error code should be CLEAN_EXIT.  you might just manually break this one up into the parts of the macro since it's a special case.

lib/dpdk/dpdkCore.cpp
```diff
@@ -310,7 +310,7 @@ int dpdkCore::lcore_rx(void* args) {
         if (core->handlers[port] == nullptr) {
             // This is the one place (static member function) where normal logging does work.
             fprintf(stderr, "No valid handler provided for port %d", port);
-            raise(SIGINT);
+            FATAL_ERROR("No valid handler provided for port %d", port);
```
andrerenard: Ah this may not work.  The lcore thread is another special case.  Basically for a somewhat complex reason the logging doesn't quite work as you'd expect here.  Which is why it uses `fprintf(stderr, ...);` above.   Generally I think you should just ignore cases were the error messages are printed with `printf` and leave them as they are now with the `raise()`.

kotekan/kotekan.cpp
```diff
@@ -453,8 +453,9 @@ int main(int argc, char** argv) {
     });
 
     rest_server.register_get_callback("/kill", [&](connectionInstance& conn) {
-        INFO("/kill endpoint called, raising SIGINT to shutdown the kotekan system process.");
-        raise(SIGINT);
+        ERROR("/kill endpoint called, raising SIGINT to shutdown the kotekan system process.");
+        set_error_message("/kill endpoint called, raising SIGINT to shutdown the kotekan system process.");
```
andrerenard: I'd remove the SIGINT bit just to make the final message more to the point since this is a special (but common) case. 
```suggestion
        set_error_message("/kill endpoint called.");
```

lib/processes/bufferMerge.cpp
```diff
@@ -70,8 +70,7 @@ void bufferMerge::main_thread() {
     frameID out_frame_id(out_buf);
 
     if (get_num_producers(out_buf) != 1) {
-        ERROR("Cannot merge into a buffer with more than one producer");
-        raise(SIGINT);
+        FATAL_ERROR("Cannot merge into a buffer with more than one producer");
```
andrerenard: Just realized there is a bug here:
```suggestion
        FATAL_ERROR("Cannot merge into a buffer with more than one producer");
        return;
```

lib/processes/monitorBuffer.cpp
```diff
@@ -30,30 +30,26 @@ void monitorBuffer::main_thread() {
...
             if ((cur_time - last_arrival) > timeout && last_arrival > 1) {
-                ERROR("The buffer %s hasn't received a frame for %f seconds.", buf->buffer_name,
-                      (cur_time - last_arrival));
-                ERROR("Closing kotekan because of system timeout.");
                 for (auto& buf : buffer_container.get_buffer_map()) {
                     print_buffer_status(buf.second);
                 }
                 usleep(50000);
-                raise(SIGINT);
+                FATAL_ERROR("The buffer %s hasn't received a frame for %f seconds.\nClosing kotekan because of system timeout.", buf->buffer_name,
```
andrerenard: multiline errors with `\n` don't show up in syslog.  So better to just not use them. 
```suggestion
                FATAL_ERROR("The buffer %s hasn't received a frame for %f seconds.  Closing kotekan because of system timeout.", buf->buffer_name,
```

andrerenard:
@jrs65 Could you please check the changes @james-s-willis made to the unit test error handling?  If they look good then I'm happy to merge this in.

python/kotekan/runner.py
```diff
@@ -136,9 +138,17 @@ def run(self):
...
 
-            # Throw an exception if we don't exit cleanly
-            if p.returncode:
-                raise subprocess.CalledProcessError(p.returncode, cmd)
+            # If failure is expected just report the exit code
+            if self.expect_failure is True:
+                print("Test failed as expected with exit code: " + p.returncode)
+                self.return_code = p.returncode
+            else:
+                try:
```
jrs65: This block is a bit odd in that it raises an exception which it immediately catches and then doesn't propagate (which it should). I think you can replace the whole `else` branch with:

```
elif p.returncode:
    raise subprocess.CalledProcessError(p.returncode, cmd)
```

james-s-willis:
So I've made that change and the checks are failing here now:

```
[],"fakenetwork":{"kotekan_stage":"testDataGen","network_out_buf":"fakenetwork_buf0","num_frames":60,"stage_name":"fakenetwork","type":"tpluse"},"fakenetwork_buf0":{"frame_size":"samples_per_data_set * num_elements* num_local_freq * num_data_sets","kotekan_buffer":"standard","metadata_pool":"main_pool","num_frames":"buffer_depth"},"log_level":"info","main_pool":{"kotekan_metadata_pool":"chimeMetadata","num_metadata_objects":"30 * buffer_depth"},"max_dump_samples":3500,"num_data_sets":1,"num_elements":256,"num_frames_buffer":18,"num_gpu_frames":64,"num_local_freq":1,"rest_mode":"none","samples_per_data_set":1024,"stream_id":0,"total_frames":60,"type":"tpluse","value":153,"vis_pool":{"kotekan_metadata_pool":"visMetadata","num_metadata_objects":"30 * buffer_depth"}}
No GPS time set, using system clock.
restServer: Adding REST endpoint: /config
Creating metadata pool of type: chimeMetadata, at config tree path: /main_pool
Creating metadata pool of type: visMetadata, at config tree path: /vis_pool
Creating standard buffer named fakenetwork_buf0, with 20 frames, frame_size of 262144, and metadata pool main_pool
Failed to start kotekan with config file /tmp/tmpqmNyKu, error message: Baseband dump directory not writable.
Exiting...

FAILED

```

james-s-willis:
I have made all of the tests pass now. Do the changes look good to you @andrerenard ?

PR #459, opened 2019-07-29
Add support for URL query args

andrerenard:
Adds support for URL query args. e.g. `/my_endpoint?my_option=my_val`

Changes baseband status request API from `/baseband/<event_id>` to `/baseband?event_id=<event_id>`

### DIFF
--- a/lib/core/restServer.hpp
+++ b/lib/core/restServer.hpp
@@ -99,8 +99,8 @@ public:
      * @brief Gets the query args as a map of key value strings
      *
      * Example "/my_endpoint?val=42&myval=hello" would return a map with items:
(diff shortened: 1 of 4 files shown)


lib/core/restServer.hpp
```diff
@@ -95,6 +95,19 @@ class connectionInstance {
      */
     std::string get_uri();
 
+    /**
+     * @brief Gets the query args as a map of key value strings
+     *
+     * Example "/my_endpoint?val=42&myval=hello" would return a map with items:
+     * map[myval] == 42
```
cubranic: This would be `map["val"] == "42"` and `map["myval"] == "hello", no?

PR #467, opened 2019-07-31
PSR FRB updatable config

cherryng:
- renamed frb gain endpoint to be consistent with the `updatable_config` style
- renamed psr gain endpoint to be consistent with the `updatable_config` style
- changed psr pointing to be an `updatable_config`


nritsche:
@andrerenard @cherryng It would be easy to change this in coco, but I think the config changes here are not how the structure of updatable config blocks is normally done in kotekan. The updatable config block itself should be called `updatable_config`. Here the link from the stage's block to the updatable config block is in a block called `updatable_config` but that's not necessary.

Here is a correct example:
```
# Updatable config blocks
updatable_config:
  gains:
    kotekan_update_endpoint: "json"
    start_time: 1541039597.
    tag: "gain_20181101T023317.440691Z_cyga"
  26m_ungated:
    kotekan_update_endpoint: "json"
    enabled: false

# Kotekan stages
switch_26m_ungated:
  kotekan_stage: bufferSwitch
  in_bufs:
    - enabled: visbuf_5s_26m_ungated
  out_buf: visbuf_5s_26m_ungated_post
  updatable_config: "/updatable_config/26m_ungated"
```

andrerenard:
I think I'm a little torn on if that's a good convention, just because it adds to the length of the URLs, and I'm not sure how much it adds to readiably.  However since that's our current convention (and used in the receiver nodes), I think the best thing to do is stay with it for now.  @cherryng could you change the config file to match the convention above.

lib/hsa/hsaPulsarUpdatePhase.cpp
```diff
@@ -50,14 +53,13 @@ hsaPulsarUpdatePhase::hsaPulsarUpdatePhase(Config& config, const string& unique_
...
 
-    _source_ra = config.get<std::vector<float>>(unique_name, "source_ra");
-    _source_dec = config.get<std::vector<float>>(unique_name, "source_dec");
-    _source_scl = config.get<std::vector<int>>(unique_name, "psr_scaling");
-
-    for (int i = 0; i < _num_beams; i++) {
-        psr_coord_latest_update.ra[i] = _source_ra[i];
-        psr_coord_latest_update.dec[i] = _source_dec[i];
-        psr_coord_latest_update.scaling[i] = _source_scl[i];
+        for (int i = 0; i < _num_beams; i++) {
```
andrerenard: You shouldn't need this block at all.  The call to `configUpdater::instance().subscribe()` will call the `pulsar_grab_callback()` function on registration, which will set these values.

lib/hsa/hsaPulsarUpdatePhase.cpp
```diff
@@ -1,6 +1,9 @@
 // curl localhost:12048/pulsar_gain -X POST -H 'Content-Type: application/json' -d
 // '{"pulsar_gain_dir":["path0","path1","path2","path3","path4","path5","path6","path7","path8","path9"]}'
 
+// curl localhost:12048/pulsar_pointing/0  -X POST -H 'Content-Type: application/json' -d
```
nritsche: missing .../updatable_config/...

config/chime_science_run_gpu.yaml
```diff
@@ -491,11 +542,10 @@ gpu:
   gpu_2:
     kotekan_stage: hsaProcess
     gpu_id: 2
-    updatable_gain_psr: /pulsar_gain
-    updatable_gain_frb: /frb_gain
-    source_ra:  [53.51337,44.51337,52.1633,60.26337,46.76336,54.63837,49.01337,60.26337,55.76337,55.76337]
-    source_dec: [54.6248916,54.7248916,54.7598915,54.7248916,54.6248916,54.6448916,54.5748916,54.5248916,54.5048916,54.4748916]
-    psr_scaling: [48,48,48,48,48,48,48,48,48,48]
+    updatable_config:
```
nritsche: The convention until now would be sth like:
```
gpu_2:
    updatable_config_gain_psr: /updatable_config/pulsar_gain
    ...
```

andrerenard: I don't think that's true.   Because isn't there a function that works on all `updatable_config` block references and matches them to call backs?  Let me check.

andrerenard: This section does follow the currently documented convention: 
``` * In the config block of a stage that wants to get updates of a specific
 * updatable block, either a key "updatable_config" with the full path to the updatable
 * block as a value has to exist, or an object called "updatable_config" with
 * a list of all updatable config blocks.
 *
 * Example:
 * ```
 * my_stage:
 *     updatable_config: "/foo/bar"
 * ```
 * or
 * ```
 * my_stage:
 *     updatable_config:
 *         bar: "/foo/bar"
 *         fu: "/foo/fu"
 * ```

Comment (nritsche) [reply config/chime_science_run_gpu.yaml]: Oh, you're right. Thanks!

File: config/chime_science_run_gpu.yaml [hunk 4]
```diff
@@ -68,28 +68,81 @@ num_frb_total_beams: 1024
...
-frb_gain:
-  kotekan_update_endpoint: json
-  frb_gain_dir: /mnt/frb-baseband/GainFiles/Latest_FRB
+updatable_config:
```
andrerenard: This section needs to be merged with the `updatable_config:` section below.   yaml doesn't allow duplicate keys at the same level.

PR #488, opened 2019-09-03
Fix compile errors in osxBindCPU.cpp

cubranic:
This was probably caused by #483, which didn't update this OSX-only file for
changes in the logging API.


nritsche:
I updated `sphinxcontrib-plantuml` and it seems to be fixed.

PR #495, opened 2019-09-06
Port to python3 using py3port

### LINKED ISSUE #493: port to python3

nritsche:
Closes #493


andrerenard:
Also looks like there is now a conflict to fix.

jrs65:
It looks good to me too. You might consider applying black to the code too while you are doing a massive commit to the python code.

nritsche:
> It looks good to me too. You might consider applying black to the code too while you are doing a massive commit to the python code.

I was planning to do that in a separate PR, but I can add it here if you want.

nritsche:
This keeps breaking, need to get it in before the base changes again. Please review @jrs65 @andrerenard

jrs65:
@nritsche I think this isn't quite ready yet. I'm having some failures of the test while running them in Python 2. I'm fixing them now, I'll try and get it al pushed in the morning.

nritsche:
@jrs65 We had once decided to drop python2 support here. So if you don't think it's absolutely necessary, you can also leave the tests broken.

docs/sphinx/dev/dev_style_guide_code.rst
```diff
@@ -152,3 +152,10 @@ Disabling Code Formatting
 If you write a piece of code that you want to be excluded from auto-formatting,
 you can prepend ``// clang-format off`` or ``/* clang-format off */`` and append
 ``// clang-format on`` or ``/* clang-format on */``.
+
+Python Code Formatting
+----------------------
+
+All python code in this project should be formatted accoring to the `black code style
+<https://black.readthedocs.io/en/stable/the_black_code_style.html>_`. You can let black take care
+of that using `black -l 99 --exclude docs kotekan/root/dir` before you commit python code.
```
nritsche:
```suggestion
of that using `black --exclude docs kotekan/root/dir` before you commit python code.
```

jrs65:
I agree we should drop things swiftly. I just think it's better to have a little bit of time where they work in parallel so you can be sure that you've still got everything tested properly.

jrs65:
I don't know what this last failure is. I've seen it fail intermittently, but can't reproduce it on my laptop.

jrs65:
Scratch that. I can reproduce it on my mac.

jrs65:
Alright. Figured out the bug. I need to go back for my English test so the fix is going to have to wait for a few hours!

jrs65:
Woo. Tests passed.

PR #511, opened 2019-09-11
FRB/PSR gain update

cherryng:
- fread in kotekan stage `lib/stages/ReadGain.cpp`
- async copy in command object `lib/hsa/hsaAsyncCopyGain.cpp`
- use with config `config/frb_test.yaml`


lib/stages/ReadGain.cpp
```diff
@@ -0,0 +1,142 @@
...
+    {
+      std::unique_lock<std::mutex> lock(mux);
+      while (!update_gains) {
+	cond_var.wait(lock);//, [&]{return update_gains = true;});
+      }
+    }
+    INFO("[ReadGain] Going to update gain+++++++++++++++++update_gains={:d}", update_gains);
+
+    //Need to read it in to all gpu frames.
+    for (int f = 0; f < gain_buf->num_frames; f++) {
```
andrerenard: You don't actually need to fill all the frames here.  The command object is dealing with copying one frame from this buffer into the GPU `num_gpu_frames` times.

lib/stages/ReadGain.cpp
```diff
@@ -0,0 +1,142 @@
...
+                           bufferContainer& buffer_container) :
+  Stage(config, unique_name, buffer_container, std::bind(&ReadGain::main_thread, this)) {
+  //Apply config.
+  _num_elements = config.get<uint32_t>(unique_name, "num_elements");
+  scaling = config.get_default<float>(unique_name, "frb_scaling", 1.0);
+  vector<float> dg = {0.0, 0.0}; // re,im
+  default_gains = config.get_default<std::vector<float>>(unique_name, "frb_missing_gains", dg);
+
+  metadata_buf = get_buffer("network_buf");
+  register_consumer(metadata_buf, unique_name.c_str());
```
andrerenard: Hmm, this is actually a little tricky with the current framework.  You only need one of these buffers to get the metadata, but once you register as a consumer it will expect you to read every frame from this buffer, otherwise it will stay "full".   I think this actually requires a bit of a framework tweek, let me get back to you on this.

lib/stages/ReadGain.cpp
```diff
@@ -0,0 +1,142 @@
...
+      if (ptr_myfile == NULL) {
+	WARN("GPU Cannot open gain file {:s}", filename);
+	for (int i = 0; i < 2048; i++) {
+	  out_frame[i * 2] = default_gains[0] * scaling;
+	  out_frame[i * 2 + 1] = default_gains[1] * scaling;
+	}
+      }
+      else {
+	if (_num_elements != fread(out_frame, sizeof(float) * 2, _num_elements, ptr_myfile)) {
+	  WARN("Gain file ({:s}) wasn't long enough! Something went wrong, breaking...", filename);
```
andrerenard:
```suggestion
	  WARN("Gain file ({:s}) wasn't long enough! Something went wrong, using default gains", filename);
```

lib/hsa/hsaAsyncCopyGain.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+hsaAsyncCopyGain::~hsaAsyncCopyGain() {}
+
+int hsaAsyncCopyGain::wait_on_precondition(int gpu_frame_id) {
+    (void)gpu_frame_id;
+    
+    //Check for new gains
+    INFO("[AsyncCopyPrecon] waiting for gain_buf_id={:d} to be full; gpu_frame_id={:d}", gain_buf_precondition_id, gpu_frame_id);
+    if (first_pass) {
+      uint8_t* frame = wait_for_full_frame(gain_buf, unique_name.c_str(),  gain_buf_precondition_id);
```
andrerenard: You should run `clang-format` there are quite a few format issues in this section.

lib/hsa/hsaAsyncCopyGain.cpp
```diff
@@ -0,0 +1,97 @@
...
+      frame_to_fill_finalize = frame_to_fill;
+      filling_frame = true;
+      if (frame == NULL)
+	return -1;
+    } else {
+      //Check for new gains only if filled all gpu frames (not currently filling frame)
+      if (!filling_frame) { 
+	auto timeout = double_to_ts(0);
+	int status = wait_for_full_frame_timeout(gain_buf, unique_name.c_str(), gain_buf_precondition_id, timeout);
+	INFO("[AsyncCopyPrecon] status of gain_buf_precondition_id[{:d}]={:d} ==(0=ready 1=not)================", gain_buf_precondition_id, status);
```
andrerenard: Once you think this is working, can you please change these `INFO` statements to DEBUG statements?

lib/hsa/hsaAsyncCopyGain.cpp
```diff
@@ -0,0 +1,97 @@
...
+  //async copy in
+  if (filling_frame && frame_to_fill > 0) {
+    INFO("[AsyncCopyExe] going to async copy gain_buf_id={:d} gpu_frame_id={:d}", gain_buf_id, gpu_frame_id);
+    void* device_gain = device.get_gpu_memory_array("beamform_gain", gpu_frame_id, gain_len);
+    void* host_gain = (void*)gain_buf->frames[gain_buf_id];
+    device.async_copy_host_to_gpu(device_gain, host_gain, gain_len, precede_signal, signals[gpu_frame_id]);
+    gain_buf_id = (gain_buf_id+1)% gain_buf->num_frames;
+    frame_to_fill--;
+  }
+  return signals[gpu_frame_id];
```
andrerenard: you should return the `precede_signal` if you aren't running a command.  so `return signals[gpu_frame_id];` at the end of the `if` and then `return precede_signal` here.

lib/hsa/hsaAsyncCopyGain.cpp
```diff
@@ -0,0 +1,97 @@
...
+      
+}
+
+void hsaAsyncCopyGain::finalize_frame(int frame_id) {
+    hsaCommand::finalize_frame(frame_id);
```
andrerenard: This line can go inside the `if` statement.

lib/hsa/hsaBeamformKernel.cpp
```diff
@@ -206,8 +179,11 @@ void hsaBeamformKernel::calculate_ew_phase(float freq_now, float* host_coeff,
     }
 }
 
-
 hsa_signal_t hsaBeamformKernel::execute(int gpu_frame_id, hsa_signal_t precede_signal) {
+
+    // Unused parameter, suppress warning
+    (void)precede_signal;  //why do i need this then?
```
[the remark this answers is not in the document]
andrerenard: Remove this comment.
```suggestion
    (void)precede_signal;
```

lib/stages/ReadGain.cpp
```diff
@@ -0,0 +1,139 @@
...
+ReadGain::ReadGain(Config& config, const std::string& unique_name,
+                           bufferContainer& buffer_container) :
+  Stage(config, unique_name, buffer_container, std::bind(&ReadGain::main_thread, this)) {
+  //Apply config.
+  _num_elements = config.get<uint32_t>(unique_name, "num_elements");
+  scaling = config.get_default<float>(unique_name, "frb_scaling", 1.0);
+  vector<float> dg = {0.0, 0.0}; // re,im
+  default_gains = config.get_default<std::vector<float>>(unique_name, "frb_missing_gains", dg);
+
+  metadata_buf = get_buffer("network_buf");
```
andrerenard: Can we change this to `in_buf` just to be more standardized with other stages.  (I know we aren't very consistent with this.)

lib/stages/ReadGain.cpp
```diff
@@ -0,0 +1,139 @@
...
+      goto end_loop;
+    stream_id_t stream_id = get_stream_id_t(metadata_buf, metadata_buffer_id);
+    freq_idx = bin_number_chime(&stream_id);
+    freq_MHz = freq_from_bin(freq_idx);
+    //INFO("[CHECK ReadGain] freq_idx {:d} {:.2f}", freq_idx, freq_MHz);
+    metadata_buffer_precondition_id = (metadata_buffer_precondition_id + 1) % metadata_buf->num_frames;
+  }
+  mark_frame_empty(metadata_buf, unique_name.c_str(), metadata_buffer_id);
+  metadata_buffer_id = (metadata_buffer_id + 1) % metadata_buf->num_frames;
+  //unregister_consumer(metadata_buf, unique_name.c_str());
```
andrerenard: You can enable this by running `git merge ar/unregister` on your branch if you want to test this sooner.   I need to run a few tests on this function before I merge into `develop`

lib/stages/ReadGain.cpp
```diff
@@ -0,0 +1,139 @@
...
+    
+  while (!stop_thread) {
+    INFO("[ReadGain] update_gains={:d}-(0=false 1=true)-------------", update_gains);
```
andrerenard: Same comment with other info statements, this should be changed to debug.   Also you don't need to include `[ReadGain]` the unique_name from the config is added by the framework automatically.

lib/stages/ReadGain.cpp
```diff
@@ -0,0 +1,139 @@
...
+      std::unique_lock<std::mutex> lock(mux);
+      while (!update_gains) {
+	cond_var.wait(lock);//, [&]{return update_gains = true;});
+      }
+    }
+    INFO("[ReadGain] Going to update gain+++++++++++++++++update_gains={:d}", update_gains);
+
+    float* out_frame = (float*)wait_for_empty_frame(gain_buf, unique_name.c_str(), gain_buf_id);
+    if (out_frame == NULL) {
+      INFO("[ReadGain] waiting for frame={:d} but it is not empty", gain_buf_id);
```
andrerenard: This isn't really a correct message.   If it returns `NULL` that's because the system is shutting down.

lib/stages/ReadGain.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include "Stage.hpp"
+#include "restServer.hpp"
+
+#include <vector>
+#include <mutex>
+#include <condition_variable>
+
+using std::vector;
+
+class ReadGain : public kotekan::Stage {
```
andrerenard: Needs standard comments

lib/hsa/hsaAsyncCopyGain.cpp
```diff
@@ -0,0 +1,105 @@
...
+        // Check for new gains only if filled all gpu frames (not currently filling frame)
+        if (!filling_frame) {
+            auto timeout = double_to_ts(0);
+            int status = wait_for_full_frame_timeout(gain_buf, unique_name.c_str(),
+                                                     gain_buf_precondition_id, timeout);
+            DEBUG("status of gain_buf_precondition_id[{:d}]={:d} ==(0=ready 1=not)================",
+                  gain_buf_precondition_id, status);
+            if (status == 0) {
+                filling_frame = true;
+                frame_to_fill = gain_buf->num_frames;
```
andrerenard: You just need to fill all the GPU frames. 
```suggestion
                frame_to_fill = device.get_gpu_buffer_depth();
```

lib/hsa/hsaPulsarUpdatePhase.cpp
```diff
@@ -129,6 +114,41 @@ int hsaPulsarUpdatePhase::wait_on_precondition(int gpu_frame_id) {
...
+            return -1;
+    } else {
+        // Check for new gains only if filled all gpu frames (not currently filling frame)
+        if (!filling_frame) {
+            auto timeout = double_to_ts(0);
+            int status = wait_for_full_frame_timeout(gain_buf, unique_name.c_str(),
+                                                     gain_buf_precondition_id, timeout);
+            DEBUG("status of gain_buf_precondition_id[{:d}]={:d} ==(0=ready 1=not)",
+                  gain_buf_precondition_id, status);
+            if (status == 0) {
```
andrerenard: I don't think you are doing the right thing here.  You always copy in gains*pointings in this command object, so you don't need to copy the gains to the GPU.  All you need to do is update the gains used to the compute the pointings and then trigger an "update" to have it compute new pointings with the updated gains and switch banks.

lib/hsa/hsaAsyncCopyGain.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+    // Check for new gains
+    DEBUG("Waiting for gain_buf_id={:d} to be full; gpu_frame_id={:d}", gain_buf_precondition_id,
+          gpu_frame_id);
+    if (first_pass) {
+        uint8_t* frame =
+            wait_for_full_frame(gain_buf, unique_name.c_str(), gain_buf_precondition_id);
+        gain_buf_precondition_id = (gain_buf_precondition_id + 1) % gain_buf->num_frames;
+        first_pass = false;
+        frame_to_fill = gain_buf->num_frames;
```
andrerenard: This should be the number of frames on the GPU, not the input buffer. 
```suggestion
        frame_to_fill = device.get_gpu_buffer_depth();
```

lib/hsa/hsaAsyncCopyGain.cpp
```diff
@@ -0,0 +1,105 @@
...
+        frame_to_fill_finalize = frame_to_fill;
+        filling_frame = true;
+        if (frame == NULL)
```
andrerenard: This should happen right after the call to `wait_for_full_frame `

lib/hsa/hsaBeamformKernel.hpp
```diff
@@ -150,17 +143,12 @@ class hsaBeamformKernel : public hsaCommand {
     int32_t freq_idx;
     /// Freq in MHz
     float freq_MHz;
-
+
```
andrerenard: Extra space.  You might need to run `clang-format` again. 
```suggestion

```

lib/hsa/hsaPulsarUpdatePhase.cpp
```diff
@@ -129,6 +109,36 @@ int hsaPulsarUpdatePhase::wait_on_precondition(int gpu_frame_id) {
...
+    DEBUG("Waiting for gain_buf_id={:d} to be full; gpu_frame_id={:d}", gain_buf_id,
+          gpu_frame_id);
+    if (first_pass) {
+        uint8_t* frame =
+	  wait_for_full_frame(gain_buf, unique_name.c_str(), gain_buf_id);
+        host_gain = (float*)gain_buf->frames[gain_buf_id];
+	update_phase = true;
+        mark_frame_empty(gain_buf, unique_name.c_str(), gain_buf_id);
+	gain_buf_id = (gain_buf_id + 1) % gain_buf->num_frames;		
+        if (frame == NULL)
```
andrerenard: This should right happen after the call to `wait_for_full_frame ()`

lib/hsa/hsaPulsarUpdatePhase.cpp
```diff
@@ -129,6 +109,36 @@ int hsaPulsarUpdatePhase::wait_on_precondition(int gpu_frame_id) {
...
+            return -1;
+    } else {
+      auto timeout = double_to_ts(0);
+      int status = wait_for_full_frame_timeout(gain_buf, unique_name.c_str(),
+					       gain_buf_id, timeout);
+      DEBUG("status of gain_buf_id[{:d}]={:d} ==(0=ready 1=not)",
+	    gain_buf_id, status);
+      if (status == 0) {
+	host_gain = (float*)gain_buf->frames[gain_buf_id];
```
andrerenard: This needs to be a `memcpy()` from `gain_buf->frames[gain_buf_id]` to `host_gain` since otherwise you'll have a memory leak and possible race condition.

lib/hsa/hsaPulsarUpdatePhase.cpp
```diff
@@ -129,6 +109,36 @@ int hsaPulsarUpdatePhase::wait_on_precondition(int gpu_frame_id) {
...
+	gain_buf_id = (gain_buf_id + 1) % gain_buf->num_frames;		
+        if (frame == NULL)
+            return -1;
+    } else {
+      auto timeout = double_to_ts(0);
+      int status = wait_for_full_frame_timeout(gain_buf, unique_name.c_str(),
+					       gain_buf_id, timeout);
+      DEBUG("status of gain_buf_id[{:d}]={:d} ==(0=ready 1=not)",
+	    gain_buf_id, status);
+      if (status == 0) {
```
andrerenard: You should add `std::lock_guard<std::mutex> lock(_pulsar_lock);` after the `if (status ==0)` so that you don't update the gain at the same time as a phase calculation is happening.

lib/stages/ReadGain.cpp
```diff
@@ -0,0 +1,240 @@
...
+}
+
+void ReadGain::main_thread() {
+
+    if (first_pass) {
+        first_pass = false;
+        uint8_t* frame =
+            wait_for_full_frame(metadata_buf, unique_name.c_str(), metadata_buffer_precondition_id);
+        if (frame == NULL)
+            goto end_loop;
```
andrerenard: Avoid `goto` unless it's needed.
```suggestion
            return;
```

lib/stages/ReadGain.cpp
```diff
@@ -0,0 +1,240 @@
...
+    } catch (std::exception const& e) {
+        WARN("[PSR] Fail to read gain_dir {:s}", e.what());
+        return false;
+    }
+    return true;
+}
+
+void ReadGain::main_thread() {
+
+    if (first_pass) {
```
andrerenard: You don't need this `if (first_pass)`, this part of the `main_thread()` will only get called once since it's before the while loop.

lib/stages/ReadGain.cpp
```diff
@@ -0,0 +1,240 @@
...
+    unregister_consumer(metadata_buf, unique_name.c_str());
+
+    while (!stop_thread) {
+        DEBUG("update_frb_gains={:d} update_psr_gains={:d} (0=false 1=true)", update_gains_frb,
+              update_gains_psr);
+
+        // FRB================================================
+        {
+            std::unique_lock<std::mutex> lock(mux);
+            while (!update_gains_frb) {
```
andrerenard: The way this is structured is problematic.  Right now if the pulsar gains are updated, but you are waiting in this loop, then you'll won't update the pulsar gains until the FRB gains are updated to break out of this loop and go to the next waiting condition.    There are two solutions here: 1) you can break this into two `stages`, or 2) make the condition variable sleep on `update_gains_frb` and `update_gains_psr`, and then break the loop if either are true, then have the updates below only happen if `update_gains_frb` or `update_gains_psr` are true.  If you do 2) I would also suggest breaking the gain file read logic into two functions for FRB and Pulsar just to make it easier to follow the logic.    Also I think 2 is likely less work to implement.

lib/stages/ReadGain.cpp
```diff
@@ -0,0 +1,240 @@
...
+                cond_var.wait(lock);
+            }
+        }
+        DEBUG("Going to update frb gain update_gains={:d}", update_gains_frb);
+
+        float* out_frame_frb =
+            (float*)wait_for_empty_frame(gain_frb_buf, unique_name.c_str(), gain_frb_buf_id);
+        if (out_frame_frb == NULL) {
+            goto end_loop;
```
andrerenard: There is only one level of loop here, so `break;` will work. 
```suggestion
            break;
```

lib/stages/ReadGain.cpp
```diff
@@ -0,0 +1,240 @@
...
+            while (!update_gains_psr) {
+                cond_var.wait(lock);
+            }
+        }
+        DEBUG("Going to update psr gain update_gains={:d}", update_gains_psr);
+
+        float* out_frame_psr =
+            (float*)wait_for_empty_frame(gain_psr_buf, unique_name.c_str(), gain_psr_buf_id);
+        if (out_frame_psr == NULL) {
+            goto end_loop;
```
andrerenard:
```suggestion
            goto break;
```

lib/stages/ReadGain.cpp
```diff
@@ -0,0 +1,240 @@
...
+            std::unique_lock<std::mutex> lock(mux);
+            while (!update_gains_frb) {
+                cond_var.wait(lock);
+            }
+        }
+        DEBUG("Going to update frb gain update_gains={:d}", update_gains_frb);
+
+        float* out_frame_frb =
+            (float*)wait_for_empty_frame(gain_frb_buf, unique_name.c_str(), gain_frb_buf_id);
+        if (out_frame_frb == NULL) {
```
andrerenard: Not hugely important, but modern C++ should use `nullptr` not the old C `NULL`.
```suggestion
        if (out_frame_frb == nullptr) {
```

lib/stages/ReadGain.hpp
```diff
@@ -0,0 +1,102 @@
...
+ *     @buffer_format Array of @c floats
+ *     @buffer_metadata none
+ *
+ * @conf   num_elements         Int (default 2048). Number of elements
+ * @conf   num_pulsar           Int (default 10). Number of pulsar beams to be formed
+ * @conf   scaling              Float (default 1.0). Scaling factor on gains
+ * @conf   default_gains        Float array (default 1+1j). Default gain value if gain file is
+ * missing
+ *
+ * The gain path is registered as a subscriber to an updatable config block.
```
andrerenard: Add some details about the updatable config blocks this code expects.

lib/stages/ReadGain.hpp
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * @author Cherry Ng
+ *
+ */
+
```
andrerenard: Remove this space, it might cause issues with doxygen associating the comments to the code block.

lib/stages/ReadGain.cpp
```diff
@@ -0,0 +1,230 @@
...
+    } catch (std::exception& e) {
+        WARN("[FRB] Fail to read gain_dir {:s}", e.what());
+        return false;
+    }
+    return true;
+}
+
+bool ReadGain::update_gains_psr_callback(nlohmann::json& json) {
+    {
+        std::lock_guard<std::mutex> lock(mux);
```
andrerenard: Same problem here as above.    And actually there is another problem here.  If a gain update is called while `update_gains_psr == true` then it's possible you'll never update the gains with the latest file since in the main thread you'll set `update_gains_psr = false` before looping back.    There are three ways to fix this: 1) check if `update_gains_psr == true` and if so reject the update with a warning you cannot handle two updates at once 2) lock the entire `read_gain_psr(); update_gains_psr = false;` sections, or 3) have an update queue.   I think 1) might be the best bet for now.  I don't think we'll ever want to send updates that fast, and it's better than 2) which locks up the REST server for a while, and 3) is a bit more work...

lib/hsa/hsaPulsarUpdatePhase.cpp
cherryng: i have put the warning in `update_gains_frb_callback` and `update_gains_psr_callback`. is that OK?

lib/stages/ReadGain.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+    mark_frame_empty(metadata_buf, unique_name.c_str(), metadata_buffer_id);
+    metadata_buffer_id = (metadata_buffer_id + 1) % metadata_buf->num_frames;
+    unregister_consumer(metadata_buf, unique_name.c_str());
+
+    while (!stop_thread) {
+
+        {
+            std::unique_lock<std::mutex> lock(mux);
+            while (!update_gains_frb and !update_gains_psr) {
```
andrerenard: Most people use `&&` not `and`, and we should be consistent in this code base and use `&&`
```suggestion
            while (!update_gains_frb && !update_gains_psr) {
```

lib/stages/ReadGain.cpp
```diff
@@ -0,0 +1,240 @@
...
+    string gainpsr = config.get<std::string>(unique_name, "updatable_config/gain_psr");
+    if (gainpsr.length() > 0)
+        configUpdater::instance().subscribe(
+            gainpsr, std::bind(&ReadGain::update_gains_psr_callback, this, _1));
+}
+
+bool ReadGain::update_gains_frb_callback(nlohmann::json& json) {
+    if (update_gains_frb) {
+        WARN("[FRB] cannot handle two back-to-back gain updates, rejecting the latter");
+        return false;
```
andrerenard: Note that returning `false` here will exit kotekan.  I think given that we've listed this an a `WARN` condition.  We either need to change it for `ERROR` if we want this to kill the system, or return true, and just not apply the gains but keep running.   My impression is we don't want to cause instability, so in this case we should do the slightly bad thing and just return `true`, so we don't impact stability.   However we should be careful we don't get into a situation where it is possible to hit the condition and miss an update.

lib/stages/ReadGain.cpp
```diff
@@ -0,0 +1,240 @@
...
+    }
+    try {
+        _gain_dir_frb = json.at("frb_gain_dir");
+    } catch (std::exception& e) {
+        WARN("[FRB] Fail to read gain_dir {:s}", e.what());
+        return false;
```
andrerenard: Same here.

PR #519, opened 2019-09-17
Redesign of baseband processing flow and synchronization (fixes #497)

### LINKED ISSUE #497: Rework baseband request management
Currently, `basebandReadoutManager` holds a queue of baseband requests, with a pointer to the next request waiting to be read-out and another pointing to the next request ready to be written out. By design, only one request can be processed at a time: there is a single data buffer area that is used by both readout and writeout threads, and while the readout thread is preparing the baseband data into this buffer, the writeout thread is waiting for it to complete. Similarly, the readout thread cannot process the next request until the writeout thread has completed writing the data into an HDF5 file.

The current design also seems to contain a concurrency bug (#457, only seen on the Pathfinder), where the readout thread gets past the waiting request list instead of blocking. (basebandReadoutManager:44).

Instead, I propose:
1. share the data area for multiple events (as long as there is space), by making it a bipartite circular buffer (BipBuffer)
2. the readout thread has a queue of events waiting to be processed, and reads out their baseband data into a segment of the BipBuffer if there is one available. (If there isn't, the event is errored-out and the thread moves on to the next request in its queue.)
3. when the readout thread is done with an event, it adds a write request to the writeout thread's queue. This request includes the pointer to the event's data in the BipBuffer.
4. the writeout thread processes requests in the write queue, creating the HDF file with data from the BipBuffer, and marking that data segment available when it's done.
5. there is also a list of basebandRequestStatus that is the history of all requests since Kotekan started, and is used to return the request status data by the REST API (GET to `/baseband/event-id`). Elements in the readout and writeout queues include pointers to elements in this list, so that the readout and writeout threads can update their `state`.

cubranic:
Makes baseband dump stages operate in parallel with each other, rather than blocking at each processing step of a single baseband request.

- readout data area is shared by multiple events (as long as there is space), by making it a bipartite circular buffer (BipBuffer)
- the readout thread has a queue of requests `waiting` to be processed, and reads out their baseband data into a segment of the BipBuffer if there is one available. (If there isn't, the request is errored-out and the thread moves on to the next request in its queue.)
- when the readout thread is done with a requests, it adds it to the writeout thread's `ready` queue. This request includes the pointer to the request's data in the BipBuffer.
- the writeout thread processes requests in the `ready` queue, creating the HDF file with data from the BipBuffer, and marking that data segment available when it's done.
- there is also a list of basebandRequestStatus that is the history of all requests since Kotekan started, and is used to return the request status data by the REST API (GET to /baseband/event-id). Elements in the `waiting` and `ready` queues include pointers to elements in this list, so that the readout and writeout threads can update their state

### DIFF
--- a/lib/stages/basebandReadout.cpp
+++ b/lib/stages/basebandReadout.cpp
@@ -23,6 +23,7 @@
 
 
 using kotekan::basebandApiManager;
+using kotekan::basebandDumpData;
 using kotekan::basebandDumpStatus;
 using kotekan::basebandReadoutManager;
 using kotekan::basebandRequest;
@@ -47,7 +48,7 @@ basebandReadout::basebandReadout(Config& config, const string& unique_name,
     oldest_frame(-1),
     frame_locks(_num_frames_buffer),
     // Over allocate so we can align the memory.
-    baseband_data(std::make_unique<uint8_t[]>(_num_elements * _max_dump_samples + 15)),
+    data_buffer(_num_elements * _max_dump_samples + 15),
     readout_counter(kotekan::prometheus::Metrics::instance().add_counter(
         "kotekan_baseband_readout_total", unique_name, {"freq_id", "status"})),
     readout_in_progress_metric(kotekan::prometheus::Metrics::instance().add_gauge(
@@ -99,6 +100,7 @@ void basebandReadout::main_thread() {
     std::unique_ptr<std::thread> wt;
     std::unique_ptr<std::thread> lt;
 
+    basebandReadoutManager* mgr = nullptr;
     while (!stop_thread) {
 
         if (wait_for_full_frame(buf, unique_name.c_str(), frame_id % buf->num_frames) == nullptr) {
@@ -119,11 +121,10 @@ void basebandReadout::main_thread() {
             readout_in_progress_metric.labels({std::to_string(freq_id)}).set(0);
 
             INFO("Starting request-listening thread for freq_id: {:d}", freq_id);
-            basebandReadoutManager& mgr =
-                basebandApiManager::instance().register_readout_stage(freq_id);
-            lt = std::make_unique<std::thread>([&] { this->listen_thread(freq_id, mgr); });
+            mgr = &basebandApiManager::instance().register_readout_stage(freq_id);
+            lt = std::make_unique<std::thread>([&] { this->listen_thread(freq_id, *mgr); });
 
-            wt = std::make_unique<std::thread>([&] { this->write_thread(mgr); });
+            wt = std::make_unique<std::thread>([&] { this->write_thread(*mgr); });
         }
 
         int done_frame = add_replace_frame(frame_id);
@@ -133,7 +134,9 @@ void basebandReadout::main_thread() {
 
         frame_id++;
     }
-    ready_to_write.notify_all();
+    if (mgr) {
+        mgr->stop();
+    }
 
     if (lt) {
         lt->join();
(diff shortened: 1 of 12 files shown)


lib/stages/basebandReadout.cpp
```diff
@@ -47,7 +48,7 @@ basebandReadout::basebandReadout(Config& config, const string& unique_name,
     oldest_frame(-1),
     frame_locks(_num_frames_buffer),
     // Over allocate so we can align the memory.
-    baseband_data(std::make_unique<uint8_t[]>(_num_elements * _max_dump_samples + 15)),
+    data_buffer(_num_elements * _max_dump_samples + 15),
```
kiyo-masui: This should be `2 * (_num_elements * _max_dump_samples + 15)` right? Because we can't guarantee that a contiguous piece of memory bigger than the length / 2 will be available (even with no writes queued).

lib/stages/basebandReadout.cpp
```diff
@@ -212,54 +207,69 @@ void basebandReadout::listen_thread(const uint32_t freq_id, basebandReadoutManag
...
+        BipBufferReader reader(data_buffer);
+        auto rr = reader.access(dump_data.reservation_length);
+        if (!rr) {
+            ERROR("No matching buffer reader reservation for the baseband dump data");
+            throw std::runtime_error("Not able to get reader reservation");
+        }
+
+        // Sanity check that the read reservation points to the same memory used by the writer
+        DEBUG("Read reservation: starting at {}, length {}",
+              rr->data.data() - data_buffer.data.get(), rr->length);
```
kiyo-masui: Does this account for the possible up-to 15 byte offset of the span compared to the reservation?

lib/stages/basebandReadout.cpp
kiyo-masui: I think this should be `2 * (_num_elements * _max_dump_samples + 15)`, since even if there are no writes queued you can't guarantee that more than length/2 of contiguous memory is available.

cubranic:
@kiyo-masui re: write reservation vs read reservation

The reason why both of them exist is that they are used to move two pointers in the buffer structure safely, normally by a producer and a consumer thread. I based this on the design described in this blog post: https://andrea.lattuada.me/blog/2019/the-design-and-implementation-of-a-lock-free-ring-buffer-with-contiguous-reservations.html

The sanity checks are not normally required, I just put them in to catch any bugs in my implementation of BipBuffer and exit Kotekan, rather than silently work with the wrong data. I hope this entire section will be removed once we have more confidence in the code:
```
        // Sanity check that the read reservation points to the same memory used by the writer
        DEBUG("Read reservation: starting at {}, length {}",
              rr->data.data() - data_buffer.data.get(), rr->length);
        // Ugly workaround to lose the read-only view of the ReadReservation, for consumption by
        // `span_from_length_aligned`
        auto reader_data_span = basebandDumpData::span_from_length_aligned(
            gsl::span<uint8_t>((uint8_t*)rr->data.data(), rr->data.size()));
        DEBUG("Read data starts alignment at: {}",
              reader_data_span.data() - data_buffer.data.get());
        if (reader_data_span != dump_data.data) {
            ERROR("Baseband dump data doesn't have matching buffer writer and reader reservations "
                  "after alignment: writer ({}, {}); reader ({}, {})",
                  dump_data.data.data(), dump_data.data.length(), reader_data_span.data(),
                  reader_data_span.length());
            throw std::runtime_error("reader and writer data spans don't align.");
        }
```

lib/stages/basebandReadout.cpp
cubranic: Agreed. I changed it and pushed the updated version.

lib/stages/basebandReadout.cpp
cubranic: Reservation includes the alignment offset, that's why the sanity check has to align the read reservation before comparing it to the `data` pointer in the `basebandDumpData` structure.

lib/utils/SynchronizedQueue.hpp
```diff
@@ -0,0 +1,94 @@
+#ifndef SYNCHRONIZED_QUEUE_HPP
```
andrerenard: Needs an `@file` block for doxygen.

andrerenard:
Oh and looks like this needs to be updated with the latest develop to resolve the merge conflict.

lib/utils/SynchronizedQueue.hpp
cubranic: Done in 345f99ba101a6580d76c70f1f6d03a922eca3352

PR #524, opened 2019-09-23
[WIP] Frame view for HFB data

james-s-willis:
Could you review my `hfbFrameView` and `hfbFileRaw`? @andrerenard @jrs65 

I copied `visFrameView` and removed fields not relevant to HFB data. I did the same with `visFileRaw`.


jrs65:
Thanks @james-s-willis I should be able to have a look over it later today.

james-s-willis:
Can you review this, @jrs65 @andrerenard? So I can go ahead and create an abstract `frameView` class. I just need to identify the commonality between the `vis` and `hfb` frames. I also need to work out how it will fit in with the `visFile` class, I think we should have another call to discuss this.

lib/utils/hfbBuffer.hpp
```diff
@@ -0,0 +1,272 @@
...
+
+    /// ID of the frequency bin
+    uint32_t freq_id;
+    /// ID of the dataset (vis, gatedvisX ...), main vis dataset = 0
+    uint64_t dataset_id;
+
+    /// Number of elements for data in buffer
+    uint32_t num_elements;
+    /// Number of products for data in buffer
+    uint32_t num_prod;
```
jrs65: I would suggest breaking this one down into `num_beams` and `num_subfreq` (or similar) parameters. I can see it being very useful having that information available even if you don't directly need it to decode the frame.

lib/utils/hfbBuffer.hpp
```diff
@@ -0,0 +1,272 @@
...
+    /// The ctime of the integration frame
+    timespec ctime;
+    /// Nominal length of the frame in FPGA ticks
+    uint64_t fpga_seq_length;
+    // Amount of data that actually went into the frame (in FPGA ticks)
+    uint64_t fpga_seq_total;
+
+    /// ID of the frequency bin
+    uint32_t freq_id;
+    /// ID of the dataset (vis, gatedvisX ...), main vis dataset = 0
```
jrs65: I definitely think you should keep the dataset_id in here, but you probably want to update the comment which is fairly vis data specific.

lib/utils/hfbBuffer.cpp
```diff
@@ -0,0 +1,248 @@
+#include "hfbBuffer.hpp"
+
+#include "gpsTime.h"
+
+#include "fmt.hpp"
+
+#include <set>
+
+
+template<typename T>
```
jrs65: These two are obviously shared with `visFrameView` and so we should move to a common location. It sounds like you're on top of that one though.

Issue #509, opened 2019-09-11
Test RFI system stability


jrs65:
Full system test and review all RFI endpoints for any critical section problems.

andrerenard:
This is basically done, and so far it looks good.  The only test left is to see how it performs in a solar transit with zeroing _disabled_

PR #528, opened 2019-09-25
(Re)Enable RFI system

### LINKED ISSUE #479: Extract variance from RFI system
We want to be able to output the variance of one feed along with the SK value dump in the RFI system.   This will help us evaluate the performance of the RFI system, and tune the sigma cut. 

Ideally we'll do this before the `2019.09` release as a few hour test in the middle of this current run.

### LINKED ISSUE #486: RFI system: reorder bad input list
The RFI system will now get the bad input list in cylinder order, when it needs to be in FPGA input order.  So we need to map from cylinder order to FPGA order when we get the bad input list.

### LINKED ISSUE #506: Turn on RFI kernels
... and zeroing.

This is a config update.

### LINKED ISSUE #508: Ensure the coco endpoints exist for RFI system
We need to ensure the following endpoints exist in kotekan and have them forwarded to from coco

- [x] bad-input-list
- [x] enable/disable zeroing
- [x] enable/disable SK + variance dump
- [x] change which feed variance is dumped for

andrerenard:
Master PR for all RFI related changes to be able to re-enable it. 

Closes #506 - enable RFI kernels 
Closes #486 - reorder bad input list
~Closes #479 - extract variance from RFI system~ This was removed because of a strange issue that caused it to corrupt the N2 data. 
Closes #508 - coco endpoints (the kotekan side updates)
Needed for #509 - test stability of RFI system


config/chime_science_run_gpu.yaml
```diff
@@ -962,6 +978,26 @@ updatable_config:
       ra: 53.6197236741
       dec: 54.6433973569
       scaling: 48
+  bad_inputs:
```
nritsche: Does this not need the additional parameters that are forwarded to receiver kotekan?
```
tag: str
start_time: float
```

andrerenard: It does not currently use those (although maybe it should at some point).  However if we want it to match the receiver kotekan then I can add them?

nritsche: Aren't the `bad_inputs` sent meant to be applied in the future?

config/chime_rfi.yaml
```diff
@@ -263,55 +280,59 @@ gpu:
     updatable_rfi_zeroing: /rfi_masking/toggle
```
nritsche: Why is this called toggle if it gets a boolean?

andrerenard: oh it looks like `chime_rfi.yaml` is out of data... for now just look at `chime_science_run_gpu.yaml`.   I think you can have a "toggle" as a boolean.  According to google define:  `2. COMPUTING a key or command that toggles between two modes, such as the caps lock key.`

config/chime_science_run_gpu.yaml
andrerenard: In the RFI system they have always just been applied as they become available.  In other systems they are applied in the future since we want them to be synchronized across the array.  But it this case the bad inputs are just trying to improve the statistics of the RFI SK estimator for a given frequency, so synchronization isn't as important as reacting quickly to bad feed events which might cause an entire frequency to otherwise become zeroed.

andrerenard:
I think this branch is ready to merge.   However I'm going to hold off merging it until it's been tested on the entire cluster to check for stability on more than one node.

PR #530, opened 2019-09-25
Use random RESTserver port to parallelize tests

nritsche:


nritsche:
Once this is in develop, we should make sure Jenkins is doing 2 checks in parallel, where each check includes the builds (make runs with -j4) and the unit tests. There are 3 parallel stages for the unit tests:
- the boost tests (fast)
- the pytest unit tests running with `-n 6`
- test_vistestpattern (can't parallelize with pytest, so treated seperately)

nritsche:
I left this Jenkinsdebuggingsausage here. Please squash-merge.
```
- feat(visTestpattern): option to disable reporting back
- build(version.py): improve error handling
- build(cmake): use python3 to call version script
- fix(version.py): add verbosity to error prints
- ci(jenkins): cancel parallel checks in failure case
- feat(ci): run tests in parallel
- feat(restServer): option to use random free port
```

nritsche:
@andrerenard this is down to ~9min and ready for review

nritsche:
The `expect_failure` parameter was added by ea99f7193898182da1f7772e261ef4952bf81fdb
The reason was that the test was failing. That change basically disabled the test, but to a high cost: it was always timing out, making it take about 7min on its own.

PR #531, opened 2019-09-26
feat: rate limiting for the visRawReader

jrs65:
This adds the option for limiting the read rate for visRawReader to try
and reduce issues with gossec over NFS. It also adds the ability to
create VisRaw style files from Python space for testing.

### DIFF
--- a/lib/stages/visRawReader.cpp
+++ b/lib/stages/visRawReader.cpp
@@ -28,6 +28,7 @@
 #include <stdexcept>
 #include <sys/mman.h>
 #include <sys/stat.h>
+#include <time.h>
 #include <unistd.h>
 
 
@@ -43,6 +44,7 @@ visRawReader::visRawReader(Config& config, const string& unique_name,
 
     filename = config.get<std::string>(unique_name, "infile");
     readahead_blocks = config.get<size_t>(unique_name, "readahead_blocks");
+    max_read_rate = config.get_default<double>(unique_name, "max_read_rate", 0.0);
 
     chunked = config.exists(unique_name, "chunk_size");
     if (chunked) {
@@ -226,6 +228,7 @@ int visRawReader::position_map(int ind) {
 
 void visRawReader::main_thread() {
 
+    double start_time, end_time;
     size_t frame_id = 0;
     uint8_t* frame;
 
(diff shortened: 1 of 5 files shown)


lib/stages/visRawReader.cpp
```diff
@@ -296,5 +308,15 @@ void visRawReader::main_thread() {
...
         ind++;
+
+        // Get the end time for the loop and sleep for long enough to satisfy
+        // the max rate
+        end_time = current_time();
+        double sleep_time = min_read_time - (end_time - start_time);
+        DEBUG("Sleep time {}", sleep_time);
+        if (sleep_time > 0) {
+            auto ts = double_to_ts(sleep_time);
+            nanosleep(&ts, NULL);
```
nritsche: I guess the sleep times end up so short, that `usleep` wouldn't be precise enough here?

python/kotekan/visbuffer.py
```diff
@@ -21,6 +21,13 @@ class time_spec(ctypes.Structure):
 
     _fields_ = [("tv", ctypes.c_int64), ("tv_nsec", ctypes.c_uint64)]
 
+    @classmethod
+    def from_float(cls, v):
```
nritsche: add documentation
e.g.
```
Create a time_spec from a float.

Parameters
----------------
v : float
    The interval in seconds.
```

python/kotekan/visbuffer.py
```diff
@@ -365,6 +374,160 @@ def __init__(self, filename, mmap=False):
...
+    """
+
+    nprod = ninput * (ninput + 1) // 2
+    nev = 4
+
+    time = [{"ctime": (10.0 * i), "fpga_count": i} for i in range(ntime)]
+
+    freq = [{"centre": (800 - i * 10.0), "width": 10.0} for i in range(10)]
+
+    input_ = [(i, f"test{i:04}") for i in range(ninput)]
```
nritsche: This only works for python >= 3.7.
```suggestion
    input_ = [(i, "test{:04}".format(i)) for i in range(ninput)]
```

lib/stages/visRawReader.cpp
jrs65: Nah, I don't think they're that short, but as it was easy enough for me to get a timespec I figured why not.

PR #542, opened 2019-09-30
DPDK: Check for correct cabling

### LINKED ISSUE #487: Incorrectly cabled nodes fail to exit
There appears to be a serious problem where it is possible for a node to be cabled incorrectly in a way that won't cause kotekan to shutdown and warn about it.   To make matters worse this bug also leads the `StreamID` to get set to `[0,0,0,0]` or `bin = 0` for all the data it is generating, possibly overwriting or duplicating existing good data for `bin = 0`.

andrerenard:
Checks if the cabling (streamIDs) meets the expected pattern for a correct shuffle.  This should resolve the bug in the last run where the system created duplicate frequency products. 

Closes #487

### DIFF
--- a/lib/dpdk/iceBoardShuffle.hpp
+++ b/lib/dpdk/iceBoardShuffle.hpp
@@ -75,6 +75,13 @@ protected:
     bool advance_frames(uint64_t new_seq, bool first_time = false);
 
     /**
+     * @brief Checks that the rules for streamIDs are met.  i.e. correct cabling.
+     *
+     * @return True if cable/streamID rules are met, False otherwise.
+     */
+    bool check_stream_id();
+
+    /**
      * @brief Copies the given packet accounting for the last stage suffle.
      *
      * This means it copies the packet into 4 buffer frames, and can advance
@@ -125,6 +132,11 @@ protected:
 
     /// Frame IDs
     int out_buf_frame_ids[shuffle_size] = {0};
+
+    /// The stream_ids for all iceBoardShuffle objects.
+    /// This might be an issue in the case of multiple indepdent
+    /// shuffle operations. In that case this will need to be factored out.
+    static stream_id_t all_stream_ids[shuffle_size];
 
     // ** FPGA Second stage error counters **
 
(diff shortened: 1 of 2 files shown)


andrerenard:
Note this needs a little more testing before I merge it in.

lib/dpdk/iceBoardShuffle.hpp
```diff
@@ -308,6 +326,46 @@ inline int iceBoardShuffle::handle_packet(struct rte_mbuf* mbuf) {
...
+        if (i == port || all_stream_ids[i].crate_id == 255)
+            continue;
+
+        // Check that all the slots and links are the same.
+        if (all_stream_ids[i].slot_id != slot_id || all_stream_ids[i].link_id != link_id) {
+            FATAL_ERROR("One of the link_ids or slot_ids don't match! There is a cabling problem.");
+            return false;
+        }
+
+        // Check that we don't have the same create ID as another link
```
james-s-willis: Typo.

lib/dpdk/iceBoardShuffle.hpp
```diff
@@ -308,6 +326,46 @@ inline int iceBoardShuffle::handle_packet(struct rte_mbuf* mbuf) {
...
+        // This should be impossible unless there is an FPGA problem
+        if (all_stream_ids[i].crate_id == crate_id) {
+            FATAL_ERROR("Two of the crate_ids are the same! There is a cabling problem.");
+            return false;
+        }
+
+        // Check that all the creates are from the same group (all even/odd)
```
james-s-willis: Typo.

lib/dpdk/iceBoardShuffle.hpp
```diff
@@ -308,6 +326,46 @@ inline int iceBoardShuffle::handle_packet(struct rte_mbuf* mbuf) {
...
+        if (even != ((all_stream_ids[i].crate_id % 2) == 0)) {
+            FATAL_ERROR("The create IDs are not all even or all odd. There is a cabling problem.");
```
james-s-willis:
```suggestion
            FATAL_ERROR("The crate IDs are not all even or all odd. There is a cabling problem.");
```

PR #543, opened 2019-09-30
Subtract weight bias and accumulate RFI

jrs65:
This PR adds a number of things. The headline changes are that:

- It calculates and subtracts the bias in the weights calculated in visAccumulate. This has tests to check it is working correctly.
- It adds space for and accumulates the `rfi_total` within `visMetadata`

There are a number of ancillary things included in this one:

- It merges in the refactor of `FakeGpu` that was languishing and adds some new test patterns.
- It refactors the internals of `visAccumulate` to move the multifrequency processing out of the main loop, and moves all of the state within the individual `internalState` objects.
- It adds tests for the multifrequency buffer support (@andrerenard)


lib/stages/visAccumulate.cpp
```diff
@@ -431,32 +426,48 @@ void visAccumulate::main_thread() {
...
+        auto& frame = state.frames[freq_ind];
+
+        // Copy over the metadata
+        // TODO: CHIME
+        frame.fill_chime_metadata(metadata);
 
-    // TODO: set frequency id in some sensible generic manner
-    frame.freq_id += freq_ind;
+        // TODO: set frequency id in some sensible generic manner
+        frame.freq_id += freq_ind;
```
andrerenard: I think this line will cause problems at the moment.   In the 16-element case I had to hardcoded this to `bin_number_16_elem(stream_id)` if `num_freq_in_frame == 128` and `bin_number_chime(stread_id)` if `num_freq_in_frame == 1` since I never got around to doing the `sensible generic` thing.   It looks like the `fill_chime_metadata()` function is setting `frame.freq_id = bin_number_chime(stream_id)`, which does work in the CHIME case, but then this multi-freq code will just overwrite it with `0`.

lib/stages/visAccumulate.cpp
```diff
@@ -361,19 +344,27 @@ void visAccumulate::main_thread() {
...
-            long samples_in_frame =
-                samples_per_data_set - get_lost_timesamples(in_buf, in_frame_id);
+            int32_t samples_in_frame = samples_per_data_set - lost_in_frame;
 
             // Accumulate the weighted data into each dataset. At the moment this
             // doesn't really work if there are multiple frequencies in the same buffer..
             for (internalState& dset : enabled_gated_datasets) {
 
+                // TODO: CHIME specific frequency decoding
+                float freq_in_MHz = 800.0 - 400.0 * dset.frames[0].freq_id / 1024.0;
```
andrerenard: There is already a function for this `freq_from_bin()` in `fpag_header_functions.h`

lib/stages/visAccumulate.cpp
jrs65: I don't quite understand your comment here. It's clear that it's not going to label things correctly in the multifrequency case, which I think is fine, some of the CHIME specific stuff just needs replacing to make that work, and I think we both agree on that.

This does work fine for CHIME though, which you seem to suggest it won't. It sets the id in `fill_chime_metadata`, and then *increments* it by zero in the line below.

andrerenard: Ah yes it's `+=` not `=`, so that does work; sorry I misread that.  And yes both agree we need to replace the CHIME specific streamID mapping to make this work in the multi-frequency case.

andrerenard: Although for any system involving an ICEBoard we won't have sequential frequencies.  So when we fix this, it will need to be something which looks up the frequency in some way here.  A simple `+= counter` won't work in any of our existing multi-frequency setups.

jrs65: Kudos for attempting to read it well enough that you flagged up a small detail like this! :)

lib/stages/visAccumulate.cpp
```diff
@@ -483,33 +477,64 @@ void visAccumulate::combine_gated(visAccumulate::internalState& gate,
...
-    map_vis_triangle(input_remap, block_size, num_elements, freq_ind,
-                     [&](int32_t pi, int32_t bi, bool conj) {
-                         (void)conj;
-                         float t = state.vis2[bi];
-                         output_frame.weight[pi] = w * w / t;
-                     });
-
-    // Set the actual amount of time we accumulated for
-    output_frame.fpga_seq_total = total_samples;
+    uint32_t low_sample_cut = low_sample_fraction * num_gpu_frames * samples_per_data_set;
```
nritsche: Is there a reason you moved this calculation into the main loop?

jrs65: I didn't want to store yet another variable in the class, although I guess I could just not store `low_sample_fraction`

jrs65:
Okay all fixed according to suggestions. I'm going to merge now.

PR #544, opened 2019-10-01
Don't free a stack pointer.

andrerenard:


andrerenard:
I'm going to push this in without waiting for code review so @cubranic can get his code running.  @james-s-willis will follow this up with a version that converts this string to an `std::string` so we don't have to deal with freeing in the only some cases.

PR #552, opened 2019-10-02
feat(factory): more verbose exception msg

nritsche:
When trying to create a type that is not known, print the name.

### DIFF
--- a/lib/utils/type.cpp
+++ b/lib/utils/type.cpp
@@ -1,4 +1,5 @@
 #include "type.hpp"
+
 #include <cstdlib>
 #include <cxxabi.h>
 #include <memory>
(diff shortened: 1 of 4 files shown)


jrs65:
Yeah. Something like what's in here. I've done it before and it works pretty nicely: https://stackoverflow.com/questions/281818/unmangling-the-result-of-stdtype-infoname

nritsche:
I added it here, but I'm not sure if it works. Can someone test this?

lib/utils/type.cpp
```diff
@@ -0,0 +1,25 @@
+#include "type.hpp"
+#ifdef __GNUG__
```
jrs65: I'm pretty sure clang supports `__cxa_demangle` so you can probably change the `ifdef` here. I think only Visual Studio doesn't.

PR #555, opened 2019-10-03
Accumulate RFI total in timeDownsample

jrs65:
Also, this adds an option to throw away frames that are too old.

### DIFF
--- a/lib/stages/timeDownsample.cpp
+++ b/lib/stages/timeDownsample.cpp
@@ -2,12 +2,17 @@
 
 #include "chimeMetadata.h"
 #include "errors.h"
+#include "prometheusMetrics.hpp"
 #include "visBuffer.hpp"
 #include "visUtil.hpp"
+
+#include <time.h>
+
 
 using kotekan::bufferContainer;
 using kotekan::Config;
 using kotekan::Stage;
+using kotekan::prometheus::Metrics;
 
 REGISTER_KOTEKAN_STAGE(timeDownsample);
 
@@ -23,6 +28,7 @@ timeDownsample::timeDownsample(Config& config, const string& unique_name,
 
     // Get the number of time samples to combine
     nsamp = config.get_default<int>(unique_name, "num_samples", 2);
+    max_age = config.get_default<float>(unique_name, "max_age", 120.0);
 
     nprod = num_elements * (num_elements + 1) / 2;
 }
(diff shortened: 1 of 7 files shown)


lib/stages/timeDownsample.cpp
```diff
@@ -127,13 +136,21 @@ void timeDownsample::main_thread() {
...
+            output_frame.rfi_total += frame.rfi_total;
 
             // Move to next frame
             nframes += 1;
             mark_frame_empty(in_buf, unique_name.c_str(), frame_id);
             frame_id = (frame_id + 1) % in_buf->num_frames;
 
         } else {
+
+            timespec output_age = std::get<1>(output_frame.time) - std::get<1>(frame.time);
```
tristpinsm: I may just be confused here, but given we ended up in this block (i.e. `fpga_seq_start > wdw_end`) `frame` is from a later time than `output_frame` so this quantity will always be negative and we will never enter the `if` statement. Should there be a separate `else if` for the case where the frame is too old?

jrs65: Ooops. Good point. They should be the other way around!

jrs65: And this is why we do code reviews :)

tristpinsm:
Following up on the timeout case, it means frames that are older than the timeout will get skipped and not trigger the end of the accumulation. Could this potentially lead to a deadlock where the output frame is never marked full?

lib/stages/timeDownsample.cpp
```diff
@@ -127,13 +136,21 @@ void timeDownsample::main_thread() {
...
             nframes += 1;
             mark_frame_empty(in_buf, unique_name.c_str(), frame_id);
             frame_id = (frame_id + 1) % in_buf->num_frames;
 
         } else {
+
+            timespec output_age = std::get<1>(frame.time) - std::get<1>(output_frame.time);
+            if (ts_to_double(output_age) > max_age) {
+                skipped_frame_counter.labels({std::to_string(output_frame.freq_id), "age"}).inc();
+                continue;
```
andrerenard: I might be missing something but are you missing: 
```
mark_frame_empty(in_buf, unique_name.c_str(), frame_id);
frame_id = (frame_id + 1) % in_buf->num_frames;
```
?

jrs65: No. The logic is a little confusing, but I don't need that. I did need a `num_frames = 0` though. Or so I've figured out with Tristan.

PR #556, opened 2019-10-04
Accumulation fix

jrs65:


lib/stages/visAccumulate.cpp
```diff
@@ -280,8 +280,7 @@ void visAccumulate::main_thread() {
             break;
 
         int32_t* input = (int32_t*)in_frame;
-        uint32_t frame_fpga = get_fpga_seq_num(in_buf, in_frame_id);
-        uint32_t frame_count = frame_fpga / samples_per_data_set;
+        uint32_t frame_count = (get_fpga_seq_num(in_buf, in_frame_id) / samples_per_data_set);
```
andrerenard: any reason not to make this a `uint64_t` to avoid this hidden cast?

jrs65: I could do indeed. I'll push that quickly now.

PR #557, opened 2019-10-05
Live ringmaps

tristpinsm:
This probably doesn't need to be in for this deployment, but it will make keeping the live ringmaps up to date easier to have it merged eventually. I don't remember if anything was preventing this from being merged, and I probably need to clean things up to pass the CI.


tristpinsm:
this is passing now, and fairly disconnected from the rest of `kotekan` so this might be a good time to get it in.

lib/stages/ringmap.cpp
```diff
@@ -0,0 +1,586 @@
...
+    auto res = times_map.find(t.ctime);
+    if (res == times_map.end()) {
+        // No entry for this time
+        WARN("Could not find this time in ringmap. Discarding.");
+        return -1;
+    }
+    return res->second;
+}
+
+redundantStack::redundantStack(Config& config, const string& unique_name,
```
jrs65: It looks like this has a large amount in common with the `baselineCompression` stage, and so I think we can probably merge these two. I don't think you need to do much beyond teach that stage how to work from the stack dataset and not just the prod dataset, and then write a new stack type definition function.

lib/stages/ringmap.cpp
```diff
@@ -0,0 +1,586 @@
...
+        stack_ctype s = stacks[i];
+        input_a = chimeFeed::from_input(inputs[prods[s.prod].input_a]);
+        input_b = chimeFeed::from_input(inputs[prods[s.prod].input_b]);
+        ns_baselines[i] = feed_sep * (input_b.feed_location - input_a.feed_location);
+        if (s.conjugate)
+            ns_baselines[i] *= -1;
+    }
+
+    // Construct matrix of phase weights for every baseline and pixel
+    for (auto f : freqs) {
```
jrs65: It looks like you don't do any form of weighting or apodisation of the visibilities before map making, which I think explains the large amount of NS sidelobes that can be seen in the images. We should probably start using something (potentially configurable). It's a reasonably straightforward change you just multiply the baselines axis by some kind of apodization function. A Blackmann window is probably a good choice here as we just want to present clean looking maps for diagnosis, and not analysis.

lib/stages/ringmap.cpp
tristpinsm: That's right. Initially I was hesitating between building this into `baselineCompression` or just emulating it and I went this way to avoid interfering with the ciritical code. But I agree it would make sense to merge them.

lib/stages/ringmap.cpp
tristpinsm: Yeah this would be really easy to implement. I think I will play around in a notebook with some different options to better wrap my head around apodization in the beamforming.

tristpinsm:
created issue #614

jrs65:
Merged. Thanks @tristpinsm !

PR #561, opened 2019-10-10
Re-enable variance extraction

james-s-willis:
Branch that re-enables variance extraction.

*THIS IS NOT INTENDED TO BE MERGED!*


andrerenard:
I'm going to close this, since we have no intention of merging, and don't plan to use this again anytime soon.  We can look back at the github history if we want to implement these changes again some day (after we resolve the N2 bug).

PR #576, opened 2019-10-29
Fix configs

james-s-willis:


andrerenard:
I will merge this without passing CI, since it requires the new CI, and is required for the new CI.

PR #579, opened 2019-10-30
Add a class comment for rawFileRead

cubranic:


lib/stages/rawFileRead.hpp
```diff
@@ -1,3 +1,9 @@
+/**
+ * @file
+ * @brief foo bar
```
nritsche: foo bar

andrerenard:
@cubranic can you quickly address the one comment and/or just accept the suggested change above, and then we'll merge this in.

PR #581, opened 2019-10-31
macOS link fix

jrs65:
Fixes linking errors on macOS Catalina due to completely empty libraries being present, and fixes an issue when yamllint isn't present when using Python 3.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -1,4 +1,8 @@
-cmake_minimum_required( VERSION 2.8 )
+# There are no symbols in this entire library, and so it is created as an
+# INTERFACE library. If any symbols are ever added it will need to be turned
+# back into a standard CMake library
+
+cmake_minimum_required( VERSION 3.0 )
 project( kotekan_libs )
 
 include_directories (${KOTEKAN_SOURCE_DIR}/include)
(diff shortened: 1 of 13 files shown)


lib/CMakeLists.txt
```diff
@@ -1,4 +1,8 @@
-cmake_minimum_required( VERSION 2.8 )
+# There are no symbols in this entire library, and so it is created as an
+# INTERFACE library. If any symbols are ever added it will need to be turned
+# back into a standard CMake library
+
+cmake_minimum_required( VERSION 3.0 )
```
andrerenard: If we are changing the require version, we should change it in all places we make this `cmake_minimum_required()` call, or possibly just remove all of these calls and only set it in the root CMakeLists.txt.  (I think that should work.)

jrs65: Okay. I can do that.

I subscribed to the opposite argument though, that you should set the minimum CMake version required for each individual file. That way you could in theory turn parts off on an older CMake build.

I think moving to a single one at the root level is probably the best option though. I'll do that one now.

jrs65: Done.

jrs65:
Okay. Tests have passed so I'm merging this now.

Issue #583, opened 2019-11-04
remove acqDatasetIdState


nritsche:
Once we have network gossec, remove `acqDatasetIdState`.

PR #584, opened 2019-11-05
Gossec comet

tristpinsm:
In preparation for network gossec, I've integrated `visTranspose` with the dataset broker.

This means that dataset IDs can now change within a raw file. I think it's fair to assume nothing else about the file changes when this happens, right?


lib/stages/visTranspose.cpp
```diff
@@ -247,6 +251,9 @@ void visTranspose::main_thread() {
             frame.fpga_seq_length == 0 ? 0. : float(frame.rfi_total) / frame.fpga_seq_length;
         strided_copy(frame.gain.data(), gain.data(), offset * num_input + ti, write_t, num_input);
 
+        // Store original dataset ID (before adding time axis)
+        dset_id[offset + ti] = base_ds_id;
```
nritsche: but why?

nritsche:
Closes #583

nritsche:
Closes #583

lib/stages/visRawReader.cpp
```diff
@@ -171,28 +172,34 @@ visRawReader::~visRawReader() {
...
-            .first;
-
-    // register it as root dataset
-    _dataset_id = dm.add_dataset(mstate_id);
+    if (use_comet) {
+        // Add time to dataset
+        state_id_t tstate_id = dm.add_state(std::make_unique<timeState>(_times)).first;
+        // Register with dataset broker
+        _dataset_id = dm.add_dataset(tstate_id, ds_id);
+    } else {
```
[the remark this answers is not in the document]
jrs65: I guess my fear is somewhat that if we are going wholeheartedly down the `comet` route then this branch that intends it to allow gossec to work without comet is just going to atrophy if it's too different to the one we use.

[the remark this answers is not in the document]
jrs65: One way of merging them somewhat could be:
- Always register a `timeState` on top of the original dataset_id.
- If comet isn't available then you need to register all the other states too (inputState, prodState, ...), if comet is available you don't need to bother.
- Rewrite the dataset_id for each frame with all these new states in there.
- In `visTranspose` you simply look up the all the information you need from the current state, and then you write the output dataset_id as being the one that the `timeState` was registered against.

I think all put together that means there is a single branch here which adds the extra states when comet is not available, but everything else works fine.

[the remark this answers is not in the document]
tristpinsm: one problem is if comet is not available, you can't register a state on top of the original dataset_id because it doesn't exist as far as the dataset manager is concerned. So I think if we remove the branch that allows it to work without comet entirely we may need to drop that functionality entirely (this would be ok, I just left this in in case we wanted to transpose files manually for some reason)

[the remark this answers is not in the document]
tristpinsm: Right now if you run this without comet it ends up writing 0 to the `dataset_id` dataset in the HDF5 file because the base dataset is the root dataset. We could avoid this if it was possible to reconstruct a dataset state from a dataset ID but at the moment I don't know how you would do this.

lib/stages/visTranspose.cpp
tristpinsm: I don't think I know what you mean...

nritsche: Why do we store the ID of the **base** dataset?

tristpinsm: the `timeState` was added by the reader just as a convenience to communicate the time axis. It isn't really part of the dataset state that describes the data, which can now potentially be changing in time.

lib/stages/visRawReader.cpp
```diff
@@ -46,6 +46,7 @@ visRawReader::visRawReader(Config& config, const string& unique_name,
     readahead_blocks = config.get<size_t>(unique_name, "readahead_blocks");
     max_read_rate = config.get_default<double>(unique_name, "max_read_rate", 0.0);
     sleep_time = config.get_default<float>(unique_name, "sleep_time", -1);
+    use_comet = config.get_default<bool>(unique_name, "use_dataset_broker", false);
```
tristpinsm: default should be true

lib/stages/visRawReader.cpp
```diff
@@ -171,28 +172,34 @@ visRawReader::~visRawReader() {
...
+        // Add time to dataset
+        state_id_t tstate_id = dm.add_state(std::make_unique<timeState>(_times)).first;
+        // Register with dataset broker
+        _dataset_id = dm.add_dataset(tstate_id, ds_id);
+    } else {
+        // Create new states: metadata, time, prod, freq, input, eigenvalue and stack.
+        state_uptr sstate = nullptr;
+        if (!_stack.empty())
+            sstate = std::make_unique<stackState>(_num_stack, std::move(_rstack));
+        state_uptr istate = std::make_unique<inputState>(_inputs, std::move(sstate));
```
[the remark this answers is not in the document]
nritsche: @tristpinsm You can merge `develop` in here now to have the new dM API.

lib/stages/bufferRecv.cpp
```diff
@@ -214,6 +214,15 @@ void bufferRecv::main_thread() {
...
     evutil_make_socket_nonblocking(listener);
 
+    // This makes is possible to reuse the same socket even if it goes into TIME_WAIT
+    // Used by gossec
+    int reuse = 1;
+    if (setsockopt(listener, SOL_SOCKET, SO_REUSEADDR, &reuse, sizeof(int)) < 0) {
+        ERROR("Could not set SO_REUSEADDR on socket");
+        close(listener);
+        return;
+    }
```
tristpinsm: @andrerenard could you comment on this change? It is needed to make it possible for `gossec` to restart an instance of `kotekan` listening on the same port without running into `TIME_WAIT`

lib/stages/visRawReader.cpp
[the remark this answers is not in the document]
tristpinsm: ok, wish me luck...

lib/stages/bufferRecv.cpp
andrerenard: To be honest I don't have a great understanding of what implications option has...   I would suggest for now we only use it if `drop_frames` is false.  I don't want to risk creating an unexpected edge case in the normal use of `bufferRecv`

lib/stages/bufferSend.cpp
```diff
@@ -131,12 +136,15 @@ void bufferSend::main_thread() {
...
+            DEBUG("Sent frame: {:s}[{:d}] to {:s}:{:d}", buf->buffer_name, frame_id, server_ip,
+                  server_port);
 
         } else {
-            WARN("Dropping frame {:s}[{:d}], because connection to {:s}:{:d} is down.",
-                 buf->buffer_name, frame_id, server_ip, server_port);
+            // Wait for connection and block
+            INFO("Waiting for connection to {:s}:{:d}...", server_ip, server_port);
+            std::unique_lock<std::mutex> connection_lock(connection_state_mutex);
+            connection_state_cv.wait(connection_lock, [&]() {return (true && connected);});
```
andrerenard: Why does this return `true && connected` isn't that the same as `connected`?

andrerenard: Also there is no way for this wait to break when `stop_thread == true`.   I think you should make this a `wait_for` with a timeout of say 1 second, and check the status of `stop_thread` as well.   Then make sure if the stop_thread condition is true break this loop so you can trigger the notify for the connection thread when calling `close_connection()` at the end of this function.

tristpinsm: ok, thanks for the feedback. I made those changes. The reason it was `true && connected` was that `connected` is an atomic type so I had to evaluate it somehow in that context.

nritsche:
This is now using kotekan's `/stop` and `/start` endpoints to restart kotekan for every new file. I think it's been running fine over the weekend (@tristpinsm please correct me).  Do you know of anything speaking against doing it like that, @andrerenard ?

lib/utils/datasetManager.cpp
```diff
@@ -132,6 +132,11 @@ datasetManager::~datasetManager() {
 }
 
 
+void datasetManager::stop() {
```
nritsche: I'm not convinced this is necessary. We do it in the destructor already.

lib/stages/visTranspose.cpp
```diff
@@ -178,23 +209,41 @@ void visTranspose::main_thread() {
...
+            ERROR("Couldn't find ancestor of dataset {}. "
+                  "Make sure there is a stage upstream in the config, that adds the dataset "
+                  "states.\nExiting...",
+                  ds_id);
+            return;
+        }
     }
 
+    // Get the original dataset ID (before adding time axis)
+    base_ds_id = base_dset(ds_id);
```
nritsche: This is a bit of a stupid question asked much too late, but what's wrong about just using the datasetID that has the timeState instead of going back to the base dataset?

lib/stages/visTranspose.cpp
```diff
@@ -223,9 +272,17 @@ void visTranspose::main_thread() {
...
+            if (frame.fpga_seq_length == 0) {
+                INFO("Got an empty frame.");
+            } else {
+                // TODO assuming that dataset ID changes here never change dataset dimensions
+                INFO("Dataset ID has changed from {} to {}. Getting base dataset ID from "
+                     "broker...",
+                     ds_id, frame.dataset_id);
+                ds_id = frame.dataset_id;
+                base_ds_id = base_dset(ds_id);
+                INFO("Got base dataset ID {}.", base_ds_id);
```
nritsche: This seems like it should be `DEBUG`, but I leave this decision to you, since you probably have a better idea of what is useful.

lib/stages/visTranspose.cpp
```diff
@@ -266,10 +332,11 @@ void visTranspose::main_thread() {
         }
 
         // Increment within read chunk
-        ti = (ti + 1) % write_t;
-        if (ti == 0)
-            fi++;
-        if (fi == write_f) {
+        // within a chunk, frequency is the fastest varying index
```
nritsche: Can you explain this change?

lib/stages/visTranspose.cpp
```diff
@@ -344,3 +415,31 @@ void visTranspose::increment_chunk() {
     write_f = f_edge ? num_freq - f_ind : chunk_f;
     write_t = t_edge ? num_time - t_ind : chunk_t;
 }
+
+dset_id_t visTranspose::base_dset(dset_id_t ds_id) {
```
nritsche: This should really be implemented in the datasetManager. Make sure to leave an issue in case we merge it like this.

lib/utils/visFileArchive.cpp
```diff
@@ -135,14 +135,19 @@ void visFileArchive::write_block(std::string name, size_t f_ind, size_t t_ind, s
                                  size_t chunk_t, const T* data) {
     // DEBUG("writing {:d} freq, {:d} times, at ({:d},{:d}).", chunk_f, chunk_t, f_ind, t_ind);
     if (name == "flags/inputs") {
+        DEBUG("writing {}...", name);
```
nritsche: should we make these `DEBUG2`?

lib/utils/visFileArchive.cpp
```diff
@@ -135,14 +135,19 @@ void visFileArchive::write_block(std::string name, size_t f_ind, size_t t_ind, s
...
         dset(name).select({0, t_ind}, {length("input"), chunk_t}).write(data);
     } else if (name == "evec") {
+        DEBUG("writing {}...", name);
```
nritsche: same here

lib/utils/visFileArchive.cpp
```diff
@@ -135,14 +135,19 @@ void visFileArchive::write_block(std::string name, size_t f_ind, size_t t_ind, s
...
         dset(name)
             .select({f_ind, 0, 0, t_ind}, {chunk_f, length("ev"), length("input"), chunk_t})
             .write(data);
-    } else if (name == "erms" || name == "flags/frac_lost" || name == "flags/frac_rfi") {
+    } else if (name == "erms" || name == "flags/frac_lost" || name == "flags/frac_rfi"
+               || name == "flags/dataset_id") {
+        DEBUG("writing {}...", name);
```
nritsche: and here

lib/utils/visFileArchive.cpp
```diff
@@ -135,14 +135,19 @@ void visFileArchive::write_block(std::string name, size_t f_ind, size_t t_ind, s
...
         dset(name).select({f_ind, t_ind}, {chunk_f, chunk_t}).write(data);
     } else {
+        DEBUG("writing {}...", name);
```
nritsche: and here

lib/utils/visFileH5.hpp
```diff
@@ -254,6 +254,10 @@ class visFileH5Fast : public visFileH5 {
 
 // These templated functions are needed in order to tell HighFive how the
 // various structs are converted into HDF5 datatypes
+const size_t DSET_ID_LEN = 33; // Length of the string used to represent dataset IDs
+struct dset_id_str {
+    char hash[DSET_ID_LEN];
```
nritsche: why does this need to be a char array inside a struct? Maybe @jrs65 can have a look at just the changes in this file

lib/utils/visFileH5.hpp
```diff
@@ -265,6 +269,17 @@ template<>
...
 DataType create_datatype<cfloat>();
+
+// \cond NO_DOC
+// Fixed length string to store dataset ID
+template<>
+inline AtomicType<dset_id_str>::AtomicType() {
+    _hid = H5Tcopy(H5T_C_S1);
+    H5Tset_size(_hid, DSET_ID_LEN);
+    // define encoding to UTF-8 by default
+    // H5Tset_cset(_hid, H5T_CSET_UTF8);
```
nritsche: can we remove this commented out line?

lib/stages/visRawReader.cpp
```diff
@@ -130,7 +132,7 @@ visRawReader::visRawReader(Config& config, const string& unique_name,
         chunk_t = std::min(chunk_t, ntime);
 
         // Number of elements in a chunked row
-        row_size = (chunk_f * chunk_t) * (nfreq / chunk_f) + (nfreq % chunk_f) * chunk_t;
+        row_size = chunk_t * nfreq;
```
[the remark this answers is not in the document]
tristpinsm: they produce the same thing but the second line is much easier to read. I don't know what I was thinking with the initial version...

lib/stages/visRawReader.cpp
```diff
@@ -212,9 +230,9 @@ int visRawReader::position_map(int ind) {
         // edges case
         int f_width = std::min(nfreq - ci * chunk_f, chunk_f);
         // time and frequency indices
-        // time is fastest varying
-        int fi = ci * chunk_f + ((ind % row_size) % (t_width * f_width)) / t_width;
-        int ti = ri * chunk_t + ((ind % row_size) % (t_width * f_width)) % t_width;
+        // frequency is fastest varying
```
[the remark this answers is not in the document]
tristpinsm: for IO performance it is better to read with frequency as the fastest varying axis, which is how the data ordered in the file

lib/stages/visRawReader.hpp
```diff
@@ -192,4 +208,61 @@ class visRawReader : public kotekan::Stage {
     double sleep_time;
 };
 
+class ensureOrdered : public kotekan::Stage {
```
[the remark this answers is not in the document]
tristpinsm: This turned out not to be used in this release because we are instead relying on the fact that the network buffers receive frames in the same order they are sent. I'm not sure what is the best thing to do with this code since it might still be useful in the future.

lib/stages/visTranspose.cpp
```diff
@@ -223,9 +272,17 @@ void visTranspose::main_thread() {
         auto frame = visFrameView(in_buf, frame_id);
 
         if (frame.dataset_id != ds_id) {
-            FATAL_ERROR("Dataset ID of incoming frames changed from {} to {}. Changing  ID "
-                        "not supported, exiting...",
-                        ds_id, frame.dataset_id);
+            if (frame.fpga_seq_length == 0) {
+                INFO("Got an empty frame.");
+            } else {
+                // TODO assuming that dataset ID changes here never change dataset dimensions
```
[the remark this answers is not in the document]
tristpinsm: this should be guaranteed by the fact that these frames come from the same file. So I should probably remove the TODO there

lib/stages/visTranspose.cpp
tristpinsm: this is the complement to the change in the incoming order in `visRawReader`

lib/utils/visFileH5.hpp
tristpinsm: it's to get around the limitations for writing fixed length strings through `HighFive`

tristpinsm:
@nritsche Thanks for the thorough review! It looks like you've identified some improvements to make and since the code freeze has passed, I will leave it for next week to address these and we can deploy them then too.

nritsche:
I keep wondering if *timeState* is the correct solution. Every timeState is used only once and sent to comet just so it can be communicated to the receiver part of comet, right? Maybe it can be reused between different acquisition types, but I don't think that justifies making it a dataset state intended for storing metadata that changes seldomly enough so we don't want to pack it into kotekan frames.

A side effect of how we are handling it here at the moment is that we are adding a dataset with a timeState to the dataset tree with every set of files read by visRawReader.

tristpinsm:
That's correct. Finding an alternate solution would probably require adding features to kotekan that would be used only for gossec. Something like a new metadata format or a different interface for inter-process communication. What are the downsides of adding (probably unused in the future) time states to the dataset tree?

jrs65:
An alternate solution could be to write a very lightweight gossec specific broker. It has a read only view of the redis database and just keeps the timestate in memory but never writes it out. You'd be able to reuse most of the existing code, just wrap/overload a few sections.

PR #586, opened 2019-11-06
Revert Jenkinsfile

andrerenard:
Enable the old Jenkins server while we fix the new one.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,8 +9,7 @@ pipeline {
       parallel {
         stage('Build kotekan without hardware specific options') {
           steps {
-            sh '''mkdir build_no_hardware
-                  cd build_no_hardware/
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -19,7 +18,8 @@ pipeline {
         }
         stage('Build CHIME kotekan') {
           steps {
-            sh '''cd build
+            sh '''mkdir chime-build
+                  cd chime-build
                   cmake -DRTE_SDK=/opt/dpdk \
```
nritsche: Do we not have to revert back to
```
cmake -DRTE_SDK=/opt/dpdk-stable-16.11.4/ \
```
?

andrerenard: No, we symlinked that so it would work on both CI servers.

PR #587, opened 2019-11-06
Handle missing frames in frb- and pulsarPostProcess stages (closes #462)

### LINKED ISSUE #462: Support missing frames in FRB/Pulsar postProcess
Currently the FRB/Pulsar postProcess codes cannot cleanly support missing frames.  i.e. they expect each frame will be exactly one unit of time in the future from the last frame `t=0,1,2,3,4,5,6,...`.  If they could support things like `t=0,1,2,5,6` then we could drop frames in the GPU/network system to avoid crashing on system slow down.

cubranic:


andrerenard:
The title and description of this PR should be updated to reflect the fact it now addresses the Pulsar system as well. 

Also could you add `valve` stages before the `postProcess` stages (and after the GPU stage) to  `chime_science_run_gpu.yaml` and `chime_science_run_gpu_no_output.yaml`?

lib/testing/TestDataGenMissingFrames.hpp
```diff
@@ -0,0 +1,54 @@
...
+ *     @buffer_metadata none, or a class derived from @c kotekanMetadata
+ *
+ * @conf   missing_frames @c Vector of UInt32 (Default: empty). Frames to drop.
+ * @conf   drop_frame_chance @c Double (Default: 0). Chance of dropping a frame if not in the @c
+ *missing_frames list.
+ *
+ * @author James Willis, Davor Cubranic
+ **/
+
+class TestDataGenMissingFrames : public kotekan::Stage {
```
andrerenard: The name of this should likely be changed to something like `TestMissingFrames` or `TestDropFrames`.   It's not really generating data.

config/chime_science_run_gpu_drop_frames.yaml
```diff
@@ -0,0 +1,3521 @@
+##########################################
+#
+# chime_science_run_gpu_drop_frames.yaml
```
andrerenard: I'm not sure this should be in the main `config/` directory, but it also shouldn't be in `config/tests` since that's for larger GPU unit tests. 

Since this looks very similar to the setup for the `valve` based drops, I would just remove this config file, but add a comment in the main config files with the drops, that the stages names can be switched `valve` to `TestDataGenMissingFrames` to run this simulation.  Otherwise there is a risk this will just become out of data with the main config if we ever want to run it again.

config/tests/pulsar_simulate.yaml
```diff
@@ -0,0 +1,264 @@
+###
+# Simple Pulsar simulator: one GPU, dumping beamforming and post-processing buffers to /tmp
```
andrerenard: We might want to have a new folder for this, or move this into the `/tests/` folder, or maybe `/tests/helper_config/`.   Evently the files in `/config/tests/` will be run automatically as a part of the GPU unit testing system @james-s-willis is working on.

lib/testing/TestDataGenMissingFrames.hpp
cubranic: Renamed to `TestDropFrames`

config/chime_science_run_gpu_drop_frames.yaml
cubranic: How about I remove it when I add valves to the science configs?

config/tests/pulsar_simulate.yaml
cubranic: Moved to `/tests/helper_config`

config/chime_science_run_gpu_drop_frames.yaml
andrerenard: Sure.

PR #598, opened 2019-11-13
Use DPDK included with Ubuntu

andrerenard:
Currently we build DPDK from source, and manually configure it, however Ubuntu provides packages for DPDK without requiring that extra work.  This PR moves to use those packages, while still keeping backwards compatibility.


docs/sphinx/compiling/dpdk.rst
```diff
@@ -4,9 +4,53 @@
...
+
+Then include ``-DUSE_DPDK=ON`` in the cmake options.
+
+Setup
+^^^^^
+
+Run ``dpdk-devbind --status`` to see the addresses of your system NICs, then
+add the NICs you want to have bound to DPDK at startup to ``/etc/dpdk/interfaces``
+for example to bind two Intel XL710 NICs to the ``uio_igb`` driver at PCIe addresses
+``0000:46:00.0`` and ``0000:48:00.0`` and the following two lines:
```
james-s-willis: "add" the following two lines

PR #600, opened 2019-11-13
Fix RFI bad inputs update problems

### LINKED ISSUE #594: Investigate extended RFI excision for some frequencies
We are seeing every now and again that the RFI system will start to excise some frequencies at a constant rate for an extended period of time. These may be as long as 12 hours. The frequencies seem to differ between occurrences of the event, and may occasionally coincide at the start with an actual high RFI rate (but not always).

james-s-willis:
Buffers updates to the bad input list, which is fed to the `hsaRfiUpdateBadInputs` command object.

Fixes a bug where two consecutive updates to the bad input list in <0.5s would result in the second update being lost. The bug was found as a result of investigating #594. 

Also applies a fix for the race condition found in #600 which closes #594


lib/hsa/hsaRfiUpdateBadInputs.cpp
```diff
@@ -51,12 +59,36 @@ hsaRfiUpdateBadInputs::~hsaRfiUpdateBadInputs() {
...
+    } else {
+        // Check for new bad inputs only if filled all gpu frames (not currently filling frame)
+        if (!filling_frame) {
+            auto timeout = double_to_ts(0);
+            int status = wait_for_full_frame_timeout(_in_buf, unique_name.c_str(),
+                                                     _in_buf_precondition_id, timeout);
+            DEBUG("status of bad inputs _in_buf_precondition_id[{:d}]={:d} (0=ready 1=not)",
+                  _in_buf_precondition_id, status);
+            if (status == 0) {
+                filling_frame = true;
```
andrerenard: There is a mix of up variables here.  These look like they are copied from the `AsyncGain `logic, but haven't been updated in the rest of the code to match the logic.   However given the bug we found in #603, this change should likely merge that in as well, and then the variables here should be updated to reflect the way that code now works.

lib/stages/bufferBadInputs.cpp
```diff
@@ -0,0 +1,81 @@
...
+    if (badInputs.length() > 0)
+        configUpdater::instance().subscribe(
+            badInputs, std::bind(&bufferBadInputs::update_bad_inputs_callback, this, std::placeholders::_1));
+}
+
+bufferBadInputs::~bufferBadInputs() {}
+
+bool bufferBadInputs::update_bad_inputs_callback(nlohmann::json& json) {
+    
+    static uint32_t out_buffer_ID = 0;
```
andrerenard: Don't use `static` in a stage unless there is very good reason, use a member variable.  Otherwise you cannot have more than one of a given stage.

lib/stages/bufferBadInputs.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+    DEBUG("update_bad_inputs_callback(): Update to bad inputs list.");
+
+    // Get the first output buffer which will always be id = 0 to start.
+    uint8_t* host_mask = wait_for_empty_frame(out_buf, unique_name.c_str(), out_buffer_ID);
+
+    try {
+      bad_inputs_cylinder = json["bad_inputs"].get<std::vector<int>>();
```
andrerenard: The formatting here doesn't follow the 4 space guidelines, use `clang-format` to fix.

config/chime_science_run_gpu.yaml
```diff
@@ -499,6 +499,25 @@ baseband:
...
+# Buffer bad input updates
+# 1) Reorder list
+# 2) Zero bad inputs mask
+# 3) Add current bad input mask
+buffer_bad_input_update:
+  updatable_config:
+    bad_inputs: /updatable_config/bad_inputs
+  buffer_bad_inputs: 
+    kotekan_stage: bufferBadInputs
+    out_buf: bad_inputs_buffer
```
andrerenard: This could be restructured like this:
```
  buffer_bad_inputs: 
    kotekan_stage: bufferBadInputs
    updatable_config:
      bad_inputs: /updatable_config/bad_inputs
    out_buf: bad_inputs_buffer
```

lib/stages/bufferBadInputs.cpp
```diff
@@ -0,0 +1,82 @@
...
+    input_remap = std::get<0>(input_reorder);
+
+    out_buf = get_buffer("out_buf");
+    register_producer(out_buf, unique_name.c_str());
+
+    out_buffer_ID = 0;
+
+    // Listen for bad input list updates
+    string badInputs = config.get<std::string>(unique_name, "updatable_config/bad_inputs");
+    if (badInputs.length() > 0)
```
andrerenard: We should always expect the subscribe to this endpoint, don't make it optional.

lib/hsa/hsaRfiUpdateBadInputs.cpp
```diff
@@ -67,14 +93,14 @@ hsa_signal_t hsaRfiUpdateBadInputs::execute(int gpu_frame_id, hsa_signal_t prece
     set_rfi_num_bad_inputs(_network_buf, _network_buf_execute_id, bad_inputs_correlator.size());
```
andrerenard: There is a subtitle, but important thing here.  `bad_inputs_correlator` should be removed from this object since we no longer update it.  However that means the frames we get will have to contain the number of bad inputs and we'll have to record that so that we can continue to update the metadata here with the right value.

lib/hsa/hsaRfiUpdateBadInputs.cpp
```diff
@@ -51,12 +55,34 @@ hsaRfiUpdateBadInputs::~hsaRfiUpdateBadInputs() {
...
+        DEBUG("hsaRfiUpdateBadInputs::wait_on_precondition: Bad inputs update.");
+        if (frame == NULL)
+            return -1;
+        _in_buf_precondition_id = (_in_buf_precondition_id + 1) % _in_buf->num_frames;
+        first_pass = false;
+        frames_to_update = device.get_gpu_buffer_depth();
+        filling_frame = true;
+    } else {
+        // Check for new bad inputs only if filled all gpu frames (not currently filling frame)
+        if (!filling_frame) {
```
andrerenard: `filling_frame` will never get set back to false.  You should use the check of `frame_copy_active` and `frames_to_update` from my branch.

lib/hsa/hsaRfiUpdateBadInputs.cpp
```diff
@@ -51,12 +55,34 @@ hsaRfiUpdateBadInputs::~hsaRfiUpdateBadInputs() {
 int hsaRfiUpdateBadInputs::wait_on_precondition(int gpu_frame_id) {
     (void)gpu_frame_id;
 
-    uint8_t* frame =
-        wait_for_full_frame(_network_buf, unique_name.c_str(), _network_buf_precondition_id);
-    if (frame == nullptr)
-        return -1;
-
-    _network_buf_precondition_id = (_network_buf_precondition_id + 1) % _network_buf->num_frames;
+    // Check for bad input updates
```
andrerenard: This needs be locked with the `update_mutex`

lib/stages/bufferBadInputs.hpp
```diff
@@ -0,0 +1,76 @@
...
+private:
+    struct Buffer* in_buf;
+    struct Buffer* out_buf;
+
+    /// Config variables
+    uint32_t _num_frames_to_integrate;
+    uint32_t _num_frb_total_beams;
+    uint32_t _num_sub_freqs;
+    uint32_t _samples_per_data_set;
+    float _good_samples_threshold;
```
andrerenard: There are a lot of member variables here that aren't used.

lib/hsa/hsaRfiUpdateBadInputs.cpp
```diff
@@ -87,64 +111,21 @@ hsa_signal_t hsaRfiUpdateBadInputs::execute(int gpu_frame_id, hsa_signal_t prece
...
+        frame_copy_active.at(frame_id) = false;
         hsaCommand::finalize_frame(frame_id);
-        // Check if we've finished loading the new bad input mask in all frames.
-        if (frames_to_update_finalize == 0) {
-            update_bad_inputs = false;
+        DEBUG("finalize_frame for gpu_frame_id={:d} using _in_buf_finalize_id={:d}", frame_id,
+              _in_buf_finalize_id);
+        frame_to_fill_finalize--;
+        if (frame_to_fill_finalize == 0) {
+            mark_frame_empty(_in_buf, unique_name.c_str(), _in_buf_finalize_id);
```
andrerenard: Since you are doing a `memcpy()` in the precondition part you can release the frame in the precondition. No need to wait for here.  That would also simply the logic.

lib/hsa/hsaRfiUpdateBadInputs.cpp
```diff
@@ -87,64 +111,21 @@ hsa_signal_t hsaRfiUpdateBadInputs::execute(int gpu_frame_id, hsa_signal_t prece
 
 void hsaRfiUpdateBadInputs::finalize_frame(int frame_id) {
     std::lock_guard<std::mutex> lock(update_mutex);
-    if (update_bad_inputs && frames_to_update_finalize > 0) {
-        frames_to_update_finalize--;
+    
+    // Only mark input empty if filling frame and
+    // no more frames to finalize.
+    if (frame_to_fill_finalize > 0 && frame_copy_active.at(frame_id)) {
```
andrerenard: We don't actually need the `frame_to_fill_finalize` condition anymore.  The `frame_copy_active.at(frame_id)` is sufficient as far as I can tell.

lib/hsa/hsaRfiUpdateBadInputs.cpp
```diff
@@ -51,30 +49,59 @@ hsaRfiUpdateBadInputs::~hsaRfiUpdateBadInputs() {
...
+        if (frame == NULL)
+            return -1;
+        first_pass = false;
+        frames_to_update = device.get_gpu_buffer_depth();
+        memcpy(host_mask, frame, input_mask_len);
+        num_bad_inputs = get_rfi_num_bad_inputs(_in_buf, _in_buf_precondition_id);
+        DEBUG("finalize_frame for gpu_frame_id={:d} using _in_buf_precondition_id={:d}",
+              gpu_frame_id, _in_buf_precondition_id);
+        mark_frame_empty(_in_buf, unique_name.c_str(), _in_buf_precondition_id);
+        _in_buf_precondition_id = (_in_buf_precondition_id + 1) % _in_buf->num_frames;
```
andrerenard: Factor out this and lines 76-82 to an inline function maybe?  The only local variable is `gpu_frame_id` by the looks of it.   Also the debug message still talks about `finalize_frame` which is no longer true.

lib/hsa/hsaRfiUpdateBadInputs.cpp
```diff
@@ -32,49 +29,77 @@ hsaRfiUpdateBadInputs::hsaRfiUpdateBadInputs(Config& config, const string& uniqu
...
-        config.get<std::string>(unique_name, "updatable_config/bad_inputs"),
-        std::bind(&hsaRfiUpdateBadInputs::update_bad_inputs_callback, this, std::placeholders::_1));
 }
 
 hsaRfiUpdateBadInputs::~hsaRfiUpdateBadInputs() {
     hsa_host_free(host_mask);
 }
 
+inline void hsaRfiUpdateBadInputs::copy_frame(int gpu_frame_id) {
+
```
andrerenard: You need a `(void)gpu_frame_id;` here.

lib/hsa/hsaRfiUpdateBadInputs.cpp
```diff
@@ -32,49 +29,77 @@ hsaRfiUpdateBadInputs::hsaRfiUpdateBadInputs(Config& config, const string& uniqu
...
+    if (first_pass) {
+        uint8_t* frame = wait_for_full_frame(_in_buf, unique_name.c_str(), _in_buf_precondition_id);
+        if (frame == NULL)
+            return -1;
+        first_pass = false;
+        copy_frame(gpu_frame_id);
+    } else {
+        // Check for new bad inputs only if all gpu frames have been updated (not currently updating
+        // frame)
+        if (frames_to_update == 0 && !frame_copy_active.at(gpu_frame_id)) {
```
andrerenard: This actually should check all elements of `frame_copy_active()` are false.  Otherwise you could have a copy happening which hasn't yet finished, but `frames_to_update` would be zero, and you'd write into `host_mask` while it was being written.  See how I did this in #603

PR #602, opened 2019-11-15
refactor(FakeVis): move patterns into separate classes

jrs65:


lib/testing/FakeVisPattern.hpp
```diff
@@ -0,0 +1,225 @@
...
+    virtual ~FakeVisPattern() = default;
+
+    /**
+     * @brief Fill the data with a test pattern.
+     *
+     * @note You can also modify the metadata (for example to set the number of
+     *       lost frames), but be careful doing this if there are multiple
+     *       frequencies in the buffer.
+     *
+     * @param  data       Data for the specific frequency in the buffer.
```
nritsche: CI failed because of the parameters listed here incorrectly.

jrs65: Ooops I've pushed a fix. Should be building now.

PR #604, opened 2019-11-15
Use ccache to speed up jenkins build

### LINKED ISSUE #532: Jenkins: speed up make
@cubranic wrote:
I use ccache on my laptop to speed up builds and it works amazingly well. https://ccache.dev/manual/3.7.1.html
it just needs -DCMAKE_CXX_COMPILER_LAUNCHER=ccache in the initial invocation to cmake, and it will effectively use ccahce $CXX to build.

nritsche:
The parallel make commands in the Jenkinsfile take about 2min with this (without any changes to the source files).

Closes #532

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,7 +4,16 @@ pipeline {
     timeout(time: 1, unit: 'HOURS')
     parallelsAlwaysFailFast()
   }
+  environment {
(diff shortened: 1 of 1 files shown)


nritsche:
The parallel `make` commands in the Jenkinsfile take about 2min with this (without any changes to the source files).

PR #606, opened 2019-11-21
Gain and flag dataset updates

jrs65:


lib/stages/receiveFlags.cpp
```diff
@@ -139,23 +155,36 @@ void receiveFlags::main_thread() {
...
+
+        std::pair<state_id_t, dset_id_t> key = {state_id, frame_out.dataset_id};
+        if (output_dataset_ids.count(key) == 0) {
+            output_dataset_ids[key] = dm.add_dataset(state_id, frame_out.dataset_id);
         }
-        ts_late = update.first - ts_frame;
+        frame_out.dataset_id = output_dataset_ids[key];
+
+        ts_late = ts_update - ts_frame;
         if (ts_late.tv_sec > 0 && ts_late.tv_nsec > 0) {
```
tristpinsm: this is probably irrelevant, but shouldn't this be `||`?

tristpinsm:
Looking through this generated one question regarding the changes I am about to make. Currently, `applyGains` listens for a gain update from the broker that triggers it to read new gains from a file. The new scheme would have this trigger it to query the calibration broker for the new gains. Wouldn't it make more sense for the calibration broker to just send the gains in its update? This would require changes to Seth's implementation, so maybe we are better off doing it with the back-and-forth update

lib/stages/receiveFlags.cpp
jrs65: Hmmmm. Actually it doesn't need to be either. `tv_nsec` is guaranteed to be `>= 0` and so the second step can be skipped.

tristpinsm:
Following up on my earlier question, I've made a sketch of the new functionality and another thing to consider is that we will have to add a base64 library to decode the gain updates. I'm going to call it a day now. I don't think it will be a ton of work to wrap this up, but we should probably decide if we want to go ahead without any changes to the broker.

jrs65:
@nritsche can you look in particular at the additions to the dataset manager? But if you fancy reviewing anything else that would be great.

lib/utils/datasetManager.hpp
```diff
@@ -500,16 +539,41 @@ class datasetManager {
...
+    // If not found, return null
+    if (!ret)
+        return nullptr;
+
+    state_id_t state_id = ret.value().second.state();
 
-    // get an update on the dataset topology (blocking)
-    update_datasets(dset);
+    {
+        std::lock_guard<std::mutex> dslock(_lock_dsets);
```
nritsche: Consequences of releasing the lock between updating the datasets and figuring out which states to request could mean that states are requested multiple times. May not be a big problem. but should keep it in mind?

lib/utils/datasetState.hpp
```diff
@@ -630,4 +630,111 @@ class acqDatasetIdState : public datasetState {
...
+        json j;
+        j["update_id"] = _update_id;
+        j["transition_interval"] = _transition_interval;
+        return j;
+    }
+
+    // The label for the gains
+    std::string _update_id;
+
+    // The length of time we (in seconds) should blend gain updates over
```
nritsche:
```suggestion
    // The length of time (in seconds) gain updates are blended over.
```
But is this for blending with the previous or the next update?

lib/utils/updateQueue.hpp
```diff
@@ -42,6 +44,17 @@ class updateQueue {
             values.pop_front();
     }
 
+    /**
+     * @brief Get the current size of the queue.
+     *
+     * This may be less than the length if not enough updates have been posted.
```
nritsche:
```suggestion
     * This may be less than the maximum size if not enough updates have been posted.
```
(?)

lib/utils/datasetManager.hpp
```diff
@@ -314,16 +316,58 @@ class datasetManager {
...
+
+    /**
+     * @brief Fingerprint a dataset for specified states.
+     *
+     * Generate a summary of the specified states present in the requested
+     * dataset. This will be unique for datasets where there one or more of the
+     * requested states differ. Datasets that share all these states will give
+     * the same fingerprint regardless of differences in any other states.
+     *
+     * The fingerprint does not depend on the order of state_types. It is also
```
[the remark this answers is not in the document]
jrs65: It does not as it uses a `std::set` (which iterates in order)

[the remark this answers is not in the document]
nritsche: Oh, alright.
You could still hash stuff individually and then XOR the hashes. Slower but would look less hacky.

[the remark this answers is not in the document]
jrs65: That would definitely be better. Do you have an idea how I can incorporate the state type in there too?

tests/boost/dataset_broker_producer2.cpp
```diff
@@ -77,6 +77,12 @@ BOOST_FIXTURE_TEST_CASE(_dataset_manager_general, CompareCTypes) {
     dset_id_t ds_id;
     ds_id.set_from_string(line);
 
+    // Add a meaningless state in here. This is designed to trigger a failure
+    // case in older versions where if the local manager already knew about a
+    // state, it didn't bother checking to see if it knew about earlier states
+    //auto state_id = dm.create_state<flagState>("flag_id1").first;
+    //ds_id = dm.add_dataset(state_id, ds_id);
```
nritsche: Let's keep these 2 lines, though

lib/utils/datasetManager.hpp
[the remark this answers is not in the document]
jrs65: I'm going to just resolve this as I'm not sure how to address it with the comment below too.

lib/utils/datasetManager.cpp
```diff
@@ -576,3 +579,67 @@ void datasetManager::force_update_callback(kotekan::connectionInstance& conn) {
...
+    // TODO: doing this via constructing a string seems like a bit of a hack, so
+    // we should look for a cleaner way to do it. The big problem is that the
+    // type names are all variable length strings.
+    std::string fs;
+
+    for (const auto& state : state_types) {
+        auto dset = closest_dataset_of_type(ds_id, state);
+        state_id_t id = dset ? dset.value().second.state() : dset_id_t::null;
+
+        fs += state + id.to_string();
```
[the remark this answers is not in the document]
jrs65: Indeed.

tests/boost/dataset_broker_producer2.cpp
jrs65: Oh, yeah. I struggled to get this to work. I think it may have been to do with the unclean exit problems. I'll try running again.

lib/utils/datasetManager.hpp
jrs65: Good point. I'll put a comment in the source.

PR #613, opened 2019-11-26
[WIP] RFI second stage

jrs65:


lib/stages/RfiFrameDrop.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+    pass_metadata(buf_src, frame_id_src, buf_dest, frame_id_dest);
+}
+
+bool RfiFrameDrop::rest_callback(nlohmann::json& update) {
+    bool enable_rfi_zero_new;
+    std::vector<std::tuple<float, size_t, float>> thresholds_new;
+
+    try {
+        enable_rfi_zero_new = update.at("enable_rfi_zero").get<bool>();
```
andrerenard: This value is also in the chimeMetadata as the `rfi_zeroed` flag.  If you don't want us to also have to push updates to this endpoint at solar transit as well.   Unless you want to be able to control this independently.

lib/stages/RfiFrameDrop.hpp
```diff
@@ -0,0 +1,108 @@
...
+    /// Prometheus metrics to export
+    kotekan::prometheus::MetricFamily<kotekan::prometheus::Counter>& _failing_frame_counter;
+    kotekan::prometheus::MetricFamily<kotekan::prometheus::Counter>& _dropped_frame_counter;
+    kotekan::prometheus::MetricFamily<kotekan::prometheus::Counter>& _frame_counter;
+
+    size_t num_elements;
+    size_t num_sub_frames;
+    size_t sk_step;
+    size_t sk_samples_per_frame;
+    size_t samples_per_sub_frame;
```
andrerenard: I think the coding standard for `_` got mixed up here.  Config dirived variables should start with `_`, and the metric counters are just member variables and should not have the starting underscore.

lib/testing/fakeGpu.cpp
```diff
@@ -92,32 +94,46 @@ void FakeGpu::main_thread() {
...
-            gsl::span<int32_t> data(output + 2 * freq_ind * nprod_gpu,
-                                    output + 2 * (freq_ind + 1) * nprod_gpu);
-            pattern->fill(data, metadata, frame_count, freq + freq_ind);
-        }
 
-        // Mark full and move onto next frame...
-        mark_frame_full(out_buf, unique_name.c_str(), frame_id++);
+        // Test if this will be dropped or not, and only fill out the buffer
+        // (and advance) if it won't
+        if (drop(gen) < drop_probability) {
```
andrerenard: I think @cubranic has created a stage that will randomly drop frames.  We might at some point want to switch to using that once it's merged into `develop` instead of having the test gen stages all have this functionality.  For now though I think this is ok to leave in.

config/chime_science_run_gpu_no_output.yaml
```diff
@@ -203,6 +203,14 @@ updatable_config:
...
           3.948826407949732e-35,
         ],]
+  second_stage_rfi:
+    kotekan_update_endpoint: json
+    thresholds:
+      - threshold: 5.0
+        fraction: 0.025
+      - threshold: 1.4
+        fraction: 0.4
+    enable_rfi_zero: false
```
andrerenard: This feels like a duplicate of `rfi_zeroing_toggle`.  I would consider either using that, or just using the metadata flag (see the other comment on this).

lib/testing/fakeGpu.cpp
cubranic: I did: `TestDropFrames` as part of #587 . It can drop specific frames or randomly with a given probability.

andrerenard:
@cubranic you need to click "Add a single comment" instead of the green "Start a review" button.  It is a little confusing, since it's tempting to just push the green button.

config/chime_science_run_gpu_no_output.yaml
jrs65: We've renamed it so they have the same name, but I want them to be separately switchable. Rick has rewritten the coco endpoint such that one endpoint will call both (for the solar noon script)

lib/testing/fakeGpu.cpp
jrs65: Hmmmm. Okay. I might switch to using that at somepoint. Thanks!

PR #617, opened 2019-11-27
Add clang build to CI server

### LINKED ISSUE #534: Jenkins: Add clang build
This might help catch differences in warnings between `gcc` and `clang`.

james-s-willis:
Adds a clang build to CI server. Closes #534.


CMakeLists.txt
```diff
@@ -223,7 +224,7 @@ set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} ${CMAKE_C_FLAGS} -ggdb -O2")
 set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE}")
 
 # Note that -lrt here is needed on some versions of CentOS
```
andrerenard:
```suggestion
# Note that -lrt here is needed on some versions of CentOS but not Clang
```

Jenkinsfile
```diff
@@ -37,6 +37,20 @@ pipeline {
                   make -j 4'''
           }
         }
+        stage('Build CHIME kotekan with Clang') {
+          steps {
+            sh '''mkdir -p chime-build-clang
+                  cd chime-build-clang
+                  export CC=/usr/lib/clang-8.0/bin/clang
+                  export CXX=/usr/lib/clang-8.0/bin/clang++
```
andrerenard: I'm tempted to just make this `CC=clang` and `CXX=clang++` then it will be upto the system to make sure the default clang version is a good one.

lib/core/errors.h
```diff
@@ -11,10 +11,14 @@
 #include <signal.h>
 
 #ifdef __cplusplus
+#ifdef __clang__
+#include <string.h>
```
andrerenard: Why is this needed?   Is there a way we can setup these includes such that it works with all compilers?

james-s-willis: If I include `cstring` compiling with `clang` I get errors:
```
/usr/bin/../lib/gcc/x86_64-linux-gnu/7.4.0/../../../../include/c++/7.4.0/cstring:100:3: error: conflicting types for 'memchr'
  memchr(void* __s, int __c, size_t __n)
  ^
/usr/include/string.h:90:14:
:/usr/bin/../lib/gcc/x86_64-linux-gnu/7.4.0/../../../../include/c++/7.4.0/cstring: 100:3: error: conflicting types for 'memchr'
  memchr(void* __s, int __c, size_t __n)
  ^
/usr/include/string.h:90:14: note: previous declaration is here
extern void *memchr (const void *__s, int __c, size_t __n)
```

andrerenard: I think the problem is that `<cstring>` and `<string.h>` are the C++ and C versions of the library respectively.    If you look above on line 9, `<string.h>` always gets included, so also including `<cstring>` breaks things.  I'm guessing compilers like `gcc` just pick one for you.  The better solution would seem to be to only include `<string.h>` in the `else` clause of the `#ifdef __cplusplus`, that way one of the two is included, and we don't have to try and also select for compilers.

PR #622, opened 2019-11-29
fix(visFileRaw): Make files group-writeable.

tristpinsm:
This is so `gossec` user can move and delete files


jrs65:
This shouldn't cause any issues and the CI server is down. I'm going to force merge it now.

PR #625, opened 2019-11-30
Add Valve stages in front of FRB and Pulsar post-process stages

cubranic:
As discussed in #587, here are the changes to add FRB/PSR Valves to the science configs. At this pint the "drop_frames" science version is kind of redundant, and instead we give directions in the science configs how to force dropping of frames by replacing `Valve`s with `TestDropFrames`.


cubranic:
I’m always confused: should I merge it at this point or wait for a go-ahead?

andrerenard:
Sorry I was just going to fix the CI server and let it process this before merging, but then I had to leave my computer for a bit.  It should start processing this soon.  (Although given it's just a config change, I'm really just doing it for good form.)

Issue #627, opened 2019-12-02
Missing samples in FRB intensity data (?)


cubranic:
There are what appear to be either missing or zero-ed out data in files produced by intensity callbacks.
![Example from Saturday, Nov 30th](https://user-images.githubusercontent.com/600891/69991123-f0f29d00-14fb-11ea-9219-702d88cbdb73.png)

![Weights](https://user-images.githubusercontent.com/600891/69991255-37e09280-14fc-11ea-8e89-51f382b2bdd9.png)

![Weights at the drop](https://user-images.githubusercontent.com/600891/69991434-82faa580-14fc-11ea-98db-d1abf8e0deb6.png)

chitrangpatel:
<img width="459" alt="Screen Shot 2019-12-02 at 3 27 24 PM" src="https://user-images.githubusercontent.com/11685590/69992815-82700800-1518-11ea-9b60-004c90bda4b0.png">

chitrangpatel:
The times of the two largest dips separated by `128` samples are:
```                                      
datetime.datetime(2019, 12, 2, 7, 25, 3, 547243)                                  datetime.datetime(2019, 12, 2, 7, 25, 3, 673072)
```

mondana:
@chitrangpatel @andrerenard @cubranic any progress on this?

chitrangpatel:
With the FRB archiver back up, I was able to look at the surrounding beams to see if all the beams show the zeroed packets. It seems like the major spike is consistent in all of them. But the others are not. 

<img width="1023" alt="Screen Shot 2019-12-03 at 3 47 19 PM" src="https://user-images.githubusercontent.com/11685590/70091214-a51e2180-15e9-11ea-9194-5385217a3165.png">

cubranic:
After the most recent restart, the drop is still present. This is in beam 0115:

<img width="467" alt="Screen Shot 2019-12-05 at 11 50 27 AM" src="https://user-images.githubusercontent.com/600891/70257047-6f3e8180-173e-11ea-82c1-dbc35a41363f.png">
<img width="510" alt="Screen Shot 2019-12-05 at 11 50 32 AM" src="https://user-images.githubusercontent.com/600891/70257048-6f3e8180-173e-11ea-9f2c-a5690b6680af.png">

[Slack discussion](https://chime-experiment.slack.com/archives/C83Q194RW/p1575564695332600)

andrerenard:
[...]

PR #629, opened 2019-12-04
Count masked FRB packets

cubranic:
- add a Prometheus metric to count masked FRB packets
- check for lost FPGA samples *after* syncing inputs


cubranic:
@andrerenard this is the metrics to help us check for those 128 ms drops that were reported in #627.

andrerenard:
CI server has crashed again, going to try to get that going first before merging.

PR #631, opened 2019-12-05
fix(kotekan.service): Set the umask so that file creation group permi…

tristpinsm:
…ssions are not overridden.


tristpinsm:
@jrs65 we should merge this soon, because when kotekan got installed earlier today it changed the service file such that we are writing without the correct permissions again.

PR #634, opened 2019-12-05
Disable the L1 destination checking if `ping_dead_threshold` is 0

cubranic:
We are seeing recurring storms of nodes marking multiple destinations dead and
then live again within a few seconds, so it's clear either there is a bug in the
current implementation or the thresholds need to be fine-tuned. Since we are
pressed to start the run, it makes sense to add a way to disable this feature
from the configuration, should we want to do so without reverting it completely.

### DIFF
--- a/config/chime_science_run_gpu.yaml
+++ b/config/chime_science_run_gpu.yaml
@@ -838,6 +838,7 @@ frb:
     #column_mode: true
     beam_offset: 0
     time_interval: 125829120
+    dead_ping_threshold: 0
     L1_node_ips:
       # Rack 1
       - 10.6.201.10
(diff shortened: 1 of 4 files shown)


cubranic:
I was going to do it in a separate PR.

cubranic:
I guess I may as well do it here...

andrerenard:
I just saw your comment, will wait.

cubranic:
Here you go. Will you do the merge?

andrerenard:
Yeah I'll merge as soon as it passes CI.

cubranic:
Thanks!

andrerenard:
@cubranic Could you open a ticket to revisit this at some point?  Along with maybe some of the logs you were seeing that show this problem.

PR #635, opened 2019-12-05
Refactor receiveFlags

nritsche:
- if receiveFlags receives and old frame, don't exit but drop (and print the timestamps in the updatequeue)
- move locked region into member function

### DIFF
--- a/lib/stages/receiveFlags.cpp
+++ b/lib/stages/receiveFlags.cpp
@@ -23,6 +23,10 @@ REGISTER_KOTEKAN_STAGE(receiveFlags);
 receiveFlags::receiveFlags(Config& config, const string& unique_name,
                            bufferContainer& buffer_container) :
     Stage(config, unique_name, buffer_container, std::bind(&receiveFlags::main_thread, this)),
+    receiveflags_late_frame_counter(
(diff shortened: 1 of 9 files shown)


lib/stages/receiveFlags.cpp
```diff
@@ -153,50 +149,53 @@ void receiveFlags::main_thread() {
...
+        mark_frame_empty(buf_in, unique_name.c_str(), frame_id_in);
+        frame_id_in = (frame_id_in + 1) % buf_in->num_frames;
 
-        if (update == nullptr) {
-            FATAL_ERROR("receiveFlags: Flags for frame {:d} with timestamp {:f} are not in memory. "
-                        "updateQueue is empty",
-                        frame_id_in, ts_to_double(ts_frame));
-            return;
+        if (success) {
+            // Mark output frame full
```
jrs65: Not required, but you could change this to use `frameID`'s.

lib/utils/updateQueue.hpp
```diff
@@ -124,4 +133,24 @@ class updateQueue {
     size_t _len;
 };
 
+// Define a custom fmt formatter that prints the timestamps
```
jrs65: Great!

PR #636, opened 2019-12-06
Add to science_gpu_no_output config settings from science_gpu introduced in #613

cubranic:


cubranic:
What are unit tests failing when I changed a science_no_output config?

PR #637, opened 2019-12-06
feat: add a stage to the config to track the output dataset IDs

jrs65:


config/chime_science_run_recv.yaml
```diff
@@ -267,6 +266,28 @@ remove_ev:
...
+    in_buf: visbuf_5s_26m_ungated_post
+  26mgated:
+    kotekan_stage: visDebug
+    in_buf: visbuf_5s_26m_gated_post
+  cal:
+    kotekan_stage: visDebug
+    in_buf: visbuf_10s_cal
+  timing:
+    kotekan_stage: visDebug
+    in_buf: visbuf_10s_timing_ne{
```
nritsche:
```suggestion
    in_buf: visbuf_10s_timing_ne
```

andrerenard:
```suggestion
    in_buf: visbuf_10s_timing_ne
```

jrs65: Ooops. Stupid vim.

jrs65: I think you and @nritsche both caught the same thing. It's vim's fault.

PR #641, opened 2019-12-07
fix(applygain): remove check on num_threads vs buffer size

nritsche:


jrs65:
We should also remove this for `baselineCompression`

PR #645, opened 2019-12-08
Reduce metric labels

jrs65:


jrs65:
@cubranic I tagged you for a quick review so you can see if my changes make sense. There's a few metrics I want to be able to break down by frequency or dataset_id, but not both at the same time, so I created two metrics one for each. My understanding from here (https://www.robustperception.io/on-the-naming-of-things) is that while that's generally discouraged, the one exception is when you need to do it for performance reasons.

lib/stages/visWriter.hpp
```diff
@@ -89,8 +89,10 @@
  * @metric kotekan_viswriter_write_time_seconds
  *         The write time of the HDF5 writer. An exponential moving average over ~10
  *         samples.
- * @metric kotekan_viswriter_dropped_frame_total
- *         The number of frames dropped while attempting to write.
+ * @metric kotekan_viswriter_late_frame_total
+ *         The number of frames dropped while attempting to write as they are too late.
+ * @metric kotekan_viswriter_bad_dataset_frame_total
```
cubranic: I know you expect this not to happen too often, but let's keep an eye for how it behaves in the running system, just so it doesn't blow up out of control. At worst, we can block it from being indexed by Prometheus, but still have it available if you `curl` the metrics from an individual node.

cubranic:
It seems like we could use something like OpenTracing precisely for this kind of per-dataset tracing...

Issue #640, opened 2019-12-06
Check for valid baseband destination path when receiving a request


cubranic:
Right now, if frb-baseband is not mounted, a baseband request is still processed until `BasebandReadout` tries to create a file in the request's path and fails. Instead, `BasebandApiManager` can check that the `file_path` exists, and return an error if it doesn't, saving us the trouble of reading out the data.

andrerenard:
Closed by #647

PR #647, opened 2019-12-10
Check for baseband destination directory before reading out the samples (#640)

cubranic:
Because `BasebandApiManager` is not a Stage, we don't have access to the running
configuration to construct the destination relative to `base_dir`, so the error
handling has to be done in the BasebandReadout. This is a little less ideal,
because this way we can't respond with an HTTP error as part of the request, but
still prevents using up the time to read out the samples that cannot be saved to
the requested destination.


cubranic:
@andrerenard I added the warning.

PR #656, opened 2019-12-14
refactor(BranchPrediction): multiple definitions of (un)likely

### LINKED ISSUE #599: The `likely`/`unlikely` annotations are defined in five different source files
We should probably only define it once in a header, rather than in each individual file.

../lib/testing/pulsarSimProcess.cpp
../lib/stages/pulsarPostProcess.cpp
../lib/stages/frbPostProcess.cpp
../lib/stages/beamformingPostProcess.cpp
../lib/hsa/hsaPulsarUpdatePhase.cpp

nritsche:
Closes #599

### DIFF
--- a/lib/hsa/hsaPulsarUpdatePhase.cpp
+++ b/lib/hsa/hsaPulsarUpdatePhase.cpp
@@ -8,15 +8,12 @@
 #include "configUpdater.hpp"
 #include "hsaBase.h"
 #include "restServer.hpp"
+#include "visUtil.hpp"
 
 #include <math.h>
 #include <signal.h>
 #include <string>
 #include <time.h>
-#include <utils/visUtil.hpp>
-
-#define likely(x) __builtin_expect(!!(x), 1)
-#define unlikely(x) __builtin_expect(!!(x), 0)
 
 #define PI 3.14159265
 #define light 299792458.
--- a/lib/stages/frbPostProcess.cpp
+++ b/lib/stages/frbPostProcess.cpp
@@ -1,7 +1,3 @@
-// TODO Where do these live?
-#define likely(x) __builtin_expect(!!(x), 1)
-#define unlikely(x) __builtin_expect(!!(x), 0)
-
 #include "frbPostProcess.hpp"
 
 #include <immintrin.h>
--- a/lib/stages/beamformingPostProcess.cpp
+++ b/lib/stages/beamformingPostProcess.cpp
@@ -7,10 +7,7 @@
 
 using std::string;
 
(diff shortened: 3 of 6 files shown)


nritsche:
(CI tests fixed) @andrerenard I think this is ready. Some places imported `rte_branch_prediction.h` through `dpdkCore` and I decided to leave it that way.

lib/hsa/hsaPulsarUpdatePhase.cpp
```diff
@@ -8,15 +8,12 @@
 #include "configUpdater.hpp"
 #include "hsaBase.h"
 #include "restServer.hpp"
+#include "visUtil.hpp"
```
cubranic: Ah, was the include "utils/visUtil.hpp" incorrect then? Or would both have worked?

lib/stages/frbPostProcess.cpp
```diff
@@ -1,7 +1,3 @@
-// TODO Where do these live?
```
cubranic: I can't figure out which `#include` frbPostProcess gets the defines from?

andrerenard: `frbPostProcess` doesn't actually use `[un]likely` so it just doesn't need it.

lib/hsa/hsaPulsarUpdatePhase.cpp
andrerenard: This also doesn't have any `[un]likely` uses it in, but I don't really know where the `visUtil.hpp` is coming from.  @nritsche was this included by mistake?

lib/stages/frbPostProcess.cpp
cubranic: Hah, you're right! I remembered its use from pulsarPostProcess, and never noticed that it's not used in frbPostProcess even though the two are so similar. It probably should be, to mark the `if (startup)` case in line 48, but that's a different issue. :-)

lib/hsa/hsaPulsarUpdatePhase.cpp
nritsche: I moved the visUtil import from the system includes to the project icludes. I guess both works, but style.

nritsche:
@cubranic Both "testing/pulsarSimProcess.cpp" and "frbPostProcess.cpp" were not using the defines.

lib/hsa/hsaPulsarUpdatePhase.cpp
andrerenard: Ah I see.

PR #659, opened 2019-12-17
CI: include-what-you-use

nritsche:
This adds a cmake option `IWYU` that runs `include-what-you-use` while compiling. Due to iwyu's awkward usage, the output of this is checked in a stage after the parallel builds. CI will fail if iwyu makes suggestions.
Running iwyu on kotekan is slow and it doesn't get cached. ~Therefor it's rather fortunate, that cmake doesn't run it again if a build is already cached. But that also means, one can make a build pass by restarting the CI build after it had failed because of iwyu.~ Making jenkins run it on the base build adds about 8min to the CI build.

- [x] write documentation for `lint.sh`
- [x] how to use as pre-commit hook
- [x] extend iwyu to full (clang) chime build
- [x] remove boost test builds where not needed
- [x] --no-fwd-decls
- [x] fixes for full CHIME build
- [x] ~speed up travis CI~ run iwyu in github action
- [x] update iwyu on magenta (0.11)

For now this only uses iwyu for source files included by the base kotekan build. To be continued...

Summary for squash:
```
refactor: include-what-you-use

* run iwyu in github action for branches master and develop
* (README): add badge indicating check passing/failing for `develop` branch
* Add gcc option -Wzero-as-null-pointer-constant
* (cmake): mark external library includes with SYSTEM keyword
* add tools/lint.sh: runs clang-format, black and (optionally) iwyu
*  remove using from headers: IWYU revealed that many places got e.g. using std:string from various
headers. IWYU would also add includes to make this more obvious.
* (Config): remove barely used template instantiation, because it didn't compile anymore (huh?) and was only used once anyways
* (restClient): use iovec from libevent
* (restServer): use iovec from libevent
* remove duplicate of e_time() and use e_time from util.h everywhere.
* (cmake): add option IWYU
* (jenkins): use ccache for C
* enable clang-format for C files (*.c and *.h)
```


nritsche:
refactor: include-what-you-use

* run iwyu in Jenkins
* Add gcc option -Wzero-as-null-pointer-constant
* (cmake): mark external library includes with SYSTEM keyword
* add tools/lint.sh: runs clang-format, black and iwyu
*  remove using from headers: IWYU revealed that many places got e.g. using std:string from various
headers. IWYU would also add includes to make this more obvious.
* (Config): remove barely used template instanciation, because it didn't compile anymore and was only used once anyways
* (rest*): use iovec from libevent
* remove duplicate of e_time() and use e_time from util.h everywhere.
* (cmake): add option IWYU
* (jenkins): use ccache for C

andrerenard:
@nritsche  Do you know how well `iwyu` parallelizes?   I don't like that we are adding 8 minutes to the CI process, but maybe we could give it more cores to run on and/or get a system with more cores?

andrerenard:
Ah I see you have increased the number of cores for that build.   Let me see about getting a server with more cores for us to use at some point...

nritsche:
> Ah I see you have increased the number of cores for that build. Let me see about getting a server with more cores for us to use at some point...

I am very open to disabling this slow check before merging.

lib/gpu/gpuProcess.hpp
```diff
@@ -4,17 +4,26 @@
...
+#include "Stage.hpp"           // for Stage
+#include "bufferContainer.hpp" // for bufferContainer
 
-#include "json.hpp"
+#include <stdint.h> // for uint32_t
+#include <string>   // for string
+#include <thread>   // for thread
+#include <vector>   // for vector
+
+class gpuCommand;
```
andrerenard: I'm not sure I understand what's going on with these forward declarations (here and in other places).   Why forward declare and not include?

lib/hsa/hsaBeamformUpchan.cpp
```diff
@@ -1,11 +1,21 @@
 #include "hsaBeamformUpchan.hpp"
 
+#include "Config.hpp"             // for Config
+#include "gpuCommand.hpp"         // for gpuCommandType, gpuCommandType::KERNEL
+#include "hsaDeviceInterface.hpp" // for hsaDeviceInterface
+
+#include <string.h> // for memcpy, memset
+
+namespace kotekan {
+class bufferContainer;
```
andrerenard: I realise we aren't actually making use of `bufferContainer` here, but it seems odd to forward declare in a CPP file when we have the header.

andrerenard:
As for disabling the long test.  I'd vote that we disable, since waiting for the CI is a huge pain, but let's plan to re-enable the test once we have a more powerful CI computer.   And this will still be a huge amount of much needed clean-up!

lib/gpu/gpuProcess.hpp
nritsche: iwyu does that because it's supposed to compile faster (https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md#why-forward-declare). There's an option to disable forward declares in a newer version of iwyu.

lib/hsa/hsaBeamformUpchan.cpp
nritsche: I also don't understand this. What happened in some cases was:
1. iwyu changes include to fwd declare -> doesn't compile
2. I revert the change and force it to keep the include
3. iwyu insists on having both (might be a bug?)

lib/gpu/gpuProcess.hpp
andrerenard: I would be tempted to disable the forward declares option (adding `--no_fwd_decls`).  It seems like the current state will just cause problems when we try to use something referenced only as a forward declare.  And I think it will also cause issues with some IDEs autocomplete and documentation lookup.  Which is basically to say, I'm willing to trade a little compile time for not having to edit the includes when I use something that's only forward declared.

nritsche:
- [ ] --no-fwd-decls
- [ ] speed up travis CI

nritsche:
On Ubuntu 18.04 I managed to run iwyu 0.12 (has --no_fwd_decls) by installing from https://launchpad.net/ubuntu/focal/amd64/iwyu/8.0-3
I had to help this version of iwyu where to find builtin headers:
```
FLAGS="-isystem /usr/lib/llvm-9/lib/clang/9.0.0/include/"  CXXFLAGS="-isystem /usr/lib/llvm-9/lib/clang/9.0.0/include/" CXX=clang++ CC=clang cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ..
```

iwyu command changed to
```
iwyu_tool -j 4 -p . -- -Xiwyu --no_fwd_decls -Xiwyu --mapping_file=/home/rick/kotekan/iwyu.kotekan.imp -Xiwyu --max_line_length=100
```

nritsche:
@andrerenard If you approved now, I think I would be able to merge without fixing the iwyu check. I also added a badge to the kotekan README showing if iwyu currently passes in `develop`. This way I thought we can tell people to feel free to ignore the test if they don't want to wait 20min.

docs/sphinx/dev/dev_style_guide_code.rst
```diff
@@ -140,28 +140,27 @@ For more details, compare the `kotekan clang-format file
...
-So it doesn't happen again, you should check if there is `clang-format`
-integration for your favourite editor and point it at kotekan's `.clang-format
-file <https://github.com/kotekan/kotekan/blob/master/.clang-format>`_.
 
+Automatic code formatting
+-------------------------
 
-Disabling Code Formatting
-^^^^^^^^^^^^^^^^^^^^^^^^^
+.. note:: To do all code linting in one, use `tools/lint.sh <https://github.com/kotekan/kotekan/blob/master/tools/lint.sh>`_. To use the sript as a commit hook, copy it to ``.git/hooks/pre-commit``.
```
nritsche:
```suggestion
.. note:: To do all code linting in one, use `tools/lint.sh <https://github.com/kotekan/kotekan/blob/master/tools/lint.sh>`_. To use the script as a commit hook, copy it to ``.git/hooks/pre-commit``.
```

tools/iwyu/README.md
```diff
@@ -0,0 +1,14 @@
+The include-what-you-use mapping files are from the
+[iwyu repository](https://github.com/include-what-you-use/include-what-you-use) version
+`0.13`. The file kotekan CI uses to reference all mapping files is `iwyu.kotekan.imp` in the
+root directory of this repository.
+It can be applied by running
+```
+cd build
+export CXX=clang++
+cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ..
+iwyu_tool -p . -- --max_line_length=100 --mapping_file=/full/path/to/iwyu.kotekan.imp > iwyu.out
```
nritsche:
```suggestion
iwyu_tool -p . -- -Xiwyu --no_fwd_decls -Xiwyu --max_line_length=100 -Xiwyu--mapping_file=/full/path/to/iwyu.kotekan.imp | tee iwyu.out
```

tools/iwyu/README.md
```diff
@@ -0,0 +1,14 @@
...
+python2 /usr/bin/fix_include --comments < iwyu.out
```
nritsche:
```suggestion
python2 /usr/bin/fix_include --nosafe_headers  --comments < iwyu.out
```

PR #661, opened 2019-12-19
config(recv): mask out the last eight feeds on each cylinder.

jrs65:
The final eight feeds at each of the cylinder seem to have substantially
different beams to the interior feeds. To try and improve the stack,
this PR flags out these end feeds.


config/chime_science_run_recv.yaml
```diff
@@ -210,7 +210,17 @@ stacking:
   out_buf: visbuf_10s_stack
   stack_type: chime_in_cyl
   exclude_inputs: [
-    46, 142, 688, 944, 960, 1058, 1166, 1225, 1314, 1521, 1543, 2032, 2034
+    # These are non-CHIME feeds we want to exclude (26m, noise source channels, etc.)
+      46,  142,  688,  944,  960, 1058, 1166, 1225, 1314, 1521, 2032, 2034
```
ssiegelx: Do you need a comma after 2034?

jrs65: Ooops. Thanks!

PR #665, opened 2020-01-07
New framework endpoints

andrerenard:
Adds new framework endpoints to get more information about the running state of `kotekan`:

* `/pipeline_dot` Adds an endpoint that generates a graphviz `dot` file, which can be turned into a pipeline graph.   For example: 
```
curl localhost:12048/pipeline_dot | dot -Tpdf -o pipeline.pdf
```
* `/buffers` Dumps the list of buffers and their current state in JSON, intended to be used by a higher level tool to visualize the system state.


andrerenard:
Example of a complex dot generated pipeline:

pipeline

lib/core/kotekanMode.cpp
```diff
@@ -122,4 +133,95 @@ void kotekanMode::stop_stages() {
...
+    const std::string prefix = "    ";
+    std::string dot =
+        "# This is a DOT formatted pipeline graph, use the graphviz package to plot.\n";
+    dot += "digraph pipeline {\n";
+
+    // Setup buffer nodes
+    for (auto& buf : buffer_container.get_buffer_map()) {
+        dot += fmt::format(
+            "{:s}\"{:s}\" [label=<{:s}<BR/>{:d}/{:d} ({:.1f}%)> shape=ellipse, color=blue];\n",
+            prefix, buf.first, buf.first, get_num_full_frames(buf.second), buf.second->num_frames,
```
james-s-willis, line 198: I'm guessing this just repeats the label of the buffer at the start?
`buf.first, buf.first`

james-s-willis (approved):
As far as I can tell this looks good to me.

lib/core/kotekanMode.cpp
andrerenard: Yep the first instance is the node name for use in the edge pairs entries, and then the next instance is creating the label for the node.

Issue #274, opened 2018-11-12
Add const qualifier to Config::get* methods and KotekanProcess constructors


jrs65:
There doesn't seem to be any good reason that all the get methods on the Config class can't be marked as `const` as they really shouldn't have any side effects. That would also allow all the KotekanProcess constructors to receive a `const Config&`.

Steps:
- [x] Mark `Config::get*` methods as `const`
- [ ] Update process constructs to get a `const Config&`

andrerenard:
It looks like at least in the science run config we still have the following endpoints which update the config: 
```
"/frb/update_beam_offset",
"/gpu/gpu_0/frb/update_EW_beam/0",
"/gpu/gpu_0/frb/update_NS_beam/0",
"/gpu/gpu_1/frb/update_EW_beam/1",
"/gpu/gpu_1/frb/update_NS_beam/1",
"/gpu/gpu_2/frb/update_EW_beam/2",
"/gpu/gpu_2/frb/update_NS_beam/2",
"/gpu/gpu_3/frb/update_EW_beam/3",
"/gpu/gpu_3/frb/update_NS_beam/3",
"/rfi_bad_input_finder/gpu_0/rfi_bad_input_finder",
"/rfi_bad_input_finder/gpu_1/rfi_bad_input_finder",
"/rfi_bad_input_finder/gpu_2/rfi_bad_input_finder",
"/rfi_bad_input_finder/gpu_3/rfi_bad_input_finder",
"/rfi_broadcast/gpu_0/change_params",
"/rfi_broadcast/gpu_1/change_params",
"/rfi_broadcast/gpu_2/change_params",
"/rfi_broadcast/gpu_3/change_params",
```

PR #668, opened 2020-01-17
refactor(Config): mark some member functions const

nritsche:
in preparation of #274


andrerenard:
Thanks for doing this!

PR #669, opened 2020-01-18
refactor(RfiFrameDrop): shared_ptr instead of copy updates

### LINKED ISSUE #638: RFI second stage: use shared_ptr to communicate new updates
Instead of copying out the thresholds while holding a lock, use a shared pointer.

nritsche:
Allows much shorter mutex protected section in main thread.

Closes #638

### DIFF
--- a/lib/stages/RfiFrameDrop.cpp
+++ b/lib/stages/RfiFrameDrop.cpp
@@ -69,10 +69,8 @@ void RfiFrameDrop::main_thread() {
 
     std::vector<float> sk_delta(sk_samples_per_frame);
 
-    // shared pointers to receive updates from the callback
-    std::shared_ptr<bool> enable_rfi_zero;
-    std::shared_ptr<std::vector<size_t>> sk_exceeds;
-    std::shared_ptr<std::vector<std::tuple<float, size_t, float>>> thresholds;
+    // shared pointer to receive updates from the callback
+    std::shared_ptr<ThresholdData> thresholds;
 
     while (!stop_thread) {
         // Fetch the input buffers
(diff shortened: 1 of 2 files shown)


lib/stages/RfiFrameDrop.hpp
```diff
@@ -86,16 +87,16 @@ class RfiFrameDrop : public kotekan::Stage {
     Buffer* _buf_out;
 
     /// Thresholds
-    std::vector<std::tuple<float, size_t, float>> _thresholds;
+    std::shared_ptr<std::vector<std::tuple<float, size_t, float>>> _thresholds;
 
     /// Toggle RFI zeroing
-    bool _enable_rfi_zero;
+    std::shared_ptr<bool> _enable_rfi_zero;
```
jrs65: Just use a `std::atomic<bool>`

lib/stages/RfiFrameDrop.hpp
```diff
@@ -86,16 +87,16 @@ class RfiFrameDrop : public kotekan::Stage {
     Buffer* _buf_out;
 
     /// Thresholds
-    std::vector<std::tuple<float, size_t, float>> _thresholds;
+    std::shared_ptr<std::vector<std::tuple<float, size_t, float>>> _thresholds;
```
jrs65: You might consider binding both `_sk_exceeds` and `thresholds` into a single new struct, then you only need a single shared_ptr to point to all the threshold related state. That would eliminate any theoretical potential of inconsistency.

lib/stages/RfiFrameDrop.cpp
```diff
@@ -106,82 +112,84 @@ void RfiFrameDrop::main_thread() {
...
 
+        // Point the shared pointers to the newest updates
         {
-            // Calculate the scaling to turn kurtosis value into sigma
-            size_t num_inputs = num_elements - metadata_vis->rfi_num_bad_inputs;
-            float sigma_scale = sqrt((num_inputs * (sk_step - 1) * (sk_step + 2) * (sk_step + 3))
-                                     / (4.0 * sk_step * sk_step));
-
-            // Disable updates on enable_rfi_zero and thresholds while accessing these
             std::lock_guard<std::mutex> guard(lock_updatables);
```
jrs65: Rather than locking here just use a `std::atomic_load` (https://en.cppreference.com/w/cpp/memory/shared_ptr/atomic)

lib/stages/RfiFrameDrop.cpp
```diff
@@ -229,7 +237,7 @@ bool RfiFrameDrop::rest_enable_callback(nlohmann::json& update) {
         _enable_rfi_zero = enable_rfi_zero_new;
```
jrs65: Lock shouldn't be necessary with a `std::atomic<bool>`.

lib/stages/RfiFrameDrop.cpp
```diff
@@ -276,17 +282,18 @@ bool RfiFrameDrop::rest_thresholds_callback(nlohmann::json& update) {
             return false;
         }
 
-        thresholds_new.push_back({threshold, (size_t)(fraction * sk_samples_per_frame), fraction});
+        thresholds_new->push_back({threshold, (size_t)(fraction * sk_samples_per_frame), fraction});
     }
 
     {
+        // make the updates available to the main thread by shared pointers
```
jrs65: Similarly use a `std::atomic_store` here to eliminate the lock.

PR #675, opened 2020-01-22
Fix #667: use the right name for "dead_ping_threshold" in the config

### LINKED ISSUE #667: Fix disabled L1 destination check
PR #634 introduced a bug whereby the config and the code used different names for the parameter denoting the "node dead" threshold:  “dead_ping_threshold” in the config vs “ping_dead_threshold” in the code. It should be the latter.

cubranic:
(Note that this is based off `master`, in case we want to deploy it as a quick fix during February power shutdown.)

### DIFF
--- a/config/chime_science_run_gpu.yaml
+++ b/config/chime_science_run_gpu.yaml
@@ -960,7 +960,7 @@ frb:
     #column_mode: true
     beam_offset: 0
     time_interval: 125829120
-    dead_ping_threshold: 0
+    ping_dead_threshold: 0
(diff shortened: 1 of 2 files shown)


andrerenard:
@cubranic since we've decided to only apply this in the release, can you either adjust the base to `develop` (I tried just changing the github target, but now it's also pulling in `master` changes), or close this and open a new PR to fix this in `develop`?

cubranic:
@andrerenard I rebased it onto latest develop, there is only one commit.

PR #677, opened 2020-01-27
fix(test_applygains): wait for frame duration to ensure update arrives

jrs65:


nritsche:
Thanks :pray: ! I'm going to change the commit type to `test` (see https://github.com/chime-experiment/Pipeline/blob/master/CONTRIBUTING.md#type) when merging.

jrs65:
Sure, that makes sense.

jrs65:
I think it's built four times now without failure so I'm going to merge it.

PR #678, opened 2020-01-28
Set number of beams with config option

### LINKED ISSUE #674: Support 2 extra beams in the pulsar beamformer
In theory there is enough headroom in the system to support 2 more beams in the pulsar beamformer (bringing the total to 12).   However this will require some code modifications to make scale the number of beams without breaking any hardcoded assumptions. 

This is tentatively targeted for the March release, but low enough priority that it could slip to the next one. 

Separately this will require work on site to add network capacity to support the extra 2 beams.  It might be possible to transport them on the existing trunk links (it would bring the total to 38.4Gb/s), and then it would only need two extra cables to connect from the VSOP-Nexus to whatever systems we use to record this extra beam data.  If that doesn't work though, we will need to set-up another link between GPU-N Nexus and the VSOP Nexus (possibly requiring another pair of single-mode transceivers).

Also we should break apart the `pulsar_gain` update endpoint from one endpoint for all beam gains, to one endpoint for each beam.  This will prevent us from having to coordinate with the Pulsar team to push different gains to those 2 new beams.   This will also require the pulsar team to adjust their gain update system to call 10 endpoints instead of currently just the one.  And it will also naturally require a number of changes in `coco`'s config to add the extra beams and adjust how the gains are applied. 

According to @arunnaidu123 if the beam pointing is set to `0 ra, +90 dec` then the phase offset that gets applied should be constant, so we aren't planning to make any changes to the pointing endpoints that might explicitly disable phase application.

james-s-willis:
Allows the no. of pulsar beams to be set in the config file. Closes #674.


config/chime_psr_verification.yaml
```diff
@@ -18,7 +18,7 @@ log_level: info
 num_elements: 2048
 num_local_freq: 1
 num_data_sets: 1
-samples_per_data_set: 49152
+samples_per_data_set: 3840
```
andrerenard: Why was this switched?  Just to improve performance?

config/chime_psr_verification.yaml
```diff
@@ -184,38 +250,35 @@ gpu:
...
+          - . #/mnt/frb-baseband/GainFiles/Latest_PSR
+          - . #/mnt/frb-baseband/GainFiles/Latest_PSR
+          - . #/mnt/frb-baseband/GainFiles/Latest_PSR
+          - . #/mnt/frb-baseband/GainFiles/Latest_PSR
+          - . #/mnt/frb-baseband/GainFiles/Latest_PSR
+          - . #/mnt/frb-baseband/GainFiles/Latest_PSR
+          - . #/mnt/frb-baseband/GainFiles/Latest_PSR
+          - . #/mnt/frb-baseband/GainFiles/Latest_PSR
+          - . #/mnt/frb-baseband/GainFiles/Latest_PSR
+          - . #/mnt/frb-baseband/GainFiles/Latest_PSR
```
andrerenard: Why is all this commented?

lib/stages/ReadGain.cpp
```diff
@@ -103,11 +103,12 @@ bool ReadGain::update_gains_psr_callback(nlohmann::json& json) {
...
     try {
         _gain_dir_psr = json.at("pulsar_gain_dir").get<std::vector<string>>();
-        INFO("[PSR] Updating gains from {:s} {:s} {:s} {:s} {:s} {:s} {:s} {:s} {:s} {:s}",
-             _gain_dir_psr[0].c_str(), _gain_dir_psr[1].c_str(), _gain_dir_psr[2].c_str(),
-             _gain_dir_psr[3].c_str(), _gain_dir_psr[4].c_str(), _gain_dir_psr[5].c_str(),
-             _gain_dir_psr[6].c_str(), _gain_dir_psr[7].c_str(), _gain_dir_psr[8].c_str(),
-             _gain_dir_psr[9].c_str());
+        string output_msg = "[PSR] Updating gains from ";
+        for (int i = 0; i < _num_beams; i++) {
+            output_msg += _gain_dir_psr[i].c_str();
```
andrerenard:
```suggestion
            output_msg += _gain_dir_psr[i];
```
I don't think this needs `.c_str()` since you are appending to an `std::string`

config/chime_psr_verification.yaml
james-s-willis: That was just so I could run it quickly when checking the accuracy of answers.

andrerenard:
@james-s-willis can you resolve the conflicts that showed up here?  And then I'll do a final review pass, and we can merge.

lib/hsa/hsaBeamformPulsar.cpp
```diff
@@ -23,29 +22,18 @@ hsaBeamformPulsar::hsaBeamformPulsar(Config& config, const std::string& unique_n
     command_type = gpuCommandType::KERNEL;
 
     _num_elements = config.get<int32_t>(unique_name, "num_elements");
-    _num_pulsar = config.get<int32_t>(unique_name, "num_beams");
+    _num_beams = config.get<int32_t>(unique_name, "num_beams");
```
andrerenard: Can we add a check that `_num_beams <= MAX_NUM_BEAMS` and throw a runtime exception if not?

lib/stages/pulsarNetworkProcess.hpp
```diff
@@ -98,6 +98,9 @@ class pulsarNetworkProcess : public kotekan::Stage {
 
     /// array of socket ids
     int* socket_ids;
+
+    /// Number of pulsar beams
```
andrerenard:
```suggestion
    /// Number of tracking (pulsar) beams
```

andrerenard:
Also I would like to have more descriptive names for the PRs.  I'll rename it to "Set number of beams with config option" now.

PR #679, opened 2020-02-06
Pulsar gain calculation hotfix for 26m passthrough

andrerenard:
* Clamp a variable which was resulting in `NAN` in the case of a static pointing
* Disable tracking for beams with when `scaling == 1`


andrerenard:
I think the only correct way to do that long term is to add a new parameter to the pointing endpoint, something like: `tracking_beam:<bool>`  I can imagine other cases we might want `scaling != 1` but still disable the tracking phase.   However that will require config, coco, and scheduler changes that I don't want to attempt in a hotfix.

emmanuelfonseca:
while I'm not a reviewer I'd say this looks good to me. technically the `acos(std::clamp())` bits are new and will "affect" all beams, but if I've followed the latest messages right (apologies as I've been facilitating the ongoing busy week) then this still changes nothing since they're meant to catch the NaNs produced from the static-beam forming. I'll keep an eye on data over the weekend but I believe things should appear as normal, besides the triggering of the static beam when the scaling factor is set to 1.

andrerenard:
@emmanuelfonseca yes that's correct.  The `clamp` is only to limit the range of the values to the valid ranges of `acos` and `asin` so if those values ever happened in the normal beams they would result in fully railled data anyway.  So for normally beams if anything this change should only be an improvement, although it looks like it's only in this edge case that it would show up.

PR #682, opened 2020-02-12
Abstracted frameView.

james-s-willis:
I've made changes to abstract the `frameView`. Could you review the changes to make sure they align with our discussion earlier today? Note: I have removed the `visFrameView(Buffer* buf, int frame_id, visFrameView frame_to_copy)` constructor as it used metadata arguments from `frame_to_copy` to construct the new view and Andre thinks we should be copying the frame explicitly with `visFrameView::copy_frame` instead. @andrerenard @jrs65


lib/stages/EigenVisIter.cpp
```diff
@@ -163,8 +163,12 @@ void EigenVisIter::main_thread() {
             break;
         }
         allocate_new_metadata_object(out_buf, output_frame_id);
-        auto output_frame = visFrameView(out_buf, output_frame_id, input_frame.num_elements,
-                                         input_frame.num_prod, _num_eigenvectors);
+        visMetadata* metadata = (visMetadata*)out_buf->metadata[output_frame_id]->metadata;
+        metadata->num_elements = input_frame.num_elements;
+        metadata->num_prod = input_frame.num_prod;
+        metadata->num_ev = _num_eigenvectors;
```
andrerenard: This pattern shows up a lot, can we make this into a function.

lib/core/bufferFactory.cpp
```diff
@@ -100,8 +100,7 @@ struct Buffer* bufferFactory::new_buffer(const string& type_name, const string&
             num_prod = num_elements * (num_elements + 1) / 2;
         }
 
-        auto layout = visFrameView::calculate_buffer_layout(num_elements, num_prod, num_ev);
-        uint32_t frame_size = layout.first;
+        size_t frame_size = visFrameView::calculate_frame_size(num_elements, num_prod, num_ev);
```
andrerenard: I'd be tempted to push the `config.get<>()` logic above into the `calculate_frame_size()` function.  i.e. keep implementation logic out of the factory.

lib/utils/frameView.cpp
```diff
@@ -0,0 +1,27 @@
+#include "frameView.hpp"
```
andrerenard: I keep forgetting to say this, but we really should follow our own standards and use `PascalCase` for these files/classes.

lib/utils/frameView.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+protected:
+    // References to the buffer and metadata we are viewing
+    Buffer* const buffer;
+    const int id;
+
+    // Pointer to frame data. In theory this is redundant as it can be derived
+    // from buffer and id, but it's nice for brevity
+    uint8_t* const _frame;
+};
```
andrerenard: I like this simple base class!

lib/utils/frameView.hpp
```diff
@@ -0,0 +1,76 @@
...
+     * @param buf      The buffer the frame is in.
+     * @param frame_id The id of the frame to read.
+     */
+    frameView(Buffer* buf, int frame_id);
+
+    /**
+     * @brief Read only access to the frame data.
+     * @returns The data.
+     **/
+    const uint8_t* data() const {
```
andrerenard: hmm, I'm not sure about the `const` here.  i.e. I'm not sure we need to be that restrictive, since I could imagine a producer wanting a way to do a `void *` write to the frame (say from a preformatted file or network socket).   I might also rename it `raw_data()` to make it clear this is a pure C style unprotected pointer.

lib/utils/frameView.hpp
```diff
@@ -0,0 +1,76 @@
...
+        return _frame;
+    }
+
```
andrerenard: It seems like it might be worth having a `get_raw_frame_size()` function.  Although it would just query the `buffer` so I'm not sure how important that abstraction is.

lib/testing/FakeVis.cpp
```diff
@@ -274,8 +278,8 @@ void ReplaceVis::main_thread() {
         auto input_frame = visFrameView(in_buf, input_frame_id);
 
         // Copy input frame to output frame and create view
-        allocate_new_metadata_object(out_buf, output_frame_id);
-        auto output_frame = visFrameView(out_buf, output_frame_id, input_frame);
+        auto output_frame =
+            visFrameView::copy_frame(in_buf, input_frame_id, out_buf, output_frame_id);
```
andrerenard: I like this change, it's more self-descriptive.

james-s-willis:
@jrs65 have you had a chance to look at this yet?

james-s-willis:
> This is basically defunct, right? I'm adding a comment, just to get it out of my "Review Requests" list!

Ah yes, I am basing all of my new PRs from this one. So I kept it around.

andrerenard:
Can we also close this @james-s-willis ?

james-s-willis:
@andrerenard I'm still using this one to create my new PRs

andrerenard:
I think you can still close it, and just use the closed PR for reference.  Github shouldn't delete the branch as long as there are outstanding unmerged commits.

PR #685, opened 2020-02-15
Add a development tutorial

anjakefala:
Adds a basic development tutorial with producer, consumer, and producer/consumer example stages.


nritsche:
This is looking really nice :+1: :+1: :+1:
Conceptually, I think it could be more in line with kotekan's design to explain how to write a stage, then how to make it consume and produce buffers (instead of introducing the concept as exclusive consumer and producer stages).

anjakefala:
@nritsche You are right! Especially considering that a stage can both consume and produce buffers.... :sweat_smile:

andrerenard:
If possible I'd like to pick this up again.  My comments for moving forward:
- It would be good to have both a producer and consumer in the tutorial.  (With a later extension to a stage which is both a producer and consumer - and which does something interesting - e.g. calls fftw or implements a dot product).  
- I'd like to have an example configuration file, and commands for running it.  The python test framework is worth including as an extra step to the tutorial, but it would be good to get people running the binary directly and quickly seeing results. 
- Having a few diagrams, to show the data flow from producer -> buffer -> consumer, and how that maps to the configuration would be very helpful.  See for example the slides I gave for CASPER 2019.  
- I think `frameViews` are worth having, but as their own section.   
Give that I think a rough outline might be:
1. Introduction to kotekan - theory of operation (stages, buffers, config files).
2. Example producer and consumer `stages`.  Ideally with a configuration that people can just run out of the box and see results.  
3. Example of a `stage` which is both a producer and consumer (doing something simple, but useful)
4. Example of a `frameView` to give a structured datatype to the frames past simply casting the frame pointer.  You can update the stages in the previous steps to then also make use of the `frameView`
5. Python unit tests framework, with example of testing stage.

james-s-willis:
@andrerenard I've added an example stage that computes the dot product. Along with documentation on how to build it and run it. Could you take a look?

lib/stages/SampleStage.cpp
```diff
@@ -0,0 +1,67 @@
...
+using kotekan::bufferContainer;
+using kotekan::Config;
+using kotekan::Stage;
+
+// Register the stage with abstract factory.
+// Needed for Kotekan to be able to find the stage.
+REGISTER_KOTEKAN_STAGE(SampleStage);
+
+/* Constructor for the stage
+    TODO: Annotate the arguments everywhere
```
andrerenard, line 19: I would include these annotations in the header using the doxygen standard to keep things consistent.

lib/stages/DotProduct.cpp
```diff
@@ -0,0 +1,107 @@
...
+            break;
+        if (frame_b_ptr == nullptr)
+            break;
+
+        // Wait for new output buffer
+        uint8_t* out_frame_ptr = wait_for_empty_frame(out_buf, unique_name.c_str(), out_frame_id);
+        if (out_frame_ptr == nullptr)
+            break;
+
+        // Get pointers to float values
```
andrerenard, line 83:
```suggestion
        // Cast pointers to float arrays
```

lib/stages/DotProduct.cpp
```diff
@@ -0,0 +1,107 @@
...
+    // Register as a producer of out_buf
+    out_buf = get_buffer("out_buf");
+    register_producer(out_buf, unique_name.c_str());
+
+    // Get the no. of elements in each frame
+    _num_elements = in_a_buf->frame_size / sizeof(float);
+
+    // Ensure the input and output buffers are the same
+    // length for the dot product
+    assert(in_a_buf->frame_size == in_b_buf->frame_size
```
andrerenard, line 43: Use an `if` statement here, and then throw an `std::runtime_error` with a message if the size rules fail.

lib/stages/DotProduct.cpp
```diff
@@ -0,0 +1,107 @@
...
+using kotekan::Config;
+using kotekan::Stage;
+
+// Register the stage with abstract factory.
+// Needed for Kotekan to be able to find the stage.
+REGISTER_KOTEKAN_STAGE(DotProduct);
+
+/* Constructor for the stage
+    TODO: Annotate the arguments everywhere
+    Note, that you can use the macro STAGE_CONSTRUCTOR(DotProduct)
```
andrerenard, line 20: I would use the `STAGE_CONSTRUCTOR` in one of the two example stages, so people can see it in action.

lib/stages/DotProduct.cpp
```diff
@@ -0,0 +1,107 @@
...
+        for (uint32_t i = 0; i < _num_elements; i++) {
+            output[i] = a[i] * b[i];
+        }
+
+        // Logging
+        INFO("Input buffer {:s}[{:d}]: {:f}, ..., {:f}, ..., {:f}", in_a_buf->buffer_name,
+             in_a_frame_id, a[0], a[_num_elements / 2], a[_num_elements - 1]);
+        INFO("Input buffer {:s}[{:d}]: {:f}, ..., {:f}, ..., {:f}", in_b_buf->buffer_name,
+             in_b_frame_id, b[0], b[_num_elements / 2], b[_num_elements - 1]);
+        INFO("Output buffer {:s}[{:d}]: {:f}, ..., {:f}, ..., {:f}", out_buf->buffer_name,
```
andrerenard, line 98: Actually it might be worth making the `SampleStage` do this output, then people could see how data moves between stages.

lib/stages/DotProduct.cpp
```diff
@@ -0,0 +1,107 @@
+#include "DotProduct.hpp"
+
+#include "StageFactory.hpp"
+#include "errors.h"
+#include "visUtil.hpp" // for frameID, modulo
+
+/* Load the classes we will be using
+ * TODO: Annotate the includes and usings */
```
andrerenard, line 8: I don't think you need to do this.

config/examples/dot_product.yaml
```diff
@@ -0,0 +1,60 @@
...
+  num_frames: 10
+  frame_size: 128 * sizeof_float
+  kotekan_buffer: standard
+  metadata_pool: main_pool
+
+# Fill input buffers with data
+data_gen:
+  data_gen_a:
+    # Fills each element of a each frame with a constant value
+    kotekan_stage: testDataGenFloat
```
andrerenard, line 40: `testDataGenFloat` is a fairly complex stage.  For someone just learning it's a lot to try and understand.  It might be worth having three stages here: `ExampleProducer`, `ExampleDotProduct` and `ExampleConsumer`.  Then replace the `hexDump` stage with `SampleStage` -> `ExampleConsumer`.

config/examples/dot_product.yaml
```diff
@@ -0,0 +1,60 @@
...
+  input_b_buffer:
+    kotekan_buffer: standard
+    metadata_pool: main_pool
+
+# Construct output buffer
+output_buffer:
+  num_frames: 10
+  frame_size: 128 * sizeof_float
+  kotekan_buffer: standard
+  metadata_pool: main_pool
```
andrerenard, line 34: I believe you can actually set the metadata pool to `none`, which makes this example simpler - and avoids CHIME specific code.

docs/sphinx/dev/dev_dot_product_example.rst
```diff
@@ -0,0 +1,67 @@
+********************
+Dot Product Example
+********************
+The following example outlines how to create a dot product between two vectors: A and B, using a `kotekan` pipeline.
+
```
andrerenard, line 5: I would generate a graph of what this pipeline looks like visually.  e.g. with the existing dot system or the new javascript debug tool.  I think it's helpful to get a visual description of the pipeline along with the written info.

docs/sphinx/dev/dev_consumer_stage_tutorial.rst
```diff
@@ -0,0 +1,71 @@
...
+
+Now, let us create the header file.
+
+.. literalinclude:: ../../../lib/stages/SampleStage.hpp
+
+To let the compiler know about the stage, add it to lib/stages/CMakeLists.txt.
+
+The compiling instructions can then be found here: http://lwlab.dunlap.utoronto.ca/kotekan/compiling/general.html.
+
+Finally, let us write some basic system tests, so that we can make sure our stage is behaving as expected while we work on it.
```
andrerenard, line 41: I would break up creating a unit test into its own page.  The unit test framework is a bit too complex for what should be a very simple example.   Instead just describe how to run the code here, and move the unit test details into another section.

andrerenard: For this to work we also need to have a very simple yaml file with say the `ExampleProducer` and renamed `ExampleConsumer`.   I would also include a diagram of this very simple pipeline in the text here.

andrerenard (changes requested):
This is looking great.  Although I do have some comments I'd like to address below.

andrerenard:
While I'd like to see this expanded to include a `FrameView` example, for now just having a basic non-FrameView set of examples would be really great, so let's limit the scope to just basic kotekan buffers without views or even metadata for now.   We might also consider moving the unit tests part of this PR into another PR, just to focus in on making the concepts of a `producer -> consumer` and `producer -> consumer/producer` -> consumer` pipeline as simple and clear as possible.

config/examples/dot_product.yaml
james-s-willis: When I set `metadata_pool: none` I get the following error:
```
Buffer: input_a_buffer Registering producer: /data_gen/data_gen_a
Buffer: input_b_buffer Registering producer: /data_gen/data_gen_b
Registering consumer /dot_product for buffer input_a_buffer
Registering consumer /dot_product for buffer input_b_buffer
Buffer: output_buffer Registering producer: /dot_product
Registering consumer /screen_dump for buffer output_buffer
restServer: Adding GET endpoint: /buffers
restServer: Adding GET endpoint: /pipeline_dot
Starting kotekan_stage: /data_gen/data_gen_a...
/data_gen/data_gen_a: Setting thread affinity
No metadata pool on input_a_buffer but metadata was needed by a producer
Starting kotekan_stage: /data_gen/data_gen_b...
/data_gen/data_gen_b: Setting thread affinity
Segmentation fault (core dumped)
```

lib/stages/DotProduct.cpp
james-s-willis: What do you mean by this? Make the `SampleStage` print the data inside one buffer? Or for it to take 2 buffers and print their data?

dstndstn:
I'm wondering what the plan is for this PR.  I didn't notice it and I started reinventing this -- it would be a great improvement to have a partial tutorial merged in and deployed to the documentation site, even if it's not perfect!  I would consider working on this PR if you give me a wish-list of things to improve.  Thanks!

andrerenard:
@dstndstn That would be amazing!  I've been looking for someone to pick this up again, and I'd love to get into docs.   Let me revisit this PR, and make a list of minimal changes to get it merged.  Perhaps we could talk about it for ~10 minutes or so after the BX-engine call on Monday morning?  That might be the fastest way to cover some of details.

docs/sphinx/dev/dev_dot_product_example.rst
```diff
@@ -0,0 +1,74 @@
...
+.. literalinclude:: ../../../lib/stages/ExampleDotProduct.cpp
+    :language: c++
+
+Compile Stage
+-------------
+To compile the stage add the source file to `lib/stages/CMakeLists.txt`:
+
+.. literalinclude:: ../../../lib/stages/CMakeLists.txt
+    :lines: 73-78
```
andrerenard: These lines could change over time, I think it's better to just directly copy in some code.   Also I think this should go in the `Stage Tutorial` part of the documentation, since it's really something that needs to be done when adding a new stage, not writing a pipeline.

docs/sphinx/dev/dev_dot_product_example.rst
```diff
@@ -0,0 +1,74 @@
...
+    :linenos:
+
+Move to the `build` directory, call `cmake` to create the `Makefile` and compile `kotekan`:
+
+.. code-block:: bash
+
+    cd build
+    cmake -DCMAKE_BUILD_TYPE=Debug -DWITH_TESTS=ON ..
```
andrerenard: The `-DWITH_TESTS=ON` option adds boost dependencies and adds a lot of build time.   I think for this example, we shouldn't include it.

lib/stages/ExampleDotProduct.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+    if (in_a_buf->frame_size != out_buf->frame_size) {
+        throw std::runtime_error(
+            fmt::format(fmt("Input frame size does not match output frame size. {:d} != {:d}"),
+                        in_a_buf->frame_size, out_buf->frame_size));
+    }
+
+    // Load options that can be set in config
+    // unique_name_for_stage, name_of_config, default_value_if_not_set
+    //_offset = config.get_default<int32_t>(unique_name, "offset", 0);
```
andrerenard: Remove
```suggestion
```

lib/stages/ExampleDotProduct.cpp
```diff
@@ -0,0 +1,96 @@
...
+    // Until the thread is stopped
+    while (!stop_thread) {
+
+        // Acquire input frames
+        uint8_t* frame_a_ptr = wait_for_full_frame(in_a_buf, unique_name.c_str(), in_a_frame_id);
+        uint8_t* frame_b_ptr = wait_for_full_frame(in_b_buf, unique_name.c_str(), in_b_frame_id);
+
+        // A null frame is returned on shutdown
+        if (frame_a_ptr == nullptr)
+            break;
```
andrerenard, line 64: The check should happen immediately after the `wait_for_full_frame()` call.

lib/stages/ExampleDotProduct.hpp
```diff
@@ -0,0 +1,44 @@
...
+#include "util.h"
+
+/**
+ * @class ExampleDotProduct
+ * @brief A stage to compute the dot product between two vectors: A and B, which are represented by
+ * two buffers. The result is written to an output buffer.
+ *
+ * @par Buffers
+ * @buffer in_a_buf The input buffer representing vector A.
+ *      @buffer_format standard
```
andrerenard:
```suggestion
 *      @buffer_format Array of floats
```
And same for the other buffers below.

lib/stages/ExampleDotProduct.hpp
```diff
@@ -0,0 +1,44 @@
...
+ *      @buffer_metadata any
```
andrerenard:
```suggestion
 *      @buffer_metadata none
```
Let's either say these stages don't have metadata, or add an `ExampleMetadata` type and show how it is used in the stages.

lib/stages/ExampleProducer.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+private:
+    // Output buffer
+    Buffer* out_buf;
+
+    // Initialised value
+    float _init_value;
+
+    // Length of frame
+    uint32_t _num_elements;
```
andrerenard: I don't like the use of `num_elements` here, that has a specific meaning in other parts of the system.   I also wouldn't make this a member variable, it can just be local to `main_thread()` as something like `frame_length`.

lib/stages/ExampleProducer.hpp
```diff
@@ -0,0 +1,52 @@
...
+     * @brief Constructor for the stage
+     *   Note: you can use the macro STAGE_CONSTRUCTOR(ExampleProducer)
+     *   if your constructor does not need additional customisation
+     *   and you wish to hide the complexity.
+     */
+    ExampleProducer(kotekan::Config& config, const std::string& unique_name,
+                    kotekan::bufferContainer& buffer_container);
+
+    /**
+     * @brief Deconstructor - what happens when Kotekan shuts down.
```
andrerenard:
```suggestion
     * @brief Deconstructor - Called on shutdown, after @c main_thread has exited.
```

lib/stages/ExampleConsumer.hpp
```diff
@@ -0,0 +1,48 @@
...
+ *      @buffer_metadata any
+ *
+ */
+class ExampleConsumer : public kotekan::Stage {
+public:
+    /**
+     * @brief Constructor for the stage
+     *   Note: you can use the macro STAGE_CONSTRUCTOR(ExampleConsumer)
+     *   if your constructor does not need additional customisation
+     *   and you wish to hide the complexity.
```
andrerenard, line 25: I think this node should move to the CPP file, it doesn't apply to the header file.  We should also setup the examples such that the first stage in the tutorial doesn't include the macro, but later stages do.   Or perhaps use the macro everywhere, but in a commend show the expanded version.

docs/sphinx/dev/dev_dot_product_example.rst
```diff
@@ -0,0 +1,74 @@
+********************
+Dot Product Example
```
andrerenard: I'm tempted to rename this just `Example Pipeline`.  The fact it's a dot product shouldn't really be important to the reader, who very likely wants to just learn how to do something like it.

docs/sphinx/dev/dev_dot_product_example.rst
```diff
@@ -0,0 +1,74 @@
...
+.. literalinclude:: ../../../lib/stages/ExampleDotProduct.cpp
+    :language: c++
+
+Compile Stage
+-------------
+To compile the stage add the source file to `lib/stages/CMakeLists.txt`:
+
+.. literalinclude:: ../../../lib/stages/CMakeLists.txt
+    :lines: 73-78
+    :linenos:
```
andrerenard: This feels like it belongs in another section.  I'd like this page to focus on the config file, and running the pipeline only.

config/examples/dot_product.yaml
```diff
@@ -0,0 +1,62 @@
...
+sizeof_float: 4
+
+cpu_monitor:
+    enabled: true 
+    track_length: 2 # save last 2 mins cpu usage.
+
+# Create pool of metadata for each buffer
+main_pool:
+  kotekan_metadata_pool: chimeMetadata
+  num_metadata_objects: 30
```
andrerenard: If we want to include the concept of metadata in this tutorial, then it should be done with an `ExampleMetadata` type not the CHIME Metadata, and we should do something with it in the example stages.  For simplicity, I think we can just remove it entirely for now.  Along with the lines `metadata_pool: main_pool` below.  The buffer will default to no metadata if a pool isn't given.

config/examples/dot_product.yaml
```diff
@@ -0,0 +1,62 @@
...
+cpu_affinity: [0,1,2,3]
+num_local_freq: 1
+samples_per_data_set: 4608
+
+# Constants
+sizeof_float: 4
+
+cpu_monitor:
+    enabled: true 
+    track_length: 2 # save last 2 mins cpu usage.
```
andrerenard: We should remove this entirely, it's just confusing to new users.

config/examples/dot_product.yaml
```diff
@@ -0,0 +1,62 @@
+---
+type: config
+log_level: DEBUG
+cpu_affinity: [0,1,2,3]
+num_local_freq: 1
+samples_per_data_set: 4608
```
andrerenard: Strange length?  Let's make it simple.
```suggestion
samples_per_data_set: 1024
```

config/examples/dot_product.yaml
```diff
@@ -0,0 +1,62 @@
+---
+type: config
+log_level: DEBUG
+cpu_affinity: [0,1,2,3]
+num_local_freq: 1
```
andrerenard: Unused. 
```suggestion
```

config/examples/dot_product.yaml
```diff
@@ -0,0 +1,62 @@
+---
+type: config
+log_level: DEBUG
+cpu_affinity: [0,1,2,3]
```
andrerenard: Some simple comments where would be good in case the reader is going out of order and just looking at the file.

andrerenard (changes requested):
Here are some high level thoughts on this:
It appears this PR is currently trying to get at three different areas.  
1. Writing stages and building a basic pipeline with them
2. Creating a python unit test
3. Updating some of the notes around the pipeline viewer tool

To simplify things, I would move 2. into another PR, since I think that needs a bit of work still and adds too much complexity for this introduction.

The changes around 3 can likely just stay here since they are fairly simple. 

The structure of the tutorial needs a little work.  There is the `Consumer Stage tutorial` and the `Dot Product Example`, seem to be a mix of writing stages and configs.   I think we should really have something like a `Stage tutorial` or `Writing new stages` which covers all three of common stage type: consumer, producer, and consumer/producer examples.  Once you cover the first case it's easy to expand to the other two cases - and I think can all be in one document.  e.g. start with what's there, add just build on it to add the `ExampleProducer` and `ExampleDotProduct` 

For the dot product example, that should really just be called something like `Example Pipeline Config File`, or `Writing a Pipeline Configuration File`, and it should be made clear it's the followup to the `Stage Tutorial`.   

Smaller comments inline.

andrerenard:
Also we should leave all the advanced concepts out of this tutorial for now.  e.g. FrameViews, Updatable Config, Metadata, the Prometheus system, etc. that should be covered as other chapters of the Developers Guide, or as advanced tutorials to add later.

dstndstn:
I looked at the IWYU complaints, but the example code style seems to be in line with the rest of the codebase -- eg the include patterns are very similar in
https://github.com/kotekan/kotekan/tree/develop/lib/stages
and
https://github.com/kotekan/kotekan/blob/develop/lib/stages/BadInputFlag.cpp

so is it really a blocker for merging?

andrerenard (approved):
This is looking really great.  The `example_test.py` feels a bit out of place, since it's using more advanced `vis` stages with FrameViews.  However, I don't see any reason not to merge this now, since it's super useful to have the basic tutorial, and we can add to it from here with the more advanced ideas that lead into `example_test.py`.

andrerenard:
I think the IWYU issues may be the result of a version changes in one of the docker images.  I think that might actually get resolved in another PR, for now I agree, we can just merge this in.

PR #686, opened 2020-02-24
Shared Buffer for Kotekan that has a time and frequency index

anjakefala:
An expansion of: https://github.com/kotekan/kotekan/pull/688. PR comments from there also apply to here.

This code will create two shared memory regions:

    fname_access_record (of size ntime * uint64_t)
        keeps a record of fname_buf's access
    fname_buf (of size ntime * frame_size)
    - each region will contain frame metadata, frame data, and then a valid bit

It will also create a semaphore (sem_name) which will control access to fname_access_record.

The process:

fname_access_record is locked. -1 is written to `time_ind` to represent that that index is invalid. fname_access_record is unlocked.
Current frame metadata, and data (along with a 'valid bit') is written at `time_ind` and `freq_ind` of fname_buf.
fname_access_record is locked. The `fpga_seq_start` for the written frame is written to `time_ind` to represent when that region became valid. fname_access_record is unlocked.

For `time_ind`:
* if `time_ind` is less than the current indexed `min_time`, it is dropped.
* if `time_ind` is not indexed and between `min_time` and `max_time`, it is dropped.
* if `time_ind` is indexed in the buffer, then the frame is inserted at the appropriate location.
* if `time_ind` is not indexed and greater than `max_time`, all of the frames at `min_time` is dropped, and that shared memory region is now `time_ind`.

Todo list (added by Richard):

- [x] Add class documentation
- [x] Remove the `sem_name` parameter and rename `fname_buf` -> `fname`
- [x] Fix potential segfault. We need to check that `freq_state` (in ::main_thread) before using it. If it's null we should probably throw a FATAL_ERROR.
- [x] validate(): ignore invalid frames (filter all data with `[visraw.valid_frames.astype(np.bool)]`)
- [x] Add note that the buffer interface guarantees that the input and prod datasetStates won't change
- [x] add option to VisRaw to get index_maps from comet's states
  - input
  - prod
  - freq
  - stack
  - additionally get update_ids from gains/flags
- [x] run validation test with a buffer size of 1024 freqs * 256 time samples and 135 inputs to have a frame size comparable to reality
- [x] Anywhere you would print an ERROR and return should probably be a FATAL_ERROR if we cannot recover from it
- [x] VisRaw: flatten `.time` (has unnecessary frequency dimension
- [x] check that the data size has not changed; throw a FATAL_ERROR if it has
- [x] Using prometheus to export the length of time that you wait to acquire the semaphore. Then we can alert of dangerous slow downs.
- [ ] make structured data variables, that should not change over runtime, const or macro variables (This is a larger polish item, that should be added to a list of clean-ups)
- [x] Rename params `fname` -> `name`; `sem_wait_time` -> `wait_time`; `nsamples` -> `num_samples`
- [x] add_sample returns a bool, but this is ignored
- [x] Remove gating from critical states

- [ ] Update diagram and move to better location
- [x] fix unit tests: occasionally times out in github action (runs 5h then cancels without giving access to the logs)


lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+    // Set up the structure of the access record shared memory
+    size_t access_record_size = sizeof(uint64_t);
+
+    record_addr = assign_memory<uint64_t*>(fname_access_record, ntime * access_record_size, record_addr);
+
+    // Get properties of stream from first frame and datasetManager
+    std::map<uint32_t, uint32_t> freq_id_map;
+    auto& dm = datasetManager::instance();
+    auto sstate_fut = std::async(&datasetManager::dataset_state<stackState>, &dm, frame.dataset_id);
```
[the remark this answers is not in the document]
jrs65: You should be continually checking that the frame length hasn't changed though. That's a FATAL_ERROR condition.

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,295 @@
...
+    nfreq = fstate->get_freqs().size();
+    size_t ninput = istate->get_inputs().size();
+    size_t nvis = sstate ? sstate->get_num_stack() : pstate->get_prods().size();
+
+    // Set the alignment (in kB)
+    size_t alignment = 4; // Align on page boundaries
+
+    // Calculate the ring buffer stricture
+    auto layout = visFrameView::calculate_buffer_layout(ninput, nvis, num_ev);
+    data_size = layout.first;
```
jrs65: I think this is the only reason you have done all the dM stuff above. You can do this in three other ways:
- Look if up from the buffer. This is not ideal because often the buffer length is just some number I made up that I know if larger than the frame size.
- Make the same `calculate_buffer_layout` call you do above, but fetch the `ninput`, `nvis` and `num_ev` arguments from the incoming data (they're exposed by the `visFrameView`)
- Last, and probably the best option, add a method (or expose a reference) to the data size on `visFrameView` directly. This you could do by adding a public `const size_t& data_size` to `visFrameView` and then in the constructors initializers list adding `data_size(buffer_layout.first)`.

jrs65:
Okay. I've added a bunch of specific review comments. Here are some more general comments...

First, I would move to a single shared memory region. It seems less flaky to me. I could be argued down from that though.

Second, you should include the information that the readers will need to decode the structure of the buffer. In the most generic case that is: number of times, number of frequencies, metadata size and data size. The latter might make sense to break out into the relevant structural params (number of inputs, prods and eigenvalues). I would recommend defining a `struct` to contain them along with a flag that specifies if the writer is still going (like we discussed yesterday). This should be the *first* entry in the shared memory region, so it can be read without needing to understand the structure. After that you should place the access record, and then the data section.

jrs65:
Another comment. You probably should do some validation important aspects of the incoming data haven't changed. At the very least you should check the data size has not changed (and if you write the number of inputs, prods etc. into the buffer as discussed above, you should check those individually).

An extra check you can do which would be good, would be to check to see that the important dataset states haven't not changed. That should be done with the `fingerprint` routine (https://github.com/kotekan/kotekan/blob/develop/lib/utils/datasetManager.hpp#L281). You give it a list of important states, it gives you a hash of them, and then if they change you throw a FATAL_ERROR.

anjakefala:
Other things from the other PR:
* add documentation
* what are we doing upon shut-down
* this comment: https://github.com/kotekan/kotekan/pull/688/files#r388632823

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+    if ( vis_time_ind_map.count(t) != 0) {
+        // if the time is already indexed, write to memory at that location
+        write_to_memory(frame, vis_time_ind_map.at(t), freq_ind);
+        return true;
+    }
+    else if (vis_time_ind_map.size() < ntime) {
+        // if there is still empty space in the shared buffer
+        // add an additional time index
+        cur_pos = (cur_pos + 1) % ntime;
```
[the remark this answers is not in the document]
anjakefala: Using modulo<int> per discussed!

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+    size_t offset = (time_ind * nfreq + freq_ind) * frame_size;
+
+    // each time, we block off an entire time_ind
+
+    CHECK(sem_wait(sem));
+    memcpy(record_addr + time_ind, &in_progress, sizeof(in_progress));
+    CHECK(sem_post(sem));
+
+    memcpy(buf_addr + offset, frame.metadata(), metadata_size);
```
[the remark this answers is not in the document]
nritsche: agree (see https://github.com/kotekan/kotekan/pull/688#discussion_r388643538)

anjakefala:
*  sem_wait() can potentially lock forever if the reader does something funny. We could set a timeout and do sth in case it gets hit (like dropping frames, print or exit).

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,295 @@
...
+    // Set up the structure of the ring buffer shared memory
+    // Get one frame for reference
+    wait_for_full_frame(in_buf, unique_name.c_str(), frame_id);
+
+    auto frame = visFrameView(in_buf, frame_id);
+
+    // Set up the structure of the access record shared memory
+    size_t access_record_size = sizeof(uint64_t);
+
+    record_addr = assign_memory<uint64_t*>(fname_access_record, ntime * access_record_size, record_addr);
```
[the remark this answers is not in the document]
nritsche: In C I would argue for `void *` (like malloc). In C++ that requires casting, overriding type-checking.
If we merge the two shared memory files, this doesn't need to be generic anymore anyways.

nritsche:
I think having a single shared memory region only is a good idea. I wonder if we should model the content of this region, though, instead of implicitly describing it in pointer arithmetic.

nritsche:
> Second, you should include the information that the readers will need to decode the structure of the buffer. In the most generic case that is: number of times, number of frequencies, metadata size and data size. The latter might make sense to break out into the relevant structural params (number of inputs, prods and eigenvalues). I would recommend defining a `struct` to contain them along with a flag that specifies if the writer is still going (like we discussed yesterday). This should be the _first_ entry in the shared memory region, so it can be read without needing to understand the structure. After that you should place the access record, and then the data section.

Shouldn't it be enough to write the dataset ID, so that the reader can access the "relevant structural params" via comet?

anjakefala:
> So at minimum the number of time samples, frequencies, and the metadata, data   and frame sizes.                                                               

> The writer should throw a fatal error if it detects a dataset ID change that    would cause any of those to change.

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,340 @@
...
+    if (sem_post(sem) == -1) {
+        FATAL_ERROR("Failed to release semaphore {}", sem_name);
+        return;
+    }
+
+    memcpy(buf_write_pos + 4, frame.metadata(), metadata_size);
+    memcpy(buf_write_pos + metadata_size + 4, frame.data(), data_size);
+    memcpy(buf_write_pos, &one, 4);
+
+    uint64_t fpga_seq = frame.metadata()->fpga_seq_start;
```
nritsche: Richard: "oops. I said to use this but I was wrong. Rick has pointed out the error of my ways, and I think we should just go back to having an incrementing counter here."

anjakefala: Hah!

lib/stages/visSharedMemWriter.cpp
```diff
@@ -300,7 +298,10 @@ void visSharedMemWriter::main_thread() {
     frame_size = _member_alignment(data_size + metadata_size + 4, alignment * 1024);
 
     // memory_size should be ntime * nfreq * file_frame_size (data + metadata)
-    buf_addr = assign_memory(fname_buf, ntime * nfreq * frame_size);
+    buf_addr = assign_memory(fname_buf, (ntime * access_record_size) + (ntime * nfreq * frame_size));
```
nritsche: - this is missing the valid field, I think
- We need an access record per time per frequency

tests/test_vissharedmemwriter.py
```diff
@@ -0,0 +1,102 @@
...
+
+import pytest
+import posix_ipc
+import mmap
+import os
+
+from kotekan import runner, shared_memory_buffer
+
+sem_name = "kotekan"
+fname_access_record = "calBufferAccessRecord"
```
anjakefala: We can now erase this! They are all one.

tests/test_vissharedmemwriter.py
```diff
@@ -0,0 +1,102 @@
...
+    memory = posix_ipc.SharedMemory(fname_buf)
+    mapfile = mmap.mmap(memory.fd, memory.size, prot=mmap.PROT_READ)
+    os.close(memory.fd)
+    yield mapfile
+    mapfile.close()
+    posix_ipc.unlink_shared_memory(fname_buf)
+
+
+@pytest.fixture(scope="module")
+def mem_map_access_record():
```
anjakefala: And we do not need this

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+    // each time, we block off an entire time_ind
+
+    CHECK(sem_wait(sem));
+    memcpy(record_addr + time_ind, &in_progress, sizeof(in_progress));
+    CHECK(sem_post(sem));
+
+    memcpy(buf_addr + offset, frame.metadata(), metadata_size);
+    memcpy(buf_addr + offset + metadata_size, frame.data(), data_size);
+    memcpy(buf_addr + offset + metadata_size + data_size, &one, 1);
```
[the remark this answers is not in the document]
nritsche: What's the advantage of using a 4byte data type over using a char plus padding (more obvious)?

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,350 @@
...
+    else {
+        INFO("Dropping integration as buffer (FPGA count: {:d}) arrived too late (only accepting new times greater than {:d})", t.fpga_count, max_time.fpga_count);
+        return false;
+    }
+}
+
+void visSharedMemWriter::reset_memory(uint32_t time_ind) {
+
+    uint8_t *buf_write_pos = buf_addr + (time_ind * nfreq * frame_size);
+    uint64_t *record_write_pos = record_addr + time_ind;
```
nritsche: This needs to be changed to reflect a 2 dimensional access record (per time, per frequency)

lib/stages/visSharedMemWriter.hpp
```diff
@@ -0,0 +1,80 @@
...
+    uint64_t* structured_data_addr;
+    uint64_t *access_record_addr;
+    uint8_t *buf_addr;
+    size_t ntime;
+
+    const uint8_t valid = 1;
+    size_t valid_size = 4;
+    const int64_t in_progress = -1;
+    uint64_t num_writes = 0;
+    size_t structured_data_num = 6;
```
nritsche: can this be const?

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,393 @@
...
+}
+
+void visSharedMemWriter::main_thread() {
+    INFO("Reached main thread");
+
+    frameID frame_id(in_buf);
+
+    // The current position in the ring buffer of the most recent time sample
+    // from 0 -> ntime
+    cur_pos = modulo<int>(ntime);
```
nritsche: I don't think this is equivalent to `cur_pos = 0`. Instead make `cur_pos` be of type `modulo<int>` (or `modulo<size_t>`?) in `visSharedMemWriter.hpp`

nritsche: note that you have to cast it back to int in line 111

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,389 @@
...
+    // calculate the time index for time sample t, add the frame for time sample t at position frequency index
+    //
+    // curr_pos always points to the ring buffer index for the most recent time
+
+    if ( vis_time_ind_map.count(t) != 0) {
+        // if the time is already indexed, write to memory at that location
+        write_to_memory(frame, vis_time_ind_map.at(t), freq_ind);
+        return true;
+    }
+    else if (vis_time_ind_map.size() == 0) {
```
nritsche: not sure, but can this branch be removed if you swap lines 118 and 119?

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,389 @@
...
+        // fd is no longer needed
+        if (close(fd) == -1) {
+            throw std::runtime_error(
+                fmt::format(fmt("Failed to close file descriptor for shm {:s}: {:s}."), shm_name, strerror(errno)));
+        }
+
+        return addr;
+}
+
+bool visSharedMemWriter::add_sample(const visFrameView& frame, time_ctype t, uint32_t freq_ind) {
```
nritsche: the mix of `return` and `else` in this method confused me for a good 20 minutes :spaghetti:

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,389 @@
...
+    if (sem_post(sem) == -1) {
+        FATAL_ERROR("Failed to post semaphore {}", sem_name);
+        return;
+    }
+
+    INFO("Resetting ring buffer memory at position time_ind: {}\n", time_ind);
+    // set the full time_ind to 0 in the ring buffer
+    std::vector<char> zeros(nfreq * frame_size, 0);
+    tmp = zeros.data();
+    memcpy(buf_write_pos, &tmp, sizeof(tmp));
```
nritsche: see the comment on the memcpy above

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,389 @@
...
+        // this data is older than anything else in the map, so we should
+        // just drop it
+        INFO("Dropping integration as buffer (FPGA count: {:d}) arrived too late (minimum in pool {:d})\n", t.fpga_count, min_time.fpga_count);
+        return false;
+    }
+
+    else if (t > max_time) {
+        // we need to drop the oldest time
+        cur_pos++;
+        reset_memory(cur_pos);
```
nritsche: why is this necessary?

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,389 @@
...
+}
+
+void visSharedMemWriter::reset_memory(uint32_t time_ind) {
+
+    // resets all memory at time_ind to 0s
+
+    uint8_t *buf_write_pos = buf_addr + (time_ind * nfreq * frame_size);
+    uint64_t *access_record_write_pos = access_record_addr + (time_ind * nfreq * access_record_size);
+
+    INFO("Resetting access_record memory at position time_ind: {}\n", time_ind);
```
nritsche: this should not be on info level

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,389 @@
...
+    }
+    std::vector<char> invalid_vector(nfreq * access_record_size, invalid);
+    char* tmp = invalid_vector.data();
+    memcpy(access_record_write_pos, &tmp, sizeof(tmp));
+    if (sem_post(sem) == -1) {
+        FATAL_ERROR("Failed to post semaphore {}", sem_name);
+        return;
+    }
+
+    INFO("Resetting ring buffer memory at position time_ind: {}\n", time_ind);
```
nritsche: this should not be on info level

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,389 @@
...
+    INFO("Memory reset\n");
+}
+
+void visSharedMemWriter::write_to_memory(const visFrameView& frame, uint32_t time_ind, uint32_t freq_ind) {
+    // write frame to ring buffer at time_ind and freq_ind
+
+    uint8_t *buf_write_pos = buf_addr + ((time_ind * nfreq + freq_ind) * frame_size);
+    uint64_t *access_record_write_pos = access_record_addr + (time_ind * nfreq + freq_ind);
+
+    INFO("Writing ringbuffer to time_ind {} and freq_ind {}\n", time_ind, freq_ind);
```
nritsche: this shouldn't be on info level

anjakefala:
A comment from Richard that I just noticed:

On data size in the structured data:

> The latter might make sense to break out into the relevant structural params (number of inputs, prods and eigenvalues)

lib/stages/visSharedMemWriter.cpp
```diff
@@ -353,9 +350,10 @@ void visSharedMemWriter::main_thread() {
     memcpy(structured_data_addr + 5, &data_size, sizeof(data_size));
 
     // initially set the address records with -1
-    std::vector<char> invalid_vector(ntime * nfreq * access_record_size, invalid);
-    char* tmp = invalid_vector.data();
-    memcpy(access_record_addr, &tmp, sizeof(tmp));
+    // std::fill_n(array, 100, -1);
+    int64_t invalid_vector [ntime * nfreq];
+    std::fill_n(invalid_vector, ntime*nfreq, invalid);
+    memcpy(access_record_addr, invalid_vector, ntime * nfreq * sizeof *invalid_vector);
```
nritsche: typo?

lib/stages/visSharedMemWriter.cpp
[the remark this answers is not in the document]
anjakefala: You want me to re-get the frame size for each frame, and check that it has not changed?

lib/stages/visSharedMemWriter.cpp
anjakefala: We want to wipe out the entire old time sample, before we write in a frame for the new time.

We have `time 0`. We received new `time 10`. We want to erase all frames at `time 0` before we replace it with a frame with `time 10` and `freq 200`.

anjakefala: Does that make sense?

My write_to_memory function assumes writing to a specific `time_ind` and `freq_ind`. `reset_memory` wipes an entire `time_ind` in one go.

lib/stages/visSharedMemWriter.cpp
anjakefala: No! Apparently `sizeof *invalid_vector` is equivalent to `sizeof(invalid_vector[0])`. But the latter is definitely more clear, so I think I am going to change to the more clear variant.

lib/stages/visSharedMemWriter.cpp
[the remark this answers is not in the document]
jrs65: The shared buffer can't change structure midway while kotekan is running, but what you get on the buffers can (in theory, though I don't know of any actual place it does). You should guard against this error somehow. Each frames metadata contains the structural parameters, so I don't think you really need to check them directly (plus if we add more of them you'll need to add more checks here), you just need to check that the frame size is the same as what you've written into the shared buffer. You must make sure that the `get_data_size` function does not call the dM at any point though.

lib/stages/VisSharedMemWriter.cpp
```diff
@@ -0,0 +1,390 @@
...
+        throw std::runtime_error("Could not write to shared memory.");
+    }
+
+    // Count the eigenvalue index
+    size_t num_ev;
+    if (eigenvalue_state) {
+        num_ev = eigenvalue_state->get_num_ev();
+    } else {
+        num_ev = 0;
+    }
```
jrs65: Every one of the parameters you are looking up in here are exposed already on the `frame` object, so you can (and should) remove all the dM calls here.

lib/stages/visSharedMemWriter.cpp
nritsche: ooooh, i see. Yes, make sense, thanks!

lib/stages/VisSharedMemWriter.cpp
anjakefala: Can you review? https://github.com/kotekan/kotekan/pull/686/commits/0ff3351f19975d122de8d09b7e3fcd89efd26ed1

lib/stages/VisSharedMemWriter.cpp
```diff
@@ -72,6 +69,28 @@ VisSharedMemWriter::~VisSharedMemWriter() {
     }
 }
 
+bool VisSharedMemWriter::wait_for_semaphore() {
+    // handles timed waits for semaphores
+    // does a standard wait if the system clock is not accessible
+
+    struct timespec ts;
+    if (clock_gettime(CLOCK_REALTIME, &ts) == -1){
+        DEBUG("Failed to get system time.\n Not using timed semaphores.");
```
nritsche: Should this be `INFO` level? Or do we expect this to happen?

nritsche: Could print `std::strerror(errno)` here to find out what's going on: https://linux.die.net/man/3/clock_gettime

lib/stages/VisSharedMemWriter.cpp
```diff
@@ -72,6 +69,28 @@ VisSharedMemWriter::~VisSharedMemWriter() {
...
+        if (sem_wait(sem) == -1) {
+            FATAL_ERROR("Failed to acquire semaphore {}", _sem_name);
+            return false;
+        }
+        return true;
+    }
+
+    ts.tv_sec += 2;
+    if (sem_timedwait(sem, &ts) == -1) {
+        DEBUG("sem_timedwait() timed out\n");
```
nritsche: This should be a higher log level. Maybe even `ERROR`?

lib/stages/VisSharedMemWriter.cpp
```diff
@@ -201,8 +220,8 @@ void VisSharedMemWriter::write_to_memory(const visFrameView& frame, uint32_t tim
 
     // notify that time_ind and freq_ind are being written to, by setting that
     // location to invalid in the access record
-    if (sem_wait(sem) == -1) {
-        FATAL_ERROR("Failed to acquire semaphore {}", _sem_name);
+    if (wait_for_semaphore() == false) {
+        DEBUG("Failed to acquire semaphore {}. Dropping frame.\n", _sem_name);
```
nritsche: If this happens, a reader is not releasing the semaphore. Higher log level? Also, should we publish a prometheus metric counting dropped frames here?

lib/stages/VisSharedMemWriter.cpp
```diff
@@ -221,8 +240,8 @@ void VisSharedMemWriter::write_to_memory(const visFrameView& frame, uint32_t tim
     uint64_t fpga_seq = frame.metadata()->fpga_seq_start;
 
 
-    if (sem_wait(sem) == -1) {
-        FATAL_ERROR("Failed to acquire semaphore {}", _sem_name);
+    if (wait_for_semaphore() == false) {
+        FATAL_ERROR("Failed to acquire semaphore {} before metadata was updated.\n", _sem_name);
```
nritsche: FATAL_ERROR exits kotekan. This way we would notice for sure that there's a problem with the shared memory. It should be in line with the error handling above, though. I'd vote for `ERROR` log and a prometheus metric.

lib/stages/VisSharedMemWriter.cpp
```diff
@@ -265,7 +284,7 @@ void VisSharedMemWriter::main_thread() {
     DEBUG("Semaphore created.\n");
 
     // Acquire semaphore until shared memory is created
-    if (sem_wait(sem) == -1) {
+    if (wait_for_semaphore() == false) {
         FATAL_ERROR("Failed to acquire semaphore {}", _sem_name);
```
nritsche: same as above

lib/stages/VisSharedMemWriter.cpp
anjakefala: Oh, I have not integrated prometheus metrics at all! That would be a good stretch-item to do.

lib/stages/VisSharedMemWriter.cpp
anjakefala: I made that a FATAL_ERROR because the ring buffer got modified, but the metadata did not get updated. That seemed pretty bad. However, worst case scenario in that outcome is that the calibration broker never finds out that that data is available.

python/kotekan/shared_memory_buffer.py
```diff
@@ -0,0 +1,539 @@
...
+            FMT_UINT64_T, self.shared_mem.read(SIZE_UINT64_T)
+        )[0]
+        return (
+            num_writes,
+            num_time,
+            num_freq,
+            size_frame,
+            size_meta,
+            size_data,
+        )
```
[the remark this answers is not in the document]
jrs65: You also can get rid of the obscure macro like size definitions.

python/kotekan/shared_memory_buffer.py
```diff
@@ -0,0 +1,539 @@
...
+
+    def _access_record(self):
+        with self.semaphore:
+            record = np.ndarray(
+                (self.num_time, self.num_freq),
+                np.int64,
+                self.shared_mem,
+                self.pos_access_record,
+                order="C",
+            )
```
jrs65: I think you need to explicitly make a copy in here while holding the semaphore. I'm pretty sure this just creates a view of the underlying memory.

python/kotekan/shared_memory_buffer.py
```diff
@@ -0,0 +1,539 @@
...
+                                )
+                        logger.debug(
+                            "Setting time index map [{}] to {}.".format(
+                                access_record[t][f_i], t_i
+                            )
+                        )
+                        self._time_index_map[access_record[t][f_i]] = t_i
+
+                    # copy the value
+                    # TODO: eliminate extra copy
```
jrs65: You're probably good as I'm pretty sure this doesn't actually copy.

python/kotekan/shared_memory_buffer.py
```diff
@@ -0,0 +1,539 @@
...
+            self.num_freq,
+            num_elements=7,
+            num_stack=28,
+            num_ev=0,
+        )
+
+    def _copy_from_shm(self, times, access_record):
+        # copy data updates within the last n time slots
+        for t in times:
+            for f_i in range(self.num_freq):
```
jrs65: Looping over times and frequencies is going to be very slow. Possibly 200k individual copies. I would recommend figuring out which time slots have changed, and copying over the whole time slot. That'll save you a factor of 1024 in copies.

python/kotekan/shared_memory_buffer.py
```diff
@@ -0,0 +1,539 @@
...
+            self.size_frame,
+            len(idxs),
+            self.num_freq,
+            num_elements=7,
+            num_stack=28,
+            num_ev=0,
+        )
+
+    def _copy_from_shm(self, times, access_record):
+        # copy data updates within the last n time slots
```
jrs65: This doesn't quite do what the comment suggests. It copies out the data in the specified slots.

python/kotekan/shared_memory_buffer.py
jrs65: Yeah. This does not copy. The numpy docs are not super clear, but here's a quick experiment:
```
>>> a = np.arange(10)
>>> print(np.ndarray(3, np.int64, a, 8, order="C").flags)
  C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False
```
Note that `OWNDATA` is False.

python/kotekan/shared_memory_buffer.py
```diff
@@ -0,0 +1,539 @@
...
+
+        self.shared_mem.close()
+        try:
+            posix_ipc.unlink_shared_memory(self.shared_mem_name)
+        except posix_ipc.ExistentialError:
+            logger.debug(
+                "Shared memory file did not exist when trying to unlink from it."
+            )
+
+    def read_last(self, n: int):
```
jrs65: This code needs to run in the cal broker which is Python 2. So no type annotations I'm afraid.

python/kotekan/shared_memory_buffer.py
```diff
@@ -0,0 +1,539 @@
...
+        access_record_after_copy = self._access_record()
+        for t in times:
+            for f in range(self.num_freq):
+                if access_record_after_copy[t][f] != access_record[t][f]:
+                    logger.debug(
+                        "Data at t={}, f={} became invalid while reading it.".format(
+                            t, f
+                        )
+                    )
+                    self._data[self._time_index_map[t], f] = None
```
jrs65: This doesn't work, and is very inefficient. For one, you can't assign `None` to elements of `self._data`. I'm not sure how that passes your tests, so you might want to make sure your tests cover this.

This would be much better off done as something like:
```
self._data[...] = np.where(access_record_after_copy == access_record, 1, 0)
```

Now the `[...]` is because that doesn't work either with things as they currently are. I'll write the comment more above, but it probably makes a lot of sense to use a dtype that exposes the valid field, that way this could be `self._data["valid"] = ...`

python/kotekan/shared_memory_buffer.py
```diff
@@ -0,0 +1,539 @@
...
+        if not self._time_index_map:
+            return None
+
+        # sort time index map
+        last_ts, idxs = list(zip(*sorted(self._time_index_map.items())))
+
+        # select last n time slots
+        idxs = idxs[-n:]
+        return VisRaw.from_nparray(
+            self._data[idxs, :],
```
jrs65: This is going to create your third copy of the data, and *must* be able to be avoided. A complex slice like this always generates a copy, and we can't afford the memory. I think the same very much applies for the `read_since` branch.

python/kotekan/shared_memory_buffer.py
```diff
@@ -0,0 +1,539 @@
...
+        self.pos_access_record = self.size_structural_data
+
+        self.size_data = self.len_data * self.size_frame
+        self.pos_data = self.pos_access_record + self.size_access_record
+
+        self._initial_validation()
+
+        os.close(shared_mem.fd)
+
+        self._data = np.ndarray(
```
jrs65: Given that you don't have the vis structural parameters here (i.e. num_inputs) etc, you can't create a full view, but you can create a datatype which exposes three parts: the explicit valid field, and then you can just have `np.void` regions for the metadata and data sections. That will allow you to easily manipulate the valid field as suggested below.

lib/stages/VisSharedMemWriter.cpp
```diff
@@ -0,0 +1,379 @@
...
+    uint64_t fpga_seq = frame.metadata()->fpga_seq_start;
+
+
+    if (wait_for_semaphore() == false) {
+        ERROR("Failed to acquire semaphore {} before metadata was updated.\n", _sem_name);
+        return;
+    }
+
+    DEBUG("Writing fpga_seq {} to time index {}\n", fpga_seq, time_ind);
+    memcpy(access_record_write_pos, &fpga_seq, sizeof(fpga_seq));
```
jrs65: Again, if you're writing a scalar. Don't use `memcpy`. I really am starting to think you'd be better using a span.

lib/stages/VisSharedMemWriter.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+    // update num_writes
+    num_writes++;
+    memcpy(structured_data_addr, &num_writes, sizeof(num_writes));
```
jrs65: Again. Just assign directly.

Definitely looks like you've been bought off by the National Memcpy Council.

lib/stages/VisSharedMemWriter.cpp
```diff
@@ -0,0 +1,379 @@
...
+    // add valid_size amount of padding
+    memcpy(buf_write_pos + valid_size, frame.metadata(), metadata_size);
+    memcpy(buf_write_pos + metadata_size + valid_size, frame.data(), data_size);
+    memcpy(buf_write_pos, &valid, sizeof(valid));
+
+    // Document the fpga sequence counter for that frame in the access record
+    uint64_t fpga_seq = frame.metadata()->fpga_seq_start;
+
+
+    if (wait_for_semaphore() == false) {
```
jrs65: You should probably think what the correct behaviour is here.

lib/stages/VisSharedMemWriter.cpp
```diff
@@ -0,0 +1,379 @@
...
+    if (sem == SEM_FAILED) {
+        FATAL_ERROR("Failed to create semaphore {}", _sem_name);
+        return;
+    }
+
+    DEBUG("Semaphore created.\n");
+
+    // Acquire semaphore until shared memory is created
+    if (wait_for_semaphore() == false) {
+        ERROR("Failed to acquire semaphore {}", _sem_name);
```
jrs65: Anywhere you would print an `ERROR` and return should probably be a `FATAL_ERROR`, as you've already decided you can't recover from it.

lib/stages/VisSharedMemWriter.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+    // memory_size should be _ntime * nfreq * file_frame_size (data + metadata)
+    buf_addr = assign_memory(_fname_buf, (structured_data_size * structured_data_num) + (_ntime * nfreq * access_record_size) + (_ntime * nfreq * frame_size));
+
+    // The elements contained in the structured data and access record are each 64 bytes
+    structured_data_addr = (uint64_t*) buf_addr;
+    access_record_addr = (int64_t*) (structured_data_addr + structured_data_num);
+    buf_addr += (structured_data_size * structured_data_num) + (_ntime * nfreq * access_record_size);
+
+    // Record structure of data
```
jrs65: Structs are a language feature for a reason. Define one for the structural data, and then just copy the whole struct. Not only will your code be cleaner, but defining the struct in the header gives a clear statement about what the interface is.

lib/stages/VisSharedMemWriter.cpp
```diff
@@ -0,0 +1,379 @@
...
+    memcpy(structured_data_addr + 2, &nfreq, sizeof(nfreq));
+    memcpy(structured_data_addr + 3, &frame_size, sizeof(frame_size));
+    memcpy(structured_data_addr + 4, &metadata_size, sizeof(metadata_size));
+    memcpy(structured_data_addr + 5, &data_size, sizeof(data_size));
+
+    // initially set the address records with -1
+    // std::fill_n(array, 100, -1);
+    int64_t invalid_vector [_ntime * nfreq];
+    std::fill_n(invalid_vector, _ntime*nfreq, invalid);
+    memcpy(access_record_addr, invalid_vector, _ntime * nfreq * sizeof(invalid_vector[0]));
```
jrs65: See above comment.
```suggestion
    std::fill_n(access_record_addr, _ntime*nfreq, invalid);
```

lib/stages/VisSharedMemWriter.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+
+        // Wait for the buffer to be filled with data
+        if (wait_for_full_frame(in_buf, unique_name.c_str(), frame_id) == nullptr) {
+            break;
+        }
+
+        // Get a view of the current frame
+        auto frame = visFrameView(in_buf, frame_id);
+
```
jrs65: I'd like there to be some code in here which creates a fingerprint of a set of specified dataset states (inputs, prods, stack, freq, metadata + specified others) and asserts that it hasn't changed. That's done in a few places (e.g. visWriter), so you can crib from there, but the general pattern is.

- First frame:
  * calculate the fingerprint with respect to the interesting states and save it
  * insert the dataset_id of the frame into a set.
- All other frames:
  * check if the dataset_id is in the set:
    * if it is, do nothing and write the data
    * if it is not:
      * calculate the fingerprint of the new dataset, and check that it matches the saved fingerprint
      * if it does, add the dataset_id to the set, and write the data.
      * if it does not, throw an error as the buffer is now receiving data in a way you guaranteed it would.

lib/stages/VisSharedMemWriter.hpp
```diff
@@ -0,0 +1,105 @@
+#ifndef VISSHAREDMEMWRITER_HPP
```
jrs65: This needs a module level docstring.

lib/stages/VisSharedMemWriter.hpp
```diff
@@ -0,0 +1,105 @@
...
+#include <string>               // for string, operator+
+#include "Config.hpp"           // for Config
+#include "Stage.hpp"            // for Stage
+#include "buffer.h"             // for Buffer
+#include "bufferContainer.hpp"  // for bufferContainer
+#include "visBuffer.hpp"        // for visFrameView
+#include "visUtil.hpp"          // for time_ctype
+
+class VisSharedMemWriter : public kotekan::Stage {
+
```
jrs65: This needs a comprehensive class level docstring.

lib/stages/VisSharedMemWriter.hpp
jrs65: This is probably the correct place to add a write up of how the buffer scheme works and how all the data is packed, etc.

lib/stages/VisSharedMemWriter.hpp
```diff
@@ -0,0 +1,105 @@
...
+inline void check_remove(std::string fname) {
+    // Check if we need to remove anything
+    if (access(fname.c_str(), F_OK) != 0)
+        return;
+    // Remove
+    if (remove(fname.c_str()) != 0) {
+        if (errno != ENOENT)
+            throw std::runtime_error("Could not remove file " + fname);
+    }
+}
```
jrs65: This is only used twice, both in the cpp file for this class, and neither time is in a critical loop, so this should just be an internal, non-inline function, in the cpp file.

lib/utils/visBuffer.hpp
```diff
@@ -243,6 +243,14 @@ class visFrameView {
...
     }
 
+    /**
+     * @brief Size of the frame data.
+     * @returns The data size.
+     **/
+    size_t get_data_size() const {
+        return buffer_layout.first;
+    }
+
```
jrs65: Remove me.

python/kotekan/shared_memory_buffer.py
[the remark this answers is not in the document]
jrs65: I noted that @anjakefala  did not define a struct for this. She should, so assume we need something to match that.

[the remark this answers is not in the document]
nritsche: `ctypes.Structure` is indeed much nicer. The problem was that `ctypes.Structure.from_buffer` doesn't work with read-only access. But now I found `ctypes.Structure.from_buffer_copy`.

lib/stages/VisSharedMemWriter.cpp
```diff
@@ -0,0 +1,379 @@
...
+    if (clock_gettime(CLOCK_REALTIME, &ts) == -1){
+        WARN("Failed to get system time. {:d} ({:s})\n Not using timed semaphores.", errno, std::strerror(errno));
+        if (sem_wait(sem) == -1) {
+            ERROR("Failed to acquire semaphore {}", _sem_name);
+            return false;
+        }
+        return true;
+    }
+
+    ts.tv_sec += 2;
```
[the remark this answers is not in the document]
anjakefala: Hi! It is most definitely arbitrary. I asked Rick what would be a reasonable amount, and he suggested 1 second. I then bumped that to 2 seconds. 

I think it should definitely be configurable, that would make modifying it not require an additional deployment. I will make that adjustment. I think it would be hard to determine what it should be without running tests where readers are reading the buffer as kotekan is writing to it.

The behaviour depends on at which point it fails to claim the lock. If the frame has not been writing to the buffer yet, it drops the frame, and moves to the next one. If the frame had been written to the buffer, but that update has not been recorded to the access record yet, the access record does not get updated, and that frame may as well "be considered dropped". 

Do you have an alternative suggested behaviour?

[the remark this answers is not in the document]
anjakefala: I am currently cleaning up all of the logic that determines timeout behaviour, to ensure it is properly logged and consistent.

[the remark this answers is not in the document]
jrs65: To be honest, it's not obvious to me that your should have a time wait at all.

Data comes in at ~100 frames / second, although you'll probably get it within one ~5s burst every 10s. Let's say that the reader blocks things for 10s. That means that I'm going to take a frame, wait for 2s, drop it and then repeat for 5 frames. At the same time I have data backing up in the system which is going to back up for 1000 frames all the way up to the network receive and cause *all* incoming data to get dropped.

So in that sense your timed wait is too long, to prevent a scenario like this you need to make it be around (1/(inbound rate)) ~10ms. However, it's very plausible that the reader might block for that long, and there is enough buffer depth (and the processing rate fast enough), that a 10ms slow down doesn't matter. In that sense, 10ms is too short. 

In all I think this argument suggests that dropping data based on this timed rate isn't the right thing to be doing.

I would suggest doing this:

- Using the timed wait with a really long timeout (more like 2 minutes), and if you wait that long throw a FATAL_ERROR.
- Using prometheus to export the length of time that you wait to acquire the semaphore. Then we can alert of dangerous slow downs.
- Adding a Valve stage before your buffer process in the config to prevent thing backing up.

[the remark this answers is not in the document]
jrs65: It would be good to get @nritsche's input here too.

python/kotekan/shared_memory_buffer.py
nritsche: Here I actually want a copy. If I get a view, it can change after I validated the access record. But doesn't
```
self._data[t_i, f_i] = tmp[:]
```
copy?

lib/stages/VisSharedMemWriter.cpp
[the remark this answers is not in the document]
anjakefala: That work has been pushed.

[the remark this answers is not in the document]
anjakefala: Oh, I had an out-dated page opened and have not seen your comments!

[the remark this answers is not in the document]
nritsche: This make sense. And the valve stage provides a prometheus metric, right?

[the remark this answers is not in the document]
jrs65: Yup! We'll need to ensure there's a large buffer depth in front of it, but we have a boat load of memory on recv2 as well now, so that's easy.

python/kotekan/shared_memory_buffer.py
jrs65: Yup. That does copy. Sorry, if this wasn't clear, I was just inferring (perhaps mistakenly) that the `TODO` above was essentially about whether `tmp` creates and extra copy, and I'm saying that you're good, as it creates a view, and then the assignment below creates a copy.

python/kotekan/shared_memory_buffer.py
nritsche: The chance of this happening in my tests is very low. I have to write a test for it still and then fix it.

python/kotekan/shared_memory_buffer.py
nritsche: Do you have a suggestion?

jrs65: I'm not completely sure what the best way to do this is. I mentioned this elsewhere, but there's essentially three different reads we'll be doing, of which there are only really two different types:
- Infrequent reads of the whole buffer
- Frequency reads of a small number of recents samples

Currently what you're doing is fine for the second class, but not for the first. If we have a hard requirement that we can't do an extra copy for the first case, then that essentially means you must expose the internal copy.

I think what might work is refactoring along this line.

- First ignore the `read_since` case, it's less important, and not obvious how to fit into what I'm about to say. If we can figure out a way then great, if not, then we drop it.
- Rather than separating what's returned by the `read_*` calls and the internal copy you always return the internal copy directly.
- There is no `.read_last(n)` method, `n` is a parameter supplied at construction, and there is an `.update()` method.
- The internal copy is a ring buffer of length `n` in time. When `.update()` is called, we just update the ring buffer with any new data.
- A (read-only) view of the ring buffer is returned directly by the `.update()` call. This means what the caller gets is a contiguous view of the latest n samples (albeit a cyclic one). You could in theory provide a wrapper which re-indexes the cyclic axis.
- In the basic scenario above you would need to open the buffer again for every type of view of the buffer that you want. This may be done more elegantly by having one class which opens and manages the buffer, and that class has a `.view(n)` method which creates a separate `BufferView` object which holds the internal copy and had the `.update()` method to update it from the shared memory buffer.

python/kotekan/visbuffer.py
```diff
@@ -279,12 +279,13 @@ class VisRaw(object):
 
     Parameters
     ----------
-    filename : str
-        Name of file to open.
-    mode : str, optional
-        Mode to open file in. Defaults to read only.
-    mmap : bool, optional
-        Use an mmap to open the file to avoid loading it all into memory.
+    TODO
```
jrs65: You should probably fill in TODO

python/kotekan/visbuffer.py
```diff
@@ -299,38 +300,236 @@ class VisRaw(object):
...
+        time = np.ndarray(
+            shape=(num_time, num_freq),
+            dtype=[
+                (native_str("fpga_count"), np.uint64),
+                (native_str("ctime"), time_spec),
+            ],
+        )
+        for t in range(num_time):
+            for f in range(num_freq):
+                time[t, f] = (fpga_seq[t, f], ctime[t, f])
```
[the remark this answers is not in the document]
nritsche: This doesn't work anymore, because `metadata["ctime"]` has to be converted from `time_spec` to float. I don't think there's a way to still do it like this.

python/kotekan/shared_memory_buffer.py
```diff
@@ -0,0 +1,461 @@
...
+    def _max_ts_per_freq(self, access_record):
+        # get the most recent timestamp for each time slot
+        last_ts = [None] * self.num_time
+        for t in range(self.num_time):
+            for f in range(self.num_freq):
+                if (
+                    last_ts[t] is None and access_record[t, f] != self.invalid_value
+                ) or (last_ts[t] is None or access_record[t, f] > last_ts[t]):
+                    last_ts[t] = access_record[t, f]
+        return last_ts
```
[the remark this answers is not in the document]
nritsche: Thanks!

lib/stages/VisSharedMemWriter.cpp
```diff
@@ -51,8 +53,9 @@ void check_remove(std::string fname) {
 
 VisSharedMemWriter::VisSharedMemWriter(Config& config, const std::string& unique_name,
                                        bufferContainer& buffer_container) :
-    Stage(config, unique_name, buffer_container,
-          std::bind(&VisSharedMemWriter::main_thread, this)) {
+    Stage(config, unique_name, buffer_container, std::bind(&VisSharedMemWriter::main_thread, this)),
+    dropped_frame_counter(Metrics::instance().add_counter(
+        "kotekan_vissharedmemwriter_dropped_frame_total", unique_name, {"fpga_count"})) {
```
jrs65, line 58: `fpga_count` is probably not a great label to use here. Essentially each label creates an entirely new timestream  and so each event dropping frames will be in its own series. Additionally prometheus is basically giving us the time axis anyway, so we don't need to generate our own version of one (although we will lose some time resolution as it pulls even minute or so).

I would recommend dropping the FPGA count label, but adding a freq_id one similar to what visWriter does.

jrs65: You should additionally add a `reason` label to differentiate the two different reasons data can get dropped.

lib/stages/VisSharedMemWriter.cpp
```diff
@@ -192,6 +195,7 @@ bool VisSharedMemWriter::add_sample(const visFrameView& frame, time_ctype t, uin
         INFO("Dropping integration as buffer (FPGA count: {:d}) arrived too late (minimum in pool "
              "{:d})\n",
              t.fpga_count, min_time.fpga_count);
+        dropped_frame_counter.labels({std::to_string(t.fpga_count)}).inc();
```
jrs65, line 198: This should probably get a `reason=late` label.

lib/stages/VisSharedMemWriter.cpp
```diff
@@ -211,6 +215,7 @@ bool VisSharedMemWriter::add_sample(const visFrameView& frame, time_ctype t, uin
         INFO("Dropping integration as buffer (FPGA count: {:d}) arrived too late (only accepting "
              "new times greater than {:d})\n",
              t.fpga_count, max_time.fpga_count);
+        dropped_frame_counter.labels({std::to_string(t.fpga_count)}).inc();
```
jrs65, line 218: Not sure exactly what to call this reason. `reason=order`? As I guess the issue is fundamentally that some data came out of order here, and we got some data from a timesample ahead of this one, before getting anything from this time.

.github/workflows/main.yml
```diff
@@ -80,6 +80,12 @@ jobs:
     steps:
     - uses: actions/checkout@v1
 
+    - name: free disk space
+      run: |
+          sudo apt clean
```
nritsche, line 85: I thought you would do this in the Dockerfile. That would decrease the size of the image but not make the CI build slower

anjakefala: I do not think I understand the difference between the two choices, in terms of why one makes the CI build slower, and not the other.

tests/test_vissharedmemwriter.py
```diff
@@ -0,0 +1,177 @@
...
+
+    ## Test Structured Data
+    num_writes = struct.unpack("<Q", memory_map_buf.read(8))[0]
+    num_time = struct.unpack("<Q", memory_map_buf.read(8))[0]
+    num_freq = struct.unpack("<Q", memory_map_buf.read(8))[0]
+    size_frame = struct.unpack("<Q", memory_map_buf.read(8))[0]
+    size_frame_meta = struct.unpack("<Q", memory_map_buf.read(8))[0]
+    size_frame_data = struct.unpack("<Q", memory_map_buf.read(8))[0]
+
+    assert num_writes == 0
```
jrs65, line 114: I would have thought this should increase through the run.

tests/test_vissharedmemreaderwriter.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+@pytest.fixture()
+def comet_broker():
+    broker_path = shutil.which("comet")
+    if not broker_path:
+        pytest.skip(
+            "Make sure PYTHONPATH is set to where the comet dataset broker is installed."
+        )
+
```
jrs65, line 39: This should test to see if `redis` is available too. See `has_redis` in `test_dataset_broker.py`

tests/test_vissharedmem_validation.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+@pytest.fixture()
+def comet_broker():
+    broker_path = shutil.which("comet")
+    if not broker_path:
+        pytest.skip(
+            "Make sure PYTHONPATH is set to where the comet dataset broker is installed."
+        )
+
```
jrs65, line 37: Same here as in another comment. You should check to see if redis is actually available.

lib/stages/VisSharedMemWriter.cpp
```diff
@@ -121,10 +123,14 @@ void VisSharedMemWriter::wait_for_semaphore() {
...
 
     ts.tv_sec += _wait_time;
+    double start_time = current_time();
+
     if (sem_timedwait(sem, &ts) == -1) {
         FATAL_ERROR("sem_timedwait() timed out");
         return;
     }
+
+    access_record_wait_time_seconds.labels({_name}).set(current_time() - start_time);
```
nritsche, line 133: We can't guarantee, the metric is scraped every time it changes in this line. So wouldn't it make more sense to accumulate the time?

anjakefala: If it is a total accumulated wait time, how do we assess which baseline time is worth our attention?

lib/stages/VisSharedMemWriter.cpp
```diff
@@ -55,7 +55,9 @@ VisSharedMemWriter::VisSharedMemWriter(Config& config, const std::string& unique
                                        bufferContainer& buffer_container) :
     Stage(config, unique_name, buffer_container, std::bind(&VisSharedMemWriter::main_thread, this)),
     dropped_frame_counter(Metrics::instance().add_counter(
-        "kotekan_vissharedmemwriter_dropped_frame_total", unique_name, {"freq_id", "reason"})) {
+        "kotekan_vissharedmemwriter_dropped_frame_total", unique_name, {"freq_id", "reason"})),
+    access_record_wait_time_seconds(Metrics::instance().add_gauge(
+        "kotekan_access_record_wait_time_seconds", unique_name, {"shm_name"})) {
```
nritsche, line 60: I think this one should also be prefixed with `kotekan_vissharedmemwriter_` (see https://github.com/kotekan/kotekan/blob/9be2a6f0a0bc237d6e29f4e5cdb93d4869c5edf7/lib/core/prometheusMetrics.hpp#L305)

python/kotekan/testing.py
```diff
@@ -0,0 +1,504 @@
...
+                else:
+                    logger.info(msg)
+
+    def _validate_time(self, visraw, r):
+        """Validation of all things time."""
+        self._validate_age(visraw, r)
+        self._validate_missing(visraw, r)
+
+    def _validate_missing(self, visraw, r):
+        """Check for missing frames and if this was to be expected."""
```
jrs65, line 434: It would be good to have an explanation of what this test is trying to do. I'm actually not sure, and the description is pretty vague.

tests/conftest.py
```diff
@@ -21,6 +22,17 @@ def pytest_addoption(parser):
...
+def pytest_runtest_call(item):
+    df = subprocess.Popen(["df", "-h"], stdout=subprocess.PIPE)
+    output = df.communicate()[0]
+    print(output.decode())
+    yield
+    df = subprocess.Popen(["df", "-h"], stdout=subprocess.PIPE)
+    output = df.communicate()[0]
+    print(output.decode())
+
+
```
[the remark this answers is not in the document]
anjakefala: I had removed it already! I think you were looking at an older version.

tests/test_vissharedmemwriter.py
anjakefala: The stage closes by the time this test runs!

python/kotekan/visbuffer.py
```diff
@@ -449,36 +512,111 @@ def from_buffer(cls, buffer, size_frame, num_time, num_freq):
...
-            dtype=[
-                (native_str("fpga_count"), np.uint64),
-                (native_str("ctime"), time_spec),
-            ],
+            dtype=[("fpga_count", np.uint64), ("ctime", np.float64),],
         )
+
+        # flatten time index map (we only need one value per time slot, but we have one per
+        # frame/frequency)
         for t in range(num_time):
```
jrs65, line 532: Ideally this would be a vectorized numpy version, but again I guess time is too short to fix this one.

tests/test_vissharedmemwriter.py
nritsche: But then num_writes should be higher than 0, shouldn't it?

anjakefala: No, because num_writes is set to 0 when the stage is shut down.

jrs65: It should set `wait: True` for the `FakeVis` params which means it will produce frames and the correct rate and should not shut down.

PR #688, opened 2020-03-04
Shared buffer for Kotekan that is sequentially written to

anjakefala:
This code will create two shared memory regions:
- fname_access_record (of size ntime * uint64_t)
     - keeps a record of fname_buf's access
- fname_buf (of size ntime * frame_size)
      - each region will contain a valid byte, frame metadata, frame data

It will also create a semaphore (sem_name) which will control access to fname_access_record.

The process:

Buffer is writing to index `i`.
`fname_access_record` is locked.  -1 is written to index `i` to represent that that index is invalid. `fname_access_record` is unlocked.
Current frame metadata, and data (along with a 'valid bit') is written to index `i` of `fname_buf`.
`fname_access_record` is locked. The current timestamp is written to index `i` to represent when that region became valid. `fname_access_record` is unlocked.

When the end of the shared memory region is reached, index `i` resets to 0.

### DIFF
--- a/lib/stages/visSharedMemWriter.cpp
+++ b/lib/stages/visSharedMemWriter.cpp
@@ -50,10 +50,6 @@ visSharedMemWriter::visSharedMemWriter(Config& config, const std::string& unique
         // Acquire semaphore until shared memory is created
         CHECK(sem_wait(sem));
 
-        // memory_size should be ntime * nfreq * file_frame_size (data + metadata)
-        record_addr = assign_memory<uint64_t*>(fname_access_record, ntime * 64, record_addr);
-        buf_addr = assign_memory<uint8_t*>(fname_buf, ntime * 8, buf_addr);
-        INFO("Created the shared memory segments\n");
 
 }
 
@@ -88,6 +84,27 @@ T visSharedMemWriter::assign_memory(std::string shm_name, int shm_size, T addr) 
         return addr;
 }
 
+void visSharedMemWriter::write_to_memory(const visFrameView& frame, size_t index) {
+
+        CHECK(sem_wait(sem));
+        memcpy(record_addr + index, &in_progress, access_record_size);
+        CHECK(sem_post(sem));
+
+        memcpy(buf_addr + index * frame_size, &ONE, sizeof(uint8_t));
+        memcpy(buf_addr + index * frame_size + 1, frame.metadata(), metadata_size);
+        memcpy(buf_addr + index * frame_size + metadata_size + 1, frame.data(), data_size);
+
+        struct timeval timestamp;
+
+        gettimeofday(&timestamp, nullptr);
+        uint64_t time_us = timestamp.tv_sec * 1000000 + timestamp.tv_usec;
+        INFO("Data written to location at {} us", time_us);
+
+        CHECK(sem_wait(sem));
+        memcpy(record_addr + index, &time_us, access_record_size);
+        CHECK(sem_post(sem));
+}
+
 
 void visSharedMemWriter::main_thread() {
     INFO("Reached main thread");
@@ -97,10 +114,7 @@ void visSharedMemWriter::main_thread() {
     size_t i = 0;
 
     // Set up the structure of the access record shared memory
-    struct timeval timestamp;
-    uint64_t time_us;
-    size_t access_record_size = sizeof(time_us);
-    const uint64_t in_progress = -1;
+    access_record_size = sizeof(uint64_t);
 
     record_addr = assign_memory<uint64_t*>(fname_access_record, ntime * access_record_size, record_addr);
 
@@ -108,38 +122,24 @@ void visSharedMemWriter::main_thread() {
     // Get one frame for reference
     wait_for_full_frame(in_buf, unique_name.c_str(), frame_id);
 
-    // Get a view of the current frame
-    // TODO: this frame is not getting written anywhere yet.
     auto frame = visFrameView(in_buf, frame_id);
 
-    // dset_id_t ds_id -> frame.dataset_id
-    // Frequency IDs that we are expecting
-    std::map<uint64_t, uint64_t> freq_id_map;
-    auto& dm = datasetManager::instance();
+    // Use the view to figure out the structure of the ring buffer
+    metadata_size = sizeof(visMetadata);
+    data_size = sizeof(uint8_t);
+    frame_size = data_size + metadata_size + sizeof(ONE);
 
-    // Build the frequency index
-    auto fstate_fut = std::async(&datasetManager::dataset_state<freqState>, &dm, frame.dataset_id);
-    const freqState* fstate = fstate_fut.get();
-    uint ind = 0;
-    for (auto& f : fstate->get_freqs())
-        freq_id_map[f.first] = ind++;
-
-
-    // Figure out the structure of the ring buffer
-    const uint8_t ONE = 1;
-    size_t nfreq = fstate->get_freqs().size();
-    size_t metadata_size = sizeof(visMetadata);
-    size_t data_size = sizeof(uint8_t);
-    size_t frame_size = data_size + metadata_size + sizeof(ONE);
-
-    // memory_size should be ntime * nfreq * file_frame_size (data + metadata)
-    buf_addr = assign_memory<uint8_t*>(fname_buf, ntime * nfreq * frame_size, buf_addr);
+    // memory_size should be ntime * file_frame_size (data + metadata)
+    buf_addr = assign_memory<uint8_t*>(fname_buf, ntime * frame_size, buf_addr);
     INFO("Created the shared memory segments\n");
     CHECK(sem_post(sem));
 
+    write_to_memory(frame, i);
+
+    ++i;
+
     // gets called once when kotekan is running
     while (!stop_thread) {
-
 
         // Wait for the buffer to be filled with data
         if (wait_for_full_frame(in_buf, unique_name.c_str(), frame_id) == nullptr) {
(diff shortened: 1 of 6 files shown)


lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,188 @@
...
+        register_consumer(in_buf, unique_name.c_str());
+
+        // Check if any of the old buffer files exist
+        DEBUG("Checking for and removing old buffer files...");
+        check_remove(root_path + "sem." + sem_name);
+        check_remove(root_path + fname_access_record);
+        check_remove(root_path + fname_buf);
+
+        // Create the semaphore, and gain first access to it
+        sem = sem_open(
```
jrs65: It might make sense to create the semaphore at the same time you create the shared mem. The state is a little confusing if one can exist without the other.

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,188 @@
...
+        INFO("Resized to %ld bytes\n", (long) shm_size);
+
+        addr = (T) mmap(nullptr, shm_size, (PROT_READ | PROT_WRITE), MAP_SHARED, fd, 0);
+        if (addr == MAP_FAILED) {
+            throw std::runtime_error(
+                fmt::format(fmt("Failed to map shm {:s} to memory: {:s}."), shm_name, strerror(errno)));
+        }
+
+        // fd is no longer needed
+        CHECK(close(fd));
```
nritsche: this is mixing exit(1) and throwing exceptions in the error handling in this method. I'm not sure, but from looking at a few other stages I conclude that you can't throw in the main loop (as in they wouldn't be caught). Instead they use `FATAL_ERROR(); return;`

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+        return addr;
+}
+
+void visSharedMemWriter::write_to_memory(const visFrameView& frame, size_t index) {
+
+        CHECK(sem_wait(sem));
```
nritsche: Some comments to explain what this is all about would be nice (only if we actually plan to merge this code).

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,188 @@
...
+        memcpy(record_addr + index, &in_progress, access_record_size);
+        CHECK(sem_post(sem));
+
+        memcpy(buf_addr + index * frame_size, &ONE, sizeof(uint8_t));
```
jrs65: Okay one actually important point about the valid byte now, although this comment only applies to the time-freq version. You should write the valid byte *last*.

That's because on the time-freq version you can probably do something like this:

- Entirely new time sample arrives:
  - Find where it's going to go.
  - Lock that sample in the metadata, mark is as invalid and release the lock.
  - Zero all the data in that slot
  - Lock the metadata, mark as valid (with new timestamp), and release.
  - Now you can write the data lock free *provided* you write the valid field last, i.e. you only write it when you can guarantee everything else for that freq was written. I think the valid field needs to be small enough, and well aligned such that setting it is atomic, which I think both a single byte, or 4 byte int is fine for.
* Any frequency arrives within an already open sample:
  * Simply write it in, again writing the valid field last.

I'm pretty sure in this scheme the reader will always see only consistent data because for the valid byte to be non-zero the data must have been correctly written already, and on Intel architectures the ordering of memory writes is guaranteed to be seen the same way on all processors.

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+    // Set up the structure of the ring buffer shared memory
+    // Get one frame for reference
+    wait_for_full_frame(in_buf, unique_name.c_str(), frame_id);
+
+    auto frame = visFrameView(in_buf, frame_id);
+
+    // Use the view to figure out the structure of the ring buffer
+    metadata_size = sizeof(visMetadata);
+    data_size = sizeof(uint8_t);
```
nritsche: Is this correct?

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+void visSharedMemWriter::write_to_memory(const visFrameView& frame, size_t index) {
+
+        CHECK(sem_wait(sem));
+        memcpy(record_addr + index, &in_progress, access_record_size);
+        CHECK(sem_post(sem));
+
+        memcpy(buf_addr + index * frame_size, &ONE, sizeof(uint8_t));
+        memcpy(buf_addr + index * frame_size + 1, frame.metadata(), metadata_size);
+        memcpy(buf_addr + index * frame_size + metadata_size + 1, frame.data(), data_size);
```
nritsche: This looks fine, but maybe becomes a bit easier to read if you keep a pointer to the start of the buffer and one to the current writing position to replace all the operations here, eg:
```
memcpy(write_pos, &one, sizeof(uint8_t));
write_pos += sizeof(one);
memcpy(write_pos, frame.metadata(), metadata_size);
write_pos += metadata_size;
...
```

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+        struct timeval timestamp;
+
+        gettimeofday(&timestamp, nullptr);
+        uint64_t time_us = timestamp.tv_sec * 1000000 + timestamp.tv_usec;
+        INFO("Data written to location at {} us", time_us);
+
+        CHECK(sem_wait(sem));
+        memcpy(record_addr + index, &time_us, access_record_size);
```
nritsche: Same as above, you could keep a pointer to the write position to the record (and one to the start of the record).

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,188 @@
...
+    metadata_size = sizeof(visMetadata);
+    data_size = sizeof(uint8_t);
+    frame_size = data_size + metadata_size + sizeof(ONE);
+
+    // memory_size should be ntime * file_frame_size (data + metadata)
+    buf_addr = assign_memory<uint8_t*>(fname_buf, ntime * frame_size, buf_addr);
+    INFO("Created the shared memory segments\n");
+    CHECK(sem_post(sem));
+
+    write_to_memory(frame, i);
```
nritsche: This first frame is going to be written to the buffer twice, if I'm not wrong.

lib/stages/visSharedMemWriter.cpp
```diff
@@ -0,0 +1,188 @@
...
+        memcpy(record_addr + index, &in_progress, access_record_size);
+        CHECK(sem_post(sem));
+
+        memcpy(buf_addr + index * frame_size, &ONE, sizeof(uint8_t));
+        memcpy(buf_addr + index * frame_size + 1, frame.metadata(), metadata_size);
+        memcpy(buf_addr + index * frame_size + metadata_size + 1, frame.data(), data_size);
+
+        struct timeval timestamp;
+
+        gettimeofday(&timestamp, nullptr);
```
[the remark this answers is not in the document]
anjakefala: Anja Boskovic  14:49
What is the FPGA sequence number of each timesample?

Richard Shaw  14:50
It's the timestamp given to each frame by the correlator.
It's fpga_seq in the metadata I think.
It's really just the number of F-engine clock ticks that have happened.
I think there's one every 2.56 microseconds.
All frequencies at the same time have the same fpga sequence number.

Anja Boskovic  14:57
Currently, each time index has a single timestamp, recording when anything in that time index has been recently updated
should each frequency and time index have their own timestamp?

Richard Shaw  14:57
I don't think so.

Anja Boskovic  14:57
(basically, I made the address record 1-D)
But then that would mean we only update the record address with a timestamp when the first frequency gets inserted for a time index
is that okay?

Richard Shaw  14:58
One per time sample (i.e. one to cover all frequencies in a timesample)
I mean you could update it, but I don't think you need to.
Provided you do the scheme where you zero everything when you get the first new frequency, and always write the valid field last when writing any other entry. (edited) 
I think the only thing you need to do maintain consistency is make sure you mark as invalid while zeroing.
And lock the metadata while change it's state (i.e. updating timestamp, marking invalid etc).

Richard Shaw  15:06
Does that help ?

Anja Boskovic  15:06
> I think the only thing you need to do maintain consistency is make sure you mark as invalid while zeroing.

I will incorporate this.

> always write the valid field last when writing any other entry.

And this!

> Provided you do the scheme where you zero everything when you get the first new frequency

I do the zero-ing when I get the first frequency for a new time index. I zero that entire time index. (edited) 

Richard Shaw  15:07
Great. That’s the right approach. It saves you marking things as longer than necessary. 
I think this scheme means that we can read all of the most recent data, without having to block off the most current sample.

lib/stages/visSharedMemWriter.cpp
anjakefala: CHECK() was my attempt to incorporate simple error handling without having a lot of boiler plate cluttering! 

I will instead use FATA_ERROR(); return

lib/stages/visSharedMemWriter.cpp
anjakefala: Nope! Fixed it.

lib/stages/visSharedMemWriter.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+    template<typename T>
+    T assign_memory(std::string shm_name, int shm_size, T addr);
+
+    void write_to_memory(const visFrameView& frame, size_t index);
+
+    std::string root_path, sem_name, fname_access_record, fname_buf;
+};
+
+inline void check_remove(std::string fname) {
```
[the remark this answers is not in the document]
anjakefala: I copied this from Tristan, and he had placed it in the header file! No deeper reason than that. General helper function. It should probably be factored out into a utils, I think it is useful.

anjakefala:
Conversations have been moved to: https://github.com/kotekan/kotekan/pull/686

PR #689, opened 2020-03-05
16 element mode (128 freqs per stream)

kiyo-masui:
Allows baseband readout to support `num_local_freq != 1`.

Not ready at all ready to merge, I just need to make the PR so I can comment on the diff.


lib/stages/basebandReadout.cpp
```diff
@@ -141,19 +145,33 @@ void basebandReadout::main_thread() {
...
-            wt = std::make_unique<std::thread>([&] { this->writeout_thread(*mgr); });
+            INFO("slot_id:{:d}",stream_id.slot_id);            
+            INFO("link_id:{:d}",stream_id.link_id);            
+            uint32_t freq_id=0;
+		for(int freqidx=0; freqidx < _num_local_freq;freqidx++){
+		    // XXX Map stream IDs to freq IDs with bin_number() (for Pathfinder) or bin_number_chime()
+		    // XXX Use num_local_freqs to figure out if we're running on CHIME or PF
+		    if (_num_local_freq == 1){
+			freq_id = bin_number_chime(&stream_id);
+		    } else { // more than one freq per stream
```
kiyo-masui: `bin_number` only works for the pathfinder configuration right? I think there are 3 valid cases: chime, pathfinder, and 16 input.

lib/stages/basebandReadout.cpp
```diff
@@ -128,8 +130,10 @@ void basebandReadout::main_thread() {
 
     std::unique_ptr<std::thread> wt;
     std::unique_ptr<std::thread> lt;
+    basebandReadoutManager *mgrs[_num_local_freq];
```
kiyo-masui: The pointers in this array have to be set to `nullptr` (I don't think that is guaranteed automatically).

lib/stages/basebandReadout.cpp
```diff
@@ -163,8 +181,14 @@ void basebandReadout::main_thread() {
 
         frame_id++;
     }
-    if (mgr) {
-        mgr->stop();
+    if (mgrs[0]) {
+        //mgr->stop();
+        for(int freqidx = 0; freqidx < _num_local_freq; freqidx++){
+            if (mgrs[freqidx]){
+                mgrs[freqidx]->stop(); // stop all the managers that were started
```
kiyo-masui: With can segfault if `mgrs[freqidx]` is an invalid pointer that is not null.

lib/stages/basebandReadout.cpp
```diff
@@ -141,19 +145,33 @@ void basebandReadout::main_thread() {
...
+		    mgrs[freqidx] = &(basebandApiManager::instance().register_readout_stage(freq_ids[freqidx]));//TODO: Check register_readout_stage
+		    DEBUG("Initialize baseband metrics for freq_id: {:d}", freq_id);
+		    readout_counter.labels({std::to_string(freq_id), "done"});
+		    readout_counter.labels({std::to_string(freq_id), "error"});
+		    readout_counter.labels({std::to_string(freq_id), "no_data"});
+		    readout_in_progress_metric.labels({std::to_string(freq_id)}).set(0);
+		    INFO("Starting request-listening thread for freq_id: {:d}", freq_id);
+		}
+            mgr = mgrs[_num_local_freq-1];//&basebandApiManager::instance().register_readout_stage(freq_id);
+	    INFO("freq_id: {:d}",freq_id);
```
kiyo-masui: indentation

lib/stages/basebandReadout.cpp
```diff
@@ -175,167 +199,230 @@ void basebandReadout::main_thread() {
...
+        request_no_data_counters[freqidx] = &(readout_counter.labels({std::to_string(freq_ids[freqidx]), "no_data"}));
+    }
+
+    //TODO: Ask davor about request_no_data_counter()
 
     while (!stop_thread) {
         // Code that listens and waits for triggers and fills in trigger parameters.
         // Latency is *key* here. We want to call get_data within 100ms
         // of L4 sending the trigger.
+        next_requests[0] = mgrs[0]->get_next_waiting_request();
```
kiyo-masui: Where is `next_requests[]` declared?

lib/stages/basebandReadout.cpp
```diff
@@ -175,167 +199,230 @@ void basebandReadout::main_thread() {
...
-            // yet that can change the dump_status object
-            const basebandRequest request = dump_status.request;
-            // std::time_t tt = std::chrono::system_clock::to_time_t(request.received);
-            const uint64_t event_id = request.event_id;
-            INFO("Received baseband dump request for event {:d}: {:d} samples starting at count "
+
+                INFO("Non-null request for freq_ids[idx]: {:d}", freq_ids[freqidx]);
+                }
+            
+            // Do just the last one individually
```
kiyo-masui: Why is this here? Looks like this happens again in the loop just below.

lib/stages/basebandReadout.cpp
leungcalvin: Yes, I am implementing this in fpga_header_functions.c as `bin_number_multifreq(num_local_freq, stream_id, freqidx)`.

andrerenard:
It might be worth asking @cubranic for a quick review of this as well.

lib/stages/basebandReadout.cpp
cubranic: Doesn't the typing protect you? It's a pointer to `basebandReadoutManager`, so you can't assign anything to it except for a readout manager. (Unless what you assign were freed.)

Which leads me to an important question: where exactly are these created?

lib/stages/basebandReadout.cpp
cubranic: Line 201, it looks like an array of `std::unique_ptr<basebandReadoutManager::requestStatusMutex>`. Why is it a pointer to mutex?

lib/stages/basebandReadout.cpp
cubranic: (Unless you're referring to `if (mgr[0])`, which is indeed incorrect and will segfault if `mgr`itself is null. And I don't know if it's guaranteed to be zeroed when the array is created, and so perhaps can be an invalid value until something is assigned.)

And to answer my own question, it's created in (current) line 156. But I don't like the way this is done with arrays -- why not a vector so it handles the empty case cleanly and can't have invalid pointers.

cubranic:
I should wait with my comments until I have a clear idea of your design: can you briefly explain how this is intended to work? @leungcalvin 

My starting understanding is this: you have the basebandReadout main thread create a readoutManager for each frequency in the first packet it receives, and then in the listen thread ask for the next request from the first one (which is the manager for the lowest frequency and so has the least dispersion delay). When there is an available request, iterate through all its managers in order to process all associated frequencies?

leungcalvin:
> I should wait with my comments until I have a clear idea of your design: can you briefly explain how this is intended to work? @leungcalvin
> 
> My starting understanding is this: you have the basebandReadout main thread create a readoutManager for each frequency in the first packet it receives, and then in the listen thread ask for the next request from the first one (which is the manager for the lowest frequency and so has the least dispersion delay). When there is an available request, iterate through all its managers in order to process all associated frequencies?

Hi Davor--yes, this is the idea. We check the highest-frequency manager for a next_request, and then simply assume they have all arrived (they should all arrive roughly synchronously but we are most in danger of losing the first manager/request because the burst arrives at high frequency first). To my knowledge this logic hasn't caused many segfaults but I realize it's not ideal. Is there a better architecture that would allow us to process them as they arrive?

Then we process the managers, calling `get_data()` starting with the high frequencies and a corresponding memory index offset of `freq_idx * num_elements` corresponding to the frequency index (since our data is packed in time, freq, element format) Once we've copied the data out of the buffer, we just loop over everything in `writeout_thread()`.

lib/stages/basebandReadout.cpp
leungcalvin: Just because there's no default constructor for a mutex. I need the mutexes to not go out of scope over the entire readout_thread(), right?

andrerenard (changes requested):
Can you revert all the changes in `external/` to the `develop` versions?  It seems like maybe the auto formatter, or the wrong version of the auto formatter was run on those files, and it's generating a lot of unrelated changes.

leungcalvin:
> Can you revert all the changes in `external/` to the `develop` versions? It seems like maybe the auto formatter, or the wrong version of the auto formatter was run on those files, and it's generating a lot of unrelated changes.

leungcalvin:
> Can you revert all the changes in `external/` to the `develop` versions? It seems like maybe the auto formatter, or the wrong version of the auto formatter was run on those files, and it's generating a lot of unrelated changes.

Done.

andrerenard (changes requested):
Initial pass review, mostly for coding standards.

lib/stages/basebandReadout.cpp
```diff
@@ -138,19 +145,24 @@ void basebandReadout::main_thread() {
...
-            INFO("Starting request-listening thread for freq_id: {:d}", freq_id);
-            mgr = &basebandApiManager::instance().register_readout_stage(freq_id);
-            lt = std::make_unique<std::thread>([&] { this->readout_thread(freq_id, *mgr); });
-
-            wt = std::make_unique<std::thread>([&] { this->writeout_thread(*mgr); });
+            for (int freqidx = 0; freqidx < _num_local_freq; freqidx++) {
+                freq_id = tel.to_freq_id(buf, buf_frame, freqidx);
+                freq_ids[freqidx] = freq_id;
+                mgrs.push_back(&(basebandApiManager::instance().register_readout_stage(
+                    _board_id * 1048576
```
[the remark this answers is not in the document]
leungcalvin: moved this and a description to basebandApiManager.hpp

lib/stages/basebandReadout.cpp
```diff
@@ -95,12 +98,13 @@ basebandReadout::basebandReadout(Config& config, const std::string& unique_name,
...
 
     // Memcopy byte alignments assume the following.
-    if (_num_elements % 128) {
+    if (_num_local_freq != 128 and _num_elements % 128) {
         throw std::runtime_error("num_elements must be multiple of 128");
     }
 
     register_consumer(buf, unique_name.c_str());
 
+    INFO("Registered consumer baseband!!");
```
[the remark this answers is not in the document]
leungcalvin: done

lib/stages/basebandReadout.cpp
```diff
@@ -95,12 +98,13 @@ basebandReadout::basebandReadout(Config& config, const std::string& unique_name,
     }
 
     // Memcopy byte alignments assume the following.
-    if (_num_elements % 128) {
+    if (_num_local_freq != 128 and _num_elements % 128) {
```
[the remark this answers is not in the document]
leungcalvin: done

lib/stages/basebandReadout.cpp
```diff
@@ -172,167 +194,223 @@ void basebandReadout::main_thread() {
...
+                    duration_cast<milliseconds>(system_clock::now().time_since_epoch());
+                // INFO("memcpy() call: {:d} = {:d}-{:d} ms", ms_after.count() - ms_before.count(),
+                //       ms_after.count(), ms_before.count());
             }
         }
     }
 }
 
-void basebandReadout::writeout_thread(basebandReadoutManager& mgr) {
+void basebandReadout::writeout_thread(std::vector<basebandReadoutManager*> mgrs) {
```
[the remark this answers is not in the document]
leungcalvin: done

lib/stages/basebandReadout.cpp
```diff
@@ -160,8 +172,18 @@ void basebandReadout::main_thread() {
 
         frame_id++;
     }
-    if (mgr) {
-        mgr->stop();
+    if (mgrs[0] != nullptr) {
```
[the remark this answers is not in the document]
leungcalvin: done

andrerenard:
Looks like we are still failing to build on MacOS.  I'll try to disable parts of the code on my MacOS system, and hopefully I can track down what's breaking here...

lib/stages/basebandReadout.cpp
```diff
@@ -171,149 +185,192 @@ void basebandReadout::main_thread() {
     }
 }
 
-void basebandReadout::readout_thread(const uint32_t freq_id, basebandReadoutManager& mgr) {
-    auto& request_no_data_counter = readout_counter.labels({std::to_string(freq_id), "no_data"});
+void basebandReadout::readout_thread(const uint32_t freq_ids[],
+                                     std::vector<basebandReadoutManager*>& mgrs) {
+    std::unique_ptr<basebandReadoutManager::requestStatusMutex> next_requests[_num_local_freq];
+    std::shared_ptr<basebandReadoutManager::requestStatusMutex> next_request;
+    kotekan::prometheus::Counter* request_no_data_counters[_num_local_freq];
```
andrerenard, line 192: This was covered in more detail in slack, but these two arrays should likely just be converted to use vectors, since that's a safer access model anyway. And will get around the crazy MacOS clang bug.

lib/stages/basebandReadout.cpp
```diff
@@ -171,149 +185,192 @@ void basebandReadout::main_thread() {
...
-
-            if (data.status == basebandDumpData::Status::Ok) {
-                INFO("Captured {:d} samples for event {:d} and freq {:d}.", data.data_length_fpga,
-                     data.event_id, data.freq_id);
-
-                // we are done copying the samples into the readout buffer
-                mgr.ready({dump_status, data});
+                milliseconds ms_after =
+                    duration_cast<milliseconds>(system_clock::now().time_since_epoch());
+                // INFO("memcpy() call: {:d} = {:d}-{:d} ms", ms_after.count() - ms_before.count(),
```
andrerenard, line 305: Because this was removed (I'm not sure why?). This is generating unused variable warnings.  If we want to add it back we should make this DEBUG, and wrap these variables and calls in `#ifdef DEBUGGING`.

andrerenard (changes requested):

andrerenard:
One side comment, when I was testing the MacOS build problems, I had to merge in the latest develop.   That generated a few conflicts in `basebandReadout.cpp` related to changes @cubranic made in the last release (removing the directory checks).  It would be good for someone else to double check that the code still looks good after the merge.

cubranic:
@leungcalvin Can you remind me how the data arrives within frames in this mode? Instead of one frequency per stream, there will be multiple ones (I guess 128 in PF?), packed one after another for a given `t`?

leungcalvin:
> @leungcalvin Can you remind me how the data arrives within frames in this mode? Instead of one frequency per stream, there will be multiple ones (I guess 128 in PF?), packed one after another for a given `t`?

Yes. From slowest-varying to fastest-varying, the indices are time, frequency, element. We support shuffle256 at the Pathfinder where there are 8 frequencies per stream, since there are 256 elements, and shuffle16 at GBO where there are 128 frequencies per stream, since there are 16 elements.

lib/stages/basebandReadout.cpp
leungcalvin: I don't think we actually want these INFO statements back or the timing variables--this was to quickly test the speed of memcpy() vs nt_memcpy() and I never changed it because things were working and Kiyo mumbled something once about print statements affecting the stability of multi-threaded processes, but that's not a good reason for having it in there.

.github/workflows/main.yml
```diff
@@ -452,6 +452,10 @@ jobs:
         -DWITH_TESTS=ON ..
         make -j 2
 
+    - uses: actions/checkout@v2
+    - name: Setup tmate session
```
cubranic, line 456: Do we want to leave this action? I thought it's only for interactive debugging and shouldn't be included in automated tests?

lib/stages/basebandReadout.cpp
```diff
@@ -470,13 +527,32 @@ basebandDumpData basebandReadout::get_data(uint64_t event_id, int64_t trigger_st
...
+            using namespace std::chrono;
+            milliseconds ms_before =
+                duration_cast<milliseconds>(system_clock::now().time_since_epoch());
+            nt_memcpy(&dump.data[data_ind_start * _num_elements],
+                      &buf_data[frame_ind_start * _num_elements],
+                      (frame_ind_end - frame_ind_start) * _num_elements);
+            milliseconds ms_after =
+                duration_cast<milliseconds>(system_clock::now().time_since_epoch());
+
+        } else if (_num_local_freq > 1) {
```
cubranic, line 542: Does this have to be `else if` and not just `else`?

lib/stages/basebandReadout.cpp
```diff
@@ -470,13 +527,32 @@ basebandDumpData basebandReadout::get_data(uint64_t event_id, int64_t trigger_st
...
+            using namespace std::chrono;
+            milliseconds ms_before =
+                duration_cast<milliseconds>(system_clock::now().time_since_epoch());
+            // INFO("Right before memcpy()");
+            for (int timeidx = 0; timeidx < (frame_ind_end - frame_ind_start); timeidx++) {
+                memcpy(&dump.data[(data_ind_start + timeidx) * _num_elements],
```
cubranic, line 548: why `memcpy` instead of `nt_memcpy` as in the single-frequency case?

lib/stages/basebandReadout.cpp
```diff
@@ -470,13 +527,32 @@ basebandDumpData basebandReadout::get_data(uint64_t event_id, int64_t trigger_st
...
+                       &buf_data[((frame_ind_start + timeidx) * _num_local_freq + freqidx)
+                                 * _num_elements],
+                       _num_elements);
+            }
+            milliseconds ms_after =
+                duration_cast<milliseconds>(system_clock::now().time_since_epoch());
```
cubranic, line 554: This time-keeping is not really used, is it?

cubranic (commented):
Most of my comments are on minor details, but one thing I'm really not clear about, but is making me uneasy right now, is whether there can be duplicate freq_ids in certain telescope configurations, which is why that BOARD_SPACING is used to register the baseband readout. If in that one place, why not everywhere where frequency ids are made visible to the world? (I mean the metric labels, but also -- especially -- the written-out files. The files encode the frequency index in their file name, in `freq_id` attribute, and also convert it to the real frequency and add that as the `freq` attribute.

.github/workflows/main.yml
leungcalvin: That's right. I will remove this.

lib/stages/basebandReadout.cpp
leungcalvin: Yes, that's right. I will remove it.

lib/stages/basebandReadout.cpp
leungcalvin: I think `nt_memcpy` assumes there is an integer multiple of 128 baseband data points to copy per call. This is not true in 16-element mode, where the block sizes are 16.

cubranic: OK

leungcalvin:
> Most of my comments are on minor details, but one thing I'm really not clear about, but is making me uneasy right now, is whether there can be duplicate freq_ids in certain telescope configurations, which is why that BOARD_SPACING is used to register the baseband readout. If in that one place, why not everywhere where frequency ids are made visible to the world? (I mean the metric labels, but also -- especially -- the written-out files. The files encode the frequency index in their file name, in `freq_id` attribute, and also convert it to the real frequency and add that as the `freq` attribute.

Hi Davor, sorry for the late reply. I would say we actually understand  exactly when duplicate `freq_id`'s occur: only in the corner case when you're recording data from multiple boards on the same Kotekan instance AND when you do not want to install a backplane. If we had a backplane, and ran `shuffle256` the boards would be distinguished by their slot id and there would not be identical stream id's. BOARD_SPACING is a workaround for when the backplane is not there to perform that function for us, and that will only happen for the GBO/TONE outrigger: for Allenby and other outriggers we will use `shuffle128` and `shuffle256` with the help of a backplane, and we will not need to discriminate boards with this workaround. Does that help?

lib/stages/basebandReadout.cpp
leungcalvin: No, we can use `else`.

lib/stages/basebandReadout.cpp
```diff
@@ -137,19 +144,24 @@ void basebandReadout::main_thread() {
...
+                freq_id = tel.to_freq_id(buf, buf_frame, freqidx);
+                freq_ids[freqidx] = freq_id;
+                mgrs.push_back(&(basebandApiManager::instance().register_readout_stage(
+                    _board_id * BOARD_SPACING
+                    + freq_ids[freqidx]))); // _board_id is used in 16 element mode to distinguish
+                                            // between identical boards running without a backplane.
+                                            // We will identify data managers by their board_id as
+                                            // well as their freq_id =< 1024.
+                DEBUG("Initialize baseband metrics for freq_id: {:d}", freq_id);
+                readout_counter.labels({std::to_string(freq_id), "done"});
```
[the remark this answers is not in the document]
leungcalvin: Yes, we can do this. To facilitate this we can use a nicer number (e.g. 1000000) as the `BOARD_SPACING` constant so things are more readable.

andrerenard:
Depends on: #927

lib/core/basebandApiManager.cpp
```diff
@@ -226,19 +226,26 @@ void basebandApiManager::handle_request_callback(connectionInstance& conn, json&
...
-            const uint32_t freq_id = element.first;
+            // Break readout_id into a freq_id and a board_id
+            const uint32_t freq_id = element.first % BOARD_SPACING;
+            const uint32_t board_id = element.first / BOARD_SPACING;
             auto& readout_entry = element.second;
 
-            const std::string readout_file_name =
+            std::string readout_file_name =
                 fmt::format(fmt("baseband_{:d}_{:d}.h5"), event_id, freq_id);
+            if (board_id != 0) {
```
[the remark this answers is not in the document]
cubranic: @leungcalvin See above on file naming.

andrerenard:
I'm going to close this PR, since the changes in #927 and #972 have largely rewritten the system this PR was based on top of.

PR #691, opened 2020-03-10
fix(configUpdater): search up-tree for updatable_conf options

nritsche:


nritsche:
> This looks good to me, but can we make sure the formatting changes here are also reflected in the science run config, mostly to make the diff just involves the operational changes.

I think so, too. But should we directly add a yamllint CI check when doing this?

andrerenard:
Yes a yamllint check is something I've been meaning to add to the CI process.  It might just require a fair bit of clean up of some of the other config files.  And when we do we should add it to the "pre_commit" clean up script as well.

james-s-willis:
Looks good to me. I've tested it on `csDg9` too and it runs fine.

james-s-willis:
@nritsche I think the unit tests just need updating

andrerenard:
I'm afraid I just realised I need this for a test tomorrow.  I'm going to pull it in without passing CI, since the CI server seems to be down again...

jrs65:
@andrerenard @nritsche can we please not just merge things when the CI server is down? Or at the very least can someone please run the tests themselves? This PR breaks almost all of the Python unit tests and now I'm having to figure out how to fix it.

PR #696, opened 2020-03-17
fix(config): add valve process before RFI flagging in GPU configs

### LINKED ISSUE #666: Add valve after rfi_vis_buffer_* before RfiDropFrame
Currently `gpu_output_n2_buffer_*` has a `valve` stage before being read by `RfiDropFrame`, but the buffers ` rfi_vis_buffer_*` don't have a `valve` stage, so it's possible that back pressure in the N2 post process pipeline can cause a backup in the GPU stage resulting in a system crash.

This addresses an issue were pulsar gaiting updates causes some nodes to crash.

jrs65:
Fixes #666

### DIFF
--- a/config/chime_science_run_gpu.yaml
+++ b/config/chime_science_run_gpu.yaml
@@ -458,6 +458,19 @@ gpu_rfi_output_buffers:
   gpu_rfi_output_buffer_3:
     kotekan_buffer: standard
 
+gpu_rfi_valve_buffers:
+  num_frames: buffer_depth
+  frame_size: sizeof_float * num_local_freq * samples_per_data_set / sk_step
+  metadata_pool: main_pool
+  rfi_valve_buffer_0:
+    kotekan_buffer: standard
+  rfi_valve_buffer_1:
+    kotekan_buffer: standard
+  rfi_valve_buffer_2:
+    kotekan_buffer: standard
+  rfi_valve_buffer_3:
+    kotekan_buffer: standard
+
 gpu_rfi_mask_output_buffers:
   num_frames: buffer_depth
   frame_size: num_local_freq * samples_per_data_set / sk_step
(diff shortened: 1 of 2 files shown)


config/chime_science_run_gpu.yaml
```diff
@@ -410,6 +410,14 @@ gpu_rfi_output_buffers:
...
   gpu_rfi_output_buffer_3:
     kotekan_buffer: standard
+  rfi_value_buffer_0:
+    kotekan_buffer: standard
+  rfi_value_buffer_1:
+    kotekan_buffer: standard
+  rfi_value_buffer_2:
+    kotekan_buffer: standard
+  rfi_value_buffer_3:
+    kotekan_buffer: standard
```
andrerenard: I would set the `num_frames` value for these to be something much smaller.  Maybe just: `num_frames: buffer_depth`

config/chime_science_run_gpu.yaml
```diff
@@ -458,6 +456,17 @@ gpu_rfi_output_buffers:
   gpu_rfi_output_buffer_3:
     kotekan_buffer: standard
 
+gpu_rfi_valve_buffers:
+  num_frames: buffer_depth
```
andrerenard, line 460: Frame size is missing here.

config/chime_science_run_gpu_no_output.yaml
```diff
@@ -459,6 +459,14 @@ gpu_rfi_output_buffers:
...
   gpu_rfi_output_buffer_3:
     kotekan_buffer: standard
+  rfi_value_buffer_0:
+    kotekan_buffer: standard
+  rfi_value_buffer_1:
+    kotekan_buffer: standard
+  rfi_value_buffer_2:
+    kotekan_buffer: standard
+  rfi_value_buffer_3:
+    kotekan_buffer: standard
```
andrerenard, line 469: This needs to be updated to match the production config.

config/chime_science_run_gpu.yaml
jrs65: Yeah. I found this one out pretty rudely last night.

config/chime_science_run_gpu_no_output.yaml
jrs65: Ah good point.

PR #699, opened 2020-03-19
Run CI on Github actions

jrs65:
Okay this is ready to go. @andrerenard I had to add a few options to the build scripts to make it all work: one to disable mlocking the buffer memory; and one to explicitly set the build architecture, so it would be great if you could look at those. @nritsche if you could check everything else.

Once this is reviewed and merged, I think we can just change the Jenkins webhook status to inactive and it will stop reporting.

There's a few things it doesn't yet do, which we should address in subsequent changes:
- The baseband readout tests are disabled as they didn't work trivially
- All of the tests involving comet are disabled. This should be easy to fix, probably by just installing redis into the container.
- It doesn't built the docs
- It doesn't do a full CHIME kotekan build because of updates to ROCm that need some of @andrerenard's attention.
- It doesn't run the kernel tests (and probably never will).

Additionally there's a few changes that could make things speedier:
- Figure out a way to only run the build-docker job when the Dockerfile's have changed. Although it doesn't actually do anything it pointlessly burns 90s fetching and pushing the images.
- See if there's a way of speeding up the building of the boost_tests. It turns out to add a significant amount of time to the ccache'd build time to build and link the boost tests.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,13 @@
-name: include-what-you-use
+# NOTE:
+# When we build kotekan on the Github VM we need to take into account a few
+# limitations: first we can't lock very much memory so we need to disable
+# mlock the buffers. Second, the VMs run across a range of CPU
+# architectures (from Haswell to Skylake-AVX512 as of 2020/03). Because our
+# use of ccache shares compiled objects we must target our build for the
+# lowest architecture available or we will occasionally crash with SIGILL
+# when a newer instruction is called than is available.
+
+name: kotekan-ci-test
 on:
   pull_request:
     branches:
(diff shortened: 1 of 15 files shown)


.github/workflows/main.yml
```diff
@@ -17,10 +17,10 @@ jobs:
...
-       docker build --cache-from=rocm/dev-ubuntu-18.04,docker.pkg.github.com/kotekan/kotekan/kotekan:latest -f tools/iwyu/docker/Dockerfile -t docker.pkg.github.com/kotekan/kotekan/kotekan:latest .
-       docker push docker.pkg.github.com/kotekan/kotekan/kotekan:latest
-   - name: Run iwyu
-     run: |
-       docker run --rm docker.pkg.github.com/kotekan/kotekan/kotekan:latest /code/tools/iwyu/docker/iwyu.sh
+       docker login docker.pkg.github.com -u ${GITHUB_REPOSITORY} -p "${{ secrets.GITHUB_TOKEN }}" | while read line ; do echo "$(date)| $line"; done;
+       docker pull --disable-content-trust docker.pkg.github.com/kotekan/kotekan/kotekan:latest | while read line ; do echo "$(date)| $line"; done;
+       docker build --cache-from=rocm/dev-ubuntu-18.04,docker.pkg.github.com/kotekan/kotekan/kotekan:latest -f tools/iwyu/docker/Dockerfile -t docker.pkg.github.com/kotekan/kotekan/kotekan:latest . | while read line ; do echo "$(date)| $line"; done;
+       docker push docker.pkg.github.com/kotekan/kotekan/kotekan:latest | while read line ; do echo "$(date)| $line"; done;
+  #  - name: Run iwyu
```
nritsche: We could still do this, just remove `develop` from the branches above and limit it on PRs/pushes towards master as you suggested.

Jenkinsfile
```diff
@@ -1,138 +0,0 @@
-pipeline {
```
nritsche: Is there a reason to remove this? We can continue running that, just not make the result block any PR.

jrs65: If you expect something to fail, I don't think there's any point in running it at all - it's just noise you have to sift through.

I'm also not done with this PR yet, at the moment all I'm doing is trying to figure out how to make the docker build faster!

jrs65: I've re-added the file. I think we can disable the webhooks once it's merged so the status doesn't get checked.

.github/workflows/main.yml
```diff
@@ -9,18 +18,169 @@ on:
...
+            -DARCH=haswell \
+            -DNO_MEMLOCK=ON \
+            -DUSE_OMP=ON \
+            -DBOOST_TESTS=OFF \
+            -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache ..
+        docker run "${OPTS[@]}" make -j 2
+        docker run "${OPTS[@]}" ccache -s
+
+
+  iwyu:
```
nritsche: did you test this one by changing the branch to develop temporarily?

tests/test_baseband_readout.py
```diff
@@ -14,11 +15,19 @@
...
 
+def is_docker():
+    path = '/proc/self/cgroup'
+    return (
+        os.path.exists('/.dockerenv') or
+        os.path.isfile(path) and any('docker' in line for line in open(path))
+    )
+
+if is_docker():
+    pytest.skip("Does not work in Github Actions docker run.", allow_module_level=True)
```
nritsche: Would it be worth it to see if Jenkins can survive a single build to run the tests disabled here?

tests/test_dataset_broker.py
```diff
@@ -32,6 +45,8 @@ def test_produce_consume():
         pytest.skip(
             "Make sure PYTHONPATH is set to where the comet dataset broker is installed."
         )
+    if not has_redis():
+        pytest.skip("Redis is not available and so comet will fail.")
```
nritsche: This is nice! But is it especially difficult to get redis to run on github? https://github.com/zhulik/redis-action

.github/workflows/main.yml
jrs65: No I haven't. And I probably won't. Having IWYU work properly is not something I fancy debugging, I've basically hit the amount of time I'm going to work on this.

tests/test_dataset_broker.py
jrs65: Again. I'm basically done with this. This is 100% better than the current Jenkins version, so I'm happy to turn off a few tests to make things work. We can address this in later pull requests.

tests/test_baseband_readout.py
jrs65: Nope. Not interested in wasting my time debugging Jenkins either.

tests/test_dataset_broker.py
jrs65: Agree that in theory this is easy, but then it ends up being hours of change, push, build, debug cycles. Plus, I think it's easiest to just install redis into one of the docker images, then you don't have to mess about with bridging the networks.

.github/workflows/main.yml
```diff
@@ -9,18 +18,169 @@ on:
...
+          ccache-full-build
+
+    - name: Build kotekan
+      run: |
+        OPTS=(--rm --mount type=bind,src=$(pwd),target=/code/kotekan -w /code/kotekan/build/ ${IMG_CORE})
+        docker login docker.pkg.github.com -u ${GITHUB_REPOSITORY} -p "${{ secrets.GITHUB_TOKEN }}"
+        docker run "${OPTS[@]}" ccache -s
+        docker run "${OPTS[@]}" \
+          cmake -DCMAKE_BUILD_TYPE=Debug \
+            -DUSE_DPDK=ON -DRTE_TARGET=x86_64-native-linuxapp-gcc \
```
andrerenard: If you are using the Ubuntu `dpdk` package you don't need the `-DRTE_TARGET=x86_64-native-linuxapp-gcc`

.github/workflows/main.yml
```diff
@@ -9,18 +18,169 @@ on:
...
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: Configure kotekan
+      run: |
+        OPTS=(--rm --mount type=bind,src=$(pwd),target=/code/kotekan -w /code/kotekan/build/ ${IMG_IWYU})
+        docker login docker.pkg.github.com -u ${GITHUB_REPOSITORY} -p "${{ secrets.GITHUB_TOKEN }}"
+        docker run "${OPTS[@]}" \
+          cmake -DCMAKE_BUILD_TYPE=Debug \
+            -DUSE_DPDK=ON -DRTE_TARGET=x86_64-native-linuxapp-gcc
```
andrerenard: Same here.

CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ option(USE_DPDK "Enable DPDK Framework" OFF)
 option(USE_HDF5 "Build HDF5 output stages" OFF)
 option(USE_OMP "Enable OpenMP" OFF)
 option(USE_OLD_ROCM "Build for ROCm versions 2.3 or older" OFF)
+option(NO_MEMLOCK "Do not lock buffer memory (useful when running in Docker)" OFF)
```
andrerenard: Long term I'd prefer if this was a config file option instead of something we had to bake into the binary.  i.e. it could be an optional `memlock: <default:true>` parameter in the `bufferFactory`.  However I also cannot really see needing this outside of a docker container, so it might not be worth the work to restructure at this point.

cmake/FindLAPACKE.cmake
```diff
@@ -0,0 +1,380 @@
...
+#
+# Distributed under the OSI-approved BSD License (the "License");
+# see accompanying file MORSE-Copyright.txt for details.
+#
+# This software is distributed WITHOUT ANY WARRANTY; without even the
+# implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the License for more information.
+#=============================================================================
+# (To distribute this file outside of Morse, substitute the full
+#  License text for the above reference.)
```
andrerenard: We need to include the full license as request here.  Also can we add to the `LICENSE` file, see for example the `FindHCC.cmake` section.

andrerenard:
We should also remember to open a few tickets to address the limitations of this new system which you've noted in the description.

.github/workflows/main.yml
jrs65: Oh thanks!

cmake/FindLAPACKE.cmake
jrs65: Oh. Good point. I'll try and fix that one now.

CMakeLists.txt
jrs65: Yeah. I wondered about doing that, but as you say it would be quite a restructure.

jrs65:
> We should also remember to open a few tickets to address the limitations of this new system which you've noted in the description.

@nritsche's already done that one!

andrerenard:
> > We should also remember to open a few tickets to address the limitations of this new system which you've noted in the description.
> 
> @nritsche's already done that one!

Nicely done @nritsche !  Referencing here: #700

nritsche:
Thanks @jrs65 for the latest addition. I think the dataset-stuff-tests are still skipped, though, because comet is not installed. This should probably always be an up-to-date version, so maybe install it in the action rather than the Dockerfile.

tests/test_dataset_broker.py
jrs65: I've actually just enabled these again. Fingers crossed they pass!

cmake/FindLAPACKE.cmake
jrs65: Done.

jrs65:
> Thanks @jrs65 for the latest addition. I think the dataset-stuff-tests are still skipped, though, because comet is not installed. This should probably always be an up-to-date version, so maybe install it in the action rather than the Dockerfile.

Comet is installed, there was just a bug in my `has_redis` implementation. Hopefully it should build now.

tests/test_dataset_broker.py
jrs65: 🍾

PR #703, opened 2020-03-25
Multithreaded L1 ping

cubranic:
Split L1 pinging into having separate threads for sending/scheduling and receiving pings, so that waiting for a response doesn't delay sending out the next scheduled ping.

I also revised the scheduling logic a bit: 

- For the first "ping_interval" (default 6 min), the nodes are pinged every "quick_ping_interval" (5s)
until they first respond. This should help to detect live nodes during startup,
so that a dropped response packet doesn't mean a 6-minute wait until the node is
checked again.
- When a node is due for a check:
  - send a ping if it's been at least than `(ping_interval - quick_ping_interval)` since it last responded; and schedule a check in "quick_ping_interval", to recover from potential lost packets
  - if a response is more recent, skip sending a new ping and schedule the next check in "ping_interval"
  - if the node has not responded in "ping_interval" but less than "ping_interval + dead_threshold" (6 min + 30 s), keep pinging every "quick_ping_interval"
  - otherwise, the node is dead and we will only ping once every "ping_interval"

- the receiving thread just marks the node "live" (so the data sending thread start sending the data immediately) and records the time of the response, for use by the sending thread's scheduling logic


lib/stages/frbNetworkProcess.cpp
```diff
@@ -265,7 +278,11 @@ void frbNetworkProcess::main_thread() {
         count++;
     }
     ping_cv.notify_all();
-    ping_thread.join();
+
+    if (send_ping_thread.joinable()) {
+        send_ping_thread.join();
+        receive_ping_thread.join();
```
andrerenard, line 284: I think you should check if this thread is joinable as well.

cubranic: Done in 6ac68e8a0dd6345827afb077558e048b28cd6bfe

andrerenard:
I was about to merge this but realised this PR doesn't re-enable the ping check.  Which I think we should do here if we intend to use this feature.

cubranic:
Good point! Re-enabled. The current value in the config is the same as the default value, should I have  just removed the entry instead?

andrerenard:
I think it's fine to leave it in the config, sometimes that's actually nicer if you later want to change it (which could happen here).

PR #705, opened 2020-03-25
Build kotekan documentation in Github actions

james-s-willis:


jrs65:
@james-s-willis I think the earlier build failed, not because there was anything wrong with the Dockerfile, but because a build I was running rebuilt the image without Doxygen. We may need to do something smarter in the long run with regards to caching the Docker images.

james-s-willis:
> @james-s-willis I think the earlier build failed, not because there was anything wrong with the Dockerfile, but because a build I was running rebuilt the image without Doxygen. We may need to do something smarter in the long run with regards to caching the Docker images.

@jrs65 the reason it wasn't passing is because this happens:

```
116 s | 73 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
116 s | Need to get 496 MB of archives.
116 s | After this operation, 4405 kB of additional disk space will be used.
116 s | Do you want to continue? [Y/n] Abort.
The command '/bin/sh -c apt-get update &&     apt-get install -y software-properties-common &&     add-apt-repository ppa:deadsnakes/ppa &&     apt-get update &&     apt-get upgrade &&     apt-get install -y python3.7 python3-setuptools python3-pip python3.7-distutils python3.7-dev                        build-essential git coreutils ccache pkg-config                        gcc cmake clang-9 clang-format-8                        rocm-opencl-dev rocm-opencl                        dpdk dpdk-dev                        libhdf5-serial-dev libboost-test-dev libevent-dev libssl-dev                        wget' returned a non-zero code: 1
```

and doxygen doesn't get installed. But the docker image build stage still reports a success. In my previous commit (8450822) I was trying to get it so that this kind of issue reports a failure.

jrs65:
Okay. Interesting. Thanks. Well it all seems to be working now.

PR #708, opened 2020-03-30
Pull gains over network

jrs65:
Replaces #697


jrs65:
@cubranic tagged you to look at the (minor) changes to `SynchronizedQueue.hpp`.

@nritsche tagged you to look at the `updateQueue.hpp` changes and the `applyGains.*` changes. 

@tristpinsm tagged you to look at the `applyGains.*` changes.

Thanks!

lib/stages/applyGains.cpp
```diff
@@ -224,96 +240,23 @@ void applyGains::apply_thread() {
 
         // Get the frequency index of this ID. The map will have been set in the first
         // validate_frame
```
tristpinsm: this comment is no longer accurate. 
```suggestion
        // Get the frequency index of this ID. The map will have been set by initialise
```

lib/stages/applyGains.cpp
```diff
@@ -381,6 +323,53 @@ void applyGains::apply_thread() {
...
+            break;
+
+        auto [update_id, transition_interval, new_ts] = *update;
+
+        std::optional<applyGains::GainData> gain_data;
+        if (read_from_file) {
+            INFO("Reading gains from file...");
+            gain_data = read_gain_file(update_id);
+        } else {
+            INFO("Fetching gains...");
```
nritsche: If this happens every 10min, that's a lot of not-so-useful output.

lib/stages/applyGains.cpp
```diff
@@ -78,15 +83,18 @@ applyGains::applyGains(Config& config, const std::string& unique_name,
                                     "to be equal or greater than one (is "
                                     + std::to_string(num_kept_updates) + ").");
 
-    // Get the path to gains directory
+    // Get the calibration broker gains endpoint
     gains_dir = config.get<std::string>(unique_name, "gains_dir");
```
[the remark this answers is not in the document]
jrs65: Good point.

lib/utils/updateQueue.hpp
```diff
@@ -101,36 +108,43 @@ class updateQueue {
...
     };
 
     /**
      * @brief Get all updates stored by the queue and their timestamps.
      *
      * @return A const reference to an std::deque holding all updates and their timestamps.
      */
-    const std::deque<std::pair<timespec, T>>& get_all_updates() const {
+    const std::deque<std::pair<timespec, std::shared_ptr<const T>>>& get_all_updates() const {
         return values;
```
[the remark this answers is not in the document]
jrs65: I agree on `.size()`.

I think to make this one work I need to return a copy of the `deque` not a reference (and grab the lock),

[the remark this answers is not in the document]
nritsche: Oh. Yeah.

lib/stages/applyGains.cpp
```diff
@@ -148,55 +154,59 @@ bool applyGains::receive_update(nlohmann::json& json) {
...
-        INFO("Setting thread affinity");
-        for (auto& i : config.get<std::vector<int>>(unique_name, "cpu_affinity"))
-            CPU_SET(i, &cpuset);
+    // Fetch the first frame to initialise various parameters
+    initialise();
 
-        pthread_setaffinity_np(thread_handles[i].native_handle(), sizeof(cpu_set_t), &cpuset);
+    // Sleep briefly here. This is to give the fetch_thread chance to spin up
+    // and read the initial gains
+    std::this_thread::sleep_for(0.5s);
```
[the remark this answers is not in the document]
jrs65: I don't think we need to anything special here. If it doesn't manage to fetch the update in time all that happens is that when processing incoming frames it sees there is no update in the queue and skips the frame (and warns that it is going so). I think that's fine, this only occurs at start up and the gains should generally be read much quicker than that anyway.

lib/stages/applyGains.cpp
jrs65: Errr. I guess the issue is that it might happen every 10s, so it's even worse. That is a good point so I'll turn this down to DEBUG.

jrs65:
> Glad to see this all wrapped up! I found a few comments that are out of date, but nothing very substantial.

Thanks for all your work on it @tristpinsm !

lib/stages/applyGains.cpp
```diff
@@ -381,6 +325,53 @@ void applyGains::apply_thread() {
...
+        // update gains
+        state_id_t state_id = dm.create_state<gainState>(update_id, transition_interval).first;
+        GainUpdate gain_update{std::move(gain_data.value()), transition_interval, state_id};
+
+        {
+            // Lock mutex exclusively while we update FIFO
+            std::lock_guard<std::shared_mutex> lock(gain_mtx);
+            gains_fifo.insert(double_to_ts(new_ts), std::move(gain_update));
+        }
+        INFO("Updated gains to {:s}.", update_id);
```
nritsche, line 370: I guess this one is okay to stay here, even if it's every 10s?

jrs65: Gah. Dang. Didn't see that comment. Given we're not going to 10s gains yet I'll leave it in for the moment. It's very likely we'll need to do some rewriting of this stage anyway for that.

PR #710, opened 2020-04-02
[WIP] Jrs/replay buffer

jrs65:
@tristpinsm this is the WIP branch that does a few refactors. Note, that I haven't tested my changes yet. It seems that the only unit test that uses `visRawReader` is for `visTranspose` so I guess they'll need to be changed together before they pass.

### DIFF
--- a/lib/stages/visRawReader.cpp
+++ b/lib/stages/visRawReader.cpp
@@ -272,7 +272,7 @@ int visRawReader::position_map(int ind) {
 void visRawReader::main_thread() {
 
     double start_time, end_time;
-    size_t frame_id = 0;
+    frameID frame_id(out_buf);
     uint8_t* frame;
 
     size_t ind = 0, read_ind = 0, file_ind;
@@ -357,8 +357,7 @@ void visRawReader::main_thread() {
             DEBUG("madvise failed: {:s}", strerror(errno));
 
         // Release the frame and advance all the counters
-        mark_frame_full(out_buf, unique_name.c_str(), frame_id);
-        frame_id = (frame_id + 1) % out_buf->num_frames;
+        mark_frame_full(out_buf, unique_name.c_str(), frame_id++);
         read_ind++;
         ind++;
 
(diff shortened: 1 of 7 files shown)


lib/stages/visRawReader.cpp
```diff
@@ -303,64 +322,52 @@ void visRawReader::main_thread() {
...
             frame.fpga_seq_length = 0;
+            frame.fpga_seq_total = 0;
+
             DEBUG("visRawReader: Reading empty frame: {:d}", frame_id);
         }
 
         // Set the dataset ID to the updated value
-        ((visMetadata*)(out_buf->metadata[frame_id]->metadata))->dataset_id = out_dset_id;
+        dset_id_t& ds_id = ((visMetadata*)(out_buf->metadata[frame_id]->metadata))->dataset_id;
+        ds_id = get_dataset_state(ds_id);
```
tristpinsm, line 353: I guess this means it will looks through the keys of the `ds_in_file` map for every frame. This is probably a small overhead, but I don't have a clear idea of how many different dataset IDs there can be in one file.

tristpinsm (commented):
This looks good to me so far. It's also made the handling of dataset IDs quite a bit clearer... After so many iterations it had gotten pretty messy.

tristpinsm:
Should I patch `visTranspose` on top of this branch and make a PR onto here then? I could also try and extend the unit test a bit to include empty frames and check the dataset IDs.

lib/stages/visRawReader.cpp
jrs65: Yeah. I figured it was worth the very small amount of overhead to clarify the handling. Lookups in a `std::map` are `O(log N)` so pretty cheap, and at most we might have 10s of datasets per file, and 100s if we go to very fast cadence updates.

jrs65:
> Should I patch `visTranspose` on top of this branch and make a PR onto here then? I could also try and extend the unit test a bit to include empty frames and check the dataset IDs.

Maybe a PR on top of this make sense. Extending the unit tests while you do so would be great.

jrs65:
@andrerenard this is still a thing. I'll try and review it's final status more in the near future.

andrerenard:
Closing due to age, and similar work in current code.

PR #711, opened 2020-04-02
Write out null dataset IDs for empty frames

tristpinsm:


lib/stages/visRawReader.cpp
```diff
@@ -351,6 +351,7 @@ void visRawReader::main_thread() {
         // Set the dataset ID to the updated value
         dset_id_t& ds_id = ((visMetadata*)(out_buf->metadata[frame_id]->metadata))->dataset_id;
         ds_id = get_dataset_state(ds_id);
+        ((visMetadata*)(out_buf->metadata[frame_id]->metadata))->dataset_id = ds_id;
```
jrs65, line 350: I don't get it. Isn't this achieved by using a reference to this two line above?

lib/stages/visTranspose.cpp
```diff
@@ -312,13 +300,30 @@ void visTranspose::main_thread() {
             frame.fpga_seq_length == 0 ? 0. : float(frame.rfi_total) / frame.fpga_seq_length;
         strided_copy(frame.gain.data(), gain.data(), offset * num_input + ti, write_t, num_input);
 
-        // Store original dataset ID (before adding time axis)
-        std::string dset_id_str = fmt::format("{}", base_ds_id);
-        if (dset_id_str.length() != DSET_ID_LEN - 1) {
+        // Parse the dataset ID
+        ds_id = frame.dataset_id;
+        if (frame.fpga_seq_length == 0) {
```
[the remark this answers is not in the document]
tristpinsm: My thinking was that you might in principle have an actual frame come in with `fpga_seq_total=0` for whatever reason (even though in reality this probably can never happen), and I was using `fpga_seq_length=0` as a flag to indicate that this frame never existed and was produced by `visRawReader` as an empty frame.

But I'm happy to change it to the standard check.

lib/stages/visRawReader.cpp
tristpinsm: Hmm, I guess that is right. I had to go through a few iterations before I got things to work, and I thought this was one problem along the way, not noticing it is a reference. I'll remove this commit and check it still works.

PR #712, opened 2020-04-03
Correct for packet loss in RFI SK estimation

andrerenard:
Corrects for the effect of packet loss on the RFI SK estimation.  

This PR also updates how some of the HFB stages operate to make them more generic and uses them to help with the RFI SK correction.


lib/stages/compressLostSamples.cpp
```diff
@@ -28,50 +27,57 @@ compressLostSamples::compressLostSamples(Config& config_, const std::string& uni
...
-                                                  / (NUM_SETS_OF_SUB_FREQS * _factor_upchan)] = 1;
-                    total_lost_samples += NUM_SETS_OF_SUB_FREQS * _factor_upchan;
-                    break;
-                }
+        for (uint32_t sample = 0; sample < _samples_per_data_set; sample += _compression_factor) {
+            // assert(sample/_compression_factor < (uint32_t)out_buf->frame_size);
+            out_frame[sample / _compression_factor] = 0;
+            for (uint32_t sub_index = 0; sub_index < _compression_factor; sub_index++) {
+                out_frame[sample / _compression_factor] += in_frame[sample + sub_index];
+            }
```
james-s-willis, line 76: Why not have:
```
               if(_zero_all_in_group && in_frame[sample + sub_index]) {
                    out_frame[sample / _compression_factor] = 1;
                    total_lost_samples += _compression_factor;
                    break;
                }
                else {
                    out_frame[sample / _compression_factor] += in_frame[sample + sub_index];
                }
            }
            if (!_zero_all_in_group) {                                                                                                                                                                               
                total_lost_samples += out_frame[sample / _compression_factor];
            }
```
which would mean you could exit the loop early if you find any non-zero values in the lost samples buffer. Slight optimisation but it might not matter as it introduces more `if` statements.

config/chime_science_run_gpu.yaml
```diff
@@ -1551,6 +1590,13 @@ rfi_sk_record:
     kotekan_stage: rfiRecord
     rfi_in: gpu_rfi_output_buffer_3
 
+rfi_pkt_loss_record:
+  total_links: 1
+  updatable_config: "/updatable_config/rfi_pkt_loss_record"
+  gpu_0:
+    kotekan_stage: rfiRecord
+    rfi_in: compressed_lost_samples_buf
```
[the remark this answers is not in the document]
andrerenard: Yes thanks for catching that.

lib/stages/compressLostSamples.cpp
andrerenard: Yeah that seems like a reasonable optimization.

PR #713, opened 2020-04-06
feat(InputSubset): a task for selecting data from a subset of inputs

jrs65:


lib/stages/InputSubset.cpp
```diff
@@ -0,0 +1,197 @@
...
+    register_consumer(in_buf, unique_name.c_str());
+    out_buf = get_buffer("out_buf");
+    register_producer(out_buf, unique_name.c_str());
+}
+
+void InputSubset::change_dataset_state(dset_id_t ds_id) {
+
+    auto& dm = datasetManager::instance();
+
+    auto fprint = dm.fingerprint(ds_id, {"inputs", "products", "state"});
```
tristpinsm, line 53: I couldn't figure out what the `"state"` dataset is for.

jrs65: Ooops. Should be `stack`!

jrs65: Good catch. This line combined with the check in the branch below that there is no `stackState` is mean to guarantee you never get any stacked data coming through, this would definitely have broken it.

lib/stages/InputSubset.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+    // If we enter here, that means we are processing the first frame, so we
+    // need to calculate the actual input and visibility indices we should copy
+    // out
+    if (fingerprint == fingerprint_t::null) {
```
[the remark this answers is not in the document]
jrs65: Makes sense!

Is this the only style comment? You made it sound like there were masses!

PR #724, opened 2020-04-15
Fix rfi_receiver.py script when started during solar transit

james-s-willis:


python/scripts/rfi_receiver/rfi_receiver.py
```diff
@@ -811,9 +812,10 @@ def rfi_zeroing():
         if time_to_transit_s < 0.5 * 3600 or time_to_transit_s > 23.5 * 3600:
 
             # Calculate time until end of solar transit window
-            downtime_s = time_to_transit_s
             if time_to_transit_s > 23.5 * 3600:
                 downtime_s = 24 * 3600 - time_to_transit_s
+            else:
+                downtime_s = time_to_transit_s + 1800
```
andrerenard, line 818: I think this is correct, but could we turn some of these constants into variables based off `downtime_m` and make that a config parameter.  If we ever want to adjust that value it's going to be hard to sort out what should happen here.

PR #726, opened 2020-04-16
Enable bias correction in SK estimation.

james-s-willis:
* Adds variance data output for the average of all (good) feeds.
* Uses the variance to correct the bias in the SK estimation caused by low bit depth sampling. 
* Cleans up some of the kernel names and removes unused files.


config/chime_science_run_gpu_no_output.yaml
```diff
@@ -737,7 +737,7 @@ gpu:
     #Copy out queue
     - name: hsaRfiBadInputOutput
     - name: hsaRfiOutput
-    #- name: hsaRfiVarOutput
+    - name: hsaRfiVarOutput
```
andrerenard, line 740: This command object isn't really used by the current config, since the name/size of the variance array hasn't been updated in it.  The size won't actually change, but the output will come from the `rfi_chime_input_sum` since there needs to be a reduction across inputs (with masking).   For this stage of the testing I'd suggest commenting it out.

lib/hsa/hsaRfiTimeSum.cpp
```diff
@@ -73,21 +73,21 @@ hsa_signal_t hsaRfiTimeSum::execute(int gpu_frame_id, hsa_signal_t precede_signa
...
     // Initialize arguments
     memset(&args, 0, sizeof(args));
     // Set argumnets to correct values
     args.input = device.get_gpu_memory_array("input", gpu_frame_id, input_frame_len);
-    args.output = device.get_gpu_memory("timesum", output_frame_len);
-    // args.output_var =
-    //    device.get_gpu_memory_array("rfi_output_var", gpu_frame_id, output_var_frame_len);
+    args.output = device.get_gpu_memory("time_sum", output_frame_len);
+     args.output_var =
+        device.get_gpu_memory_array("rfi_output_var", gpu_frame_id, output_var_frame_len);
```
andrerenard, line 87: since this is now an intermediate product, it can just be a `get_gpu_memory()` call.

andrerenard: And it should be renamed something like `rfi_time_sum_var`, since it hasn't been input summed yet.

lib/hsa/kernels/rfi_chime_input_sum.cl
```diff
@@ -31,18 +33,21 @@ rfi_chime_inputsum(
...
     //Declare Local Memory
     __local float sq_power_across_input[256];
+    __local float var_across_input[256];
     // Compute index in input array
     uint base_index = gx + gy * num_elements + gz * num_elements * gy_size;
     sq_power_across_input[lx] = input_mask[lx]*input[base_index];
     // Partial sum if more than 256 inputs
     for(int i = 1; i < num_elements/lx_size; i++) {
         sq_power_across_input[lx] += input_mask[lx + i * lx_size] * input[base_index + i * lx_size];
+        var_across_input[lx] += input_mask[lx + i * lx_size] * input_var[base_index + i * lx_size];
```
andrerenard, line 43: `var_across_input[lx]` is never initialized to a starting value, so you are just adding to whatever is in memory.  You need a line like the `sq_power_across_input[lx] = input_mask[lx]*input[base_index];` one above.

lib/hsa/kernels/rfi_chime_time_sum.cl
```diff
@@ -60,24 +60,10 @@ rfi_chime_timesum(
...
-    //        case 3:
-    //            output_var[gy] = power_across_time.s3;
-    //            break;
-    //    }
-    //}
+    //Output the variance, normalising with the number of good inputs 
+    output_var[address + 0] = power_across_time.s0 / (num_elements - num_bad_inputs);
+    output_var[address + 1] = power_across_time.s1 / (num_elements - num_bad_inputs);
+    output_var[address + 2] = power_across_time.s2 / (num_elements - num_bad_inputs);
+    output_var[address + 3] = power_across_time.s3 / (num_elements - num_bad_inputs);
```
andrerenard, line 67: This is a strange place to renormalize for bad inputs since you are only summing over a few inputs here.   I would factor this out and renormalize in the next kernel after the sum in elements.

lib/hsa/kernels/rfi_chime_input_sum.cl
james-s-willis: Ah thanks, that'll be it. I never noticed that

lib/hsa/kernels/rfi_chime_input_sum.cl
```diff
@@ -55,16 +66,35 @@ rfi_chime_inputsum(
...
+            const float var = var_across_input[0] * one_over_256;
+            const float var2 = var * var;
+            const float var_sqrt = sqrt((double)var);
+            const float sk_correction = bias_coeffs[0] * var2 * var_sqrt + 
+                                        bias_coeffs[1] * var2 +
+                                        bias_coeffs[2] * var * var_sqrt +
+                                        bias_coeffs[3] * var +
+                                        bias_coeffs[4] * var_sqrt +
+                                        bias_coeffs[5]; 
+            const float bias = 1.0 - sk_correction;
```
jrs65, line 85: Just change the coefficient definitions to remove this one, i.e. change your def of `bias_coeffs[5]` to subtract 1 from it, and remove this line.

lib/hsa/kernels/rfi_chime_input_sum.cl
```diff
@@ -55,16 +66,35 @@ rfi_chime_inputsum(
...
+
+            // Correct SK for truncation bias
+            SK += bias * trunc_bias_switch;
```
jrs65, line 88: Like this. 

Note you need to invert the sign on the actual SK correction, but that's actually better as a bias would typically be defined such that you *subtract* to correct it.

```suggestion
            const float bias = bias_coeffs[0] * var2 * var_sqrt + 
                               bias_coeffs[1] * var2 +
                               bias_coeffs[2] * var * var_sqrt +
                               bias_coeffs[3] * var +
                               bias_coeffs[4] * var_sqrt +
                               bias_coeffs[5]; // You'll need to change the definition of this coeff above to subtract 1 from it

            // Correct SK for truncation bias
            SK -= bias * trunc_bias_switch;
```

lib/hsa/kernels/rfi_chime_input_sum.cl
```diff
@@ -55,16 +66,35 @@ rfi_chime_inputsum(
...
+            output_var[address] = -1.0;
             output_mask[address] = 1;
         } else {
             float new_sq_power_across_input = sq_power_across_input[0] * cf * cf;
             float SK = ((n + 1) / (n - 1)) * ((new_sq_power_across_input / (n * N)) - 1);
+
+            // Calculate the truncation bias correction for the SK value
+            const float var = var_across_input[0] * one_over_256;
+            const float var2 = var * var;
+            const float var_sqrt = sqrt((double)var);
```
jrs65, line 78: This is called the standard deviation (typical abbreviation `stddev`) or the `rms`. It makes more sense to call it that than `var_sqrt`.

lib/hsa/kernels/rfi_chime_input_sum.cl
```diff
@@ -55,16 +66,35 @@ rfi_chime_inputsum(
...
+            const float sk_correction = bias_coeffs[0] * var2 * var_sqrt + 
+                                        bias_coeffs[1] * var2 +
+                                        bias_coeffs[2] * var * var_sqrt +
+                                        bias_coeffs[3] * var +
+                                        bias_coeffs[4] * var_sqrt +
+                                        bias_coeffs[5];
```
jrs65, line 84: Two comments here. First I think you'd be better of reversing the order of `bias_coeffs`. There's two reasons for that i) there's a direct correspondence between the index and the power you are raising the rms to; ii) if we need to add terms to make things more accurate (i.e. a 6th order term) you don't need to rewrite this expression.

Second, evaluating like this is a little clunky. You might consider writing this with a for loop (and thus building up the powers as you go), or writing it as a nested series for multiply-add statements, e.g. (with the ordering of bias_coeffs reverse):
```C
// either ...
float sk_correction = 0;
float rms_pow = 1;
for (int i = 0; i < NCOEFF; i++) {
    sk_correction += bias_coeffs[i] * rms_pow;
    rms_pow *= rms;
}
/// ... or ...
const float sk_correction = bias_coeffs[0] + rms * (bias_coeffs[1] + rms * (bias_coeffs[2] + rms * (bias_coeffs[3] + rms * ...)
```

lib/hsa/kernels/rfi_chime_input_sum.cl
jrs65, line 78: Arash might be able to provide you a polynomial fit against var instead of the rms, which would save the sqrt.

jrs65: Arash is trying to do a fit for this one now. I think he's going to put the results in chime-experiment/RTPipeline#73

lib/hsa/kernels/rfi_chime_input_sum.cl
```diff
@@ -55,16 +67,33 @@ rfi_chime_inputsum(
...
             output[address] = -1.0;
+            output_var[address] = -1.0;
             output_mask[address] = 1;
         } else {
             float new_sq_power_across_input = sq_power_across_input[0] * cf * cf;
             float SK = ((n + 1) / (n - 1)) * ((new_sq_power_across_input / (n * N)) - 1);
+
+            // Calculate the truncation bias correction for the SK value
+            const float var = var_across_input[0] / (n * N);
+            const float rms = sqrt((double)var);
```
andrerenard, line 78:
```suggestion
            const float rms = sqrt((float)var);
```
Doubles are very slow on this GPU, stick to floats unless we really need to.

lib/hsa/kernels/rfi_chime_input_sum.cl
```diff
@@ -55,16 +67,33 @@ rfi_chime_inputsum(
...
+            float sk_correction = 0.f;
+            float rms_pow = 1.f;
+            for(int i=0; i<NUM_COEFFS; i++) {
+                sk_correction += bias_coeffs[i] * rms_pow;
+                rms_pow *= rms;
+            }
+
+            // Correct SK for truncation bias
+            SK -= sk_correction * trunc_bias_switch;
```
andrerenard, line 87: I would do an `if` on the `trunc_bias_switch`  then you can control if we need to do extra work in this kernel or not.  Since all the work units will follow the same `if` path, there is shouldn't be a performance hit doing it that way.

PR #729, opened 2020-04-20
CMake modernize & cleanup

### LINKED ISSUE #701: Modernize CMake build files
The current set of CMake files is a barely understandable, poorly structured horror show of bad practices. We should refactor these to make them logical, consistent and use modern conventions.

Suggestions for what we should do:
- [x] Build proper INTERFACE library targets, use `target_include_directories` and then `target_link_libraries` will automatically take care of including the right source files.
- [ ] Use `FindPackage` and `target_link_libraries` rather than separate source includes for dependencies.
- [x] Consistent spacing conventions.

Any more things? (edit, and add above)

This seems like the kind of task that @nritsche seems to derive unreasonable amounts of enjoyment from.

nritsche:
Closes #701 

- Use cmake_format for style cleanup
- Add style guide to docs
  - Please complain early about these (but they match our C++ style guide)
    - line length max: 100
    - indentation: 4 spaces (not tab)
- [x] remove `ar/fix_whole_archive_linking` from github action trigger list


lib/gpu/CMakeLists.txt
```diff
@@ -4,3 +4,9 @@ add_library(kotekan_gpu gpuDeviceInterface.cpp gpuEventContainer.cpp gpuProcess.
 
 target_link_libraries(kotekan_gpu PRIVATE libexternal kotekan_libs)
 target_include_directories(kotekan_gpu PUBLIC .)
+
+target_link_libraries(
+    kotekan_gpu
+    PRIVATE ${LIBEVENT_BASE} ${LIBEVENT_CORE} ${LIBEVENT_PTHREADS} ${LIBEVENT_EXTRA} event)
```
andrerenard, line 10: I don't think you need to link the libraries, it's just the includes that are missing.   However given the include didn't work below.  I don't think this is a case of cmake not cross linking this correctly from kotekan_libs, but a case of MacOS having a different path that's no being picked up by FindLIBEVENT correctly.   It would a good idea to check what `${LIBEVENT_INCLUDE_DIR}` is getting set too.  Also if you need access to a MacOS system, the one in LWLab that we used for Jenkins should still be working, or let me know if you'd like me to troubleshoot on my laptop.

nritsche: If you have a moment to have a look on your MacOS laptop, that would be great!

CMakeLists.txt
```diff
@@ -1,18 +1,33 @@
...
-add_definitions(/D__STDC_FORMAT_MACROS)
+# Require at least c++17 support from the compiler
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+# this makes sure we use -std=c++17, not -std=gnu++17
+set(CMAKE_CXX_EXTENSIONS OFF)
 
-# This quiets a noisy cmake warning on newer versions,
-# triggered by some of our CMake/FindXYZ.cmake scripts.
+# set C standart to gnu99
```
jrs65, line 15:
```suggestion
# set C standard to gnu99
```

lib/stages/CMakeLists.txt
```diff
@@ -1,131 +1,115 @@
...
-# HDF5 stuff
-if (${USE_HDF5})
-  set (KOTEKAN_PROCESS_LIB_SOURCES ${KOTEKAN_PROCESS_LIB_SOURCES} basebandReadout.cpp applyGains.cpp)
-endif ()
-
-# HDF5 and OpenMP
-if (${USE_HDF5} AND ${USE_OMP})
-  set (KOTEKAN_PROCESS_LIB_SOURCES ${KOTEKAN_PROCESS_LIB_SOURCES} visTranspose.cpp)
+if(${USE_FFTW})
+    set(KOTEKAN_PROCESS_LIB_SOURCES ${KOTEKAN_PROCESS_LIB_SOURCES} fftwEngine.cpp)
```
jrs65, line 76: It seems odd that this one still needs an explicit `set`. Is that correct? If so, maybe add a comment about it.

jrs65 (approved):
This is looking fantastic. Far cleaner! 👍 👍 

I had two small comments, but nothing major.

lib/stages/CMakeLists.txt
nritsche: Looks like I forgot this one

nritsche:
@andrerenard I think this is ready for review. I'm still going to run this on a test node for a bit.
Since the commit sausage is a mess, this can be a squash commit.
- [ ] Write summary once reviewed

nritsche:
build(cmake): modernize & cleanup
* ci(lint.sh): add cmakelint
* docs(styleguide): add cmake style & autolint
* build(cmake): don't set compiler flags manually
  * As per https://cliutils.gitlab.io/modern-cmake/chapters/features/cpp11.html 
    don't set any CMAKE_*_FLAGS. Instead let cmake figure out and maintain
    selection of the correct compiler flags.
* ci(github): turn on cmake warnings and warnings are errors
* build(cmake): add CCHACHE option
  * If -DCCHACHE passed to cmake, it will try to find and configure ccache
    for the build.
* build(cmake): bump minimum cmake version to 3.8.2
  * We are using C++17 features. We should use a cmake version released
    after C++17.
* build(cmake): remove unused C formatting definition
* build(cmake): move external library includes to target level
  * From the superproject directory level, move to target level where
    needed.
* build(cmake): let cmake decide if compiler supports options
* build(cmake): set link_library propagation keyword
  * Set PRIVATE where library not used in header files.
* build(cmake): move SSL include to target level
* refactor(applyGains): use mutex to protect ts_frame
  * Instead of making ts_frame atomic, protect it with a mutex.
    This let's us remove -latomic linker option.
* build(cmake): remove -latomic linker option
  * Now unused and doesn't exist for AppleClang
* build(cmake): remove libdl linking
  * kotekan executable doesn't use dl library
* ci(github): turn on CMAKE_LINK_WHAT_YOU_USE warnings
* build(cmake): remove static glibc&libstdc++ linking
* build(cmake): set C standard using set_property
* build(cmake): activate -Wsuggest-override
* build(cmake): add cmake option -DWERROR (default OFF)
* refactor(visCompression): move utilities to utils/Stack
* build(cmake): eliminate KOTEKAN_*_SOURCES variables
  *  remove all variables listing sources
  * replace with add_library, target_sources
* build(cmake): move warnings and optimization to the top
* build(cmake): link numa where needed only
* build(cmake): don't link lrt on MacOS/with clang
* build(cmake): set module path once per project
* build(cmake): consistent use of target_include_directories and target_link_libraries
  * remove all project-wide linking and include cmake commands
  * replace with commands per-target
* build(cmake) replace _DBOOST_TESTS with -DWITH_TESTS
  * also affects lib/testing
* style(visBuffer): mark VisFrameView::data_size override
* style(HFBFrameView): mark data_size override
* build(cmake) Fix path issue with libevent on MacOS
* build(cmake) Fix issue linking to homebrew libs on MacOS >= 10.14
* build(cmake) Fix OpenCL linking issue
* build(cmake) Stop boost warnings from breaking MacOS build
* tests(bip_buffer) Remove unneeded lambda capture

Closes #701

CMakeLists.txt
```diff
@@ -29,374 +44,351 @@ option(NO_MEMLOCK "Do not lock buffer memory (useful when running in Docker)" OF
 option(SUPERDEBUG "Enable extra debugging with no optimisation" OFF)
 option(SANITIZE "Enable clang sanitizers for testing" OFF)
 option(COMPILE_DOCS "Use Sphinx to compile documentation" OFF)
-option(BOOST_TESTS "Compile Boost C++ unit tests" OFF)
+option(WITH_TESTS "Compile testing library and boost C++ unit tests" OFF)
 option(_GOSSEC "Set buffer status messages to DEBUG log level" OFF)
 option(IWYU "Enable include-what-you-use and print suggestions to stderr" OFF)
+option(CCACHE "Use ccache to speed up the build" OFF)
+option(WERROR "Warnings are errors" OFF)
```
andrerenard, line 51: Should we not have this on by default?

nritsche: The idea is that we turn on a lot of warnings, but only make them errors on the CI server. This way we can't merge stuff that generates warnings, we also see our local compiler spit out all of them, but we never get stopped in e.g. the deployment process, when suddenly a compiler decides to warn about something new.
I remember the last example being very annoying once in the past. Also this was the recommended way by one of the cmake people in a talk I saw somewhere...

cmake/FindLAPACKE.cmake
```diff
@@ -7,185 +7,194 @@
...
-#   LAPACKE_INCLUDE_DIRS - list of required include directories
-#   LAPACKE_LIBRARIES - list of libraries to be linked
-#   LAPACKE_VERSION_MAJOR - major version number
-#   LAPACKE_VERSION_MINOR - minor version number
-#   LAPACKE_VERSION_PATCH - patch version number
-#   LAPACKE_VERSION_STRING - version number as a string (ex: "0.2.18")
+# LAPACKE_FOUND - set to true if the library is found LAPACKE_INCLUDE_DIRS - list of required
+# include directories LAPACKE_LIBRARIES - list of libraries to be linked LAPACKE_VERSION_MAJOR -
+# major version number LAPACKE_VERSION_MINOR - minor version number LAPACKE_VERSION_PATCH - patch
+# version number LAPACKE_VERSION_STRING - version number as a string (ex: "0.2.18")
```
andrerenard, line 20: I don't love how this system reflows comments.   There are a lot of cases like this were the original spacing was much nicer to read.

lib/dpdk/CMakeLists.txt
```diff
@@ -1,23 +1,15 @@
 project(kotekan_libs)
```
andrerenard, line 1:
```suggestion
project(kotekan_dpdk)
```

lib/dpdk/CMakeLists.txt
```diff
@@ -1,23 +1,15 @@
...
-include_directories(SYSTEM ${DPDK_INCLUDE_DIR})
-target_link_libraries(kotekan_dpdk ${DPDK_LIBRARIES})
-
-target_link_libraries(kotekan_dpdk libexternal)
+target_include_directories(kotekan_dpdk SYSTEM PRIVATE ${DPDK_INCLUDE_DIR})
+target_link_libraries(kotekan_dpdk PRIVATE ${DPDK_LIBRARIES} ${NUMA_LIBRARY} libexternal
+                                           kotekan_libs)
+target_include_directories(kotekan_dpdk PUBLIC .)
 
 add_definitions(-DWITH_DPDK)
```
andrerenard, line 13:
```suggestion
```
Doesn't look like this is actually used anywhere.

lib/dpdk/CMakeLists.txt
```diff
@@ -1,23 +1,15 @@
...
 
-if(NOT ${NUMA_FOUND})
-    message( FATAL_ERROR "ERROR: libnuma is required for DPDK.  Please install libnuma")
-endif()
+find_package(NUMA REQUIRED)
```
andrerenard, line 15: I think it works because cmake is odd, but maybe move this above line 9 since it's using a variable defined by this call.

lib/hcc/CMakeLists.txt
```diff
@@ -1,26 +1,24 @@
-project( libkotekan_hcc )
+project(libkotekan_hcc)
```
andrerenard, line 1:
```suggestion
project(kotekan_hcc)
```
Consistency with other libraries.

lib/hsa/CMakeLists.txt
```diff
@@ -1,69 +1,54 @@
...
-      hsaTrackingBeamformOutput.cpp
-      hsaTrackingUpdatePhase.cpp
-    )
-
-add_library ( kotekan_hsa
-              ${KOTEKAN_HSA_SOURCES})
-
-target_link_libraries( kotekan_hsa libexternal )
-target_link_libraries( kotekan_hsa kotekan_utils )
+project(libkotekan_hsa)
```
andrerenard, line 1:
```suggestion
project(kotekan_hsa)
```

lib/hsa/kernels/CMakeLists.txt
```diff
@@ -1,140 +1,134 @@
-# NOTE THAT THESE ARE HARDCODED FOR FIJI ARCHITECTURE!
-# SHOULD ADD A FLAG OR SOMETHING TO ALLOW OTHERS.
+# NOTE THAT THESE ARE HARDCODED FOR FIJI ARCHITECTURE! SHOULD ADD A FLAG OR SOMETHING TO ALLOW
+# OTHERS.
 
-
-# Find the HSACO compilers
-# Find the AMD clang
-# This should really test something about the clang to
+# Find the HSACO compilers Find the AMD clang This should really test something about the clang to
```
andrerenard, line 4: Another reflow issue.

tools/docker/Dockerfile
```diff
@@ -42,7 +42,8 @@ RUN python3.7 -m pip install --upgrade pip==20.2.2 && \
     python3.7 -m pip install --no-cache-dir numpy==1.19.1 && \
     python3.7 -m pip install --no-cache-dir pkgconfig==1.5.1 && \
     python3.7 -m pip install --no-cache-dir --upgrade cython==0.29.21 && \
-    python3.7 -m pip install --no-cache-dir black==19.10b0
+    python3.7 -m pip install --no-cache-dir black==19.10b0 &&\
+    python3.7 -m pip install --no-cache-dir cmake_format
```
andrerenard, line 46: should we lock the `cmake_format` version?

andrerenard (changes requested):
Looks great!  Just a few small comments inline.

CMakeLists.txt
andrerenard: You make a good point, let's leave it `OFF` then.

cmake/FindLAPACKE.cmake
nritsche: This was actually just missing `*` to be a bullet point list, see https://github.com/kotekan/kotekan/blob/4ed7042c08d31588d51028301f6ab0dbe45ed8f5/tools/cmake_format_config.py#L107

nritsche:
Thanks for the thorough review, @andrerenard. I will have to run CI twice to get and set the version of `cmake_format`.

andrerenard (approved):

PR #736, opened 2020-04-23
Clang and MacOS CI builds

nritsche:
Adds 2 builds to github action CI:
- full CHIME build using clang
- macOS build outside of dockerimage: slower because needs to install stuff
  - takes 15min with an empty ccache
  - and 4min after ccache has been filled

This is the compiler the mac build uses:
```
 -- The C compiler identification is AppleClang 11.0.0.11000033
-- The CXX compiler identification is AppleClang 11.0.0.11000033
-- Check for working C compiler: /Applications/Xcode_11.3.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc
-- Check for working C compiler: /Applications/Xcode_11.3.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc - works
```
Let me know if that's what we want.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -174,6 +174,111 @@ jobs:
         docker run "${OPTS[@]}" ccache -s
 
 
+  # Build a full CHIME version with clang
+  build-clang:
+    runs-on: ubuntu-latest
+    needs: build-docker
+
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: Cache ccache files
+      uses: actions/cache@v1
+      with:
+        path: .ccache
+        key: ccache-clang-build-${{ github.sha }}
+        restore-keys: |
+          ccache-clang-build
+
+    - name: Build kotekan
+      env:
+        CC: clang
+        CXX: clang++
+      run: |
+        OPTS=(--rm --mount type=bind,src=$(pwd),target=/code/kotekan -w /code/kotekan/build/ ${IMG_CORE})
+        docker login docker.pkg.github.com -u ${GITHUB_REPOSITORY} -p "${{ secrets.GITHUB_TOKEN }}"
+        docker run "${OPTS[@]}" ccache -s
+        docker run "${OPTS[@]}" cmake \
+        -DUSE_DPDK=ON \
+        -DCMAKE_BUILD_TYPE=Debug \
+        -DUSE_HDF5=ON \
+        -DHIGHFIVE_PATH=/code/build/HighFive \
+        -DUSE_LAPACK=ON \
+        -DBLAZE_PATH=/code/build/blaze \
+        -DARCH=haswell \
+        -DNO_MEMLOCK=ON \
+        -DUSE_OMP=ON \
+        -DBOOST_TESTS=ON \
+        -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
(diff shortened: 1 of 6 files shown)


nritsche:
I'm not sure if I broke the ccache. The unit test build took quite long the last time.
I gave each of the 5 ccache caches 1GB max size. That's how much github says they give us before they throw stuff out.

jrs65:
@nritsche while you are doing this you might consider turning off Debug mode. The object files are far smaller and so the ccache will be smaller and survive longer in githubs cache.

nritsche:
> @nritsche while you are doing this you might consider turning off Debug mode. The object files are far smaller and so the ccache will be smaller and survive longer in githubs cache.

I would have th

nritsche:
oups

jrs65:
> This looks great, thanks @nritsche . I think it makes sense to use the default Apple provided Xcode version of `clang` since that's what most people will have by default.

@andrerenard, @cubranic  what do you use for building on your machine? I think we're probably the main mac using kotekan devs. I usually use a recent clang installed via homebrew (to get OpenMP support), so this tests exactly what I use.

nritsche:
Can I rebase this instead of squash? @andrerenard @jrs65

andrerenard:
@nritsche sure.

cubranic:
So I used to compile using a Homebrew `gcc-7`, in order to keep in line with the site computers. But then after an OS and Xcode upgrade there was a problem with gcc finding some header or other. So I switched to clang, and I think I just copied Richard's CC settings. Which I have totally forgotten, but it appears to be using Clang-9 that I installed with Homebrew, as it's using `ccache /usr/local/opt/llvm/bin/clang++` when I run `make VERBOSE=1`.

jrs65:
Okay. Looks like two votes for clang-9, one marginal vote for XCode clang, so I guess we should just stick with clang-9 as Rick has in there? @andrerenard ?

andrerenard:
I'm ok with `clang-9` since I'm guessing it'll be close to the default Apple clang, but this PR is using the Xcode clang at the moment. Would adding a `brew install` for `clang-9` make this take longer, or will it remember the image?

cubranic:
Mine was not a vote for clang-9! It just happens to be what I use right now, but please give it 0 weight. I’d also be perfectly happy to use the standard Xcode clang if you tell me how to.

andrerenard:
I still have a preference towards the Xcode clang, just on the basis of simplifying the default install process for people setting up MacOS builds for the first time.  Unless we actually need `clang-9` to enable kotekan features - which doesn't appear to be the case at least in this PR unless something is missing.

jrs65:
> I still have a preference towards the Xcode clang, just on the basis of simplifying the default install process for people setting up MacOS builds for the first time. Unless we actually need `clang-9` to enable kotekan features - which doesn't appear to be the case at least in this PR unless something is missing.

I don't care that much either then, and it currently works fine, so let's stick with Xcode clang.

nritsche:
> so I guess we should just stick with clang-9 as Rick has in there?

> so let's stick with Xcode clang.

I'm happy I am not the only one confused by the names MacOS has for compilers and what the github action actually uses here.

React with :mask: if you vote for merging this as is.

PR #737, opened 2020-04-24
Receiver Config: Keep more gain updates (for late frames)

nritsche:
- increase `num_kept_updates` to ~`15`~ `24` (was `5`) for receiveFlags and applyGains
- add debug output for case where this was not enough for a frame arriving too late still

The log addition is tested and produces sth like this now:
```
/applyGains_test: No gains update is as old as the currently processed frame.
/applyGains_test: Frame frequency ID: 250
Frame timestamp: 1.5e+09
Update queue: 1500000010.000000 1500000015.000000 
```


lib/stages/applyGains.cpp
```diff
@@ -656,6 +656,16 @@ applyGains::calculate_gain(double timestamp, uint32_t freq_ind, std::vector<cflo
...
         WARN("No gains update is as old as the currently processed frame.");
+
+        // Look up frequency ID and print more info about late frame
+        auto it = freq_map.begin();
+        for (; it != freq_map.end(); ++it) {
+            if (it->second == freq_ind)
+                break;
+        }
+        WARN("Frame frequency ID: {}\nFrame timestamp: {}\nUpdate queue: {}", it->first, timestamp,
+             gains_fifo);
```
jrs65, line 667: If for some reason the `freq_ind` isn't found this would attempt to dereference a non existent entry. For that reason I don't think it makes sense to have the WARN outside of the loop, so you might as well move it inside and then you get to use a structured binding range based for, and rejoice in the wonders of C++17. Something like:
```suggestion
        for (const auto& [fm_id, fm_ind] : freq_map) {
            if (fm_ind == freq_ind) {
                WARN(
                    "Frame frequency ID: {}\nFrame timestamp: {}\nUpdate queue: {}",
                    fm_id, timestamp, gains_fifo
                );
                break;
            }
        }

```

nritsche:
The config change was also done by hotfix #745 , so we can drop commit 9ca911d here once develop is up to date with master. Afterwards this PR still has an improvement to debug logging.

andrerenard:
Master has been merged back into `develop`, @nritsche can you update this and then we can merge it in?

lib/stages/applyGains.cpp
```diff
@@ -683,6 +683,17 @@ applyGains::calculate_gain(double timestamp, uint32_t freq_ind, std::vector<cflo
 
     if (update_new == nullptr) {
         WARN("No gains update is as old as the currently processed frame.");
+
+        // Look up frequency ID and print more info about late frame
```
jrs65, line 687:
```suggestion
        // Look up frequency ID as this method only gets passed the index,
        // and print more info about late frame 
```

andrerenard:
@nritsche can you check @jrs65's one change request this morning, and then we should merge this in.

jrs65:
Looks good.

PR #738, opened 2020-05-01
Created stage to delay frames.

james-s-willis:


lib/stages/bufferDelay.cpp
```diff
@@ -0,0 +1,104 @@
...
+        // If the in buffer is holding the right amount of frames, release one
+        if (in_frame_hold_ctr == _num_frames_to_hold) {
+          if (_copy_frame) {
+            allocate_new_metadata_object(out_buf, out_frame_id);
+            copy_metadata(in_buf, in_frame_release_id, out_buf, out_frame_id);
+            std::memcpy(output_frame, in_buf->frames[in_frame_release_id], in_buf->frame_size);
+          } else {
+            pass_metadata(in_buf, in_frame_release_id, out_buf, out_frame_id);
+            swap_frames(in_buf, in_frame_release_id, out_buf, out_frame_id);
+          }
```
andrerenard, line 78: This could use more checks to make sure this these are valid operations, see: `Valve::copy_frame`

lib/stages/bufferDelay.cpp
```diff
@@ -0,0 +1,104 @@
...
+    //frameID in_frame_id(in_buf);
+    uint32_t in_frame_id = 0;
+    uint32_t in_frame_hold_ctr = 0;
+    uint32_t in_frame_release_id = 0;
+    uint32_t out_frame_id = 0;
+
+    uint8_t* output_frame =
+      wait_for_empty_frame(out_buf, unique_name.c_str(), out_frame_id);
+    if (output_frame == nullptr)
+      return; // Shutdown condition
```
andrerenard, line 59: I don't see why you need to get an output frame here.  You can just get it as needed in the loop below as needed.

lib/stages/bufferDelay.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include "visUtil.hpp"
+
+#include <string> // for string
+#include <tuple>  // for tuple
+#include <vector> // for vector
+
+/**
+ * @brief Delays release of frames until the buffer is sufficiently full.
+ *
+ *
```
andrerenard, line 17: Missing @config section

lib/stages/bufferDelay.cpp
```diff
@@ -0,0 +1,104 @@
...
+    while (!stop_thread) {
+
+        uint8_t* input_frame = wait_for_full_frame(in_buf, unique_name.c_str(), in_frame_id);
+        if (input_frame == nullptr)
+            return;
+        
+        INFO("Got new input frame {:d}", in_frame_id);
+
+        // If the in buffer is holding the right amount of frames, release one
+        if (in_frame_hold_ctr == _num_frames_to_hold) {
```
andrerenard, line 70: I think you can make this logic a little easier to follow with something like 
```
if (in_frame_hold_ctr >= _num_frames_to_hold) 
   copy_frame
else
   in_frame_hold_ctr++
```
Then you don't have to subtract and add every time.

config/chime_science_run_recv.yaml
```diff
@@ -191,11 +194,18 @@ count_check:
     kotekan_stage: countCheck
     in_buf: visbuf_5s_26m_gated
 
+delay_frames:
+  num_frames_to_hold: 3072
```
andrerenard, line 198: You need to set `visbuf_10s_flags` to have `> 3072` frames for this config to work.  It might be worth asking Richard to check if this how he wanted to delay things here.

config/chime_science_run_recv.yaml
```diff
@@ -47,6 +47,9 @@ vis_buffer:
   visbuf_10s_flags:
     kotekan_buffer: vis
     buffer_depth: 384  # Before slow stage
+  visbuf_10s_flags_delay:
+    kotekan_buffer: vis
+    buffer_depth: 384  # Before slow stage
```
andrerenard, line 52: It's not clear we need this many frames in the output buffer?   Also we shouldn't commit this change until we get more memory in `recv2` or this will just fail to run.

lib/stages/bufferDelay.cpp
```diff
@@ -0,0 +1,116 @@
...
+            output_frame = wait_for_empty_frame(out_buf, unique_name.c_str(), out_frame_id);
+            if (output_frame == nullptr)
+                return;
+
+            // Buffer sizes must match exactly
+            if (in_buf->frame_size != out_buf->frame_size) {
+                throw std::runtime_error(fmt::format(
+                    fmt("Buffer sizes must match for direct copy (src {:d} != dest {:d})."),
+                    in_buf->frame_size, out_buf->frame_size));
+            }
```
andrerenard, line 79: This could be moved to the constructor, the values don't depend on the frames.

lib/stages/bufferDelay.cpp
```diff
@@ -0,0 +1,116 @@
...
+                    throw std::runtime_error(fmt::format(
+                        fmt("Metadata sizes must match for direct copy (src {:d} != dest {:d})."),
+                        in_buf->metadata[in_frame_release_id]->metadata_size,
+                        out_buf->metadata[out_frame_id]->metadata_size));
+                }
+                copy_metadata(in_buf, in_frame_release_id, out_buf, out_frame_id);
+                std::memcpy(output_frame, in_buf->frames[in_frame_release_id], in_buf->frame_size);
+            } else {
+                pass_metadata(in_buf, in_frame_release_id, out_buf, out_frame_id);
+                swap_frames(in_buf, in_frame_release_id, out_buf, out_frame_id);
```
andrerenard, line 95: There should be a check if this is possible.  I.e. are we the only consumer of `in_buf`, otherwise swapping frames will violate the memory model here.  If we cannot `swap_frames()` then it should do the copy above.

lib/stages/bufferDelay.cpp
```diff
@@ -0,0 +1,116 @@
...
+                  in_frame_id, in_frame_hold_ctr, in_frame_release_id);
+
+            mark_frame_full(out_buf, unique_name.c_str(), out_frame_id);
+            out_frame_id++;
+
+            // Release the input buffer frame.
+            mark_frame_empty(in_buf, unique_name.c_str(), in_frame_release_id);
+            in_frame_release_id++;
+        } else
+            in_frame_hold_ctr++;
```
andrerenard, line 109:
```suggestion
        } else {
            in_frame_hold_ctr++;
        }
```
I don't think this is against the style guide, but I don't like mixing the use (or not) of `{}` in the same statement.

config/chime_science_run_recv.yaml
james-s-willis: Ah yes, it needs to be a bit bigger than 30s. @jrs65 how many frames should we hold here? 3072 frames is 30 seconds, so do we want 4096 frames or 3584?

jrs65: I'm trying to figure that one out now. I need to know how much memory we are going to put in there first. Will get back to this soon.

config/chime_science_run_recv.yaml
```diff
@@ -45,6 +45,9 @@ vis_buffer:
     kotekan_buffer: vis
     buffer_depth: 384  # Before slow stage
   visbuf_10s_flags:
+    kotekan_buffer: vis
+    buffer_depth: 3200 # Before slow stage
```
jrs65, line 49:
```suggestion
    buffer_depth: 4096 # Before slow stage
```

PR #741, opened 2020-05-07
Check if the new gain update requires a new dataset ID. 

### LINKED ISSUE #740: applyGains should only register a new dataset when told to by the cal broker
At the moment, the applyGains process will register a new gainState and change the dataset_id every time it receives a new update from the cal broker. We need to change this such that in the json update there is a parameter which tells it whether an update requires a new gainState to be registered. Only if that is set to be true will it register a new id, when it is set to be false it uses the existing gainState and doesn't register a new dataset_id, and thus doesn't change the output dataset_id.

james-s-willis:
If the gain update does not require a dataset ID, this will use the current one. Fixes #740 .

### DIFF
--- a/lib/stages/applyGains.cpp
+++ b/lib/stages/applyGains.cpp
@@ -118,6 +118,7 @@ bool applyGains::receive_update(json& json) {
     std::string gains_path;
     std::string update_id;
     double transition_interval;
+    bool new_state;
 
     // receive new gains timestamp ("start_time" might move to "start_time")
     try {
@@ -161,8 +162,20 @@ bool applyGains::receive_update(json& json) {
         return false;
     }
 
+    // Should we register a new dataset?
+    try {
+      if (!json.at("new_state").is_boolean())
+        throw std::invalid_argument(fmt::format(fmt("applyGains: received bad gains "
+                "new_state: {:s}"),
+              json.at("new_state").dump()));
+      new_state = json.at("new_state").get<bool>();
+    } catch (std::exception& e) {
+        WARN("Failure reading 'new_state' from update: {:s}", e.what());
+        return false;
+    }
+
     // Signal to fetch thread to get gains from broker
-    update_fetch_queue.put({update_id, transition_interval, new_ts});
+    update_fetch_queue.put({update_id, transition_interval, new_ts, new_state});
 
     INFO("Received gain update with tag {:s}.", update_id);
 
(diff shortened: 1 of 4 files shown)


james-s-willis:
@jrs65 I am unsure if the dataset ID I use when `new_state` is `false` is correct? I just want to use the current one.

lib/stages/applyGains.cpp
```diff
@@ -363,7 +376,15 @@ void applyGains::fetch_thread() {
...
 
         // update gains
-        state_id_t state_id = dm.create_state<gainState>(update_id, transition_interval).first;
+        state_id_t state_id;
+        if (new_state) {
+            state_id = dm.create_state<gainState>(update_id, transition_interval).first;
+        }
+        // Use the current dataset ID
+        else {
+            const auto update = gains_fifo.get_update(double_to_ts(new_ts)).second;
```
nritsche, line 385: update shadows a variable from the next outer scope (it confused me for about one minute :smile: )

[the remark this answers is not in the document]
nritsche: Oh, I see. There's always guaranteed to be an initial one.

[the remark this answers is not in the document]
james-s-willis: That's a good point actually. How is there always guaranteed to be an initial one?

james-s-willis: Oh, nice catch! I did not realise that.

james-s-willis: Actually this has me thinking, will this line actually get the last update added to `gains_fifo`? As I am passing `new_ts` to `get_update()`.

nritsche: > Actually this has me thinking, will this line actually get the last update added to `gains_fifo`? As I am passing `new_ts` to `get_update()`.

```
/**
     * @brief Fetch the latest update for a frame with the given timestamp.
     *
     * Finds the update from the queue that should be applied to a frame
     * with the given timestamp.
     *
     * @param  timestamp  The timestamp of a frame.
     *
     * @returns  The value of the most recent update from before the given timestamp
     *           and the timestamp associated to the update. If the queue is empty, or
     *           all updates are in the future, a nullptr is returned as update.
     */
```
It will get the last update (with `new_state == true`) that came in with a timestamp before the timestamp of the new update. I guess that makes sense :shrug:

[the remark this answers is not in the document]
nritsche: That's because `configUpdater` requires the static config to have initial values for all updatable config parts. When anything subscribes to an updatable part of the config, it receives an update callback with the initial values. 
Now that I'm thinking about it, this only works because you set the initial value of `new_state` in the config to `True`. Even though a false here doesn't make any sense, the code should be robust against a `False` there.

james-s-willis:
@jrs65 I agree that the 3rd option is probably the quickest option, given we are testing this soon. I'm not sure how much work is involved with option 2 in the `cal_broker`.

james-s-willis:
@jrs65 should I just go ahead with this change?

PR #742, opened 2020-05-11
Renamed visFrameView to VisFrameView.

james-s-willis:
I'm creating this PR to break up #682.


andrerenard:
The CI failed on a python test, so I'm trying to rerun it now.

jrs65:
Gah... I appreciate I didn't get around to reviewing this with the F2F and everything going on, but why would you rename `VisFrameView` but not `visMetadata` or even the filename `visBuffer`.[hc]pp? Now it's just an awkward mix.

andrerenard:
@james-s-willis could you quickly deal with the renaming for those other functions?  it's a little annoying it'll be in two tickets, but ideally we need to do this all classes, so we'll likely have a few of those renamings anyway.   But I agree with @jrs65 it would be nice to have some self-consistency within the `vis*` components.

james-s-willis:
@andrerenard Yeah sure I can do the renaming in #743.

andrerenard:
Can we put that in it's own PR?  We might not merge #743 for this deployment, and it would be nice if the next release was self-consistent here.

james-s-willis:
Ah, yeah sure

PR #744, opened 2020-05-14
Use the "node_id" as the VDIF thread_id in the Pulsar stream

cubranic:


cubranic:
I'm not sure why VDIF thread_id was being recalculated and set for every new frame, but it seems like something that should not change while Kotekan is running, no?

cubranic:
Are all tests failing in visTranspose? My change shouldn't affect anything there. I'm not even sure what the error was, maybe this?
```
/visTranspose_test: Failed to set thread affinity for /visTranspose_test, error code 22
```

cubranic:
If error code just prints errno, then 22 would be EINVAL, I think. Or is it the return value of the process?

If errno, EINVAL really depends on the function where it happens -- "cpusetsize is smaller than the size of the affinity mask used by the kernel", but it could also be a mismatched between the mask and existing/allowed CPUs.

cubranic:
I dont' think this is the right approach, let me review the code.

Issue #748, opened 2020-05-19
Update rfi_receiver.py for dynamic frequency mapping


andrerenard:
With the changes to the dynamic frequency mapping, it looks like `rfi_receiver.py` will need to be updated.  Either by getting the mapping table from `coco` or `fpga_master` or updating the packets from `kotekan` to include the frequency bins and not just the stream ID.

james-s-willis:
Fixed in #751

PR #751, opened 2020-05-21
Add frequency bins to message sent to RFI receiver script

james-s-willis:
Solves #748.


python/scripts/rfi_receiver/rfi_receiver.py
```diff
@@ -486,7 +489,9 @@ def bad_input_listener(thread_id):
     RFIHeaderSize = app.config["chime_rfi_header_size"]
     mode = app.mode
     firstPacket = True
-    PacketSize = RFIHeaderSize + 4 * local_freq * num_elements
```
james-s-willis, line 489: @andrerenard I think this line was originally wrong, as the `fault_counter` being sent by `kotekan` in `rfiBadInputFinder.cpp` is a byte array not an integer array. So it should be:
`PacketSize = RFIHeaderSize + local_freq * num_elements`

What do you think?

andrerenard: I agree.  This looks like it's incorrect in the original code.

python/scripts/rfi_receiver/rfi_receiver.py
```diff
@@ -332,7 +327,10 @@ def data_listener(thread_id):
     mode = app.mode
     firstPacket = True
     vdifPacketSize = global_freq * 4 + RFIHeaderSize
-    chimePacketSize = RFIHeaderSize + 4 * local_freq
+    # CHIME packet: RFIHeader + frequency_bins[local_freq] + rfi_avg[local_freq]
+    chimePacketSize = RFIHeaderSize + 4 * local_freq + 4 * local_freq
```
andrerenard, line 331: Currently the size and shape of the packet both depend on both sizes having the same config.  Could we add some checks on this side to make sure the contents of the header match what this config is set to (e.g. for local_freq, etc?

james-s-willis: So there already exists a `HeaderCheck` function on line 253.

james-s-willis: Is that what you mean?

james-s-willis:
Thanks for reviewing. What's the best way to test this?

python/scripts/rfi_receiver/rfi_receiver.py
andrerenard: Ah perfect!

PR #752, opened 2020-05-21
Track the time and success of gain updates (fix #687)

### LINKED ISSUE #687: Add a metric to track the last time of gain update
- to ReadStage?
- with attribute for the kind of gain being updated (Cosmo/FRB/Pulsar)
- its value would be current Unix time (seconds since epoch), so it should be a counter

cubranic:
This PR adds two metrics for the ReadGain stage:

- `kotekan_gains_last_update_success` is a 0/1 flag to indicate the success of last update (1 means the gains files were read successfully and sent down the pipeline)
- `kotekan_gains_last_update_timestamp` records the timestamp of the last update, whether it successfully read the gains files or had to apply the default gains. ("Time" here means the current time when the update was applied, not the time it was generated by the calibration broker.)

Both metrics include a label `type` with values "frb" and "pulsar", to keep their respective updates separate.

### DIFF
--- a/lib/stages/ReadGain.cpp
+++ b/lib/stages/ReadGain.cpp
@@ -46,7 +46,11 @@ REGISTER_KOTEKAN_STAGE(ReadGain);
 
 ReadGain::ReadGain(Config& config, const std::string& unique_name,
                    bufferContainer& buffer_container) :
-    Stage(config, unique_name, buffer_container, std::bind(&ReadGain::main_thread, this)) {
+    Stage(config, unique_name, buffer_container, std::bind(&ReadGain::main_thread, this)),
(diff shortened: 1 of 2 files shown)


cubranic:
@andrerenard the PR is approved and CI passes, but the "merge" button is greyed out. Maybe because I'm not allowed to? If so, could you merge it? (Please, use the description from the PR as the body of the squashed commit.)

andrerenard:
I think it's greyed out because it's out of date with the base branch - and the new CI only checks the branch, so the rule is it also needs to be up to date with the base branch.  I will tell it to update the branch, and then as soon as the CI passes I'll merge it.

PR #753, opened 2020-05-22
Support dynamic frequency mapping

andrerenard:
* Adds the CHIME spesific stream to frequency bin mapping code.
* Updates the configs to enforce the required frequency mapping tables.


lib/utils/CHIMETelescope.hpp
```diff
@@ -11,10 +11,42 @@
 
 /**
  * @brief A telescope class to represent CHIME.
+ *
+ * @conf    require_frequency_map   Require a frequency map in the config file, or generate
+ *                                  and error an exit if none exists. Default: false
```
james-s-willis, line 16:
```suggestion
 *                                  an error and exit if none exists. Default: false
```

lib/utils/CHIMETelescope.hpp
```diff
@@ -11,10 +11,42 @@
...
+ * @conf    allow_default_map       If @c require_frequency_map is set to false then if true
```
james-s-willis, line 17:
```suggestion
 * @conf    allow_default_map       If @c require_frequency_map is set to false and allow_default_map is true
```

lib/utils/CHIMETelescope.cpp
```diff
@@ -3,16 +3,121 @@
...
+
+            if (i % 256 == 0) {
+                stream_id.unused = (stream_id.unused + 1) % 4;
+            }
+            if (i % 32 == 0) {
+                stream_id.link_id = (stream_id.link_id + 1) % 8;
+            }
+            if (i % 16 == 0) {
+                stream_id.crate_id = (stream_id.crate_id + 1) % 2;
+            }
```
jrs65, line 66: I find this logic hard to follow, with elements only incrementing conditionally, and only after the first round is set. Isn't it essentially equivalent to this:
```suggestion

            stream_id.slot_id = i % 16; // If I'm unsure about a part of this, it's this line
            stream_id.unused = (i / 256) % 4;
            stream_id.link_id = (i / 32) % 8;
            stream_id.crate_id = (i / 16) % 2;
            stream_id.unused = (i / 256) % 4;

            DEBUG2("Adding table entry: stream_id: {:d} : crate: {:d}, "
                   "slot: {:d}, link: {:d}, unused: {:d} = {:d}",
                   ice_encode_stream_id(stream_id).id, stream_id.crate_id, stream_id.slot_id,
                   stream_id.link_id, stream_id.unused, i);
            frequency_table[ice_encode_stream_id(stream_id).id] = i;
```

config/chime_science_run_recv.yaml
```diff
@@ -28,6 +28,10 @@ num_ev: 4
 # other stages (11 and 14, twinned with 3 and 6 respectively)
 cpu_affinity: [7, 11, 14, 15]
 
+telescope:
+  name: chime
+  allow_default_map: false
```
jrs65, line 33: The parameter name is a bit ambiguous. `allow_default_frequency_map` would be better, at the expense of being absurdly long. Your call!

lib/utils/CHIMETelescope.cpp
andrerenard: I agree, this is much cleaner.  I've tested it, and it does produce the right values for all bin numbers.

config/chime_science_run_recv.yaml
andrerenard: Changed.

PR #755, opened 2020-05-22
Make the thread_id a "node_id" in the 4 packet case in PulsarPostProcess

andrerenard:


lib/stages/pulsarPostProcess.cpp
```diff
@@ -249,7 +258,7 @@ void pulsarPostProcess::main_thread() {
 
                 // Fill the headers of the new buffer
                 fill_headers((unsigned char*)out_frame, &vdif_header, fpga_seq_num, &time_now,
-                             psr_coord, (uint16_t*)freq_ids);
+                             psr_coord, (uint16_t*)thread_ids);
```
cubranic, line 261: Why is this case necessary? it's a `uint16_t[]`, which I thought can be passed as a pointer?

andrerenard: I agree, not necessary.  I removed those casts.

PR #760, opened 2020-05-28
feat(visAccumulate): use initial FPGA dataset_id as base dataset

jrs65:


lib/stages/visAccumulate.hpp
```diff
@@ -157,6 +162,7 @@ class visAccumulate : public kotekan::Stage {
     size_t num_gpu_frames;
     size_t minimum_samples;
     float max_age;
+    dset_id_t fpga_dataset;
```
nritsche, line 165: Config variables should start with an underscore (http://lwlab.dunlap.utoronto.ca/kotekan/dev/dev_style_guide_code.html#variables).

lib/stages/visAccumulate.cpp
```diff
@@ -63,6 +63,7 @@ visAccumulate::visAccumulate(Config& config, const std::string& unique_name,
     block_size = config.get<size_t>(unique_name, "block_size");
     samples_per_data_set = config.get<size_t>(unique_name, "samples_per_data_set");
     max_age = config.get_default<float>(unique_name, "max_age", 60.0);
+    fpga_dataset = config.get_default<dset_id_t>(unique_name, "fpga_dataset", dset_id_t::null);
```
nritsche, line 66: The dataset ID is in the config under `fpga_dataset/id`, so I think this should be

```suggestion
    fpga_dataset = config.get_default<dset_id_t>("fpga_dataset", "id", dset_id_t::null);
```

To test if this line works, we could add
```
"fpga_dataset": {
      "id": "d3c527df45f258dbd88c6ea7f56eadba"
}
```
to `accumulate_params` in `tests/test_accumulate.py`

nritsche: This is from `coco status`:
```
cluster:
...
  fpga-dataset:
    id: d3c527df45f258dbd88c6ea7f56eadba
```

lib/stages/visAccumulate.hpp
jrs65: Indeed. But none of the existing ones do, so I'm going for consistency rather than adherence to the guidelines in here.

jrs65:
> LGTM! Thanks for adding the test

Anything to make you happy!

PR #763, opened 2020-05-29
Fixes rfi_receiver.py to lower rate of debug statements

james-s-willis:
The packet rate is calculated as follows:
- Input into `rfiBroadcast.cpp` comes directly from the GPU
- Frame size 49152 / 256 = 192, 2.56us per sample
- Each sample in the frame is summed over and sent to `rfi_receiver.py` script via UDP packets
- This gives a packet every ~0.126s per node

We set the frequency bins to be printed every 150 packets to get a cadence of ~19s


python/scripts/rfi_receiver/rfi_receiver.py
```diff
@@ -359,27 +360,35 @@ def data_listener(thread_id):
...
 
+        # A packet is received roughly every ~0.126s
         if packet != "":
 
+            # Print every ~6s
             if packetCounter % (50 * len(stream_dict) + 1) == 0:
                 logger.info(
                     "Thread id %d: Receiving Packets from %d Streams"
                     % (thread_id, len(stream_dict))
                 )
```
andrerenard, line 371: I'm not sure this message adds any value, the second message indirectly gives the number of streams by way of the list - or you could just include the number of streams in the second message.

andrerenard:
When you address the comments, also update this branch with develop so it can be merged.

python/scripts/rfi_receiver/rfi_receiver.py
```diff
@@ -359,27 +360,28 @@ def data_listener(thread_id):
...
         if packet != "":
 
-            if packetCounter % (50 * len(stream_dict) + 1) == 0:
-                logger.info(
-                    "Thread id %d: Receiving Packets from %d Streams"
-                    % (thread_id, len(stream_dict))
+            # Print frequency bins received every ~19s
+            if packetCounter % (150 * len(stream_dict) + 1) == 0:
+                logger.debug(
+                        "data_listener: Thread id %d, %d Streams: Receiving frequency bins: %s"
```
andrerenard, line 369:
```suggestion
                        "data_listener: Thread id: %d, Streams: %d, Receiving frequency bins: %s"
```

andrerenard: Also you might need to run black after this...

PR #766, opened 2020-06-02
Export no. of samples and flagged samples as prometheus metrics in rfiBroadcast

james-s-willis:


lib/stages/rfiBroadcast.cpp
```diff
@@ -228,6 +233,12 @@ void rfiBroadcast::main_thread() {
             mask_percent_metric.labels({std::to_string(current_freq_bin)})
                 .set(perc_zeroed.average());
 
+            // Increment the prometheus metrics for the total number of samples and the total number
+            // of flagged samples
+            sample_counter.labels({std::to_string(current_freq_bin)})
+                .inc((double)_samples_per_data_set / _sk_step);
```
jrs65, line 239: The double cast seems out of place here.

lib/stages/rfiBroadcast.cpp
```diff
@@ -46,7 +46,12 @@ REGISTER_KOTEKAN_STAGE(rfiBroadcast);
 
 rfiBroadcast::rfiBroadcast(Config& config, const std::string& unique_name,
                            bufferContainer& buffer_container) :
-    Stage(config, unique_name, buffer_container, std::bind(&rfiBroadcast::main_thread, this)) {
+    Stage(config, unique_name, buffer_container, std::bind(&rfiBroadcast::main_thread, this)),
+    sample_counter(Metrics::instance().add_counter("kotekan_rfibroadcast_sample_total", unique_name,
+                                                   {"freq_id"})),
+    flagged_sample_counter(Metrics::instance().add_counter(
+        "kotekan_rfibroadcast_flagged_sample_total", unique_name, {"freq_id"})) {
```
jrs65, line 53: To align this with `RfiFrameDrop` (i.e. the second stage) you might call this `kotekan_rfibroadcast_dropped_sample_total`.

lib/stages/rfiBroadcast.cpp
james-s-willis: Do you not need it for rounding errors?

jrs65: I think `_samples_per_data_set` is always and exactly multiple of `_sk_step` otherwise none of the RFI kernels would work. This just casts it do a double, does a division and then will implicitly cast it back to an int.

Also this quantity never changes so I would define it as a `const uint64_t` at the top of the main thread.

jrs65: At the very least I would pull it to the top and make the cast back to an int deliberate. In the case they don't exactly divide, it's not case what this value should actually be, so at least if you do that you can make it clear.

PR #778, opened 2020-06-08
Revert a mistaken rename of a baseband dataset attribute

cubranic:


jrs65:
@andrerenard do we want to merge this before the tagging?

andrerenard:
Yes, I think we should merge it.

mondana:
Can we envision a test suite that could have caught this? Or at least write down a validation test that can be run during the last stages of the deployment week?

PR #779, opened 2020-06-09
Fix logic to enable/disable RFI zeroing.

james-s-willis:
Also cleans up the code to make it easier to follow.

### DIFF
--- a/python/scripts/rfi_receiver/rfi_receiver.py
+++ b/python/scripts/rfi_receiver/rfi_receiver.py
@@ -830,7 +830,7 @@ def toggle_rfi_zeroing(url, headers, zeroing_on):
                 return True
     except:
         logger.info(
-            "RFI Solar Transit Toggle: Failure to contact Comet, is it running?"
+            "RFI Solar Transit Toggle: Failure to contact coco, is it running?"
         )
     return False
 
--- a/python/scripts/rfi_receiver/rfi_receiver.py
+++ b/python/scripts/rfi_receiver/rfi_receiver.py
@@ -808,26 +808,16 @@ def toggle_rfi_zeroing(url, headers, zeroing_on):
     payload = {"rfi_zeroing": zeroing_on}
     try:
         r = requests.post(url, data=json.dumps(payload), headers=headers)
-        if zeroing_on:
-            if not r.ok:
-                logger.info(
-                    "RFI Solar Transit Toggle: Failed to turn RFI zeroing on. Something went wrong in the request."
-                )
-            else:
-                logger.info(
-                    "RFI Solar Transit Toggle: Successfully turned RFI zeroing on."
-                )
-                return True
+        state = "on" if zeroing_on else "off"
+        if not r.ok:
+            logger.error(
+                f"RFI Solar Transit Toggle: Failed to turn RFI zeroing {state}. Something went wrong in the request."
+            )
         else:
-            if not r.ok:
-                logger.info(
-                    "RFI Solar Transit Toggle: Failed to turn RFI zeroing off. Something went wrong in the request."
-                )
-            else:
-                logger.info(
-                    "RFI Solar Transit Toggle: Successfully turned RFI zeroing off."
-                )
-                return True
+            logger.info(
+                f"RFI Solar Transit Toggle: Successfully turned RFI zeroing {state}."
+            )
+            return True
     except:
         logger.info(
             "RFI Solar Transit Toggle: Failure to contact coco, is it running?"
--- a/python/scripts/rfi_receiver/rfi_receiver.py
+++ b/python/scripts/rfi_receiver/rfi_receiver.py
@@ -896,6 +896,7 @@ def rfi_zeroing():
             )
 
             time.sleep(5)
+            continue
 
         # Sleep until end of transit window or until the next one occurs
         time.sleep(downtime_s)
--- a/python/scripts/rfi_receiver/rfi_receiver.py
+++ b/python/scripts/rfi_receiver/rfi_receiver.py
@@ -850,7 +850,7 @@ def rfi_zeroing():
 
         # Get the *nearest* transit which we need to determine if we are still in the window
         time_to_nearest_transit = (
-            ephemeris.solar_transit(time_now - 12 * 3600) - time_now + 12 * 3600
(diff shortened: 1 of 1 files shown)


python/scripts/rfi_receiver/rfi_receiver.py
```diff
@@ -822,18 +855,21 @@ def rfi_zeroing():
...
+        t_diff = (
+            datetime.datetime.utcfromtimestamp(t_transit) - datetime.datetime.utcnow()
+        )
 
         time_to_transit_s = abs(t_diff.total_seconds())
 
         # Check if we are in the transit window, if so set downtime accordingly
         if time_to_transit_s < half_downtime_s:
+            toggle_rfi_zeroing(url, headers, False)
             downtime_s = time_to_transit_s + half_downtime_s
```
jrs65, line 870: This is wrong I think.

Consider what happens if you enter this before the transit window has ended, but after the transit itself.

In that case t_diff is negative, and you should only sleep for what's remaining in the window (i.e. `half_downtime_s + t_diff.total_seconds()`). But because of the abs in the definition of time_to_transit_s it will sleep too long.
That kind of mistake could very much give the asymmetry in the window.

jrs65, line 870: I think better logic in here is something along the lines of this:
```python
time_now = ephemeris.ensure_unix(datetime.utcnow())

# Get the *next* transit in the future
time_to_next_transit = ephemeris.solar_transit(time_now) - time_now

# Get the *nearest* transit which we need to determine if we are still in the window
time_to_nearest_transit = ephemeris.solar_transit(time_now - 12 * 3600) - time_now + 12 * 3600

# Check if we are within the current transit window and wait until the end of it
if abs(time_to_nearest_transit) < half_window_s:
    set_rfi_zeroing(False)
    downtime_s = half_window + time_to_nearest_transit
# Otherwise, we wait until the start of the next transit window
else:
    set_rfi_zeroing(True)
    downtime_s = time_to_next_transit - half_window_s
```

Also note that call the window width `half_downtime_s` is confusing because `downtime_s` used elsewhere seems to have a different meaning, i.e. the first is the length of half of the window, but the latter is the time time to wait until the next transition. The name implies the former is just half the latter, which it isn't.

python/scripts/rfi_receiver/rfi_receiver.py
```diff
@@ -802,6 +802,39 @@ def http_server2():
     httpd.serve_forever()
 
 
+# Sends message to coco to turn RFI zeroing on/off
+def toggle_rfi_zeroing(url, headers, zeroing_on):
```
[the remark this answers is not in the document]
jrs65: Also, do you need to supply the url and headers to this? Can't you set them globally?

james-s-willis:
@andrerenard @jrs65 I think I have addressed all of your comments now

python/scripts/rfi_receiver/rfi_receiver.py
```diff
@@ -810,76 +836,63 @@ def rfi_zeroing():
...
+        # Get the *nearest* transit which we need to determine if we are still in the window
+        time_to_nearest_transit = (
+            ephemeris.solar_transit(time_now - 12 * 3600) - time_now
+        )
 
-            # Time until half_downtime_s before next solar transit in seconds
-            sleep_time_s = abs(time_to_transit_s - half_downtime_s)
+        # Check if we are within the current transit window and wait until the end of it
+        if abs(time_to_nearest_transit) < half_window_s:
+            set_rfi_zeroing(False)
```
andrerenard, line 858: This will happen again right after this `if` statement.

python/scripts/rfi_receiver/rfi_receiver.py
```diff
@@ -810,76 +836,63 @@ def rfi_zeroing():
...
+            downtime_s = half_window_s + time_to_nearest_transit
+        # Otherwise, we wait until the start of the next transit window
+        else:
+            set_rfi_zeroing(True)
```
andrerenard, line 862: If we check if the query worked below, should we not check here as well? - and consider what should happen if we land in that case.

python/scripts/rfi_receiver/rfi_receiver.py
```diff
@@ -810,76 +836,63 @@ def rfi_zeroing():
...
+            downtime_s = time_to_next_transit - half_window_s
```
jrs65, line 863: Github is being annoying so I can't actually make a suggestion block big enough, but I agree with Andre that this end of thing method is still a bit funky. You should only need to set the state once. Something like this probably works:


```suggestion
       log.debug("Next solar transit ...")
       log.debug("Nearest solar transit ...")

        # Check if we are within the current transit window and wait until the end of it
        if abs(time_to_nearest_transit) < half_window_s:
            new_zeroing_state = False
            downtime_s = half_window_s + time_to_nearest_transit
            log.info("Within transit window, disabling zeroing and waiting until the end") 
        # Otherwise, we wait until the start of the next transit window
        else:
            new_zeroing_state = True
            downtime_s = time_to_next_transit - half_window_s
            log.info("Outside transit window, enabling zeroing and waiting until next window") 

        success = set_rfi_zeroing(new_zeroing_state)
        
        if not success:
            # Figure out what to do in here, probably sleep for a few seconds,
            # and then just continue, which will basically force it to try again
            pass
            
        time.sleep(downtime_s)
```
And then remove everything else below in the function

james-s-willis:
@andrerenard @jrs65 I've updated the script now.

python/scripts/rfi_receiver/rfi_receiver.py
```diff
@@ -810,109 +836,69 @@ def rfi_zeroing():
...
+        if not success:
 
             logger.info(
-                "RFI Solar Transit Toggle: Solar transit ended. Waking up to turn RFI zeroing back on."
+                "RFI Solar Transit Toggle: Failed to set new RFI zeroing state. Will wait for a few seconds and try again."
             )
 
-            # Payload
-            payload = {"rfi_zeroing": True}
+            time.sleep(5)
```
andrerenard, line 898:
```suggestion
            time.sleep(5)
            continue
```
At the moment this doesn't do anything, since you'll get into the next sleep and never set the state.  I think either rerunning the loop, or exiting (and letting systemd restart it) should do the trick though.

PR #786, opened 2020-07-15
Create raw HFB files from frame view

james-s-willis:


andrerenard:
@james-s-willis can we close this one?

PR #787, opened 2020-07-15
Create HFB frameview from abstract FrameView (new)

james-s-willis:


lib/metadata/hfbMetadata.hpp
```diff
@@ -2,14 +2,11 @@
...
+#include "dataset.hpp" // for dset_id_t
 #include "metadata.h"
 
 #include <sys/time.h>
 
-#ifdef __cplusplus
-extern "C" {
-#endif
-
 #pragma pack()
```
jrs65, line 10: Why change the packing? Is there a good reason to not just use the standard C packing. And if you're going to change it from the default, I think it's at least good to be explicit. I think `#pragma pack()` with not argument sets 8 byte alignment.

If you're worried about wasted space in the struct, you can get rid of it much more transparently by reordering the entries.

lib/metadata/hfbMetadata.hpp
```diff
@@ -29,6 +26,12 @@ struct hfbMetadata {
...
     uint32_t compressed_data_size;
+    /// The number of beams in the data.
+    uint32_t num_beams;
+    /// The number of sub-frequencies in the data.
+    uint32_t num_subfreq;
+    /// ID of the dataset
+    dset_id_t dataset_id;
 };
 
 // Helper functions to save lots of pointer work
```
jrs65, line 37: Do you still need all these C style getters and setters? I think with the FrameView you probably don't need them and could just remove them.

lib/metadata/hfbMetadata.hpp
```diff
@@ -2,14 +2,11 @@
...
 
 #include <sys/time.h>
 
-#ifdef __cplusplus
-extern "C" {
-#endif
-
 #pragma pack()
 
 struct hfbMetadata {
```
jrs65, line 12: Probably should uppercase this for consistency with everything else, e.g. `HFBMetadata`

lib/utils/HfbFrameView.hpp
```diff
@@ -0,0 +1,230 @@
...
+ *with the ability to interact with the data and metadata. Structural parameters can only be set at
+ * creation, everything else is returned as a reference or pointer so can be
+ * modified at will.
+ *
+ * @todo This may want changing to use reference wrappers instead of bare
+ *       references.
+ *
+ * @author James Willis
+ **/
+class HfbFrameView : public FrameView {
```
jrs65, line 48: As HFB is an acronym, should this be `HFBFrameView`?

lib/utils/HfbFrameView.hpp
```diff
@@ -0,0 +1,230 @@
...
+#include <tuple>    // for tuple
+#include <utility>  // for pair
+
+
+/**
+ * @brief The fields within the hfbBuffer.
+ *
+ * Use this enum to refer to the fields.
+ **/
+enum class hfbField { hfb };
```
jrs65, line 32: Whatever you choose for the comment below, this should probably be consistent?

lib/utils/HfbFrameView.hpp
```diff
@@ -0,0 +1,230 @@
...
+ * @brief Provide a structured view of a hyperfine beam buffer.
+ *
+ * This class inherits from the FrameView base class and sets up a view on a hyperfine beam buffer
+ *with the ability to interact with the data and metadata. Structural parameters can only be set at
+ * creation, everything else is returned as a reference or pointer so can be
+ * modified at will.
+ *
+ * @todo This may want changing to use reference wrappers instead of bare
+ *       references.
+ *
```
jrs65, line 45: Is this still a valid todo? I don't see what the point of the change is, so it seems like it might be a legacy thing.

lib/utils/HfbFrameView.hpp
```diff
@@ -0,0 +1,230 @@
...
+    uint32_t& num_samples_integrated;
+    /// Number of samples expected
+    uint32_t& num_samples_expected;
+    /// A reference to the frequency ID.
+    uint32_t& freq_id;
+    /// A reference to the dataset ID.
+    dset_id_t& dataset_id;
+
+    /// View of the hyperfine beam data.
+    const gsl::span<float> hfb;
```
jrs65, line 227: Is there/should there be a weight dataset for the HFB data? I'm not sure quite how I didn't think about this before, but it seems we could and maybe should, calculate the fast timescale variance of the data (similarly to what's done for the visibility system).

It's not clear that should go in within this commit, but you might want to add space for it into the frame format, and into a field for the frame view, even if it's just initialised to zero for now.

lib/metadata/hfbMetadata.hpp
james-s-willis: I had a problem compiling with the `extern "C"`. I've just tried it again and it compiles okay now, so I'll revert this change.

lib/metadata/hfbMetadata.hpp
james-s-willis: So I use these in `integrateHFBData.cpp` instead of directly accessing the metadata:

https://github.com/kotekan/kotekan/blob/ca1a367a7e8fcd55aac1117df7bd62eafdc5ec77/lib/stages/integrateHFBData.cpp#L201-L224

lib/metadata/hfbMetadata.hpp
james-s-willis: I am now getting this error in the MacOS build:
```
../../../lib/metadata/HFBMetadata.h:93:18: error: 'get_dataset_id' has C-linkage specified, but returns user-defined type 'dset_id_t' (aka 'Hash') which is incompatible with C [-Werror,-Wreturn-type-c-linkage]
inline dset_id_t get_dataset_id(struct Buffer* buf, int ID) {
                 ^
1 error generated.
```

have you seen this before? I think this is why I changed the file to C++ in the first place.

jrs65: Yeah, I think this definitely needs to be C++. However, I don't know why you need to declare is as `extern "C"`. Do you call it from any C code?

My initial comment though, is why you change the struct packing and alignment by using the `#pragma pack()` line? That wasn't a change you made this round, but I don't see why you need to do it at all, and I'd suggest removing it entirely if there isn't a good reason.

james-s-willis: Ah okay, I will make it `.hpp` file again. The `extern "C"` and `#pragma pack()` just came from copying the `chimeMetadata.h` so I will remove them.

jrs65: I guess that turns into a question for @andrerenard. What's the point of changing the default struct packing in chimeMetadata? Are you trying to align things in memory more efficiently by using the (implicit) 8 byte alignment of members.

andrerenard: The main reason for this was trying to force the compiler's hand in the packing scheme to avoid possibly different packings on different systems.  However in practice this might not isn't really an issue for our use cases, and in theory not forcing packing can be faster.

jrs65: Okay. That makes some sense, although I can't imagine trying to run kotekan on any platform which didn't have the standard System V x64 ABI, so I don't think it's really helping. I guess it could be removed in chimeMetadata, but then we'd be unable to load any historic raw dumps we had lying around.

I think at least for the HFB case we should definitely remove this one though.

james-s-willis:
It finally passes. Could you accept this, @andrerenard ?

PR #792, opened 2020-07-24
feat(datasetManager): simplify dataset update

nritsche:
Also see https://github.com/chime-experiment/comet/pull/86

The endpoint /update-datasets does not require nor return any list of
known roots nor timestamps anymore. For this, the update_datasets()
method was simplified here.

It also doesn't return anything but the datasets between the given one
and its root dataset anymore. It doesn't base the set of returned
datasets on any timestamp, but always returns the whole branch all the
way to the root. This does not require any changes in the datsetManager.


nritsche:
> Also looks great. I wonder if we should also lower the timeout on the operations. 200s should be far too long now, and I would expect the request time to stay pretty constant into the future (rather than the current situation where it grows linearly with the amount of time the system has been running).

Good point. I agree, but that's only in `master` (https://github.com/kotekan/kotekan/commit/446de776cd3e8ec2bcd8c753d4f478086a650daf).
@andrerenard can you merge master into develop? 
Then I can revert the above commit as part of this PR.

andrerenard:
I take it this should be closed in favour of #793 ?

PR #793, opened 2020-07-28
Hotfix: feat(datasetManager): simplify dataset update

nritsche:
Replaces #792 and is based on master to be able to revert 446de77.

Also see chime-experiment/comet#84

The endpoint /update-datasets does not require nor return any list of
known roots nor timestamps anymore. For this, the update_datasets()
method was simplified here.

It also doesn't return anything but the datasets between the given one
and its root dataset anymore. It doesn't base the set of returned
datasets on any timestamp, but always returns the whole branch all the
way to the root. This does not require any changes in the datsetManager.


nritsche:
@andrerenard Can we merge this into master without the CI passing or do you have another idea?

PR #797, opened 2020-07-30
Switch to fp16 for intermediate static beam products

andrerenard:
@kvand found out that using an FP16 array instead of FP32 between the FFT Beamformer, transpose, and upchannelization results in a large speed improvement, and lowers the overall utilization on the CHIME nodes by about 2%.   

In theory this slightly increases the quantization noise, but the impact should be very minimal. 

### Typical kernel timing before change: 
```
| -------- GPU[0] Kernel timing --------
| Kernel name         |     time | utilization   |
|---------------------+----------+---------------|
| reorder             | 0.000476 | 0.3783%       |
| zero_padded_FFT512  | 0.010676 | 8.4844%       |
| transpose           | 0.004672 | 3.7127%       |
| upchannelize        | 0.005487 | 4.3609%       |
| sum_hfb             | 0.000223 | 0.1771%       |
| pulsarbf_float      | 0.005896 | 4.6855%       |
| rfi_chime_time_sum  | 0.000865 | 0.6873%       |
| rfi_bad_input       | 0.000159 | 0.1263%       |
| rfi_chime_input_sum | 3.3e-05  | 0.0261%       |
| rfi_chime_zero      | 0.000368 | 0.2928%       |
| CHIME_presum        | 0.000124 | 0.0983%       |
| CHIME_presum        | 0.000129 | 0.1027%       |
| CHIME_presum        | 0.00013  | 0.1034%       |
| CHIME_presum        | 0.000124 | 0.0988%       |
| CHIME_X             | 0.023332 | 18.5425%      |
| CHIME_X             | 0.023341 | 18.5497%      |
| CHIME_X             | 0.023366 | 18.5696%      |
| CHIME_X             | 0.023385 | 18.5848%      |
| Total:              | 0.122786 | 97.5812%      |
```
### Typical kernel timing after change: 
```
| -------- GPU[0] Kernel timing --------
| Kernel name         |     time | utilization   |
|---------------------+----------+---------------|
| reorder             | 0.000476 | 0.3780%       |
| zero_padded_FFT512  | 0.010008 | 7.9537%       |
| transpose           | 0.002517 | 2.0004%       |
| upchannelize        | 0.005417 | 4.3052%       |
| sum_hfb             | 0.000227 | 0.1807%       |
| pulsarbf_float      | 0.005907 | 4.6946%       |
| rfi_chime_time_sum  | 0.000857 | 0.6809%       |
| rfi_bad_input       | 0.000162 | 0.1291%       |
| rfi_chime_input_sum | 3.4e-05  | 0.0267%       |
| rfi_chime_zero      | 0.000303 | 0.2407%       |
| CHIME_presum        | 0.000127 | 0.1006%       |
| CHIME_presum        | 0.000125 | 0.0993%       |
| CHIME_presum        | 0.000108 | 0.0858%       |
| CHIME_presum        | 0.000106 | 0.0844%       |
| CHIME_X             | 0.023367 | 18.5706%      |
| CHIME_X             | 0.023343 | 18.5516%      |
| CHIME_X             | 0.023347 | 18.5547%      |
| CHIME_X             | 0.023389 | 18.5878%      |
| Total:              | 0.11982  | 95.2247%      |
```

### DIFF
--- a/lib/hsa/hsaBeamformKernel.cpp
+++ b/lib/hsa/hsaBeamformKernel.cpp
@@ -63,7 +63,8 @@ hsaBeamformKernel::hsaBeamformKernel(Config& config, const std::string& unique_n
     }
 
     input_frame_len = _num_elements * _num_local_freq * _samples_per_data_set;
-    output_frame_len = _num_elements * _samples_per_data_set * 2 * sizeof(float);
+    // Kernel uses fp16 complex numbers for output which are the same size as sizeof(float)
+    output_frame_len = _num_elements * _samples_per_data_set * sizeof(float);
 
 
     map_len = 256 * sizeof(int);
--- a/lib/hsa/hsaBeamformTranspose.cpp
+++ b/lib/hsa/hsaBeamformTranspose.cpp
@@ -25,8 +25,9 @@ hsaBeamformTranspose::hsaBeamformTranspose(Config& config, const std::string& un
     _num_elements = config.get<int32_t>(unique_name, "num_elements");
     _samples_per_data_set = config.get<int32_t>(unique_name, "samples_per_data_set");
 
-    beamform_frame_len = _num_elements * _samples_per_data_set * 2 * sizeof(float);
-    output_frame_len = _num_elements * (_samples_per_data_set + 32) * 2 * sizeof(float);
+    // Kernel uses fp16 complex numbers which are the same size as sizeof(float)
+    beamform_frame_len = _num_elements * _samples_per_data_set * sizeof(float);
+    output_frame_len = _num_elements * (_samples_per_data_set + 32) * sizeof(float);
 }
 
 hsaBeamformTranspose::~hsaBeamformTranspose() {}
--- a/lib/hsa/hsaBeamformUpchanHFB.cpp
+++ b/lib/hsa/hsaBeamformUpchanHFB.cpp
@@ -28,7 +28,9 @@ hsaBeamformUpchanHFB::hsaBeamformUpchanHFB(Config& config, const std::string& un
     _num_frb_total_beams = config.get<int32_t>(unique_name, "num_frb_total_beams");
     _factor_upchan = config.get<int32_t>(unique_name, "factor_upchan");
 
-    input_frame_len = _num_elements * (_samples_per_data_set + 32) * 2 * sizeof(float);
+    // Kernel uses fp16 complex numbers for input which are the same size as sizeof(float)
+    input_frame_len = _num_elements * (_samples_per_data_set + 32) * sizeof(float);
+
     output_frame_len = _num_frb_total_beams
                        * (_samples_per_data_set / _downsample_time / _downsample_freq)
                        * sizeof(float);
--- a/lib/hsa/kernels/transpose.cl
+++ b/lib/hsa/kernels/transpose.cl
@@ -2,12 +2,12 @@
 #define BLOCK_ROWS 8
 
 
-__kernel void transpose(__global float2 *input, __global float2 *output) {
+__kernel void transpose(__global float *input, __global float *output) {
   
   uint width1 = get_global_size(0);
   uint width2 = get_global_size(1)*4;
 
-  __local float2 tile[TILE_DIM][TILE_DIM];  //32x32
+  __local float tile[TILE_DIM][TILE_DIM];  //32x32
   uint x  = get_group_id(0) * TILE_DIM + get_local_id(0);
   uint y =  get_group_id(1) * TILE_DIM + get_local_id(1);
 
--- a/lib/hsa/kernels/unpack_shift_beamform_flip.cl
+++ b/lib/hsa/kernels/unpack_shift_beamform_flip.cl
@@ -13,7 +13,7 @@
 
 #define CUSTOM_BIT_REVERSE_9_BITS(index) ((( ( (((index) * 0x0802) & 0x22110) | (((index) * 0x8020)&0x88440) ) * 0x10101 ) >> 15) & 0x1FE)
 
-__kernel void zero_padded_FFT512( __global uint *data, __global uint *mapped,  __global float2 *Co, __global float2 *results_array,  __global float2 *Gain){
+__kernel void zero_padded_FFT512( __global uint *data, __global uint *mapped,  __global float2 *Co, __global float *results_array,  __global float2 *Gain){
 
   __local float2 local_data[2048];//4* 512 float2 * 2 float/float2 * 4 B/float = 16kB
     uint local_address = get_local_id(0);  //0 to 255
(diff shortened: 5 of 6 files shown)


kiyo-masui:
Okay, I am actually worried about the exponent overflowing:

We start with 4 bit data, but that is irrelevant, because we calibrate it and put it in to units of sqrt(Jy). I don't know how bright the sun is in our data... lets say 1 MJy, so the baseband data is now ~10^3.  The beamform increases this by a factor of 1024, so the data is now at the 10^6 level = 2**20.  Half-precision has 5 bits for the exponent meaning it can represent 2**-15 to 2**15... seems like we will overflow here.

Way forward:
1. Check the max brightness of the sun in calibrated cosmology data.
2. Check the gain files. By the above reasoning, they should have abs values in the 2**6 range (because they map 2**4 (uncalibrated) to 2**10 (calibrated))
3. Take out a factor of 1024 after beamforming before the transpose, and replace after the transpose before the up channelization.

I have no concerns about the precision loss due to truncating the significant.

kiyo-masui:
Also, the gains are scaled by a configurable parameter from the config file whose value is currently 0.05. That helps, but I'm not sure we want to rely on that factor too much.

kvand:
Right, thanks for catching that, Kiyo -- I hadn't considered about the overflow issue, and I think it's *both* demonstrably a problem, and likely ok:
- The 0.05 factor in the beamform is actually there exactly because of this reason... Originally it was added to keep values within the 8b range used to transmit to FRB. After the packet dynamic scaling was introduced, we tried to get rid of it and were surprised when L1 blew up. Bonsai uses (used?) f16 data data internally, and without that scaling, it overflowed and broke. We put the 0.05 back in to keep L1 happy, even though L0 and the transmit no longer needed it. So your overflow calculation empirically checks out!
- After that, there was discussion of upping L1 to f32 to avoid these issues, but that relies on getting Kendrick's attention, and I don't think it ever happened. We should confirm whether or not that happened; if L1 is still bound by f16, I think that means we don't worry much at L0 about overflowing?

kiyo-masui:
Ah, I see. However, note that L1 doesn't see quite the same numbers as the transpose, because of the up channelize. However, the up channelize only increases the levels (by a factor of 128 + a few miscellaneous factors), so I think we are good here.

Is there a place we can document this particular fragility of the system, so we don't screw ourselves at a later date?

andrerenard:
We could add a check to enforce a `frb_scaling` value of less than `0.05`, we could also check the gains to enforce that after scaling they don't exceed a maximum amplitude.  I generally like the idea of having the system crash with an error if you do something wrong.  (Obviously we should also comment in the code why this is.  Aside from that I don't actually know were this information would/should live.)

kiyo-masui:
This is instrument and calibration dependent. I think the check needs to be on the value of the gains * the scaling.

andrerenard:
What is that maximum value (`|gains * scaling| < X`)?  I can likely work it out, but if someone has already thought about all the factors it might be quicker.

kiyo-masui:
|Raw data| < 2^3 
|cal data| < |g * s| * 2^3
|beamformed| < |g * s| * 1024 * 2^3 < 2^15
|g *s| < 2^2

kiyo-masui:
I don't think that is correct, but I can't spot the mistake either...

andrerenard:
I think that's failing to account for the fact these are complex values which adds another bit to the maximum, but it still works...   If I take the maximum gain and data `(3+3i) * (7+7i) = 42`, and then `42*1024 = 43008 = (1010100000000000)base2 > 2^15` - assuming all the power is in one beam (which granted is almost impossible...).  However the limit of fp16 isn't exactly 2^15, since you get to encode extra information in the significand, so the maximum is actually +/-65504.   Given that we can actually allow for `|g*s| <= 2^2`, e.g. maximum gain of `(+/-4 +/-4i)`.  @kvand do want to double check the logic here?

kvand:
Those numbers look right to me... with g*s = 4±4i, the max possible value is 57344 in either component, so fits in an IEEE half.

I guess the next step is to check the gains and confirm they're magnitude < 3/0.05 < 60? (Assuming we don't want to fiddle with the scale factor?) Probably worth looking over the history of those values, are they stored somewhere obvious, or accessible via grafana?

I agree this is a potential pitfall down the road if the analog system response changes pathologically... The digital levels out of the F-engine are set by the background Tsys, the gains map those to sky fluxes -- for larger gains, we need a smaller number of bits exercised for a given source on sky, which I think means *higher* Tsys? Yes, I think that maps, lower F-engine "scaler" gains come from higher Tsys and require larger downstream gain numbers to get back to sky fluxes. Assuming the telescope beam directivity doesn't change substantially (it shouldn't that's a very analog E&M setup), we're basically just worried about the Tsys changing by a factor of many? Kind of seems like if that happens, railing during a solar flare is the least of our worries 😛

kvand:
Yeah, that element-wise float2 divide-by-4 confused me, too -- in the rewrite it's a whole lot more explicitly applied to each of re,im. That scaling op is a negligible cost, I agree with Kiyo that keeping the logic clear outweighs the 0.01% speedup we might get by removing it.

So it sounds like we're converging to "this is fine if bump the div-by-4 to div-by-something-bigger"?

andrerenard:
I think we are actually fine without increasing it, and just checking we don't exceed a threshold in the gains.  But if we just want to be sure this can never be a problem under any gains, increasing it would be the way to go.  Although if we do go that path we still need to decide if we want to (partially) add it back later, or require the FRB system to adjust for the different scaling in their calibration pipeline. 

I would still argue that it makes the logic cleaner if we have just one scaling (right now we have two scalar divides `/4` and the `0.05` gain scaling).  It's easy when reading the code to then miss one of them, as we did even now trying to figure out the magnitude of these values.

cherryng:
I agree that having one scaling instead of two is cleaner. Given that the current dynamic range seems fine, we could considering removing the /4, and changing the scaling factor to 0.0125. This will hopefully mean no change (or minimal) for downstream FRB processes. If one day we think the gains are too large, we can always edit the scaling factor in the config file. That is what it is there for; we won't need to submit code changes which is more disruptive.

kiyo-masui:
1. If we take out any extra factors, please replace them after the transpose. I can think of 3 or 4 different places where we would have to be careful if we messed with the absolute flux calibration.

2. I really don't think we should redistribute factors between the config and the code. You shouldn't think of the /4 as a scaling... it is just another beam-forming normalization choice. There are several of these choices in the code: we do divide by 4 when we beam form EW, but not by 256 when we beam form NS. We do divide by 3 when we down sample in time, 16 when we down sample in frequency, and 2 in polarization, but we do not divide by sqrt(128) when we upchannelize. The PFB bandpass correction is normalized to 0.8067 instead of 1.  We aren't going to fix all of these, so why not maintain continuity.

andrerenard:
This is superseded by #816

PR #809, opened 2020-08-12
Create raw HFB files from frame view (new)

james-s-willis:


lib/stages/Writer.cpp
```diff
@@ -114,76 +114,153 @@ void visWriter::main_thread() {
...
-        // Check the dataset ID hasn't changed
-        if (acqs.count(frame.dataset_id) == 0) {
-            init_acq(frame.dataset_id);
+        if (!strcmp(in_buf->buffer_type, "vis")) {
+            auto frame = VisFrameView(in_buf, frame_id);
+            write_vis_data(frame, write_time_metric, acqs_lock);
+        } else if (!strcmp(in_buf->buffer_type, "hfb")) {
+            auto frame = HFBFrameView(in_buf, frame_id);
+            write_hfb_data(frame, write_time_metric, acqs_lock);
         }
```
jrs65, line 123: I really don't like this (including the associated `write_X_data` routines). It's not clear this is any real improvement on just having separate stages for this.

The case I would like to think of is someone wanting to implement a new data/buffer type that wants to use all of this functionality, but isn't able to modify the source of this function, i.e. how can you do that in a purely pluggable format.

I can think of two vague directions that might work. One is moving as much functionality into a generic base that's templated on the frame view type, and then probably aliasing the templated types to different names to register separate processes.

Another is to move the format specific functionality into a separate class, which uses an abstract factory, and then use the buffer type to dynamically create the correct help class for writing the data.

There may be other options.

james-s-willis: Yeah this is the bit I wasn't sure how to handle correctly. So you would have a generic base class that `Writer.cpp` creates? Are there any other examples in `kotekan` that do this?

lib/stages/VisWriter.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+    INFO("Got new dataset ID={} with new fingerprint={}. Creating new acquisition.", ds_id, fp);
+
+    // If it is not known we need to initialise everything
+    acqs_fingerprint[fp] = std::make_shared<acqState>();
+    acqs[ds_id] = acqs_fingerprint.at(fp);
+    auto& acq = *(acqs.at(ds_id));
+
+    // get dataset states
+    get_dataset_state(ds_id);
```
james-s-willis, line 234: @jrs65 Does `get_dataset_state()` need to be called here? Because if not, I can move `init_acq` into `BaseWriter`.

lib/stages/VisWriter.cpp
```diff
@@ -0,0 +1,379 @@
...
+    try {
+        acq.file_bundle = std::make_unique<visFileBundle>(
+            file_type, root_path, instrument_name, metadata, chunk_id, file_length, window,
+            kotekan::logLevel(_member_log_level), ds_id, file_length);
+    } catch (std::exception& e) {
+        FATAL_ERROR("Failed creating file bundle for new acquisition: {:s}", e.what());
+    }
+}
+
+VisCalWriter::VisCalWriter(Config& config, const std::string& unique_name,
```
james-s-willis, line 254: @jrs65 I've updated the name of `visCalWriter` to `VisCalWriter`. Where is it actually used?

jrs65: It used to be used where we now use the shared memory buffer. I'm unsure whether we should keep it, or just remove it entirely (@tristpinsm thoughts?). It was a bit of a creative hack to fill a need, and I worry that keeping it around just places restrictions on what we can do with the Writers.

lib/stages/VisWriter.cpp
jrs65: I think so, it's needed to initialise information about the acquisition (which it sets within `acqs[ds_id]`). Is there a good other place to call it from?

Why can't you add `get_dataset_state` as a virtual method on `BaseWriter` and have it no-op for HFBWriter?

lib/stages/VisWriter.cpp
```diff
@@ -0,0 +1,379 @@
...
+        auto frame = VisFrameView(in_buf, frame_id);
+        write_data(frame, write_time_metric, acqs_lock);
+
+        // Mark the buffer and move on
+        mark_frame_empty(in_buf, unique_name.c_str(), frame_id++);
+
+        // Clean out any acquisitions that have been inactive long
+        close_old_acqs();
+    }
+}
```
jrs65, line 72: The main threads look super similar in both `HFBWriter` and `VisWriter`. Can you move the main thread in `BaseWriter`?

lib/stages/HFBAccumulate.cpp
```diff
@@ -43,12 +46,50 @@ integrateHFBData::integrateHFBData(Config& config_, const std::string& unique_na
...
+        std::iota(std::begin(freq_ids), std::end(freq_ids), 0);
+    }
+
+    // Create the frequency specification
+    // TODO: CHIME specific
+    std::vector<std::pair<uint32_t, freq_ctype>> freqs;
+    std::transform(std::begin(freq_ids), std::end(freq_ids), std::back_inserter(freqs),
+                   [](uint32_t id) -> std::pair<uint32_t, freq_ctype> {
+                       return {id, {800.0 - 400.0 / 1024 * id, 400.0 / 1024}};
+                   });
```
jrs65, line 74: I think all of this CHIME-only code can be replaced with the new Telescope class functionality.

lib/stages/VisWriter.cpp
tristpinsm: If it's no longer used at all I don't see any reason to keep it around. It is only good for that one task

lib/stages/VisWriter.cpp
jrs65: I think you'll need to change the API of `write_data` to take `(in_buf, frame_id)`

jrs65 (changes requested):
Looking good to me. I think we're getting very close to this being ready.

lib/stages/VisWriter.cpp
james-s-willis: I've already added `get_dataset_state` as a virtual method to `BaseWriter`and added code to the method in `HFBWriter`

lib/stages/VisWriter.cpp
james-s-willis: I've removed `VisCalWriter` and `visCalFileBundle`.

james-s-willis:
@jrs65 I've addressed your comments. I think it's close to being merged now.

jrs65 (changes requested):

james-s-willis:
@jrs65 I've addressed your latest comments. Let me know what you think.

lib/testing/FakeVis.hpp
```diff
@@ -58,9 +58,6 @@
  *                      Default is False.
  * @conf  frequencies   Array of UInt32. Definition of frequency IDs for
  *                      mode 'test_pattern_freq'.
- * @conf  dataset_id    Int. Use a fixed dataset ID and don't register
- *                      states. If not set, the dataset manager will create
- *                      the dataset ID.
  * @conf  sleep_before  Float. Sleep for this number of seconds before
```
jrs65, line 61: Why is this removed? The option still exists.

lib/utils/SystemInterface.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+const std::string get_user_name() {
+
+    std::string user(256, '\0');
+    user = (getlogin_r(&user[0], 256) == 0) ? user.c_str() : "unknown";
+
+    return user;
+}
+
+const std::string get_host_name() {
```
jrs65, line 14: These are great. Although I think *hostname* and *username* would be preferred, e.g. https://english.stackexchange.com/questions/100858/hostname-or-host-name

lib/utils/datasetState.hpp
```diff
@@ -688,4 +688,121 @@ class flagState : public datasetState {
...
+ * @author James Willis
+ */
+class subfreqState : public datasetState {
+public:
+    /**
+     * @brief Constructor
+     * @param data  The sub-frequency information as serialized by
+     *              subfreqState::to_json().
+     */
+    subfreqState(const nlohmann::json& data) {
```
jrs65, line 761: I don't think any of the other states are much better, but it would be great to document a little better what the actual state stored in here is. I gather it's subfreq index maps, but I'm not really sure what that means.

lib/utils/hfbFileRaw.cpp
```diff
@@ -0,0 +1,221 @@
...
+    if (ntime > delta_sync) {
+        flush_raw_sync(ntime - delta_sync);
+    }
+
+    times.push_back(new_time);
+
+    // Extend the file length for the new time
+#ifdef __linux__
+    fallocate(fd, 0, 0, frame_size * nfreq * num_time());
+#else
```
jrs65, line 175: Is there any possibility to merge common functionality of visFileRaw and hfbFileRaw? Either base one on the other, or have a common base class. There's a large amount of the implementation that's completely common.

python/kotekan/hfbbuffer.py
```diff
@@ -0,0 +1,779 @@
+"""Read a visBuffer dump into python.
```
jrs65, line 1: Comment needs changing.

python/scripts/read_hfb_data.py
```diff
@@ -3,85 +3,135 @@
 import matplotlib.pyplot as plt
```
jrs65, line 3: Does this script really warrant being in the repo? It looks like your own personal test script. Typically these kind of things become irrelevant and then entropy takes hold and they break, and then 12 months later you're trying to figure out what it was all for anyway, and just remove it.

python/kotekan/hfbbuffer.py
```diff
@@ -0,0 +1,779 @@
...
+        # data_struct = [
+        #     ("vis", np.complex64, self.num_stack),
+        #     ("weight", np.float32, self.num_stack),
+        #     ("flags", np.float32, self.num_elements),
+        #     ("eval", np.float32,  self.num_ev),
+        #     ("evec", np.complex64, self.num_ev * self.num_elements),
+        #     ("erms", np.float32,  1),
+        #     ("gain", np.complex64, self.num_elements),
+        # ]
+        # data_struct = np.dtype([(d[0],) + d[1:] for d in data_struct], align=True)
```
jrs65, line 603: Irrelevant copy pasta.

python/kotekan/hfbbuffer.py
```diff
@@ -0,0 +1,779 @@
...
+        -------
+        float
+        """
+        return self.tv + self.tv_nsec / 1e9
+
+
+class timeval(ctypes.Structure):
+    """Struct repr of a timeval type."""
+
+    _fields_ = [("tv_sec", ctypes.c_long), ("tv_usec", ctypes.c_long)]
```
jrs65, line 54: These are exactly the same as the definitions in `visbuffer.py`, so there should be a single implementation.

python/kotekan/hfbbuffer.py
```diff
@@ -0,0 +1,779 @@
+"""Read a visBuffer dump into python.
+"""
+# === Start Python 2/3 compatibility
+from __future__ import absolute_import, division, print_function, unicode_literals
+from future.builtins import *  # noqa  pylint: disable=W0401, W0614
+from future.builtins.disabled import *  # noqa  pylint: disable=W0401, W0614
+
```
jrs65, line 7: Generally I would suggest figuring out how to merge as much common functionality between `visbuffer.py` and `hfbbuffer.py` as it looks like there's a lot of identical stuff.

lib/stages/HFBAccumulate.cpp
```diff
@@ -43,12 +48,49 @@ integrateHFBData::integrateHFBData(Config& config_, const std::string& unique_na
...
+    const std::string weight_type = "hfb_weight_type";
+    const std::string git_tag = get_git_commit_hash();
+    const std::string instrument_name =
+        config.get_default<std::string>(unique_name, "instrument_name", "chime");
+
+    std::vector<uint32_t> freq_ids;
+
+    // Get the frequency IDs that are on this stream, check the config or just
+    // assume all CHIME channels
+    // TODO: CHIME specific
```
jrs65, line 62: I think this isn't actually CHIME specific any more.

jrs65 (changes requested):
James, your new changes look great. Here's a few more requested changes for you. Many of them are around factoring out common functionality in the `*FileRaw` implementations and in the `*buffer.py` implementations.

lib/stages/HFBWriter.cpp
```diff
@@ -0,0 +1,166 @@
...
+        // TODO: this should probably be reported to prometheus
+    } else if (acq.freq_id_map.count(freq_id) == 0) {
+        WARN("Frequency id={:d} not enabled for Writer, discarding frame", freq_id);
+
+        // Check that the number of visibilities matches what we expect
+    } else if (frame.num_beams != acq.num_beams) {
+        FATAL_ERROR("Number of beams in frame doesn't match state or file ({:d} != {:d}).",
+                    frame.num_beams, acq.num_beams);
+        return;
+
```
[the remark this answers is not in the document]
james-s-willis: I've replaced these fields with `frame_size` and moved most of `write_data` into `BaseWriter`.

lib/testing/FakeVis.hpp
james-s-willis: I thought it didn't exist because I did a search for it, but it just had a different name. I've added it back in.

python/scripts/read_hfb_data.py
james-s-willis: I prefer keeping this in the repo as I like to use it to test the raw files written.

python/kotekan/hfbbuffer.py
james-s-willis: I've moved the common stuff into `timespec.py`.

james-s-willis:
@jrs65 I've addressed your latest round of comments. However, it is now failing the unit tests in `test_inputsubset.py`:
```
/write0: Size of frame doesn't match file (248 != 2096).
```
which is the new addition to checking data validity based on `frame_size`. So it looks there's a disconnect between the config file parameters and the data written.

python/kotekan/hfbbuffer.py
jrs65: It's not just the timespec stuff I'm referring to in here. On the more trivial end there's another random function that is copied (freq_id_to_stream_id at the very end of the file), but I'm primarily thinking that the bulk of `HFBRaw` is the same as `VisRaw`, and as that's basically determined by the fact that the same writer code is used that writes metadata + data blocks into a 2D time-freq array it could be factored out into a common base class. The near term advantage of that is less code to maintain and fix, at the expense of more work now, the long term gain is that supporting new data formats that fit this time-freq pattern becomes super easy.

It might be that refactoring that is too much work for the moment, but it would be great if you could look over that and see how possible it is, and how long it might take, so we can see if we want to defer that or not.

python/scripts/read_hfb_data.py
jrs65: Okay. I'm mostly of the opinion that if it's just your own personal test script, it doesn't belong in the main repo. Mostly out of tidyness, as if we all did that the repo would be full of random scripts. Maybe @andrerenard has an opinion on this one?

jrs65 (changes requested):
I have a few comments about the Python stuff, but I think we're almost done here. Note, that I'm going to make an issue to discuss some metadata changes to do before we deploy.

jrs65:
Okay. Here's another slightly more major comments. I didn't appreciate we weren't calculating the weights in this version. Is there a reason we are not doing?

Certainly for the long run next year we should be (by the same even-odd method used in visAccumulate). Do we want to get that in this release?

james-s-willis:
@jrs65 I think this is ready to go now, what do you think?

jrs65 (approved):
Great work @james-s-willis! It's been a long-haul, but this is a really nice set of changes.

PR #810, opened 2020-08-19
Missed endpoint rename.

james-s-willis:


james-s-willis:
Yeah, I was confused at first how this bug got through CI.

PR #811, opened 2020-08-19
Read gains directory from separate endpoints.

james-s-willis:


lib/stages/ReadGain.cpp
```diff
@@ -199,34 +195,41 @@ void ReadGain::read_gain_tracking() {
...
-            }
-        } else {
-            if (_num_elements
-                != fread(&out_frame_tracking[b * _num_elements * 2], sizeof(float) * 2,
-                         _num_elements, ptr_myfile)) {
-                WARN("Gain file ({:s}) wasn't long enough! Something went wrong, using default "
-                     "gains",
-                     filename);
+    {
+        std::lock_guard<std::mutex> lock(mux);
```
andrerenard, line 199: Only lock the check and removal of the of the last beam from the queue.  Otherwise this blocks for as long as it takes the read the gain file (which could be a little while with NFS).

lib/stages/ReadGain.cpp
```diff
@@ -199,34 +195,41 @@ void ReadGain::read_gain_tracking() {
...
                 }
+                fclose(ptr_myfile);
             }
-            fclose(ptr_myfile);
-        }
-    } // end beam
+        } // end beam
+    }
     if (all_beams_successful_update) {
         gains_last_update_success_metric.labels({"tracking"}).set(1);
```
andrerenard, line 234: This metric should be per-beam now.  Otherwise a successful beam update on say beam 1, will overwrite a failure from say beam 0 before.   Alternatively it needs to track the quality of each been individually and then report this overall statistic.   However given that we are moving to each project having control over each beam, I'm not sure how useful the global statistic is.

lib/stages/ReadGain.cpp
```diff
@@ -115,28 +119,20 @@ bool ReadGain::update_gains_frb_callback(nlohmann::json& json) {
...
+                std::make_pair(beam_id, json.at("gain_dir").get<std::string>()));
+            INFO("[Tracking Beamformer] Updating gains from {:s}",
+                 _gain_dir_tracking.back().second);
+        } catch (std::exception const& e) {
+            WARN("[Tracking Beamformer] Fail to read gain_dir {:s}", e.what());
+            return false;
+        }
         update_gains_tracking = true;
     }
-    cond_var.notify_all();
```
andrerenard, line 139: You still need this notify, otherwise the loop in the main thread won't wake up to call `read_gain_tracking()`.

cubranic:
Service that applies FRB and Pulsar gains on new calibration broker output: https://github.com/CHIME-Pulsar-Timing/ch_psr_gains

andrerenard (approved):
This all looks good, as long as we've checked it on site (default gains load, and endpoints react as expected.) we can merge.

PR #814, opened 2020-08-28
Update Telescope objects

andrerenard:
* Add default frequency mapping for different ICETelescope layouts
* Move query of GPS time to telescope object (and C++ instead of Python)
* Remove CHIME specific GPS references from `kotekan.cpp`
* Allow system to query the FPGA frequency map directly from fpga_master (for testing)
* Change telescope names to match class names (inline with other `kotekan` factory generated objects`)
* (Also sneaks in a change for .gitignore) 
* Adds the metric `kotekan_vdif_lost_frames_total` to track VDIF frames lost to bad input packets.

### DIFF
--- a/lib/utils/CHIMETelescope.cpp
+++ b/lib/utils/CHIMETelescope.cpp
@@ -3,6 +3,7 @@
 #include "ICETelescope.hpp"   // for ice_stream_id_t, ice_encode_stream_id, ice_extract_stream_id
 #include "Telescope.hpp"      // for freq_id_t, REGISTER_TELESCOPE, _factory_aliasTelescope
 #include "kotekanLogging.hpp" // for ERROR, DEBUG2, WARN
+#include "restClient.hpp"
 
 #include "fmt.hpp"  // for format
 #include "json.hpp" // for json, basic_json<>::object_t, basic_json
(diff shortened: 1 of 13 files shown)


lib/utils/CHIMETelescope.cpp
```diff
@@ -15,29 +16,72 @@
...
+    set_frequency_map(fpga_freq_map_json);
+
+    if (require_gps && !gps_enabled) {
+        throw std::runtime_error("The system requires a GPS time, but none was found.");
+    }
 }
 
-void CHIMETelescope::set_frequency_map(const kotekan::Config& config, const std::string& path) {
-    nlohmann::json fpga_freq_map_json = config.get_default<nlohmann::json>("/", path, {});
+nlohmann::json CHIMETelescope::get_frequency_map(const std::string& host, const uint32_t port,
```
jrs65, line 65: I would call this method something like `fetch_frequency_map`. As `get_/set_` methods are a common pattern, you might believe that this one returns the frequency map currently set in kotekan, but it doesn't.

andrerenard: That's a very good point; I've fixed that now.   Thanks for the review!

PR #821, opened 2020-09-07
initial push for freq-encoding changes in pulsar packet

emmanuelfonseca:
this is the first in likely a series of changes to the pulsar post-processing code as we gradually move away from the VDIF structure. for starters, the "major" change i've made is to use previously-untouched parts of the current structure to encode frequency information, so that we can move away from the external map-retrieval process we've had to use since june 2020. 

for now we technically still use the old `VDIFHeader` structure definition but truly revamp to remove irrelevant features / add relevant feature once the key pulsar folks have/make time for converging on a plan.

### DIFF
--- a/lib/stages/pulsarNetworkProcess.cpp
+++ b/lib/stages/pulsarNetworkProcess.cpp
@@ -7,7 +7,7 @@
 #include "bufferContainer.hpp" // for bufferContainer
 #include "kotekanLogging.hpp"  // for FATAL_ERROR, INFO, CHECK_MEM
 #include "tx_utils.hpp"        // for add_nsec, get_vlan_from_ip, parse_chime_host_name, CLOCK_...
-#include "pulsar_functions.h"  // for PSRHeader
+#include "pulsar_functions.hpp" // for PSRHeader
 
 #include <arpa/inet.h>  // for inet_pton
 #include <atomic>       // for atomic_bool
--- a/lib/stages/pulsarNetworkProcess.cpp
+++ b/lib/stages/pulsarNetworkProcess.cpp
@@ -179,26 +179,26 @@ void pulsarNetworkProcess::main_thread() {
     clock_gettime(CLOCK_MONOTONIC, &t0);
 
     // added to take care of the missed frames
-    PSRHeader* header = reinterpret_cast<PSRHeader*>(packet_buffer);
-    int64_t psrheader_last_seconds = header->seconds;
-    int64_t psrheader_last_frame = header->data_frame;
+    PSRHeader* psr_header = reinterpret_cast<PSRHeader*>(packet_buffer);
+    int64_t psr_header_last_seconds = psr_header->seconds;
+    int64_t psr_header_last_frame = psr_header->data_frame;
 
     while (!stop_thread) {
         packet_buffer = wait_for_full_frame(in_buf, unique_name.c_str(), frame_id);
         if (packet_buffer == nullptr)
             break;
 
-        header = reinterpret_cast<PSRHeader*>(packet_buffer);
+        psr_header = reinterpret_cast<PSRHeader*>(packet_buffer);
         time_interval = 2560
-                        * (390625 * (header->seconds - psrheader_last_seconds)
-                           + 625 * (header->data_frame - psrheader_last_frame));
+                        * (390625 * (psr_header->seconds - psr_header_last_seconds)
+                           + 625 * (psr_header->data_frame - psr_header_last_frame));
 
         add_nsec(t0, time_interval);
         t1.tv_sec = t0.tv_sec;
         t1.tv_nsec = t0.tv_nsec;
 
-        psrheader_last_seconds = header->seconds;
-        psrheader_last_frame = header->data_frame;
+        psr_header_last_seconds = psr_header->seconds;
+        psr_header_last_frame = psr_header->data_frame;
 
         for (int frame = 0; frame < 80; frame++) {
             for (int beam = 0; beam < _num_beams; beam++) {
--- a/lib/stages/pulsarNetworkProcess.cpp
+++ b/lib/stages/pulsarNetworkProcess.cpp
@@ -6,8 +6,8 @@
 #include "buffer.h"             // for mark_frame_empty, wait_for_full_frame, register_consumer
 #include "bufferContainer.hpp"  // for bufferContainer
 #include "kotekanLogging.hpp"   // for FATAL_ERROR, INFO, CHECK_MEM
-#include "tx_utils.hpp"         // for add_nsec, get_vlan_from_ip, parse_chime_host_name, CLOCK_...
 #include "pulsar_functions.hpp" // for PSRHeader
+#include "tx_utils.hpp"         // for add_nsec, get_vlan_from_ip, parse_chime_host_name, CLOCK_...
 
 #include <arpa/inet.h>  // for inet_pton
 #include <atomic>       // for atomic_bool
--- a/lib/stages/pulsarNetworkProcess.cpp
+++ b/lib/stages/pulsarNetworkProcess.cpp
@@ -3,10 +3,10 @@
 #include "Config.hpp"       // for Config
 #include "StageFactory.hpp" // for REGISTER_KOTEKAN_STAGE, StageMakerTemplate
 #include "Telescope.hpp"
-#include "buffer.h"            // for mark_frame_empty, wait_for_full_frame, register_consumer
-#include "bufferContainer.hpp" // for bufferContainer
-#include "kotekanLogging.hpp"  // for FATAL_ERROR, INFO, CHECK_MEM
-#include "tx_utils.hpp"        // for add_nsec, get_vlan_from_ip, parse_chime_host_name, CLOCK_...
+#include "buffer.h"             // for mark_frame_empty, wait_for_full_frame, register_consumer
+#include "bufferContainer.hpp"  // for bufferContainer
+#include "kotekanLogging.hpp"   // for FATAL_ERROR, INFO, CHECK_MEM
+#include "tx_utils.hpp"         // for add_nsec, get_vlan_from_ip, parse_chime_host_name, CLOCK_...
 #include "pulsar_functions.hpp" // for PSRHeader
 
 #include <arpa/inet.h>  // for inet_pton
--- a/lib/stages/pulsarNetworkProcess.cpp
+++ b/lib/stages/pulsarNetworkProcess.cpp
@@ -7,7 +7,7 @@
 #include "bufferContainer.hpp" // for bufferContainer
 #include "kotekanLogging.hpp"  // for FATAL_ERROR, INFO, CHECK_MEM
 #include "tx_utils.hpp"        // for add_nsec, get_vlan_from_ip, parse_chime_host_name, CLOCK_...
-#include "vdif_functions.h"    // for VDIFHeader
+#include "pulsar_functions.h"  // for PSRHeader
 
 #include <arpa/inet.h>  // for inet_pton
 #include <atomic>       // for atomic_bool
@@ -179,26 +179,26 @@ void pulsarNetworkProcess::main_thread() {
     clock_gettime(CLOCK_MONOTONIC, &t0);
 
     // added to take care of the missed frames
-    VDIFHeader* header = reinterpret_cast<VDIFHeader*>(packet_buffer);
-    int64_t vdif_last_seconds = header->seconds;
-    int64_t vdif_last_frame = header->data_frame;
+    PSRHeader* header = reinterpret_cast<PSRHeader*>(packet_buffer);
+    int64_t psrheader_last_seconds = header->seconds;
+    int64_t psrheader_last_frame = header->data_frame;
 
     while (!stop_thread) {
         packet_buffer = wait_for_full_frame(in_buf, unique_name.c_str(), frame_id);
         if (packet_buffer == nullptr)
             break;
 
-        header = reinterpret_cast<VDIFHeader*>(packet_buffer);
+        header = reinterpret_cast<PSRHeader*>(packet_buffer);
         time_interval = 2560
-                        * (390625 * (header->seconds - vdif_last_seconds)
-                           + 625 * (header->data_frame - vdif_last_frame));
+                        * (390625 * (header->seconds - psrheader_last_seconds)
+                           + 625 * (header->data_frame - psrheader_last_frame));
 
         add_nsec(t0, time_interval);
         t1.tv_sec = t0.tv_sec;
         t1.tv_nsec = t0.tv_nsec;
 
-        vdif_last_seconds = header->seconds;
-        vdif_last_frame = header->data_frame;
+        psrheader_last_seconds = header->seconds;
+        psrheader_last_frame = header->data_frame;
 
         for (int frame = 0; frame < 80; frame++) {
             for (int beam = 0; beam < _num_beams; beam++) {
--- a/lib/stages/pulsarPostProcess.cpp
+++ b/lib/stages/pulsarPostProcess.cpp
@@ -71,9 +71,9 @@ pulsarPostProcess::~pulsarPostProcess() {
     delete[] in_frame;
 }
 
-void pulsarPostProcess::fill_headers(unsigned char* out_buf, struct PSRHeader* psr_header,
-                                     const uint64_t fpga_seq_num, struct timespec* time_now,
-                                     struct beamCoord* beam_coord, uint16_t* thread_ids) {
+void pulsarPostProcess::fill_headers(unsigned char* out_buf, PSRHeader* psr_header,
+                                     const uint64_t fpga_seq_num, timespec* time_now,
+                                     beamCoord* beam_coord, uint16_t* thread_ids) {
 
     // Get the Telescope instance and pre-calc the length of an FPGA frame
     auto& tel = Telescope::instance();
@@ -90,7 +90,7 @@ void pulsarPostProcess::fill_headers(unsigned char* out_buf, struct PSRHeader* p
             (time_now->tv_nsec / 1.e9) / (_timesamples_per_pulsar_packet * fpga_s);
 
         for (uint f = 0; f < freqloop; ++f) {
-            // load frequency indeces into thread_id and EUD3.
+            // load frequency indices into thread_id and EUD3.
             if (f == 0) {
                 psr_header->thread_id = thread_ids[f];
                 psr_header->eud3 = 0;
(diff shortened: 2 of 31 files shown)


lib/stages/pulsarPostProcess.cpp
```diff
@@ -188,17 +197,22 @@ void pulsarPostProcess::main_thread() {
...
+            // the following was commented out by E. Fonseca on 7 September 2020.
+            //if (_timesamples_per_pulsar_packet == 3125) {
+            //    thread_ids[i] = tel.to_freq_id(in_buf[i], in_buffer_ID[i]);
+            //} else if (_timesamples_per_pulsar_packet == 625) {
+            //    // In the case of 4 frequencies per packet we convert the stream_id into a
+            //    // kind of node_id that runs from 0 to 255 for the thread_id.
+            //    ice_stream_id_t stream_id = ice_get_stream_id_t(in_buf[i], in_buffer_ID[i]);
+            //    thread_ids[i] =
+            //        stream_id.crate_id * 16 + stream_id.slot_id + stream_id.link_id * 32;
+            //}
```
andrerenard, line 215: Just remove code that's no longer used.  `git` is better for tracking code changes.

lib/utils/pulsar_functions.h
```diff
@@ -0,0 +1,36 @@
...
+    uint32_t vdif_version : 3;
+    uint32_t station_id : 16;
+    uint32_t thread_id : 10;
+    uint32_t bits_depth : 5;
+    uint32_t data_type : 1;
+    uint32_t eud1 : 24;
+    uint32_t edv : 8;
+    uint32_t eud2 : 32;
+    uint32_t eud3 : 32;
+    uint32_t eud4 : 32;
```
andrerenard, line 28: Since this is no longer "VDIF" you could rename/resize these to match the data going into them.  If you don't want to do that just yet, it would be worth putting comments into here as to what data they contain.   Since people are more likely to look here than code that sets the values to figure out what they are.

lib/stages/pulsarPostProcess.cpp
```diff
@@ -145,32 +154,32 @@ void pulsarPostProcess::main_thread() {
...
-    vdif_header.station_id = (si[0] << 8) + si[1]; // Need to fomrally ask the Vdif community
-    vdif_header.thread_id = 0;                     // UD   freq
-    vdif_header.bits_depth = 3;                    // 4+4 bit so 4-1=3
-    vdif_header.data_type = 1;                     // Complex
-    vdif_header.edv = 0;
-    vdif_header.eud1 = 0; // UD: beam number [0 to 9]
-    vdif_header.eud2 = 0; //_psr_scaling from metadata
-    vdif_header.eud3 = 0; // Not used for now
-    vdif_header.eud4 = 0; // 16-b RA + 16-b Dec
+    psr_header.station_id = (si[0] << 8) + si[1]; // Need to fomrally ask the Vdif community
```
andrerenard, line 174:
```suggestion
    psr_header.station_id = (si[0] << 8) + si[1];
```
We aren't going to register this format.   We might register the new VDIF system, but that's going to be another part of the code.

lib/stages/pulsarPostProcess.cpp
```diff
@@ -85,19 +85,28 @@ void pulsarPostProcess::fill_headers(unsigned char* out_buf, struct VDIFHeader*
...
-        vdif_header->seconds = time_now->tv_sec - unix_offset;
-        vdif_header->data_frame =
+        psr_header->seconds = time_now->tv_sec - unix_offset;
+        psr_header->data_frame =
             (time_now->tv_nsec / 1.e9) / (_timesamples_per_pulsar_packet * fpga_s);
 
         for (uint f = 0; f < freqloop; ++f) {
-            vdif_header->thread_id = thread_ids[f];
+            // Before looping over pulsar beams, load freq labels into header.
+            // As of 7 September 2020, load one of four labels into thread_id,
```
andrerenard, line 94: I'm hesitant to include dates in comments.  The time this really changes is when this version code is run on site.  We should rely on the version tracking in `git` and `comet` and not comments to know when things changed.

lib/stages/pulsarPostProcess.cpp
```diff
@@ -85,19 +85,28 @@ void pulsarPostProcess::fill_headers(unsigned char* out_buf, struct VDIFHeader*
...
-            vdif_header->thread_id = thread_ids[f];
+            // Before looping over pulsar beams, load freq labels into header.
+            // As of 7 September 2020, load one of four labels into thread_id, 
+            // and pack the other three into the 32-b EUD3 attribute.
+            // TODO: redefine pulsar packet structure to not use VDIF def, instead use 
+            // structures with dimensions defined by telescope-config variables.
+            if (f == 0) {
+                psr_header->thread_id = thread_ids[f];
+            else {
+                psr_header->eud3 += (thread_ids[f] << (f - 1) * 10);
```
andrerenard, line 101: The `+=` here requires that `eud3` contains zero at the start of this loop.  However that's not a guarantee, frames are not zeroed between uses.  You can likely just zero this in the `f == 0` case, or add another case to this if for `f == 1`.

andrerenard (changes requested):
Thanks for these changes, I have a few comments inline.

lib/stages/pulsarPostProcess.cpp
```diff
@@ -85,19 +85,28 @@ void pulsarPostProcess::fill_headers(unsigned char* out_buf, struct VDIFHeader*
...
         for (uint f = 0; f < freqloop; ++f) {
-            vdif_header->thread_id = thread_ids[f];
+            // Before looping over pulsar beams, load freq labels into header.
+            // As of 7 September 2020, load one of four labels into thread_id, 
+            // and pack the other three into the 32-b EUD3 attribute.
+            // TODO: redefine pulsar packet structure to not use VDIF def, instead use 
+            // structures with dimensions defined by telescope-config variables.
+            if (f == 0) {
+                psr_header->thread_id = thread_ids[f];
+            else {
```
andrerenard, line 100:
```suggestion
            } else {
```

lib/stages/pulsarPostProcess.cpp
emmanuelfonseca: of course, just an old habit of mine; removed.

lib/stages/pulsarPostProcess.cpp
emmanuelfonseca: removed, and added a string describing the current use of `eud3`.

lib/stages/pulsarPostProcess.cpp
emmanuelfonseca: agreed totally, so i've removed this comment.

lib/stages/pulsarPostProcess.cpp
emmanuelfonseca: ah, great catch andre; added a zeroing statement within the if-statement.

lib/stages/pulsarPostProcess.cpp
emmanuelfonseca: another good catch, thanks for that.

lib/utils/pulsar_functions.h
emmanuelfonseca: good point; i've added some small doc-like strings to describe their use.

emmanuelfonseca:
@andrerenard addressed your current comments and pushed, let me know if there are other things you'd like addressed.

lib/utils/pulsar_functions.c
```diff
@@ -0,0 +1 @@
+#include "pulsar_functions.h"
```
andrerenard, line 1: You don't need this file. It's just a header include.   If you did need this for something, then it should be included in the CMakeLists.txt file, otherwise this file won't be compiled.

lib/stages/pulsarNetworkProcess.cpp
```diff
@@ -179,26 +179,26 @@ void pulsarNetworkProcess::main_thread() {
     clock_gettime(CLOCK_MONOTONIC, &t0);
 
     // added to take care of the missed frames
-    VDIFHeader* header = reinterpret_cast<VDIFHeader*>(packet_buffer);
-    int64_t vdif_last_seconds = header->seconds;
-    int64_t vdif_last_frame = header->data_frame;
+    PSRHeader* header = reinterpret_cast<PSRHeader*>(packet_buffer);
+    int64_t psrheader_last_seconds = header->seconds;
```
andrerenard, line 183: Minor point: in some places it's `psrheader*` and in others it's `psr_header*`, for consistency I'd say just make the references here `psr_header*`

andrerenard (changes requested):
Thanks for the updates, two small requests inline.  Also it looks like it isn't passing the lint checks.  You might want to run `make clang-format`, or just check what it's not accepting.

lib/utils/pulsar_functions.c
emmanuelfonseca: funny you you say this, i was also wondering about this file but for safety i copied what's done for the VDIF utils. to be clear, you're suggesting i remove `pulsar_functions.c`?

lib/stages/pulsarNetworkProcess.cpp
emmanuelfonseca: agreed, done.

lib/utils/pulsar_functions.c
andrerenard: Yes remove `pulsar_functions.c` I think `vdif_functions.c` used to have something else in it - and we've just never removed that file.

lib/utils/pulsar_functions.h
```diff
@@ -0,0 +1,41 @@
+#ifndef PULSAR_FUNCTIONS_H
+#define PULSAR_FUNCTIONS_H
+
+#include <inttypes.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
```
andrerenard, line 8: Also I don't think we need this (and the 3 corresponding lines below) anymore either.  This was a leftover from the days when the network capture code was written in C not C++.  If the code compiles without it, then just leave it out.  And we can change the name of this file to `pulsar_functions.hpp`

lib/utils/pulsar_functions.c
emmanuelfonseca: nice, will do.

lib/utils/pulsar_functions.h
emmanuelfonseca: sounds good; i removed both of the `#ifdef _cplusplus_` lines, renamed the file and pushed.

emmanuelfonseca: ah yes, going to change the include call, one moment

emmanuelfonseca:
@andrerenard finally got the `lint` tests to pass, all your previous comments have been addressed, and compilation on `cs9g1` raises no errors. feel free to suggest additional things my way.

lib/stages/pulsarPostProcess.cpp
```diff
@@ -188,17 +195,11 @@ void pulsarPostProcess::main_thread() {
         if (!new_frame_fpga_seq_num.has_value())
             return;
 
+        // Initialize data for header info, namely position and frequency labels.
+        // See fill_headers() for bit-packing of freq labels into PSRHeader.eud3.
```
cubranic, line 199: Could a doxygen comment for `PSRHeader` struct explain how the labels are packed into eud3?

lib/utils/pulsar_functions.hpp
```diff
@@ -0,0 +1,27 @@
...
+    uint32_t log_num_chan : 5;
+    uint32_t vdif_version : 3;
+    uint32_t station_id : 16;
+    uint32_t thread_id : 10; // index of first packed frequency.
+    uint32_t bits_depth : 5;
+    uint32_t data_type : 1;
+    uint32_t eud1 : 24; // UD: beam number [0 to 9]
+    uint32_t edv : 8;
+    uint32_t eud2 : 32; // _psr_scaling from metadata
+    uint32_t eud3 : 32; // 32-b number encoding other three frequency indeces.
```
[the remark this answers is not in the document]
cubranic: And: how they're encoded. (ten bits per frequency: `thread_id[3]<<20 + thread_id[2]<<10 + thread_id[1]`?)

cubranic (commented):
Minor points about variable naming and documentation. But also a proposal open to discussion: should the `eud` fields in PSRHeader struct be named more descriptively? Can @andrerenard also add his opinion on this?

lib/stages/pulsarPostProcess.cpp
```diff
@@ -145,32 +152,32 @@ void pulsarPostProcess::main_thread() {
...
-    vdif_header.bits_depth = 3;                    // 4+4 bit so 4-1=3
-    vdif_header.data_type = 1;                     // Complex
-    vdif_header.edv = 0;
-    vdif_header.eud1 = 0; // UD: beam number [0 to 9]
-    vdif_header.eud2 = 0; //_psr_scaling from metadata
-    vdif_header.eud3 = 0; // Not used for now
-    vdif_header.eud4 = 0; // 16-b RA + 16-b Dec
+    psr_header.station_id = (si[0] << 8) + si[1];
+    psr_header.thread_id = 0;  // index of first packed frequency.
+    psr_header.bits_depth = 3; // 4+4 bit so 4-1=3
```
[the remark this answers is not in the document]
emmanuelfonseca: as far as i understand things, we don't use this (or many) of the VDIF-specific fields, so we'll eventually scrap this.

lib/stages/pulsarPostProcess.cpp
```diff
@@ -145,32 +152,32 @@ void pulsarPostProcess::main_thread() {
...
-    vdif_header.station_id = (si[0] << 8) + si[1]; // Need to fomrally ask the Vdif community
-    vdif_header.thread_id = 0;                     // UD   freq
-    vdif_header.bits_depth = 3;                    // 4+4 bit so 4-1=3
-    vdif_header.data_type = 1;                     // Complex
-    vdif_header.edv = 0;
-    vdif_header.eud1 = 0; // UD: beam number [0 to 9]
-    vdif_header.eud2 = 0; //_psr_scaling from metadata
-    vdif_header.eud3 = 0; // Not used for now
-    vdif_header.eud4 = 0; // 16-b RA + 16-b Dec
+    psr_header.station_id = (si[0] << 8) + si[1];
```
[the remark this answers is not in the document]
emmanuelfonseca: i'll keep it as is for now as i'm almost certain CHIME/Pulsar doesn't care, but the key folks haven't been available to carefully check and we're planning to scrap the packet-header def in earnest over the next few months.

lib/utils/pulsar_functions.hpp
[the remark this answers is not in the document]
emmanuelfonseca: good call, will note this since we're going to overhaul things completely over the next few months and will want to remember this.

lib/stages/pulsarPostProcess.cpp
```diff
@@ -85,19 +85,26 @@ void pulsarPostProcess::fill_headers(unsigned char* out_buf, struct VDIFHeader*
...
         uint64_t fpga_now = (fpga_seq_num + _timesamples_per_pulsar_packet * i);
-        vdif_header->seconds = time_now->tv_sec - unix_offset;
-        vdif_header->data_frame =
+        psr_header->seconds = time_now->tv_sec - unix_offset;
+        psr_header->data_frame =
             (time_now->tv_nsec / 1.e9) / (_timesamples_per_pulsar_packet * fpga_s);
 
         for (uint f = 0; f < freqloop; ++f) {
-            vdif_header->thread_id = thread_ids[f];
+            // load frequency indeces into thread_id and EUD3.
```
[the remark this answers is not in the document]
emmanuelfonseca: thanks for catching that, fixed.

lib/utils/pulsar_functions.hpp
```diff
@@ -0,0 +1,27 @@
...
+    uint32_t ref_epoch : 6;
+    uint32_t unused : 2;
+    uint32_t frame_len : 24;
+    uint32_t log_num_chan : 5;
+    uint32_t vdif_version : 3;
+    uint32_t station_id : 16;
+    uint32_t thread_id : 10; // index of first packed frequency.
+    uint32_t bits_depth : 5;
+    uint32_t data_type : 1;
+    uint32_t eud1 : 24; // UD: beam number [0 to 9]
```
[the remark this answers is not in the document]
emmanuelfonseca: we're going to move away from the VDIF/`EUD` convention altogether, but over the next few months (as none of us pulsar "programmers" have had much time due to other ongoing work for pulsar and FRB). so i'll wait to make these sorts of changes for now due to time constraints, but i agree with them and they'll be made in earnest over the next ~2 months (likely to be implemented after the winter run).

lib/stages/pulsarPostProcess.cpp
```diff
@@ -71,7 +71,7 @@ pulsarPostProcess::~pulsarPostProcess() {
     delete[] in_frame;
 }
 
-void pulsarPostProcess::fill_headers(unsigned char* out_buf, struct VDIFHeader* vdif_header,
+void pulsarPostProcess::fill_headers(unsigned char* out_buf, struct PSRHeader* psr_header,
```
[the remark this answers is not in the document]
emmanuelfonseca: ah yes, good call. i've removed the explicit struct calls throughout.

lib/stages/pulsarPostProcess.cpp
```diff
@@ -85,19 +85,26 @@ void pulsarPostProcess::fill_headers(unsigned char* out_buf, struct VDIFHeader*
...
 
             for (uint32_t psr = 0; psr < _num_pulsar; ++psr) {
-                vdif_header->eud1 = psr; // beam id
-                vdif_header->eud2 = beam_coord[f].scaling[psr];
+                psr_header->eud1 = psr; // beam id
+                psr_header->eud2 = beam_coord[f].scaling[psr];
                 uint16_t ra_part = (uint16_t)(beam_coord[f].ra[psr] * 100);
                 uint16_t dec_part = (uint16_t)((beam_coord[f].dec[psr] + 90) * 100);
-                vdif_header->eud4 = ((ra_part << 16) & 0xFFFF0000) + (dec_part & 0xFFFF);
+                psr_header->eud4 = ((ra_part << 16) & 0xFFFF0000) + (dec_part & 0xFFFF);
```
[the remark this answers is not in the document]
emmanuelfonseca: ran a quick test and it appears you're correct, so i've removed the redundancy.

lib/stages/pulsarPostProcess.cpp
```diff
@@ -85,19 +85,26 @@ void pulsarPostProcess::fill_headers(unsigned char* out_buf, struct VDIFHeader*
...
+            // load frequency indeces into thread_id and EUD3.
+            if (f == 0) {
+                psr_header->thread_id = thread_ids[f];
+                psr_header->eud3 = 0;
+
+            } else {
+                psr_header->eud3 += (thread_ids[f] << (f - 1) * 10);
+            }
 
             for (uint32_t psr = 0; psr < _num_pulsar; ++psr) {
```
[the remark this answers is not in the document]
emmanuelfonseca: agreed; made this change.

lib/stages/pulsarPostProcess.cpp
emmanuelfonseca: never used `doxygen` before but i took a stab at a doxygen-readable comment.

emmanuelfonseca:
ok, the latest comments from @cubranic have been addressed and various things incorporated.

andrerenard (approved):

PR #822, opened 2020-09-09
ci(docker): build bitshuffle with -march=haswell

### LINKED ISSUE #819: CI: SIGILL in test_transpose.py::test_transpose
test_transpose.py::test_transpose started failing with SIGILL in kotekan. We added an option to run the test through gdb to get a backtrace in #818. The problem did not appear anymore when it's ran in gdb. It also doesn't appear when ran in a tmate session. So we left the test wrapped in gdb for now (06b47a7). This should be reverted once the problem is truly solved.

[log.txt](https://github.com/kotekan/kotekan/files/5170318/log.txt)

nritsche:
Closes #819

### DIFF
--- a/tests/test_accumulate.py
+++ b/tests/test_accumulate.py
@@ -72,7 +72,6 @@ def accumulate_data(tmpdir_factory):
         ),
         dump_buffer,
         accumulate_params,
-        gdb=True,
     )
 
     test.run()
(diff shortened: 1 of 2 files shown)


andrerenard (approved):
Looks great, thanks for fixing this!

PR #823, opened 2020-09-12
Add frame0_ns time to FRB L0 packet header (fix 768)

### LINKED ISSUE #768: Add GPS time to the FRB L0->L1 packet
Currently the mapping from the FPGA sequence number to GPS time is done with an out-of-band communication between the L1 nodes and `fpga_master`.

To avoid this we should add the GPS time to each of the L0->L1 packets.  This has the added benefit of allowing a free running clock on the FPGAs independent of the GPS time, should we ever want that clock configuration. 

This ticket is for adding this functionality to `kotekan`, the L1 change will be tracked separately.

cubranic:
Since this changes the header format, we bump up protocol number to 2, and will
need a corresponding change on L1.  Fixes #768


andrerenard (approved):
Thanks, this looks good!  I don't think there is anywhere else we'd want to get the `frame_0` time directly, so I agree it's not really worth adding that to API.

cubranic:
Weird, but this tiny change is causing a segmentation fault in:
```
GPU[0] device interface: Allocating GPU[0] memory: rfi_mask_output, len: 192, ptr: 0x2402de4000
Segmentation fault
```

andrerenard:
Are their still outstanding issues on this PR, or is it ready for (re)review?

cubranic:
I was hoping to fix the Python tests to work with it, but it can be reviewed now.

andrerenard (approved):
This all looks good to me.  I'll hold off merging for now so you can add the python fixes.

PR #824, opened 2020-09-12
Type name in Python 3 should be a string not bytes

cubranic:


andrerenard:
What is this trying to fix?  It seems like it's broken the boost tests in the current form.

cubranic:
I don’t see how that is possible, it changes a method in a module that is only included from `tests/test_frb_post_process.py`, and another module only used by `tests/test_psr_post_process.py`.

BTW, those tests are skipped on the CI server because it doesn’t have the captured buffer files to use as inputs, and it’s not run with the option to use the more time-consuming method. That’s why the bug was not discovered before.

andrerenard:
Looks like it's passing now, so it must have been something else going wrong with the CI.   If @nritsche can review, then we can merge this in.

nritsche:
> What is this trying to fix? It seems like it's broken the boost tests in the current form.

I also don't know what this is trying to fix.

jrs65:
It sounds like I can merge this, right @cubranic ?

jrs65:
This seems low risk. I'm going to approve and merge it just to get it off the queue.

jrs65 (approved):

cubranic:
No please wait, it hasn’t been tested recently.

jrs65:
Too late!

PR #825, opened 2020-09-18
fix(shared_memory_buffer): upon update, first check if link still exists to shared_mem

anjakefala:
If kotekan-calbuffer is restarted, then it will write the buffer to a
new location in shared memory, but the calibration broker will continue
to check old location, and will continue to see the same old data, which
it will ignore.

The SharedMemoryReader should check on every .update() call to
see if the file descriptor still has any links.

mmap does not contain a reference to the file descriptor used to create
it, and so we will need to keep the original file descriptor around.

Addresses https://github.com/chime-experiment/ch_cal/issues/2

### DIFF
--- a/python/kotekan/shared_memory_buffer.py
+++ b/python/kotekan/shared_memory_buffer.py
@@ -85,10 +85,10 @@ class SharedMemoryReader:
 
         self.shared_mem_name = shared_memory_name
         self.semaphore = posix_ipc.Semaphore(shared_memory_name)
-        shared_mem = posix_ipc.SharedMemory(shared_memory_name)
+        self.shared_mem_file = posix_ipc.SharedMemory(shared_memory_name)
 
         # 0 means entire file
-        self.shared_mem = mmap.mmap(shared_mem.fd, 0, prot=mmap.PROT_READ)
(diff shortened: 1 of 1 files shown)


nritsche (approved):
Great, thanks!
Looks good.

jrs65 (approved):
Looks great! 👍

python/kotekan/shared_memory_buffer.py
```diff
@@ -85,10 +85,10 @@ def __init__(self, shared_memory_name, view_size, comet_manager=None):
 
         self.shared_mem_name = shared_memory_name
         self.semaphore = posix_ipc.Semaphore(shared_memory_name)
-        shared_mem = posix_ipc.SharedMemory(shared_memory_name)
+        self.shared_mem_file = posix_ipc.SharedMemory(shared_memory_name)
```
anjakefala, line 88: This will throw an `ExistentialError` if `SharedMemory(shared_memory_name)` does not exist (which happens if the reader tries to instantiate, before the writer has created the SharedMemory region).

We should possibly check for `ExistentialError`s, and throw a `SharedMemoryError` with the appropriate message.

nritsche (approved):
lgtm

PR #826, opened 2020-09-22
Fix rfi band calc

james-s-willis:
I've added the fixes to the script for both `rfi_band` calculation and using NaN in the `waterfall` arrays instead of `-1`. The only issue I have is that there were commented out parts of the code which had `waterfall` initialised to `np.nan` and for some reason Jacob didn't use them...


jrs65 (approved):
These changes look good to me. However, it would be good to check to see if it actually makes the values exported to prometheus more sane. Is there a way of doing that without interrupting the current  service?

james-s-willis:
That's the issue I don't know of a way to test this without it going live on `carillon`.

jrs65:
I guess we should wait until deployment week then?

james-s-willis:
Yeah, that it was what I was thinking.

james-s-willis:
@jrs65 I've added that fix now. I think it's good to be merged.

PR #828, opened 2020-09-22
Fix LGTM warnings

nritsche:


lib/stages/RfiFrameDrop.hpp
```diff
@@ -3,8 +3,8 @@
 @brief Drops GPU frames when they are contaminated with RFI
 - RfiFrameDrop : public kotekan::Stage
 *****************************************/
-#ifndef VALVE_HPP
-#define VALVE_HPP
+#ifndef RFI_FRAME_DROP_HPP
```
jrs65, line 6: Ha oops. Mea cupla.

python/kotekan/_version.py
```diff
@@ -524,7 +524,7 @@ def get_versions():
         # versionfile_source is the relative path from the top of the source
         # tree (where the .git directory might live) to this file. Invert
         # this to find the root from __file__.
-        for i in cfg.versionfile_source.split("/"):
+        for _ in cfg.versionfile_source.split("/"):
```
jrs65, line 527: I'd be tempted to ignore this one. As this file is produced by versioneer it'll make it annoying to see what's going on if we ever update versioneer and re-gen this file.

python/requirements.txt
```diff
@@ -4,6 +4,7 @@ click
 h5py
 future
 requests
+shutil
```
jrs65, line 7: Why? It's a built in module.

jrs65 (changes requested):
Most of the LGTM suggestions seem to be pretty good, although there's a few slightly funky ones.

nritsche:
Thanks @jrs65 ! Done.

jrs65 (approved):

python/kotekan/_version.py
nritsche: I think we blackened that file btw :laughing:

andrerenard (approved):

PR #830, opened 2020-09-24
Export data quality metric

james-s-willis:
I think this is ready to review, though I'm not sure on the config file change.


lib/stages/DataQuality.cpp
```diff
@@ -0,0 +1,119 @@
...
+}
+
+DataQuality::~DataQuality() {}
+
+void DataQuality::calc_alpha_coeffs(dset_id_t ds_id) {
+
+    auto& dm = datasetManager::instance();
+
+    auto sstate_fut = std::async(&datasetManager::dataset_state<stackState>, &dm, ds_id);
+    const stackState* ss = sstate_fut.get();
```
jrs65, line 51: I think there's no point doing an async call here as you don't do anything in the middle. Might as well just do a normal blocking call.

lib/stages/DataQuality.cpp
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        // Get correct set of alpha coefficients
+        const std::vector<double>& alpha = dset_id_map.at(ds_id);
+
+        // Compute sensitivity
+        double sensitivity = 0;
+
+        for (uint32_t i = 0; i < frame.num_prod; i++) {
+            auto var = (frame.weight[i] == 0 ? 0.0 : frame.weight[i]);
```
jrs65, line 108: I'm pretty sure this should be:
```suggestion
            auto var = (frame.weight[i] == 0 ? 0.0 : 1.0 / frame.weight[i]);
```

lib/stages/DataQuality.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+void DataQuality::calc_alpha_coeffs(dset_id_t ds_id) {
+
+    auto& dm = datasetManager::instance();
+
+    auto sstate_fut = std::async(&datasetManager::dataset_state<stackState>, &dm, ds_id);
+    const stackState* ss = sstate_fut.get();
+
+    if (ss == nullptr) {
+        FATAL_ERROR("Couldn't find stackState ancestor of dataset "
```
jrs65, line 54: You should use the `fingerprint` technique in here. The way you have this implemented at the moment, you're going  to regenerate the alphas every time the dataset ID changes, which is every 10s at the moment, but stack state never changes.

In theory you could just use the stack state hash for that purpose, but we may eventually recalculate when the flags change too, so I would go straight to the fingerprint method.

lib/stages/DataQuality.hpp
```diff
@@ -0,0 +1,64 @@
...
+ *
+ *
+ * @par Buffers
+ * @buffer in_buf Visibility data.
+ *     @buffer_format VisBuffer structured
+ *     @buffer_metadata VisMetadata
+ *
+ * @conf  num_elements          Int. The number of elements (i.e. inputs) in the
+ *                              correlator data.
+ *
```
jrs65, line 32: Make sure to document the metrics exported.

jrs65 (changes requested):

james-s-willis:
@jrs65 I've addressed your review comments.

jrs65 (changes requested):
Oh oops. I write this review yesterday, but forgot to send it!

lib/stages/DataQuality.cpp
```diff
@@ -0,0 +1,126 @@
...
+REGISTER_KOTEKAN_STAGE(DataQuality);
+
+DataQuality::DataQuality(Config& config_, const std::string& unique_name,
+                         bufferContainer& buffer_container) :
+    Stage(config_, unique_name, buffer_container, std::bind(&DataQuality::main_thread, this)),
+    data_quality_metric(Metrics::instance().add_gauge("kotekan_dataquality_dataquality",
+                                                      unique_name, {"freq_id"})) {
+
+    // Apply config.
+    _num_elements = config.get<uint32_t>(unique_name, "num_elements");
```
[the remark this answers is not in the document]
jrs65: Ah okay. I think there was a mistake in my description of the algorithm. I'll fix it in the parent issue, but basically: alpha = counts**2 / sum(counts)**2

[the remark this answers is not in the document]
jrs65: So I don't think you need `_num_elements` because of that.

james-s-willis:
@jrs65 I think I have fixed it now

lib/stages/DataQuality.cpp
```diff
@@ -0,0 +1,121 @@
...
+    while (!stop_thread) {
+
+        // Get input visibilities. We assume the shape of these doesn't change.
+        if (wait_for_full_frame(in_buf, unique_name.c_str(), input_frame_id) == nullptr) {
+            break;
+        }
+
+        auto frame = VisFrameView(in_buf, input_frame_id);
+        dset_id_t ds_id = frame.dataset_id;
+        auto fprint = dm.fingerprint(ds_id, {"stack"});
```
jrs65, line 96: I would consider populating a ds_id -> fprint map here, though I don't think it's a must do. The fingerprint calculation involves some json manipulation which probably isn't great to do every single time. It's probably fine for our purposes at the moment, though if you decide not to change it we should look at the performance while testing.

jrs65 (commented):
Looking good. The only thing left is to change to the correct algorithm for the alpha as my initial version had a mistake (sorry!), and then remove `_num_elements`. Details are in a comment.

james-s-willis:
@jrs65 I've added a map of `dataset_id` -> `fingerprint` and fixed the alpha calculation.

jrs65 (approved):
Great work. Thanks @james-s-willis

andrerenard (approved):

PR #833, opened 2020-10-02
Fix config to account for weights in HFB data

james-s-willis:


config/chime_science_run_gpu.yaml
```diff
@@ -348,9 +348,10 @@ gpu_beamform_hfb_output_buffers:
   gpu_beamform_hfb_output_buffer_3:
     kotekan_buffer: standard
 
+# Factor of 2 in the frame_size comes from the weights
 cpu_hfb_output_buffers:
   num_frames: buffer_depth
-  frame_size: num_data_sets * factor_upchan * num_frb_total_beams * sizeof_float
+  frame_size: num_data_sets * factor_upchan * num_frb_total_beams * 2 * sizeof_float
```
andrerenard, line 354: Do the weights fully double the size of the data? (cc: @jrs65 )  That's quite a bit more storage than we'd estimated needing.

andrerenard (commented):
See inline.

config/chime_science_run_gpu.yaml
jrs65: For these buffers (from GPU -> accumulate stage) I don't think you need to increase the frame size.

As for the other buffers, I think for the `hfb` buffers you don't need any `frame_size` parameter at all as it's automatically calculated from the other config parameters.

james-s-willis: When I remove the `frame_size` from:
```
cpu_hfb_output_buffers:
    num_frames: buffer_depth
    #frame_size: num_data_sets * factor_upchan * num_frb_total_beams * sizeof_float
    metadata_pool: main_pool
    cpu_hfb_output_buffer_0:
        kotekan_buffer: standard
    beamform_hfb_output_buffer_0:
        metadata_pool: hfb_pool
        kotekan_buffer: standard
    beamform_hfb_output_buffer_read:
        metadata_pool: hfb_pool
        kotekan_buffer: standard
```
I get:
```
Failed to start kotekan with config file ../../config/tests/chime_frb_verification.yaml, error message: The config option: frame_size is required, but was not found in the path: /cpu_hfb_output_buffers/beamform_hfb_output_buffer_0
```

andrerenard: I think that's because you are using the generic buffer (`kotekan_buffer: standard`)

jrs65: As I said, you'll need to keep it for the GPU->accumulate buffers as they are of type `standard`. For those of type `hfb` you should be able to remove it I think.

james-s-willis:
I've added those changes and it's now working

andrerenard (approved):
This looks good, but we should keep an eye on the system performance with the reduced global buffer buffer depth.

jrs65:
Why does the global buffer depth need to be reduced? Can't we just leave it as it is? Seems like we're asking for unexpected issues with that one. I can't imagine that the new HFB buffers take enough memory we need to reduce that one.

andrerenard:
We have basically been living exactly on the edge of having enough memory for some time.  Having said that I'm also nervous to adjust the global depth.  It might be worth seeing if we can do a more targeted approach to reducing buffers which we think are least likely to back up.

jrs65:
I agree that we're close to the edge, but I don't see what's changed between this release and the one we're already running. Each HFB frame *with weights* is about 8 KB in size (4 bytes * 1024 beams * 2), so it's not like we're really adding any extra memory consumption here. Is there something else we've added that's using a lot of memory? @james-s-willis have you had any trouble memory wise in your tests?

Like you @andrerenard I'm a little quite nervous to change the global depth. It could cause us a bunch of problems and isn't the easiest to debug exactly where it's going wrong.

jrs65:
I guess to put that in context somewhat, with the weights memory usage has gone up 24 MB by my estimation, that's the size of one visibility frame. By turning the global buffer depth down to 10, you've reduced the total number of visibility frames by ~40, so reducing memory usage by nearly 1 GB. It's quite a large over correction to do in my opinion, and if it's really necessary (i.e. we were within 24 MB of the limit), you only need to reduce by a single visibility frame elsewhere to make it fit.

james-s-willis:
So as far as I know the HFB frame with weights is like you say Richard: 1 MB in size (4 bytes * 1024 beams * 128 sub freq * 2). When I run on the test node with a `buffer_depth` of `12` it always crashes with the usual buffer overflow error:

```
/monitor_2: The fraction of full frames 0.875000 (24/21) is greater than the threadhold 0.800000 for buffer: lost_samples_buffer
```

That's why I reduced the `buffer_depth` to `10`. I can't think of anything else that has increased the memory usage.

james-s-willis:
A `buffer_depth` of `11` seems fine too.

james-s-willis:
Are there any other buffers that I could reduce before lowering the `buffer_depth`?

james-s-willis:
I've reduced these buffers by one frame:
```
vis_buffers_new:
  metadata_pool: vis_pool
  num_frames: 11
  visbuf_5s_0:
    kotekan_buffer: vis
  visbuf_5s_1:
    kotekan_buffer: vis
  visbuf_5s_2:
    kotekan_buffer: vis
  visbuf_5s_3:
    kotekan_buffer: vis
  visbuf_5s_merge:
    kotekan_buffer: vis
  visbuf_10s_0:
    kotekan_buffer: vis
  visbuf_10s_1:
    kotekan_buffer: vis
  visbuf_10s_2:
    kotekan_buffer: vis
  visbuf_10s_3:
    kotekan_buffer: vis
```
and it runs fine. Is that okay to change?

jrs65:
@james-s-willis  I don't see why the length of unrelated buffers would cause the lost samples buffer to backup. Do you have a theory as to how that can happen? @andrerenard do you have any ideas?

james-s-willis:
@jrs65 the buffer error is just a side-effect of running out of memory. I saw the same problem before when I increased the number of tracking beams and thought it was over kernel utilisation but in fact `kotekan` was running out of memory.

jrs65:
Okay. So is your explanation that it doesn't explicitly get OOM killed, or have any failed allocations, but because they system is low on memory any allocations are taking so long the system backs up?

If that's true you should be able to see the system working hard at reclaiming pages, and would be a good way to confirm your explanation. You should be able to make some plots of it via grafana, as node exporter will export those metrics. I'll see you a link via slack.

andrerenard:
It's also possible the ROCm driver is failing without generating an error when it runs out of memory for internal logic - it wouldn't surprise me too much if that's what's happening.   But still worth checking the system metrics.

andrerenard:
@james-s-willis could you update this PR with the change to the buffer depth in the N2 pipeline?   I'd like to start deploying things this afternoon as soon as JF is done.  And I'm not sure things will run without this change? 
@jrs65 I'm not sure we understand the exact mechanism of failure, but we've seen this exact failure with low memory in multiple contexts (testing the extra beams, and also when the GPU nodes had too large of a journal file in RAM).  James is going to check the vmstats once the F-engine is back, but I'm quite sure this failure is from a low or out of memory situation that doesn't manage to crash things directly.   Also we know that if we try to allocate more memory than this, we do actually crash kotekan on buffer allocation.

jrs65 (approved):

PR #834, opened 2020-10-02
Roll HFB data

james-s-willis:


config/tests/chime_frb_verification.yaml
```diff
@@ -138,6 +138,11 @@ cpu_hfb_output_buffers:
     metadata_pool: main_pool
     cpu_hfb_output_buffer_0:
         kotekan_buffer: standard
+
+beamform_hfb_output_buffers:
+    num_frames: buffer_depth
+    frame_size: num_data_sets * factor_upchan * num_frb_total_beams * 2 * sizeof_float
+    metadata_pool: main_pool
```
andrerenard, line 145: I think this got into the PR by mistake?

andrerenard (approved):
This looks good to me, subject the one inline comment, but if possible it would be good to test on real data quickly to make sure it looks right.

andrerenard (approved):

PR #835, opened 2020-10-05
build(cmake): hsa kernels not generated: variable partly renamed

nritsche:
A partly rename of the variable inCL_FILEs by cmake-format broke this
cmake file.


andrerenard (approved):
I'm going to test this quickly on site before merging.

PR #837, opened 2020-10-06
build(cmake): set default build type: Debug

nritsche:


CMakeLists.txt
```diff
@@ -192,6 +192,18 @@ if(CMAKE_BUILD_TYPE MATCHES Debug OR CMAKE_BUILD_TYPE MATCHES Test)
...
+# Set a default build type if none was specified
+set(DEFAULT_BUILD_TYPE "Debug")
+if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+    message(STATUS "Setting build type to '${DEFAULT_BUILD_TYPE}' as none was specified.")
+    set(CMAKE_BUILD_TYPE
+        "${DEFAULT_BUILD_TYPE}"
+        CACHE STRING "Choose the type of build." FORCE)
+    # Set the possible values of build type for cmake-gui
+    set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release" "MinSizeRel"
+                                                 "RelWithDebInfo" "Test")
```
andrerenard, line 204: Should we not set this regardless of the outcome of the IF above?

andrerenard (changes requested):

CMakeLists.txt
andrerenard: And also I don't think we have `MinSizeRel` and `RelWithDebInfo` as options, so we shouldn't include them here.

andrerenard:
I've checked on site, and it does build `Debug` correctly if no type is given.

jrs65 (approved):

andrerenard (approved):

andrerenard (approved):

jrs65 (approved):

PR #841, opened 2020-10-07
Fix pulsar header frequency encoding

andrerenard:


emmanuelfonseca (approved):

emmanuelfonseca:
looks right to me, thanks for the quick fix @andrerenard. i'll rework this for the next round; we'll likely do away with the `timesamples_per_pulsar_packet == 3125` format, but i'll confirm this with cherry first.

cubranic (commented):
👍

PR #842, opened 2020-10-07
Change baseband dump parameters: duration limit and DM uncertainty scaling factor (closes #832)

### LINKED ISSUE #832: Change baseband dump parameters: DM uncertainty scaling factor
In a discussion of recent baseband dumps, @kiyo-masui pointed out that we currently overscale for DM uncertainty, by multiplying it by 3. Specifically [basebandApiManager:179](https://github.com/kotekan/kotekan/blob/021855bc16cfe9b1aac71317af18a433c0d84758/lib/core/basebandApiManager.cpp#L179) calculates the frequency and DM-dependent time delay as: `K_DM * (dm +/- N_DM_ERROR_TOL * dm_error) * freq_inv_sq_diff`. Here, N_DM_ERROR_TOL is 3 (see [basebandApiManager.hpp:197](https://github.com/kotekan/kotekan/blob/021855bc16cfe9b1aac71317af18a433c0d84758/lib/core/basebandApiManager.hpp#L197). Kiyo describes the intent of this code was to use 3x sigma, however the value we get from L1 does not translate that way and should be used as a straightforward range.

@andrerenard we'd like to get this change in for the deployment week. It is low risk because it makes for a more predictable per-frequency response (for us) and even if it turns out not to be optimal, we can fine-tune the request parameters more easily in the FRB callback initiator and won't need to revert the code in Kotekan after the run has started.

Blocked by #849: ~@leungcalvin noticed that currently baseband dump duration seems capped at 512 ms. That didn't match my recollection of 1s but examining the config file confirmed his hypothesis: `max_dump_samples` is currently set to 200000, which works out to 512 ms. We'd like to increase it back to 400000, i.e., 1 second.~

cubranic:
@leungcalvin noticed that currently baseband dump duration seems capped at 512 ms. That didn't match my recollection of 1s but examining the config file confirmed his hypothesis: max_dump_samples is currently set to 200000, which works out to 512 ms. We'd like to increase it back to 400000, i.e., 1 second.

In a discussion of recent baseband dumps, @kiyo-masui pointed out that we currently overscale for DM uncertainty, by multiplying it by 3. Specifically basebandApiManager:179 calculates the frequency and DM-dependent time delay as: K_DM * (dm +/- N_DM_ERROR_TOL * dm_error) * freq_inv_sq_diff. Here, N_DM_ERROR_TOL is 3 (see basebandApiManager.hpp:197. Kiyo describes the intent of this code was to use 3x sigma, however the value we get from L1 does not translate that way and should be used as a straightforward range.

closes #832

### DIFF
--- a/config/chime_science_run_gpu.yaml
+++ b/config/chime_science_run_gpu.yaml
@@ -643,7 +643,7 @@ zero_samples:
 
 # Baseband capture and dump system
 baseband:
-  max_dump_samples: 200000
+  max_dump_samples: 400000
   num_frames_buffer: baseband_buffer_depth - 10
   base_dir: /mnt/frb-baseband/
   write_throttle: 0
--- a/config/chime_science_run_gpu.yaml
+++ b/config/chime_science_run_gpu.yaml
@@ -643,7 +643,7 @@ zero_samples:
 
 # Baseband capture and dump system
 baseband:
-  max_dump_samples: 400000
+  max_dump_samples: 200000
   num_frames_buffer: baseband_buffer_depth - 10
   base_dir: /mnt/frb-baseband/
   write_throttle: 0
--- a/config/chime_science_run_gpu_no_output.yaml
+++ b/config/chime_science_run_gpu_no_output.yaml
@@ -643,7 +643,7 @@ zero_samples:
 
 # Baseband capture and dump system
 baseband:
-  max_dump_samples: 200000
+  max_dump_samples: 400000
   num_frames_buffer: baseband_buffer_depth - 10
   base_dir: /mnt/frb-baseband/
   write_throttle: 0
--- a/config/chime_science_run_gpu_no_output.yaml
+++ b/config/chime_science_run_gpu_no_output.yaml
@@ -643,7 +643,7 @@ zero_samples:
 
 # Baseband capture and dump system
 baseband:
-  max_dump_samples: 400000
+  max_dump_samples: 200000
   num_frames_buffer: baseband_buffer_depth - 10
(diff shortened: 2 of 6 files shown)


cubranic:
@andrerenard we'd like to get these two changes in for the deployment week. They're low risk because they make for a more predictable per-frequency response (for us) and even if they turn out not to be optimal, we can fine-tune those parameters more easily in the FRB callback initiator and won't need to revert the config or code in Kotekan after the run has started.

andrerenard (approved):
Does this increase in max dump size increase the memory usage?   If so we should test this on the test node to make sure we still have enough free memory to run stably.

cubranic:
Ah, I think you're right. Let me verify.

cubranic:
Now that you said it, that may be why it was lowered to 500ms from the original 1s.

cubranic:
Yep, scratch increasing the maximum dump duration. But removing the DM_e scaling by 3 should stay.

andrerenard:
I should note.  We are planning to do an optimization of the buffer depths at some point to make room for the extra beams - and in the process if we free up extra RAM we might be able to increase this size later if it's important.   How much memory are we talking about with this change?

kiyo-masui:
I think that change would require an extra 3.2 GB. The total baseband rate per node is 1 byte * 2048 inputs /2.56e-6 * 4 frequencies = 3.2GB/s. I think the staging area requires 2 * max dump size of RAM, so increasing it by 500ms requires an extra second of data.

I had forgotten about the memory cost here. Good catch. If we can optimize the memory I would prefer to increase the length of the total buffer than increasing the max dump size.

PR #843, opened 2020-10-07
Fix unintentional rounding when calculating fpga0 timestamp

cubranic:
Multiplying by `1e9` causes a silent cast to float, and then back to an int,
causing the rounding. This fix makes sure we multiply by an integer constant.

### DIFF
--- a/lib/stages/frbPostProcess.cpp
+++ b/lib/stages/frbPostProcess.cpp
@@ -131,7 +131,7 @@ void frbPostProcess::main_thread() {
     frb_header.data_nbytes = udp_packet_size - udp_header_size;
     frb_header.fpga_counts_per_sample = fpga_counts_per_sample;
     const auto fpga0_ns = tel.to_time(0);
-    frb_header.fpga0_ns = fpga0_ns.tv_sec * 1e9 + fpga0_ns.tv_nsec;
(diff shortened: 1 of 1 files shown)


cubranic:
Example code to reproduce the rounding that we saw in received L1 data:
```
// -*- flycheck-clang-language-standard: "c++17"; -*-
#include <iostream>

int main() {
    uint64_t fpga0_nano = 1601931475999999960;

    uint64_t tv_sec = 1601931475;
    uint64_t tv_nsec = 999999960;
    std::cout << "nano: " << fpga0_nano << "\n";
    std::cout << "tv_sec: " << tv_sec
              << ", tv_nsec: " << tv_nsec
              << "\n";

    uint64_t fpga0_ns = tv_sec * 1e9 + tv_nsec;
    std::cout << "fpga_ns when using 1e9: " << fpga0_ns << "\n";

    fpga0_ns = tv_sec * 1'000'000'000 + tv_nsec;
    std::cout << "fpga_ns using a long integer: " << fpga0_ns << "\n";
}
```

Output:
```
nano: 1601931475999999960
tv_sec: 1601931475, tv_nsec: 999999960
fpga_ns: 1601931476000000000
fpga_ns: 1601931475999999960
```

andrerenard (approved):
Good find!   That's quite a subtle error, we should check other places in the code that work with time we haven't made this same mistake...

PR #844, opened 2020-10-07
Increase hfbbuf_10s no. of frames to 2048 to deal with dropped frames.

james-s-willis:


andrerenard (changes requested):
You need to also increase `num_metadata_objects` to be greater than the total number of all frames, otherwise you could run out, which will trigger a shutdown.

config/chime_science_run_recv_hfb.yaml
```diff
@@ -37,7 +37,7 @@ hfb_pool:
 
 hfbbuf_10s:
   kotekan_buffer: hfb
-  num_frames: buffer_depth * 6
+  num_frames: buffer_depth * 32
```
andrerenard, line 40: I would just give this is 2048 directly.   In case we have more components and want to change their buffer depths.   This receiver buffer is a special case.

andrerenard (approved):

PR #846, opened 2020-10-07
redefinition of station_id attribute for pulsar header

emmanuelfonseca:
@andrerenard i made a small change to propagate node-ID info down to the backend, which i forgot to make beforehand. it's been confirmed that the backend get's the right frequency labels, but this new `station_id` has been intended to be used for buffering data on the backend during an observation; again, i just forgot to make this change, my mistake.

### DIFF
--- a/lib/stages/pulsarPostProcess.cpp
+++ b/lib/stages/pulsarPostProcess.cpp
@@ -167,11 +167,12 @@ void pulsarPostProcess::main_thread() {
         psr_header.log_num_chan = 3; // ln8
     }
     psr_header.vdif_version = 1;
-    psr_header.station_id = (uint16_t)(stream_id.crate_id * 16
-        + stream_id.slot_id + stream_id.link_id * 32); // effectively a GPU-node identifier.
-    psr_header.thread_id = 0;  // index of first packed frequency.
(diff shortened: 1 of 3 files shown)


lib/stages/pulsarPostProcess.cpp
```diff
@@ -149,6 +149,7 @@ void pulsarPostProcess::main_thread() {
     }
     uint64_t frame_fpga_seq_num = 0;     // sample starting the current input frame
     uint32_t current_input_location = 0; // goes from 0 to _samples_per_data_set
+    ice_stream_id_t stream_id = ice_get_stream_id_t(in_buf[0], in_buffer_ID[0]);
```
andrerenard, line 152: This needs to happen after line 193 the `sync_input_buffers()` call.  Before that call `in_buf[0]` might not have any frames with the metadata (because it hasn't made it to this part of the pipeline yet).

andrerenard (changes requested):

lib/stages/pulsarPostProcess.cpp
emmanuelfonseca: thanks for catching that @andrerenard , made that change and pushed.

andrerenard (approved):

PR #850, opened 2020-10-08
Add beam label to kotekan_gains_last_update_timestamp.

james-s-willis:


andrerenard (changes requested):
I think you need to name that label in the constructor.  And then also consider what happens in the FRB case.

cubranic:
This is not going to work as-is. The `gains_last_update_timestamp_metric` is defined with a single label, "type":
```
gains_last_update_timestamp_metric(kotekan::prometheus::Metrics::instance().add_gauge(
        "kotekan_gains_last_update_timestamp", unique_name, {"type"}));
```

When you call `gains_last_update_timestamp_metric.labels({...})`, you give it the values of the labels in order they were defined. You can't add another label that wasn't a part of the definition.

Instead, we have two options:

1. We create two separate metrics, one for FRB timestamp and one for tracking beams. The tracking beams one would have "beam" as a label. (So no more "type".)
2. Alternatively, declare `gains_last_update_timestamp_metric` as:
    ```
    gains_last_update_timestamp_metric(kotekan::prometheus::Metrics::instance().add_gauge(
        "kotekan_gains_last_update_timestamp", unique_name, {"type", "beam"}));
    ```
    and `.set()` it with `.labels({"frb", ""}).set(start_time)` for "frb" and with `.labels({"tracking", beam_label}).set(start_time)` for tracking beams.

I'm not sure which is better. I suppose the second is minimally disruptive, as existing queries using `type="frb"` will work just fine when there is a new empty label "type". It feels a bit like a hack, though.

cubranic (changes requested):
See my comment in https://github.com/kotekan/kotekan/pull/850#issuecomment-705719610

james-s-willis:
@cubranic I think that's what I have now

cubranic (approved):

cubranic:
That works @james-s-willis. I started Kotekan on the test node and ran a couple of queries against `kotekan_gains_last_update_timestamp` filtering and aggregating by `type`, and had no issues.

andrerenard (approved):

PR #851, opened 2020-10-08
Revert blaze version to v3.4

andrerenard:
When running on site with blaze version `v3.8-12-gf36b481f2` we discovered a regression in the performance of the eigenvalue decomposition.   This change reverts to the `v3.4` API and locks the version in the CI server. 

Also locks `h5py` to `2.9.0` since that broke while I was doing this PR.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -262,7 +262,7 @@ jobs:
     - name: Clone Blaze and HighFive
       run: |
         git clone https://bitbucket.org/blaze-lib/blaze.git blaze
-        cd blaze && git pull && cd ..
+        cd blaze && git checkout v3.4 && cd ..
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -247,7 +247,7 @@ jobs:
 
     - name: Build h5py from source
       run: |
-        git clone https://github.com/h5py/h5py.git h5py && cd h5py && git pull
+        git clone https://github.com/h5py/h5py.git h5py && cd h5py && git checkout 2.9.0
```
nritsche, line 250: We should also freeze the version of h5py in the Dockerfile. Otherwise it will fail next time that step is not cached.

andrerenard: Good catch, I've just done that.

jrs65 (approved):

nritsche (approved):

PR #854, opened 2020-10-09
Print iwyu diff and run on PRs

nritsche:


andrerenard (approved):
This works nicely!   Getting the diff that is.  This latest change we still need to see about.

nritsche:
It printed the diff, but then it exited with success...

andrerenard:
You could save the return code from iwyu and return that with `exit`.

andrerenard:
You might need to do something like `(exit $exit_code)` to generate the return code the way github wants it.

andrerenard:
I'm not a big fan of that auto commit.  That means we'll then have a broken commit in develop when iwyu does something wrong, and because of our branch rules means we'll still have to fix it in another PR branch.  I would just dump the diff, and then we can clean up develop in a PR manually.

PR #860, opened 2020-10-13
config: change noise source inputs

jrs65:


andrerenard (approved):
This looks good to me, as long as we are really sure we want to include `32` now.   i.e. do we think it has a good FPGA gain, etc. if we've been re-cabling things?

jrs65:
Actually the change does the opposite, it removes 32 from the config. Niko has just reconnected it to 256.

I'm not sure we've done a gain update yet.

andrerenard:
> Actually the change does the opposite, it removes 32 from the config. Niko has just reconnected it to 256.

By include here, I meant remove from the exclude list - and treat as a normal working input. 

> I'm not sure we've done a gain update yet.

We might not have to, but @ssiegelx would it make sense to do gains updates on some of these re-cabled ADCs today?

ssiegelx:
We have not calculated digital gains yet. I would like to calculate them sometime today for the new noise source inputs only.  This can be done with the F-engine running, it does not require an `fpga_master` restart.  But we do stop `kotekan-receiver` so that we are not acquiring data during a digital gain change.

PR #861, opened 2020-10-13
Fix iwyu 2020.10

nritsche:
Shaves off 2min from the CI run by skipping the docker container builds if the files didn't change.


nritsche:
Reopened as #862 against `develop`

PR #862, opened 2020-10-13
Skip docker builds if Dockerfiles didn't change

nritsche:
2min CI speedup if the Dockerfile's are unchanged

Attempts to rebuild the docker files only if they changed.
Uploads changed dockerfiles only if we are on `develop`.
-> A branch that makes changes to a Dockerfile doesn't get cached containers and always has to base it's docker builds on what is in develop.


jrs65:
@nritsche can you explain a little more the way you intend this to work. I like the idea of both shortening the build time, and of fixing the Docker image conflicts we occasionally get.

My understanding from reading your changes is that it doesn't pull or build the docker image if the Dockerfile hasn't been changed, which seems fine, but then it only pushes to the docker image cache if the branch is develop.

What I don't see is how you would ever test changes which relied on changing the Dockerfile (which definitely happens, although maybe less often now)? I think you mention that in your comments. Can you suggest a good way to actually do that?

nritsche:
> @nritsche can you explain a little more the way you intend this to work. I like the idea of both shortening the build time, and of fixing the Docker image conflicts we occasionally get.
> 
> My understanding from reading your changes is that it doesn't pull or build the docker image if the Dockerfile hasn't been changed, which seems fine, but then it only pushes to the docker image cache if the branch is develop.
> 
> What I don't see is how you would ever test changes which relied on changing the Dockerfile (which definitely happens, although maybe less often now)? I think you mention that in your comments. Can you suggest a good way to actually do that?

Oh you're right. I missed that we're pulling the containers again for every step. Maybe we can upload a version of the container for every branch.

nritsche:
Now this will tag the docker containers with the branch name (so far we tagged everything with `latest`).

If the Dockerfile's don't change, `develop` will be used (at the moment it falls back to `latest`, we will have to change the fall-back once a `develop` container is pushed).

If a Dockerfile changes, we push the result with the tag of the branch, but when initially pulling, we keep falling back to `develop` (`latest` at the moment).

nritsche:
TODO after merge:
Replace `latest` with `develop` in:
- main.yaml github action
- iwyu Dockerfile

nritsche:
@andrerenard Let's merge this *now*, so we can sort out any problems with it *before* the next deployment week, when we want PRs to go through fast.

tools/iwyu/docker/Dockerfile
```diff
@@ -1,9 +1,10 @@
+# TODO: change "latest" to "develop"
```
andrerenard, line 1: Does this need to be changed before merging?

andrerenard (approved):

andrerenard:
Looks like the python tests failed, I'm going to rerun the CI.

jrs65 (approved):
I'm going to approve this now, it looks like it solves all the problems I saw last time. The workflow file is getting increasingly complicated, we might consider how to refactor it. It might be possible to push all the docker building stuff into a custom action I guess.

tools/iwyu/docker/Dockerfile
nritsche: It needs to get changed *after* merging...

nritsche:
> Looks like the python tests failed, I'm going to rerun the CI.

It failed because caput can't find `skyfield`. I have no idea how that would suddenly happen, since skyfield is in caput's requirements file. Any idea, @jrs65 ?

jrs65:
@nritsche this seems to be the problem. We don't actually use the requirements.txt file to the install requires (https://github.com/radiocosmology/caput/blob/master/setup.py#L22).

jrs65:
Also, it used to be that we protected the Skyfield imports as it was optional, now it's basically mandatory, but it seems like the requirements that it actually uses don't reflect this. I can quickly try and fix this in caput as I'm doing stuff there anyway.

nritsche:
@andrerenard @jrs65 Can someone give me permission to rebase-merge this?

PR #865, opened 2020-10-16
Calculate the HFB weights

james-s-willis:


james-s-willis:
@jrs65 I've also tidied the code up to use the `HFBFrameview`.

lib/stages/HFBAccumulate.cpp
```diff
@@ -97,64 +97,60 @@ HFBAccumulate::HFBAccumulate(Config& config_, const std::string& unique_name,
...
+    total_lost_timesamples += in_frame.fpga_seq_start - fpga_seq_num;
+    fpga_seq_num = in_frame.fpga_seq_start;
 
     // Integrates data from the input buffer to the output buffer.
-    for (uint beam = 0; beam < _num_frb_total_beams; beam++) {
-        for (uint freq = 0; freq < _factor_upchan; freq++) {
-            sum_data[beam * _factor_upchan + freq] += input_data[beam * _factor_upchan + freq];
-        }
+    for (uint32_t i = 0; i < _num_frb_total_beams * _factor_upchan; i++) {
+        sum_data[i] += in_frame.hfb.data()[i];
```
james-s-willis, line 122: @jrs65 I had to use `in_frame.hfb.data()[i]` instead of just `in_frame.hfb[i]` as it seg faults otherwise. Any idea what could cause this?

lib/stages/HFBAccumulate.cpp
```diff
@@ -219,17 +212,49 @@ void HFBAccumulate::main_thread() {
...
+        // ... every odd sample we accumulate the squared differences into the weight dataset
+        // NOTE: this incrementally calculates the variance, but eventually
+        // output_frame.weight will hold the *inverse* variance
+        // TODO: we might need to account for packet loss in here too, but it
+        // would require some awkward rescalings
+        else {
+            for (size_t i = 0; i < _num_frb_total_beams * _factor_upchan; i++) {
+                // NOTE: avoid using the slow std::complex routines in here
+                //INFO("hfb_even[{}]: {}", i, hfb_even[i]);
+                float d = input[i] - hfb_even[i];
```
jrs65, line 245: If I had to pick one place that I thought was responsible for the incorrect weights, it is here. Both the array `input` and vector `hfb_even` are `int32_t` typed even though the underlying data is floating point. This subtraction operation will do an integer subtraction, which is then converted into a float. That will give the wrong answer, you need to do a floating point subtraction initially.

Ultimately the issue here is that none of the data is `int32_t`, if you type all the arrays and vectors appropriately this issue should disappear.

jrs65 (changes requested):

lib/stages/HFBAccumulate.cpp
```diff
@@ -163,52 +159,49 @@ void HFBAccumulate::main_thread() {
...
+            mark_frame_empty(cls_buf, unique_name.c_str(), cls_frame_id++);
             continue;
         }
         DEBUG2("Frames are synced. HFB frame: {}; Compressed Lost Samples frame: {}, diff {}",
                hfb_seq_num, cls_seq_num, hfb_seq_num - cls_seq_num);
 
-        float* sum_data = (float*)out_buf->frames[out_buffer_ID];
-        float* input_data = (float*)in_buf->frames[in_buffer_ID];
+        auto in_frame = HFBFrameView(in_buf, in_frame_id);
+        float* sum_data = (float*)out_buf->frames[out_frame_id];
```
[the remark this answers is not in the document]
james-s-willis: Ah of course, I forgot the input frames are from the GPU. And I thought there was a reason you took the data as `int32_t` in `visAccumulate`, but now I know to just take the data as a `float` in `HFBAccumulate`.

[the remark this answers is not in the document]
andrerenard: The data from the N2 kernel is an `int32_t`, but in the case of the beams they are `float`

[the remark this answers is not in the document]
james-s-willis: Ah that's why, thanks for the clarification.

andrerenard:
What's left to do on this before we can merge?

jrs65:
I think I'm meant to do a last review. I'll do that now.

lib/stages/HFBAccumulate.hpp
```diff
@@ -60,16 +64,55 @@ class HFBAccumulate : public kotekan::Stage {
...
-    void normalise_frame(float* sum_data, const uint32_t in_buffer_ID);
+    void normalise_frame(float* sum_data, const uint32_t in_frame_id);
+
+    // NOTE: Annoyingly this can't be forward declared, and defined fully externally
+    // as the std::deque needs the complete type
+    /**
+     * @class internalState
+     * @brief Hold the internal state of a gated accumulation.
+     **/
+    struct internalState {
```
jrs65, line 82: Given that we probably won't ever have gated HFB acq's you don't really need the internalState. You could just put all of it's state on the main class. Not one to do now, but probably worth noting for future cleanups.

lib/stages/HFBAccumulate.cpp
```diff
@@ -241,50 +274,65 @@ void HFBAccumulate::main_thread() {
...
-                set_fpga_seq_start_hfb(out_buf, out_buffer_ID, fpga_seq);
 
-                // Set GPS time
-                set_ctime_hfb(out_buf, out_buffer_ID, tel.to_time(fpga_seq));
+                frame.fpga_seq_start = fpga_seq;
+                frame.time = tel.to_time(fpga_seq);
+                frame.fpga_seq_total = total_timesamples - total_lost_timesamples;
+                frame.fpga_seq_length = total_timesamples;
+                frame.freq_id = tel.to_freq_id(in_buf, in_frame_id);
+                frame.dataset_id = ds_id;
```
jrs65, line 296: Note that after Rick's 2nd stage RFI changes the incoming data will have a dataset_id which needs to be used. I couldn't put a comment in the relevant in the place, but this means that the code which generates the dataset_id needs to be deferred until it receives data and can actually use that as the base for the new dataset id.

jrs65 (changes requested):
Nearly there. A few small cleanups to request, and one slightly bigger one to make sure your changes work with Rick's new changes.

andrerenard:
I think the conflicts with `develop` are blocking this from running the CI.   Can you update with `develop` and see if that helps?

andrerenard:
It looks like this is failing on the chimeMetadata includes.  We moved it from `.h` to `.hpp`

lib/stages/HFBAccumulate.cpp
james-s-willis: I think I have addressed this in my latest commit.

lib/stages/FreqSubset.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+#include <future>   // for future
+#include <map>      // for map
+#include <stdint.h> // for uint32_t
+#include <string>   // for string
+#include <utility>  // for pair
+#include <vector>   // for vector
+
+/**
+ * @class FreqSubset
```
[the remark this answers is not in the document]
james-s-willis: Can this be deferred?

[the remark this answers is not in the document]
jrs65: Definitely.

lib/stages/HFBAccumulate.hpp
```diff
@@ -27,6 +30,8 @@
  * Note: _num_frames_to_integrate cannot go below 16 frames as _num_frames_to_integrate cannot be
  * lower than the max_frames_missing
  *
+ * This stage will also calculate the within sample variance for weights.
+ *
  * The output of this stage is written to a raw file where each chunk is
  * the metadata followed by the frame and indexed by frequency ID.
  * This raw file is then transposed and compressed into a structured
```
jrs65, line 37: The docstring here doesn't list the cls buffer as an input.

jrs65 (approved):
@james-s-willis I have one comment about a missing docstring. Otherwise this is done. Nice work! 

I've clicked approve, please do fix the docstring, but no need to come back to me for another review afterwards.

lib/stages/HFBAccumulate.hpp
james-s-willis: Thanks Richard! :+1:

PR #868, opened 2020-10-22
ci(docker): change Dockerfiles to trigger rebuild

nritsche:
This should trigger the rebuild and upload of a "develop"-tagged docker container of each type.


andrerenard (approved):
Hopefully it doesn't ignore comments, and this just works.

PR #869, opened 2020-10-22
ci(docker): Pin apt versions to minor only

### LINKED ISSUE #864: Dockerfiles: use wildcard to pin apt versions
Use wildcard (`packetname=3.2.*`) to "pin" versions of packages installed with apt to a minor version.

nritsche:
Use asterix (*) to pin the versions of packages installed with apt to
the minor version only (allowing patch updates).
Where the version string had no patch version, the packet was pinned to
the major version, allowing minor updates.

Closes #864


jrs65 (approved):
LGTM and see if it fixes the new build failures!

nritsche:
I don't know whose fault it is, but it seems we can only get version `3.6Beta-14-g0c40e05-rocm-rel-3.9-17` of `rocm-opencl` now.
@andre is that okay or should we look for a way to get an older version (we had `2.0.*`)? This is where it gets it from at the moment:
`http://repo.radeon.com/rocm/apt/debian xenial/main amd64 Packages`

jrs65:
@nritsche @andrerenard it looks like we're going to keep getting test failures until this one gets merged. @andrerenard did you figure out if the new version of rocm was going to be okay?

andrerenard:
I think the ROCm version should be fine.  For my dev work with other telescopes I've been using the latest versions.   For CHIME the 3.6 version was already breaking on the HSA build because of compiler changes that are unresolved.  So at the moment that isn't building anyway.

tools/docker/Dockerfile
```diff
@@ -100,33 +100,35 @@ RUN python3.7 -m pip install msgpack==1.0.0 \
...
+    python3.7 -m pip install --no-cache-dir breathe==4.20.* \
+                                            sphinx_rtd_theme==0.5.* \
+                                            sphinxcontrib-plantuml==0.* \
                                             && \
     wget https://phoenixnap.dl.sourceforge.net/project/plantuml/plantuml.jar -P plantuml
 
 # Uncomment this to check what version got installed
 # RUN python3.7 -m pip show <package>
 # RUN apt-cache policy <package>
+RUN apt-cache policy rocm-opencl
```
andrerenard, line 130: Comment out these lines?

andrerenard (approved):

nritsche:
> I think the ROCm version should be fine. For my dev work with other telescopes I've been using the latest versions. For CHIME the 3.6 version was already breaking on the HSA build because of compiler changes that are unresolved. So at the moment that isn't building anyway.

I don't understand this comment. It's building here with 3.6, isn't it?

andrerenard:
> I don't understand this comment. It's building here with 3.6, isn't it?

We aren't building the HSA components that CHIME uses for the GPU subsystem.  Exactly because 3.6 has issues that aren't easy to resolve in a way that maintains backwards compatibility with the old version on site.

nritsche:
> > I don't understand this comment. It's building here with 3.6, isn't it?
> 
> We aren't building the HSA components that CHIME uses for the GPU subsystem. Exactly because 3.6 has issues that aren't easy to resolve in a way that maintains backwards compatibility with the old version on site.

And you're fine running 3.6 in CI even though we have an older version in production? Should we try to run that old version here? What is it?

andrerenard:
> And you're fine running 3.6 in CI even though we have an older version in production? Should we try to run that old version here? What is it?

Currently I think we just leave it has is and run with 3.6 in the CI, but without building the HSA sub components.   My impression is it's going to be nearly impossible to get things running with version 1.9 (the site version) while keeping the otherwise latest Ubuntu packages/kernel version.   I'd like to figure out how to make it build with both versions, but for now I think we keep things as is - no HSA components get built in the CI.   I know that's not ideal, but we aren't going to resolve it for this deployment.

PR #870, opened 2020-10-30
Dataset ID fix

tristpinsm:
Here is the fix for the dataset ID changes not being recorded within a file. There is also an updated unit test that checks this issue specifically. Reverting the fix and running the unit test fails.

Later today I will try and run this version on some of the data we acquired since yesterday and confirm it has resolved the issue.

### DIFF
--- a/tests/test_transpose.py
+++ b/tests/test_transpose.py
@@ -11,6 +11,9 @@ import h5py
 import glob
 import os
 import msgpack
+import base64
+from flask import Flask, jsonify, request as flask_req
+from pytest_localserver.http import WSGIServer
 
 from kotekan import runner
 from test_compression import float_allclose
@@ -21,28 +24,7 @@ if not runner.has_hdf5():
 
 NULL_DSET_ID = b"00000000000000000000000000000000"
 
-writer_params = {
-    "num_elements": 4,
-    "num_ev": 2,
-    "cadence": 5.0,
-    "total_frames": 10,  # One extra sample to ensure we actually get 256
-    "freq": [3, 50, 777, 554],
-    "chunk_size": [2, 6, 5],
-    "mode": "test_pattern_simple",
-    "test_pattern_value": [0, 0],
-    "file_type": "hdf5fast",
-    "dataset_manager": {"use_dataset_broker": False},
-}
-
-stack_params = {
-    "num_elements": 2048,
-    "num_ev": 2,
-    "cadence": 5.0,
-    "file_length": 3,
-    "freq": [3, 50, 777, 554],
-    "chunk_size": [2, 64, 3],
-    "dataset_manager": {"use_dataset_broker": False},
-}
+freq = [3, 50, 777, 554]
 
 # frequency to set lost samples
 frac_freq = -2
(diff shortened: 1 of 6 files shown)


tests/test_transpose.py
```diff
@@ -203,7 +326,11 @@ def test_transpose(transpose):
     assert np.allclose(f_tr["flags/frac_lost"][frac_freq, :], frac_lost, rtol=1e-3)
     assert np.allclose(f_tr["flags/frac_rfi"][frac_freq, :], frac_rfi, rtol=1e-3)
     assert (f_tr["flags/dataset_id"][empty_freq, :] == NULL_DSET_ID).all()
-    assert (f_tr["flags/dataset_id"][:empty_freq, :] != NULL_DSET_ID).all()
+
+    # Check dataset ID change is present
+    # expect starting ID, null ID, and at least one gain update
+    unique_dsets = np.unique(f_tr["flags/dataset_id"][:])
+    assert len(unique_dsets) >= 3
```
jrs65, line 337: Can you change this one to `== 3`? I don't see why it would ever be different from that.

tests/test_transpose.py
```diff
@@ -21,40 +24,122 @@
...
+    "freq": freq,
     "chunk_size": [2, 64, 3],
     "dataset_manager": {"use_dataset_broker": False},
 }
 
-# frequency to set lost samples
-frac_freq = -2
-frac_lost = 0.8
-frac_rfi = 0.3
+# Simulate a basic cal broker
```
jrs65, line 81: This, and a bunch of the code below looks like it's just been copied out of `test_applygains.py` can you factor it into a common fixture they both use.

jrs65 (changes requested):
@tristpinsm thanks for putting this one together so quickly. I've been over the logic a few times, and I think it seems pretty sound now. Thanks! A few suggested changes to the testing, but nothing major.

tests/test_transpose.py
tristpinsm: at first I wasn't sure if we wanted each test to be more or less independent, but if this is preferable I can import it from `test_applygains`

tristpinsm:
OK, I've tested the changes on a file from yesterday (`wind:/data/transposed/20201014T233944Z_chimestack_corr/01292791_0000.h5.h5`) and the transposed file now contains 15 unique dataset IDs:
```
In [7]: np.unique(fh["flags/dataset_id"][:])                                                                          
Out[7]: 
array([b'00000000000000000000000000000000',
       b'215740e11a5db05be4c379e82bb7ec15',
       b'27ce043718295c9fddc3f5793d514cde',
       b'4192ad00095874bc0b53fc89ff6d2efa',
       b'434be0973fb1bad65999a68276bab0c4',
       b'44b0ebfc14b2d5be62e41701c7d0495b',
       b'6325c005c62c38b3391d3ebf15a78320',
       b'6c4b452da5a9ebef17adbc0294b78fb3',
       b'88968c355ad2cad245e97f6d7051f0b0',
       b'a2548c1cfe2465d1b87270d00c2a71a4',
       b'a7058dc4a5f402209b81f99e2533464b',
       b'b565eec6efbf6067ef841e183ae4a88d',
       b'c9bf2c25dd22fe72d53f01401dd47ed9',
       b'd1ddbb870ed16ae5e6411894d622a37b',
       b'fe1a187df1ab7d1ab2e488005e99ee2f'], dtype='|S33')
```

nritsche:
macOS CI fixed here: https://github.com/kotekan/kotekan/pull/871

tristpinsm:
Do we want to merge this today or deploy a test version from this branch?

jrs65:
It looks like at the moment nothing risky has been merged into develop, so we could just merge down to develop and run that. Additionally we could also merge that down to `master` as a hotfix release, although still only deploying it on wind. Although if we did that we'd need to fight with IWYU etc, and that might not be feasible to win that battle in time.

jrs65 (approved):

PR #872, opened 2020-11-04
Gossec hotfix branch

jrs65:


andrerenard:
Given where we are in the cycle, I'm going to close this and we can merge it in with the main release tagging.

PR #873, opened 2020-11-04
Correctly mask lost samples in sum_hfb kernel.

james-s-willis:
I was only checking every 4th element of `compressed_lost_samples_buf` on whether to include the sample in the sum.

### DIFF
--- a/lib/hsa/kernels/sum_hfb.cl
+++ b/lib/hsa/kernels/sum_hfb.cl
@@ -16,27 +16,36 @@ __kernel void sum_hfb(__global float *data, __constant uint *compressed_lost_sam
 
   // Sum data across samples from global memory
   float4 freq_sum_1 = (float4)(0.f, 0.f, 0.f, 0.f), freq_sum_2 = (float4)(0.f, 0.f, 0.f, 0.f);
-  float4 data_1, data_2;
 
   for(int sample=0; sample<num_samples; sample+=4) {
 
-      if(!compressed_lost_samples_buf[sample]) {
+    float4 data_1 = (float4)(0.f, 0.f, 0.f, 0.f), data_2 = (float4)(0.f, 0.f, 0.f, 0.f);
 
-          // Load data into vectors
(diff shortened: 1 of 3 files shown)


andrerenard (changes requested):
I think to avoid all these `if` statement, you could just multiply the `1 - compressed_lost_samples_buf[sample + ]` into the results.   Also that solves the other bug I'm seeing here, which is `data_1` is basically going to keep the last value not zero in the case of a lost sample.

lib/hsa/kernels/sum_hfb.cl
```diff
@@ -20,23 +20,26 @@ __kernel void sum_hfb(__global float *data, __constant uint *compressed_lost_sam
...
-          
-          data_2.s0 = data[1024*NUM_SUB_FREQS*sample + beam*NUM_SUB_FREQS + freq + 1];
-          data_2.s1 = data[1024*NUM_SUB_FREQS*(sample + 1) + beam*NUM_SUB_FREQS + freq + 1];
-          data_2.s2 = data[1024*NUM_SUB_FREQS*(sample + 2) + beam*NUM_SUB_FREQS + freq + 1];
-          data_2.s3 = data[1024*NUM_SUB_FREQS*(sample + 3) + beam*NUM_SUB_FREQS + freq + 1];
+     // Mask data from lost samples.
+     const float mask_0 = 1.f - compressed_lost_samples_buf[sample];
+     const float mask_1 = 1.f - compressed_lost_samples_buf[sample + 1];
+     const float mask_2 = 1.f - compressed_lost_samples_buf[sample + 2];
+     const float mask_3 = 1.f - compressed_lost_samples_buf[sample + 3];
```
andrerenard, line 27: I would leave this as an integer subtraction `float mask_0 = 1 - compressed_lost_samples_buf[sample]` just too avoid any float subtraction oddities - plus I think it's slightly faster to do an int op here. 

Also I don't think you can write this as a `const`, since you do update it in the loop.  (although maybe the redeclaration might allow it...)

andrerenard (changes requested):

andrerenard (approved):

PR #875, opened 2020-11-07
Add second stage RFI dataset state

nritsche:
- promote chimeMetadata to C++
- move stream_t to Telescope.hpp
- add dataset ID to GPU frames / chimeMetadata
  - rename HFBMetadata::get_dataset_id -> get_dataset_id_frb
  - rename HFBMetadata::set_dataset_id -> set_dataset_id_frb
  - add dataset ID to chimeMetadata with getters and setters
  - read fpga_dataset from config in iceBoardShuffle
  - expect dataset ID in incoming frames in visAccumulate
- add RFIFrameDrop state keeping info:
  - frame dropping enabled
  - array of thresholds and fractions

Closes https://github.com/chime-experiment/RTPipeline/issues/72


lib/stages/visAccumulate.cpp
```diff
@@ -282,6 +273,21 @@ void visAccumulate::main_thread() {
         if (in_frame == nullptr)
             break;
 
+        // Check if dataset ID changed
+        dset_id_t ds_id_in_new = get_dataset_id(in_buf, in_frame_id);
+        if (ds_id_in_new != ds_id_in) {
+            ds_id_in = ds_id_in_new;
+
+            // Register base dataset, if not dataset ID was was set in the incoming frame,
```
jrs65, line 281:
```suggestion
            // Register base dataset, if no dataset ID was was set in the incoming frame,
```

lib/stages/visAccumulate.cpp
```diff
@@ -220,35 +218,26 @@ visAccumulate::visAccumulate(Config& config, const std::string& unique_name,
...
     // create all the states
-    datasetManager& dm = datasetManager::instance();
-    std::vector<state_id_t> base_states;
-    base_states.push_back(dm.create_state<freqState>(freqs).first);
-    base_states.push_back(dm.create_state<inputState>(inputs).first);
-    base_states.push_back(dm.create_state<prodState>(prods).first);
-    base_states.push_back(dm.create_state<eigenvalueState>(0).first);
-    base_states.push_back(
+    base_dataset_states.push_back(dm.create_state<freqState>(freqs).first);
+    base_dataset_states.push_back(dm.create_state<inputState>(inputs).first);
```
jrs65, line 230: At somepoint these should be registered in the network frontend.

lib/utils/datasetState.hpp
```diff
@@ -805,4 +805,82 @@ class subfreqState : public datasetState {
...
+        j["fractions"] = fractions;
+        return j;
+    }
+
+    /// Tells if frame dropping is enabled in the RFIFrameDrop stage.
+    bool enabled;
+
+    std::vector<float> thresholds;
+    std::vector<float> fractions;
+};
```
jrs65, line 884: I think this would be better as a vector of pairs. That way you structurally guarantee that every threshold has a fraction.

jrs65: Similarly for how this is serialised. It might make sense to more directly mirror what's in the inbound config update.

lib/stages/RfiFrameDrop.hpp
```diff
@@ -90,15 +90,16 @@ class RfiFrameDrop : public kotekan::Stage {
...
-    };
-    std::shared_ptr<ThresholdData> _thresholds;
+    /// Everything that gets changed by dynamic config updates,
+    /// including the resulting dataset state ID, is protected by a mutex.
+    /// Thresholds and fractions are kept in the dataset state directly to avoid duplicates.
+    std::mutex update_mutex;
+    std::vector<size_t> num_sk;
+    std::vector<size_t> sk_exceeds;
+    state_id_t state_id;
+    const RFIFrameDropState* state_ptr;
```
jrs65, line 100: So you keep the entire state of the thresholds in here now? Well I guess along side `num_sk`.

jrs65: Do we provide any lifetime guarantees on bare pointers that come out of the dataset manager? I'm wondering if there's theoretically any issue here that could allow us to read an invalid pointer.

I mean, I don't think there's any issue at the moment, but if this is a pattern that we think should be fine, so we need to make a guarantee in the dataset manager that the target will never be free'd.

lib/stages/RfiFrameDrop.cpp
```diff
@@ -123,78 +128,90 @@ void RfiFrameDrop::main_thread() {
...
-
         // Calculate the scaling to turn kurtosis value into sigma
         size_t num_inputs = num_elements - metadata_vis->rfi_num_bad_inputs;
         float sigma_scale = sqrt((num_inputs * (sk_step - 1) * (sk_step + 2) * (sk_step + 3))
                                  / (4.0 * sk_step * sk_step));
 
-        for (size_t ii = 0; ii < num_sub_frames; ii++) {
+        {
+            // Lock update mutex to not allow updates being processed during this critical section
+            std::lock_guard<std::mutex> lock(update_mutex);
```
jrs65, line 138: The lifetime of this lock seems to be longer than required. I think you only need to hold it long enough to get a copy of the dataset_id and the thresholds and num_sk values.

jrs65 (changes requested):

lib/stages/RfiFrameDrop.hpp
nritsche: Yes, `num_sk` and `sk_exceeds` are not part of the state,a s we discussed.

lib/utils/datasetState.hpp
```diff
@@ -805,4 +805,69 @@ class subfreqState : public datasetState {
...
+     * @return True if RFI frame-dropping enabled.
+     */
+    bool get_enabled() const {
+        return enabled;
+    }
+
+    /**
+     * @brief Get RFI frame-dropping thresholds (read only).
+     *
+     * @return Vector of pairs: Thresholds and fractions.
```
jrs65, line 852: You might want to even more clearly than this document that they are packed as `threshold, fraction` pairs, particularly for what's in the JSON serialisation.

lib/stages/RfiFrameDrop.cpp
```diff
@@ -284,21 +316,22 @@ bool RfiFrameDrop::rest_thresholds_callback(nlohmann::json& update) {
...
+    sk_exceeds.resize(thresholds_new.size(), 0);
 
     INFO("Setting new RFI excision cuts:");
-    for (auto& [threshold, t, fraction] : thresholds_new->thresholds) {
-        (void)t;
-        INFO("  added cut with threshold={}, fraction={}", threshold, fraction);
-    }
+    for (size_t i = 0; i < thresholds_new.size(); i++)
+        INFO("  added cut with threshold={}, fraction={}", thresholds_new[i].first,
+             thresholds_new[i].second);
```
jrs65, line 330: Why don't you use the structured binding range based for that was previously used?
```suggestion
for (auto& [threshold, fraction] : thresholds_new->thresholds)
    INFO("  added cut with threshold={}, fraction={}", threshold, fraction);
```

jrs65 (changes requested):
Two small changes. Nearly there.

lib/stages/RfiFrameDrop.cpp
jrs65: Also they could be `const auto&`.

jrs65 (approved):

lib/stages/RfiFrameDrop.hpp
andrerenard: This variable never gets initialized.

jrs65: @andrerenard I think it is. `RfiFrameDrop::rest_enable_callback` is implicitly called when the callbacks are bound at the end of the constructor. @nritsche should check though.

andrerenard: That could work, but neither of those call backs initialize it as far as I can tell.  And since it's `const *` it has to be set in the constructor.   Also it segfaults when I run it.

jrs65: Ah, I see now. It's initialised in one of the callbacks (`rest_thresholds_callback`) here ( https://github.com/kotekan/kotekan/blob/a57de492b013714d26669cca2b92474f63f3209d/lib/stages/RfiFrameDrop.cpp#L333), but it's actually used in the other one (`rest_enable_callback`) first.

jrs65: Oh yeah. And that one even tries to initialise using data from the previous state.

PR #877, opened 2020-11-11
fix(RingMapMaker, RedundantStack): Improve handling of datasets.

tristpinsm:
Implement the fingerprint mechanism from VisCompression to handle
frequently changing dataset IDs without regenerating products
unnecessarily.

I haven't tested this because it's not straightforward to reproduce the environment it is meant to run in, so we should find a time during the deployment to try it out. 

In the meantime, if someone else could look through the dataset state logic that would be great, since it never fails to confuse me.

Solves chime-experiment/RTPipeline#93.

### DIFF
--- a/lib/stages/RingMapMaker.cpp
+++ b/lib/stages/RingMapMaker.cpp
@@ -1,11 +1,12 @@
 #include "RingMapMaker.hpp"
 
-#include "Hash.hpp"              // for Hash, operator!=
+#include "Hash.hpp"              // for Hash, operator!=, operator<
 #include "Stack.hpp"             // for chimeFeed
 #include "StageFactory.hpp"      // for REGISTER_KOTEKAN_STAGE, StageMakerTemplate
 #include "Telescope.hpp"         // for Telescope
-#include "datasetManager.hpp"    // for datasetManager, dset_id_t, state_id_t
-#include "kotekanLogging.hpp"    // for FATAL_ERROR, WARN
+#include "dataset.hpp"           // for dataset
+#include "datasetManager.hpp"    // for datasetManager, dset_id_t, state_id_t, fingerprint_t
+#include "kotekanLogging.hpp"    // for FATAL_ERROR, WARN, INFO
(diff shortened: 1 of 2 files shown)


lib/stages/RingMapMaker.cpp
```diff
@@ -471,17 +522,19 @@ void RedundantStack::main_thread() {
...
         if (input_frame.dataset_id != input_dset_id) {
-            WARN("Input dataset ID has changed. Regenerating stack specs.");
             input_dset_id = input_frame.dataset_id;
             change_dataset_state(input_dset_id);
         }
 
-        const auto& stack_rmap = new_stack_state_ptr->get_rstack_map();
-        const auto& old_stack_map = old_stack_state_ptr->get_stack_map();
-        auto num_stack = new_stack_state_ptr->get_num_stack();
+        // Get the corresponding stack state
```
jrs65, line 529: In the if statement above, it's probably best to change the `input_frame.dataset_id != input_dset_id` clause into one like `dset_id_map.count(input_frame.dataset_id) == 0`

jrs65 (changes requested):
@tristpinsm this looks great. Thanks! It looks like you manager to apply the standard pattern we've been using for using fingerprints to respond only to relevant state changes. On reflection I can start to see some ways that we could provide a standard implementation of the pattern, we probably should have done that already.

One very minor comment, but otherwise everything looks great.

lib/stages/RingMapMaker.cpp
tristpinsm: right, thanks for catching this!

tristpinsm:
I deployed this branch for `kotekan-ringmap` on `recv1` and it has been running since last night with no issues. It hasn't been dropping frames either, so I think the fix was successful.

jrs65 (approved):

andrerenard:
@tristpinsm Are you ok to have this merged in now?   Also note, after the first restart yesterday it's likely that ch_ansible has overwritten any changes you made to the systemd file, and we will have been running what ever was in develop.

PR #878, opened 2020-11-11
Reorder time/freq axis in pulsar packets

emmanuelfonseca:
Changes the order of the data in the pulsar packets from [time, freq, pol] to [freq, time, pol]


emmanuelfonseca:
FYI, i just noticed that i apparently/mistakenly added an extra `stream_id` def. it's honestly unclear when i did that yesterday or why (the last day or two has been a bit packed on my end), but it's removed anyway.

andrerenard (changes requested):
Sorry for the delay reviewing this.   

I don't think you need to change the stepping above.   It should be sufficient to just change were the sample goes into the packet in the output index calculation line: 

This is what I came up with for the indexing, with annotations. 
```
                                // beam->packets->[freq-time-pol]
                                out_index =
                                    // Beam
                                    psr * _udp_pulsar_packet_size * _num_packet_per_stream
                                    // Packet
                                    + frame * _udp_pulsar_packet_size
                                    // Frequency
                                    + thread_id * _timesamples_per_pulsar_packet * _num_pol
                                    // Time
                                    + in_frame_location * _num_pol + p
                                    // Header offset
                                    + udp_pulsar_header_size;
```

@cubranic and/or @cherryng would you mind checking my logic here?

emmanuelfonseca:
@andrerenard i actually have your suggested change to `out_index` in with the original PR, though now that i think about i agree that the re-ordered for loops are likely redundant. shall i revert to the original for-loop structure?

andrerenard:
@emmanuelfonseca I've asked @cherryng to take a look tonight, so let's wait for her comments.

lib/stages/pulsarPostProcess.cpp
```diff
@@ -304,11 +305,11 @@ void pulsarPostProcess::main_thread() {
                                             + (in_frame_location * _num_pol + p)
                                             + udp_pulsar_header_size;
                             } else if (_timesamples_per_pulsar_packet == 625) {
-                                // beam->packets->[time-freq-pol]
+                                // beam->packets->[freq-time-pol]
                                 out_index = psr * _udp_pulsar_packet_size * _num_packet_per_stream
                                             + frame * _udp_pulsar_packet_size
-                                            + (in_frame_location * _num_gpus * _num_pol
-                                               + thread_id * _num_pol + p)
+                                            + (thread_id * _samples_per_data_set * _num_pol
```
andrerenard, line 311: Regardless of if you just revert to only making this index change or not, I'm fairly sure you need this change: 
```suggestion
                                            + (thread_id * _timesamples_per_pulsar_packet * _num_pol
```

cherryng: I agree, this should be `_timesamples_per_pulsar_packet` because we are just counting time within a packet.  `_samples_per_data_set` is a much larger number.

lib/stages/pulsarPostProcess.cpp
```diff
@@ -268,31 +268,32 @@ void pulsarPostProcess::main_thread() {
                              beam_coord, thread_ids);
             }
 
+            // now store samples into output buffer.
+            for (uint32_t thread_id = 0; thread_id < _num_gpus; ++thread_id) {
```
cherryng, line 272: I think this for loop should go back to where it was. If it is up here, `fill_headers` will be looping once every `thread_id` (freq), which is not necessary. All the 4 freq goes into one packet.

cherryng (changes requested):

lib/stages/pulsarPostProcess.cpp
emmanuelfonseca: thanks for catching this @andrerenard @cherryng, fixed in the latest commit.

lib/stages/pulsarPostProcess.cpp
emmanuelfonseca: thanks for the clairification @cherryng. i've reverted the for-loop and incrementing calls to their original ordering. now, the only change is the indexing calculation.

andrerenard (approved):
Looks great, thanks!

PR #880, opened 2020-11-16
Fix RFIFrameDrop dataset state registration

nritsche:
If the `enable` endpoint is called befor the `threshold`-one, an uninitialized state_ptr is accessed -> segfault.
Fix: The `enable` endpoint callback uses an empty vector if the state is not available yet.

I would have preferred to create an initial dummy dataset state in the constructor, but it would have to life in the datasetManager and get registered with comet.

### DIFF
--- a/lib/stages/RfiFrameDrop.cpp
+++ b/lib/stages/RfiFrameDrop.cpp
@@ -269,8 +269,17 @@ bool RfiFrameDrop::rest_enable_callback(nlohmann::json& update) {
     std::lock_guard<std::mutex> lock(update_mutex);
 
     // set enabled and state ID, but copy the rest from the last state
-    std::tie(state_id, state_ptr) =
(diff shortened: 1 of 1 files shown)


andrerenard (commented):
Could this also fail if `rest_thresholds_callback()` gets called first and tries to call: `state_ptr->get_enabled()` ?

nritsche:
> Could this also fail if `rest_thresholds_callback()` gets called first and tries to call: `state_ptr->get_enabled()` ?

Good point.

andrerenard:
Did you want me to review this as is, or were you also going to fix the other case?   Another option might just be to generate an initial state in the constructor before calling the callbacks?

nritsche:
> Did you want me to review this as is, or were you also going to fix the other case? Another option might just be to generate an initial state in the constructor before calling the callbacks?

Oups, somehow I failed to push.
I would have preferred having a dummy state, too. But it would end up in comet and chimedb.

andrerenard (approved):

PR #881, opened 2020-11-16
fix(RfiFrameDrop): mutex can't get acquired on second frame

nritsche:
There was a typo. Changed lock.release() -> lock.unlock()
release() releases the ownership over the mutex.

### DIFF
--- a/lib/stages/RfiFrameDrop.cpp
+++ b/lib/stages/RfiFrameDrop.cpp
@@ -93,6 +93,10 @@ void RfiFrameDrop::main_thread() {
     // keep track of the dataset state
     state_id_t last_state_id = state_id_t::null;
 
+    // Lock to protect dataset state and all update data (they are all changed together in either of
+    // the endpoint callbacks)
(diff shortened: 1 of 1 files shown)


lib/stages/RfiFrameDrop.cpp
```diff
@@ -136,7 +140,7 @@ void RfiFrameDrop::main_thread() {
                                  / (4.0 * sk_step * sk_step));
 
         // Lock update mutex to not allow updates being processed during this critical section
-        std::unique_lock<std::mutex> lock(update_mutex);
+        lock.lock();
```
jrs65, line 143: Why not just put this section into a block and use a lock_guard? Seems more elegant.

jrs65 (approved):
One mild suggestion, but otherwise approved.

james-s-willis (approved):
Changes look good to me

lib/stages/RfiFrameDrop.cpp
nritsche: Variable scope. I would have to declare them outside of the block and initialize inside.

PR #882, opened 2020-11-17
Add a config which can run the pipeline without the FPGAs online.

james-s-willis:


config/chime_keep_gpu_warm.yaml
```diff
@@ -0,0 +1,3792 @@
...
+    kotekan_buffer: standard
+  hfb_compressed_lost_samples_buf_3:
+    kotekan_buffer: standard
+
+
+##########################
+#### Pipeline Stages #####
+##########################
+gen_data:
+  type: random
```
andrerenard, line 599: Random will keep trying to generate a new array every time around, which will be too slow. 
```suggestion
  type: const
```

andrerenard (approved):
I approve, but since I've now written part of this, @james-s-willis can you please have a look and comment?

config/chime_keep_gpu_warm.yaml
```diff
@@ -0,0 +1,3783 @@
+##########################################
+#
+# chime_science_run_gpu.yaml
```
james-s-willis, line 3: Just needs a name change in the comment.

config/chime_keep_gpu_warm.yaml
```diff
@@ -0,0 +1,3783 @@
...
+      gain_dir: /invalid
+    7:
+      kotekan_update_endpoint: json
+      gain_dir: /invalid
+    8:
+      kotekan_update_endpoint: json
+      gain_dir: /invalid
+    9:
+      kotekan_update_endpoint: json
+      gain_dir: /invalid
```
james-s-willis, line 140: I'm guessing we don't need the extra tracking beam changes?

lib/stages/CMakeLists.txt
```diff
@@ -67,7 +67,9 @@ add_library(
     VisSharedMemWriter.cpp
     DataQuality.cpp
     BeamExtract.cpp
-    BeamInspect.cpp)
+    BeamInspect.cpp
+    testDataGen.cpp
```
james-s-willis, line 71: Is there a reason why `testDataGen` has been moved to the `stages` directory?

james-s-willis (commented):
I think it looks good. So no data is actually generated in the `network_buffers`? I guess it doesn't matter. How does this work compared to the `testDataGen` stage? Is it because you are generating the `network_buffers` all together instead of invoking the stage 4 times?

config/chime_keep_gpu_warm.yaml
andrerenard: No, and in fact better we not include them here in case I have to revert...

lib/stages/CMakeLists.txt
andrerenard: Yes, because without `-DWITH_TESTS=on` the normal kotekan builds (including the site build) won't have access too it. And since it's actually very useful, I figured we should build by default.

andrerenard:
> I think it looks good. So no data is actually generated in the `network_buffers`? I guess it doesn't matter. How does this work compared to the `testDataGen` stage? Is it because you are generating the `network_buffers` all together instead of invoking the stage 4 times?

There is still "something" in the frames, we just don't set it because we don't really need to.  I would like to add another option to make it generate simulated sky data, but that's for another day. 

As for how it's different from `testDataGen`, it is because they are all generated together.  The key here is that they are generated in sync.  The `testDataGen` can end up generating some faster than others, which can trip up the systems that detect if things are running too slow (to detect the wrong queue).

config/chime_keep_gpu_warm.yaml
andrerenard: Actually we needed them to match the `coco` configuration.

PR #885, opened 2020-11-17
Use correct getters and setters for CHIME metadata.

james-s-willis:


andrerenard (approved):
Looks good, thanks for the quick fix.

PR #886, opened 2020-11-18
Add logging to help debug delay around baseband threads' interaction

cubranic:


andrerenard:
I don't think we should actually merge this in at `INFO` level, but it is useful and easy to test in this state.

andrerenard:
@cubranic if you want to change these to `DEBUG` we can merge into `develop`, otherwise I'd suggest we close this PR.

cubranic:
OK to close.

PR #888, opened 2020-11-19
Enable two extra tracking beams

### LINKED ISSUE #801: Add extra tracking beams which can be sent over a TCP connection
We want to add additional tracking beams for other projects.   To simplify things on the x-engine nodes, we will use `bufferSend` stages to transfer the data from those additional beams to the new backends (which will also run copies of `kotekan` and reprocess the data into different formats). 

This requires two main tasks on the x-engine:

- [ ] Decouple the concepts of `num_tracking_beams` and `num_pulsar_beams`  The `pulsarPostProcess` and `pulsarNetworkProcess` can stay the same, but will use `num_pulsar_beams`, and only read the first 10 beams in the array, even if it includes more than 10 beams.
- [ ] Add a `stage` to extract the extra tracking beams, pack them into new longer buffers (about 1 second), and send them to one of the other tracking beam backends using the `bufferSend` stage.

And for convenience we should also break up the gain application endpoint see: #799

### LINKED ISSUE #813: Move beamforming 4+4-bit truncation into kernel
At the moment the tracking beamformer outputs the beams in 32+32-bit format.  The CPU then scales this in `pulsarPostProcess` down to a 4+4-bit offset encoded format along with a `scaling` factor.   To keep things symmetric, the data is clamped to 0 to 15 (or -7 to +7 after removing the offset encoding). 

Ideally this should be moved to the GPU since that should free up CPU resources, and have relatively little impact on the GPU.   It also makes it easier and less resource intensive to copy out the extra beams (beams 11+) from the GPU block output.

The one possible down side of this is it limits options that preserve more dynamic range, like with the dynamic scaling in the FRB system.  However it's not clear that's something we actually need to have here.

Note this change does require updating the `pulsarPostProcess` code to deal with the different data size/indexing.

andrerenard:
* Adds two new tracking beams. 
* New stage to extract the extra beams into their own buffers
* Send beams with kotekan bufferSend to `bongo` and the new `moose`
* Adjusts pulsar post process to deal with changes to the tracking beamformer output indexing
* Moves the beam value scaling to the GPU - which saves on CPU resources (memory and compute).  Impact on the GPU appears to be minimal <0.5% summed across all 12 beams.  

Closes #801 
Closes #813


lib/hsa/hsaTrackingUpdatePhase.cpp
```diff
@@ -68,22 +68,41 @@ hsaTrackingUpdatePhase::hsaTrackingUpdatePhase(Config& config, const std::string
...
+        (float*)hsa_host_malloc(phase_frame_len + scaling_frame_len, device.get_gpu_numa_node());
+    if (host_phase_0 == nullptr)
+        throw std::runtime_error("Could not allocate memory in hsaTrackingUpdatePhase");
+    host_scaling_0 = host_phase_0 + _num_elements * _num_beams * 2;
+
+    host_phase_1 =
+        (float*)hsa_host_malloc(phase_frame_len + scaling_frame_len, device.get_gpu_numa_node());
+    if (host_phase_1 == nullptr)
+        throw std::runtime_error("Could not allocate memory in hsaTrackingUpdatePhase");
+    host_scaling_1 = host_phase_0 + _num_elements * _num_beams * 2;
```
james-s-willis, line 95: shouldn't this be:
```
host_scaling_1 = host_phase_1 + _num_elements * _num_beams * 2;
```

lib/hsa/hsaTrackingUpdatePhase.hpp
```diff
@@ -88,12 +91,18 @@ class hsaTrackingUpdatePhase : public hsaCommand {
     bool tracking_grab_callback(nlohmann::json& json, const uint8_t beamID);
 
 private:
-    /// Length of arrray of phases, should be 2048 x 10 x 2 for complex
+    /// Length of array of phases in bytes, should be 2048 x 10 x 2 for complex
```
james-s-willis, line 94: Should the dimensions be: 2048 x 12 x 2?

lib/stages/BeamExtract.cpp
```diff
@@ -0,0 +1,98 @@
...
+        out_metadata->dataset_id = in_metadata->dataset_id;
+        out_metadata->ra = in_metadata->beam_coord.ra[_extract_beam];
+        out_metadata->dec = in_metadata->beam_coord.dec[_extract_beam];
+        out_metadata->scaling = in_metadata->beam_coord.scaling[_extract_beam];
+
+        DEBUG2("Extracted beam: {:d}, fpga_number: {:d}", _extract_beam,
+               out_metadata->fpga_seq_start);
+        DEBUG2("Some data values: {:d},{:d}", out_frame[0] & 0x0F, (out_frame[0] & 0xF0) >> 4);
+
+        mark_frame_empty(in_buf, unique_name.c_str(), in_frame_id);
```
james-s-willis, line 93: It's a minor style comment, but you could increment `in_frame_id` and `out_frame_id` inside `mark_frame_empty()`.

james-s-willis (changes requested):
Mostly minor comments, apart from the value of `host_scaling_1`. Otherwise it looks good to me :+1:

lib/hsa/hsaTrackingUpdatePhase.cpp
andrerenard: You are right, good catch!  That would have been a hard bug to find.

emmanuelfonseca (approved):

emmanuelfonseca:
i also added one minor, clarifying question but the re-definitions / moving of code once in `pulsar` looks good to me.

lib/hsa/hsaTrackingUpdatePhase.hpp
andrerenard: Yeah `num_beams` really.

lib/stages/BeamExtract.cpp
andrerenard: Done.

lib/hsa/kernels/tracking_beamformer_nbeam.cl
```diff
@@ -71,8 +74,22 @@ __kernel void trackingbf_float( __global uint *data, __global float2 *phase,  __
                       as_float(__builtin_amdgcn_mov_dpp(as_uint(sum.IM),0x102,0xf,0xf,0))+
                       as_float(__builtin_amdgcn_mov_dpp(as_uint(sum.IM),0x103,0xf,0xf,0));
 
-            if (get_local_id(0) == 0)
-                output[(t*nsamp/TS+get_group_id(2))*get_global_size(1)*2 + b*2 + get_group_id(0)] = sum;
+            if (get_local_id(0) == 0) {
```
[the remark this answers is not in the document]
andrerenard: Yes that's implementing #813 .   It makes the post processing more efficient, and critically reduces the size of the GPU output buffers by a factor of 8 (`sizeof(float2)` to `sizeof(uint8_t)`).  Which makes it much easier to add the extra beams without reducing the buffer depth elsewhere.

james-s-willis (approved):

PR #889, opened 2020-11-19
Baseband: remove check for destination path

cubranic:


andrerenard (commented):
I have one question.  If we remove this check, and the `baseband` node isn't mounted, do we run the risk of writing into the OS NFS system, or will it error out in the open/write?

cubranic:
> I have one question. If we remove this check, and the `baseband` node isn't mounted, do we run the risk of writing into the OS NFS system, or will it error out in the open/write?

The logic doesn't really change with the removal of `stat()`. Because the file is written into "file_path", which has to be created by the caller (the FRB pipeline uses `YYYY/mm/dd` and for test dumps `baseband_test/...`), we don't run the risk of writing to the OS root.

The only way to break this is to use "." as the value of the `file_path` argument, but that's always been the risk.

andrerenard (approved):
Thanks for the clarification.

PR #890, opened 2020-11-20
config(chime_science_run_recv): Update ringmap frequency list.

tristpinsm:
To generate the new list of frequencies, I created a mask from
- querying the frequencies currently active from prometheus (with 237 nodes or so, I got 944 active frequencies)
- the list of frequencies remapped to dead nodes
- setting a threshold of 1% on the excised fraction
- the static RFI mask from `ch_util`

Then I picked 64 more or less evenly spaced in the remaining set.

Here is a graph that show the chosen frequencies (orange markers) and the combined mask (blue lines).
image


jrs65 (approved):
And you even straightened out the list! Approved!

PR #891, opened 2020-11-20
Extra beam recv config template and testing stage

andrerenard:
Simple stage and config to get people started with writing the extra beam receivers.

### DIFF
--- a/lib/stages/BeamInspect.cpp
+++ b/lib/stages/BeamInspect.cpp
@@ -1,0 +1,54 @@
+#include "BeamInspect.hpp"
+
+#include "visUtil.hpp" // for frameID, modulo
+#include "BeamMetadata.hpp"
+#include "StageFactory.hpp"    // for REGISTER_KOTEKAN_STAGE
+#include "buffer.h"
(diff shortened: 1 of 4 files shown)


lib/stages/BeamInspect.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include "buffer.h"
+#include "Telescope.hpp"
+
+using kotekan::bufferContainer;
+using kotekan::Config;
+using kotekan::Stage;
+
+REGISTER_KOTEKAN_STAGE(BeamInspect);
+
+STAGE_CONSTRUCTOR(BeamInspect) {
```
james-s-willis, line 15: Didn't know this macro existed, good to know. I'll start using it in the future.

james-s-willis (approved):
Looks good to me

lib/stages/BeamInspect.cpp
andrerenard: Yeah I wrote it years ago, but always forget to use it.

PR #893, opened 2020-11-24
Revert config for extra beams

andrerenard:
This is a partial reverting of #888 to check if the problem is in having the extra beams, but not in the other changes to the scaling application.


andrerenard:
Given we seem to have resolved the extra beams issue, I will close this now.  Can be reopened later if needed.

PR #894, opened 2020-11-24
Revert extra beams entirely

andrerenard:
If we cannot find the bug in the code soon, this will fully revert the extra beams changes.


andrerenard:
Given we seem to have resolved the extra beams issue, I will close this now.  Can be reopened later if needed.

PR #899, opened 2020-11-25
config(recv): add valve processes to protect from backups in bufferSend

jrs65:


andrerenard (approved):
This looks good to me.  Just to check is there enough RAM on that node to support the extra buffers?

jrs65:
Yeah. They're tiny compared to the rest (e.g. the length 4096 *full N2* buffer!)

andrerenard (commented):
Actually might it make more sense to make the valve buffers the larger of the two in terms of splitting up the buffer depth?   The valve process generally moves things from the input buffer very quickly, but will start dropping as soon as the output buffer is full.

jrs65:
I thought about that, but the issue is that the upstream buffers all feed into the write processes. So I think you do actually want it like this.

PR #901, opened 2020-11-25
HDF5 writer for HFB data

james-s-willis:


lib/utils/HFBFileArchive.cpp
```diff
@@ -0,0 +1,368 @@
...
+}
+
+void HFBFileArchive::create_axes(const std::vector<time_ctype>& times,
+                                 const std::vector<freq_ctype>& freqs,
+                                 const std::vector<uint32_t>& beams,
+                                 const std::vector<uint32_t>& subfreqs) {
+
+    create_axis("freq", freqs);
+    create_axis("time", times);
+    create_axis("beam", beams);
```
james-s-willis, line 193: @jrs65, I don't think I need these axes, right?

jrs65 (changes requested):
I'm not finished reviewing yet. But here is a first batch of changes.

lib/stages/Transpose.cpp
```diff
@@ -0,0 +1,234 @@
...
+                    "states.\nExiting...",
+                    ds_id);
+        return;
+    }
+
+    // Get the original dataset ID (before adding time axis)
+    base_ds_id = base_dset(ds_id);
+
+    // Once the async get_dataset_state() is done, we have all the metadata to
+    // create a file.
```
jrs65, line 120: This comment doesn't seem to refer to anything.

jrs65 (changes requested):
Another batch of comments.

lib/stages/Transpose.cpp
```diff
@@ -0,0 +1,234 @@
...
+    // Get file path to write to
+    filename = config.get<std::string>(unique_name, "outfile");
+
+    // Get a timeout for communication with broker
+    timeout =
+        std::chrono::duration<float>(config.get_default<float>(unique_name, "comet_timeout", 60.));
+
+    // Collect some metadata. The rest is requested from the datasetManager,
+    // once we received the first frame.
+    metadata["archive_version"] = "3.1.0";
```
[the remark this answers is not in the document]
james-s-willis: So should this be set to `4.3.0` for both the Vis and HFB files? As it is currently set in the base class.

lib/stages/Transpose.cpp
```diff
@@ -0,0 +1,231 @@
...
+            auto [frame_size, fpga_seq_total, frame_ds_id] = get_frame_data();
+            (void)frame_size;
+
+            // Parse the dataset ID
+            if (fpga_seq_total == 0 && frame_ds_id == dset_id_t::null) {
+                DEBUG("Got an empty frame.");
+                // Empty frames have a null dataset ID
+                ds_id_str = fmt::format("{}", dset_id_t::null);
+            } else if (frame_ds_id != ds_id) {
+                // TODO assuming that dataset ID changes here never change dataset dimensions
```
jrs65, line 153: That can essentially be tested with the fingerprint, although it's not clear it's worth doing

lib/stages/Transpose.cpp
```diff
@@ -0,0 +1,231 @@
...
+        return;
+    }
+
+    // Get the original dataset ID (before adding time axis)
+    base_ds_id = base_dset(ds_id);
+
+    found_flags = std::vector<bool>(write_t, false);
+
+    // Create HDF5 file
+    create_hdf5_file();
```
jrs65, line 122: Can this fail? It's an IO operation so it presumably can, but I don't really see any way that's being dealt with.

jrs65 (changes requested):

lib/stages/Transpose.cpp
james-s-willis: So this just creates a `HFBFileArchive`:
```
void HFBTranspose::create_hdf5_file() {
    // Create HDF5 file
    file = std::unique_ptr<HFBFileArchive>(
        new HFBFileArchive(filename, metadata, times, freqs, beams, sub_freqs, chunk,
                           kotekan::logLevel(_member_log_level)));                                                                                                                                                 
}
```
There were no checks in `VisTranspose` for it. Should I check if it is `NULL`?

lib/stages/RawReader.hpp
```diff
@@ -252,22 +252,22 @@ RawReader<T>::RawReader(Config& config, const std::string& unique_name,
...
-        chunk_f = chunk_size[0];
-        if (chunk_size[0] < 1 || chunk_size[1] < 1 || chunk_size[2] < 1)
-            throw std::invalid_argument("RawReader: config: Chunk size "
-                                        "needs to be greater or equal to (1,1,1) (is ("
-                                        + std::to_string(chunk_size[0]) + ","
-                                        + std::to_string(chunk_size[1]) + ","
-                                        + std::to_string(chunk_size[2]) + ")).");
-    }
+    chunk_size = config.get_default<std::vector<int>>(unique_name, "chunk_size", {4, 64, 128});
+    chunked = true;
```
jrs65, line 256: How would you read use RawReader without chunking now?

Generally I think we shouldn't set default chunk sizes in either the transpose stages or here in the RawReader. It's not clear enough what a good default chunk size is for all cases.

jrs65 (changes requested):

lib/stages/HFBTranspose.cpp
```diff
@@ -118,18 +119,31 @@ bool HFBTranspose::get_dataset_state(dset_id_t ds_id) {
...
+    auto& tel = Telescope::instance();
+
+    sub_freqs.resize(num_subfreq);
+
+    const double freq_width = tel.freq_width(0);
+    double freq_diff = 0.5 * freq_width;
+    double freq_inc = freq_width / num_subfreq;
+    for (auto& sf : sub_freqs) {
+        sf = freq_diff;
+        freq_diff -= freq_inc;
```
jrs65, line 145: You should try and determine the sign of this (and the start frequency) from the Telescope object, as it depends on the F-engine sampling scheme. I think a decent way of doing this is to use tel::to_freq to determine if the frequencies are increasing or decreasing to determine if you are in an odd or even Nyquist zone and set the sign appropriately.

Maybe @andrerenard can comment about whether this is worth it, but I think that should be correct for most telescopes?

jrs65 (changes requested):

lib/stages/HFBTranspose.cpp
andrerenard: So for some upcoming telescopes we will be working in the first and second Nyquist zones.  So yes, I'd suggest using the telescope functions for this.

andrerenard: I wounder if we should add an option to `Telescope` or `ICETelescope` to expose the Nyquist zone?  It's a bit odd to have to infer that from the direction of the frequency bin to frequency conversion.

james-s-willis: @andrerenard, @jrs65 I've added the Nyquist zone to the `ICETelescope` and used it to determine the order of frequencies. Is this what you had in mind?

lib/stages/HFBTranspose.cpp
```diff
@@ -139,10 +139,12 @@ bool HFBTranspose::get_dataset_state(dset_id_t ds_id) {
 
     const double freq_width = tel.freq_width(0);
     double freq_diff = 0.5 * freq_width;
```
jrs65, line 141: You also need to use the sign in here.

jrs65 (changes requested):

jrs65 (approved):

PR #902, opened 2020-11-26
fix(prometheusMetrics): ensure NaN and Inf are valid Prometheus formatting

anjakefala:
If they are not, all metrics in a scrape will get dropped.

https://prometheus.io/docs/instrumenting/exposition_formats/

### DIFF
--- a/lib/core/prometheusMetrics.cpp
+++ b/lib/core/prometheusMetrics.cpp
@@ -55,13 +55,22 @@ void Gauge::set(const double value) {
 }
 
 string Gauge::to_string() {
-    return fmt::format(fmt("{:f} {:d}"), value, last_update_time_stamp);
+    std::ostringstream buf;
+    to_string(buf);
+    return buf.str();
 }
 
 std::ostringstream& Gauge::to_string(std::ostringstream& out) {
     std::lock_guard<std::mutex> lock(metric_lock);
 
-    fmt::print(out, fmt("{:f} {:d}"), value, last_update_time_stamp);
+    if (std::isnan(value)) {
+        fmt::print(out, fmt("NaN {:d}"), last_update_time_stamp);
+    } else if (std::isinf(value)) {
+        fmt::print(out, fmt("{} {:d}"), (value < 0 ? "-Inf" : "+Inf"), last_update_time_stamp);
+    } else {
+        fmt::print(out, fmt("{:f} {:d}"), value, last_update_time_stamp);
+    }
+
     return out;
 }
 
--- a/lib/core/prometheusMetrics.cpp
+++ b/lib/core/prometheusMetrics.cpp
@@ -3,11 +3,12 @@
 #include "kotekanLogging.hpp" // for ERROR_NON_OO
 #include "restServer.hpp"     // for restServer, connectionInstance
 
-#include "fmt.hpp" // for format, fmt, print
+#include "fmt.hpp" // for print, format, fmt
 
-#include <functional> // for _Bind_helper<>::type, _Placeholder, bind, _1
+#include <cmath>      // for isinf, isnan
+#include <functional> // for _Bind_helper<>::type, _Placeholder, bind, _1, placeholders
 #include <iterator>   // for begin, end
-#include <ostream>    // for operator<<, ostringstream, basic_ostream
+#include <ostream>    // for operator<<, basic_ostream
 #include <sys/time.h> // for gettimeofday, timeval
 #include <utility>    // for pair
 
--- a/tests/boost/test_prometheus_metrics.cpp
+++ b/tests/boost/test_prometheus_metrics.cpp
@@ -3,6 +3,7 @@
 #include "prometheusMetrics.hpp" // for Metrics, MetricFamily, Counter, Gauge
 
 #include <boost/test/included/unit_test.hpp> // for BOOST_PP_IIF_1, BOOST_CHECK, BOOST_PP_BOOL_2
+#include <cmath>                             // for sqrt, log
 #include <iostream>                          // for cout, ostream
 #include <string>                            // for string, allocator, basic_string, operator==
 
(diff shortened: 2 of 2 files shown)


lib/core/prometheusMetrics.cpp
```diff
@@ -55,13 +55,18 @@ void Gauge::set(const double value) {
...
+        return fmt::format(fmt("{} {:d}"), (value < 0 ? "-Inf" : "+Inf"), last_update_time_stamp);
+    }
     return fmt::format(fmt("{:f} {:d}"), value, last_update_time_stamp);
 }
 
 std::ostringstream& Gauge::to_string(std::ostringstream& out) {
     std::lock_guard<std::mutex> lock(metric_lock);
 
-    fmt::print(out, fmt("{:f} {:d}"), value, last_update_time_stamp);
+    fmt::print(out, this->to_string());
```
cubranic, line 69: I think the reason this line was duplicated in both methods is because the `to_string()` one might allocate memory for the string and then it's freed when it's just copied to `ostringstream` (which presumably already has memory allocated for it, so you're just copying bytes at the end). If anything, do it the other way around: implement the full "if" in `to_string(ostringstream&) and then in `to_string()` just do:
```
 std::ostringstream buf;
to_string(buf);
return buf.str();
```

cubranic: This is just a small optimization, it might not make a huge difference. But the "to_string()" might be just a convenience I added for debugging. In the `serialize()` method, which is what gets called on a scrape to `/metrics`, it just iterates through all known metrics and calls the `to_string(ostringstream&)` method, joining them all with newlines.

Of course, now I see that in prometheusMetrics.cpp, I do the "ostring.str()" in a number of places, which kind of defeats the purpose. I should go back and clean that up.

cubranic (changes requested):
Can you add a test for both "NaN" and infinity to `tests/boost/test_prometheus_metrics.cpp`?

lib/core/prometheusMetrics.cpp
anjakefala: Okay!

tests/boost/test_prometheus_metrics.cpp
```diff
@@ -5,6 +5,7 @@
 #include <boost/test/included/unit_test.hpp> // for BOOST_PP_IIF_1, BOOST_CHECK, BOOST_PP_BOOL_2
 #include <iostream>                          // for cout, ostream
 #include <string>                            // for string, allocator, basic_string, operator==
+#include <math.h>                            // for sqrt
```
cubranic, line 8: I think this should be `<cmath>`.

cubranic (approved):

tests/boost/test_prometheus_metrics.cpp
anjakefala: /code/kotekan/tests/boost/test_prometheus_metrics.cpp:8:10: fatal error: cmath.h: No such file or directory
 #include <cmath.h>                           // for sqrt

tests/boost/test_prometheus_metrics.cpp
```diff
@@ -5,6 +5,7 @@
 #include <boost/test/included/unit_test.hpp> // for BOOST_PP_IIF_1, BOOST_CHECK, BOOST_PP_BOOL_2
 #include <iostream>                          // for cout, ostream
 #include <string>                            // for string, allocator, basic_string, operator==
+#include <cmath>                             // for sqrt
```
cubranic, line 8: Right, this is jsut `#include <cmath>` And then you might have to refer to `sqrt` as `std::sqrt`. (Check the build to verify.) See https://en.cppreference.com/w/cpp/numeric/math/sqrt.

cubranic (approved):

tests/boost/test_prometheus_metrics.cpp
anjakefala: It also will not let me do this =)

```
../../../tests/boost/test_prometheus_metrics.cpp:37:17: error: division by zero [-Werror=div-by-zero]
     bazinf.set(1/0);
```

anjakefala:
iwyu and clang-format disagree

```
 --- a/lib/core/prometheusMetrics.cpp
+++ b/lib/core/prometheusMetrics.cpp
@@ -1,13 +1,16 @@
-#include <sys/time.h>          // for gettimeofday, timeval
-#include <functional>          // for _Bind_helper<>::type, _Placeholder, bind, _1, placeholders
-#include <iterator>            // for begin, end
-#include <ostream>             // for operator<<, basic_ostream
-#include <utility>             // for pair
-#include <cmath>               // for isinf, isnan
 #include "prometheusMetrics.hpp"
-#include "kotekanLogging.hpp"  // for ERROR_NON_OO
-#include "restServer.hpp"      // for restServer, connectionInstance
-#include "fmt.hpp"             // for print, format, fmt
+
+#include "kotekanLogging.hpp" // for ERROR_NON_OO
+#include "restServer.hpp"     // for restServer, connectionInstance
+
+#include "fmt.hpp" // for print, format, fmt
+
+#include <cmath>      // for isinf, isnan
+#include <functional> // for _Bind_helper<>::type, _Placeholder, bind, _1, placeholders
+#include <iterator>   // for begin, end
+#include <ostream>    // for operator<<, basic_ostream
+#include <sys/time.h> // for gettimeofday, timeval
+#include <utility>    // for pair
```

Who has priority?

andrerenard (approved):
This looks great, thanks for fixing this.   

Also I saw a comment about interactions of iwyu and clang-format.   Generally I apply the `iwyu` recommended changes, and then run clang-format - and that (most of the time) makes both of them happy.

anjakefala:
> This looks great, thanks for fixing this.
> 
> Also I saw a comment about interactions of iwyu and clang-format. Generally I apply the `iwyu` recommended changes, and then run clang-format - and that (most of the time) makes both of them happy.

That thankfully worked!

jrs65 (approved):
Nice! And with tests too.

cubranic:
> Nice! And with tests too.

That's how we roll baby!

PR #906, opened 2020-11-30
fix(visRawReader): clear the page cache as well as the memory map

jrs65:
If this is not done, the file will still remain in memory, potentially slowing down other operations
which need to acquire memory.

### DIFF
--- a/lib/stages/visRawReader.cpp
+++ b/lib/stages/visRawReader.cpp
@@ -376,9 +376,17 @@ void visRawReader::main_thread() {
         dset_id_t& ds_id = ((VisMetadata*)(out_buf->metadata[frame_id]->metadata))->dataset_id;
         ds_id = get_dataset_state(ds_id);
 
-        // Try and clear out the cached data as we don't need it again
+        // Try and clear out the cached data from the memory map as we don't need it again
         if (madvise(mapped_file + file_ind * file_frame_size, file_frame_size, MADV_DONTNEED) == -1)
             DEBUG("madvise failed: {:s}", strerror(errno));
+#ifdef __linux__
+        // Try and clear out the cached data from the page cache as we don't need it again
+        // NOTE: unless we do this in addition to the above madvise the kernel will try and keep as
+        // much of the file in the page cache as possible and it will fill all the available memory
+        if (posix_fadvise(fd, file_ind * file_frame_size, file_frame_size, POSIX_FADV_DONTNEED)
+            == -1)
+            DEBUG("fadvise failed: {:s}", strerror(errno));
+#endif
 
         // Release the frame and advance all the counters
         mark_frame_full(out_buf, unique_name.c_str(), frame_id++);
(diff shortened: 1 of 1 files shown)


jrs65:
As hoped, this seems to fix all the memory issues with gossec. On my test runs it seems to occupy (including page caches) a flat 3.2 GB for the full lifetime.

tristpinsm (approved):

jrs65:
@nritsche can you look at this?

lib/stages/visRawReader.cpp
```diff
@@ -376,9 +376,17 @@ void visRawReader::main_thread() {
...
+        // Try and clear out the cached data from the memory map as we don't need it again
         if (madvise(mapped_file + file_ind * file_frame_size, file_frame_size, MADV_DONTNEED) == -1)
             DEBUG("madvise failed: {:s}", strerror(errno));
+#ifdef __linux__
+        // Try and clear out the cached data from the page cache as we don't need it again
+        // NOTE: unless we do this in addition to the above madvise the kernel will try and keep as
+        // much of the file in the page cache as possible and it will fill all the available memory
+        if (posix_fadvise(fd, file_ind * file_frame_size, file_frame_size, POSIX_FADV_DONTNEED)
+            == -1)
+            DEBUG("fadvise failed: {:s}", strerror(errno));
```
nritsche, line 388: The error cases are:
```
       EBADF  The fd argument was not a valid file descriptor.

       EINVAL An invalid value was specified for advice.

       ESPIPE The specified file descriptor refers to a pipe or FIFO.
              (ESPIPE is the error specified by POSIX, but before kernel
              version 2.6.16, Linux returned EINVAL in this case.)
```
I think we would want to know about each of those, so make this at least a `WARNING`?

lib/stages/visRawReader.cpp
```diff
@@ -376,9 +376,17 @@ void visRawReader::main_thread() {
...
 
-        // Try and clear out the cached data as we don't need it again
+        // Try and clear out the cached data from the memory map as we don't need it again
         if (madvise(mapped_file + file_ind * file_frame_size, file_frame_size, MADV_DONTNEED) == -1)
             DEBUG("madvise failed: {:s}", strerror(errno));
+#ifdef __linux__
+        // Try and clear out the cached data from the page cache as we don't need it again
+        // NOTE: unless we do this in addition to the above madvise the kernel will try and keep as
+        // much of the file in the page cache as possible and it will fill all the available memory
+        if (posix_fadvise(fd, file_ind * file_frame_size, file_frame_size, POSIX_FADV_DONTNEED)
```
nritsche, line 386: Is file_frame_size page aligned?

lib/stages/visRawReader.cpp
jrs65: Fair point. I'll bump those up to WARNING

lib/stages/visRawReader.cpp
jrs65: Yes. The raw file frames are always aligned to 4kb boundaries for page alignment.

nritsche (approved):

jrs65:
Alright IWYU has passed. I'm going to merge this now.

PR #907, opened 2020-11-30
fix(buffer): remove excess logging message for full buffers

jrs65:
This should be replaced by a log level based mechanism in the future.


nritsche:
The change looks good to me, but I'm not the person using these log outputs ever, so I can't judge if removing them is okay.

andrerenard (approved):
Looks good.  @nritsche yes I'm ok with removing this for now, it's not very commonly used or all that helpful when we see it.  Mostly we notice back pressure in the `valve` stage anyway, and I'll be adding better tools for back pressure debugging soon.

PR #908, opened 2020-11-30
Templatise FreqSubset and refactor HFBAccumulate

james-s-willis:
After merging #865, there were outstanding issues raised by @jrs65 regarding code health. I have addressed those issues here:

* Templatise FreqSubset with VisFrameView and HFBFrameView
* Change dset `struct` to `HFBAccumulate` members
* Use `span` from `HFBFrameView` instead of raw `float*` when integrating frames
* Fixed a bug where `hfb1` was the wrong type (`int32_t` -> `float`), will mean that the current weights calculated are slightly off


jrs65 (approved):
Looks great. I'm amazed at how neatly the templated FreqSubset worked out, and it'll be a great example for trying to merge generic tasks in the future.

james-s-willis:
> Looks great. I'm amazed at how neatly the templated FreqSubset worked out, and it'll be a great example for trying to merge generic tasks in the future.

Thanks for reviewing, Richard. I'm just fixing up the IWYU then I'll merge this into `develop`.

PR #909, opened 2020-12-01
Correct type of primary weight accumulation vector to float.

james-s-willis:


andrerenard (approved):

jrs65 (approved):

jrs65:
Do we have any idea why the macOS builds are failing?

andrerenard:
It looks like it's trying to run two sets of CI tests for everything.   One of them appears to be passing and the other isn't...  I'm inclined to think this is a github bug of some kind.

nritsche:
> It looks like it's trying to run two sets of CI tests for everything. One of them appears to be passing and the other isn't... I'm inclined to think this is a github bug of some kind.

```
GitHub Actions has encountered an internal error when running your job.
```

andrerenard:
It's passing on one of the two MacOS tests, I'd be ok with merging this.  But we need to decide first if this is something we want to merge into the tagged release.

PR #910, opened 2020-12-01
Created RawReader base class 

james-s-willis:
I have abstracted the `visRawReader` class to `RawReader` base class and created a `HFBRawReader` that inherits from it.


lib/stages/VisRawReader.cpp
```diff
@@ -0,0 +1,112 @@
...
+    frame.freq_id = 0;
+    frame.dataset_id = dset_id_t::null;
+    frame.time = std::make_tuple(0, timespec{0, 0});
+
+    // mark frame as empty by ensuring this is 0
+    frame.fpga_seq_length = 0;
+    frame.fpga_seq_total = 0;
+
+    DEBUG("VisRawReader: Reading empty frame: {:d}", frame_id);
+}
```
jrs65, line 107: Again, not really anything you've changed, but I think creating a zero'd out frame is generic enough I'd actually consider moving it into VisFrameView itself. Plus it can probably just be done with `memset` for more efficiency in there as VisFrameView can assume it knows the layout.

Also, I'm not even sure whether the frame needs to be zeroed out as you're just going to copy the actual data into it. Although I have a feeling that @tristpinsm might need it to send empty frames for missing data....

lib/stages/VisRawReader.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+dset_id_t& VisRawReader::get_dataset_id(frameID frame_id) {
+    auto frame = VisFrameView(out_buf, frame_id);
+    return frame.dataset_id;
+}
```
jrs65, line 112: Not super important, but I feel this is crying out to use templates, or change the base FrameView class to assume that the frame has a dataset_id.

I mean, this whole scheme won't work without a dataset_id (or freq or time), and so the base `RawReader` could make stronger assumptions about what it's getting.

lib/stages/RawReader.cpp
```diff
@@ -343,37 +316,13 @@ void visRawReader::main_thread() {
...
-
-            DEBUG("visRawReader: Reading empty frame: {:d}", frame_id);
+            // Create empty frame and set structural metadata
+            create_empty_frame(frame_id);
         }
 
         // Set the dataset ID to the updated value
-        dset_id_t& ds_id = ((VisMetadata*)(out_buf->metadata[frame_id]->metadata))->dataset_id;
+        dset_id_t& ds_id = get_dataset_id(frame_id);
         ds_id = get_dataset_state(ds_id);
```
jrs65, line 326: The naming of these two functions and the way they are used makes sense individually, but in combination is somewhat confusing, i.e it implies you get a dataset_id and then assign a dataset_state to it, despite those being different types and concepts.

lib/stages/RawReader.hpp
```diff
@@ -22,9 +22,16 @@
...
 
+using nlohmann::json;
+
 /**
- * @class visRawReader
- * @brief Read and stream a raw visibility file.
+ * @class RawReader
+ * @brief Generic class to read and stream a raw file.
+ *
+ * All classes which inherit from this should provide the following API:
```
jrs65, line 31: This is another one where the hot potato landed on you, but this class could really do with a clear description of the read order. The description in the paragraph below isn't very clear at all. An example might make it clearer, also just saying *why* you might ever want to read in this chunked order.

lib/stages/RawReader.hpp
```diff
@@ -62,14 +69,14 @@
  *
  * @author Richard Shaw, Tristan Pinsonneault-Marotte, Rick Nitsche
```
jrs65, line 70: Make sure to add yourself to this lengthy list!

jrs65 (changes requested):

lib/stages/VisRawReader.cpp
tristpinsm: that's right, in gossec these frames are eventually written to disk. If it's inconvenient to zero them here, you could always modify `visTranspose` to write zeroes when it detects an empty frame.

lib/stages/RawReader.hpp
```diff
@@ -34,8 +41,8 @@
  *
  * @par Buffers
  * @buffer out_buf The data read from the raw file.
- *         @buffer_format VisBuffer structured
- *         @buffer_metadata VisMetadata
+ *         @buffer_format Buffer structured
+ *         @buffer_metadata Metadata
  *
  * @conf    readahead_blocks       Int. Number of blocks to advise OS to read ahead
  *                                 of current read.
```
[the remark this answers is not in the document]
james-s-willis: @tristpinsm, is the chunk size of length 3 because the `visTranspose` stage also requires it to be of length 3 when it is used in the `visFileArchive`?

lib/stages/VisRawReader.cpp
jrs65: Thanks for the info Tristan.

lib/stages/VisRawReader.cpp
james-s-willis: I've templatised `RawReader` to get around this. But in order to do that I had to move the functions into the header file. To stop undefined references when linking.

lib/stages/RawReader.hpp
james-s-willis: I'm not entirely sure of the motivation behind the chunking to write about it. I know it's to improve the reading efficiency. Could maybe you or @tristpinsm provide a brief summary I can add in?

lib/stages/RawReader.hpp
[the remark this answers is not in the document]
tristpinsm: yeah the only reason it is size three is because it is used by both stages

lib/stages/RawReader.hpp
tristpinsm: The reading is done in chunks for writing efficiency. The `visTranspose` stage writes to an HDF5 file with a chunked layout, and this means that IO is performed on entire chunks, so it is most efficient to write out on exact chunk boundaries. The reason for using chunking in the first place is to enable compression and try to optimise for certain IO patterns (i.e. read a few frequencies for many times).

james-s-willis: Thanks @tristpinsm!

jrs65 (changes requested):

jrs65 (approved):

PR #911, opened 2020-12-02
Add config/dpdk option to do burst captures of ADC data

andrerenard:


lib/dpdk/iceBoardStandard.hpp
```diff
@@ -152,15 +164,30 @@ inline bool iceBoardStandard::advance_frame(uint64_t new_seq, bool first_time) {
     // Advance the frame
     if (!first_time) {
         mark_frame_full(out_buf, unique_name.c_str(), out_frame_id);
-
-        // Advance frame ID
         out_frame_id = (out_frame_id + 1) % out_buf->num_frames;
+
+        // Advance the lost samples frame
+        mark_frame_full(lost_samples_buf, unique_name.c_str(), lost_samples_frame_id);
+        lost_samples_frame_id = (lost_samples_frame_id + 1) % lost_samples_buf->num_frames;
```
james-s-willis, line 171: Can we change the type on the frame IDs to `frameID`? To simplify these lines to:
```
  mark_frame_full(lost_samples_buf, unique_name.c_str(), lost_samples_frame_id++);
```

james-s-willis (changes requested):
Looks good to me :+1:

lib/dpdk/iceBoardStandard.hpp
andrerenard: We likely should, but all of the DPDK related code uses this style, so I think it should be changed all at once, and is thus out of scope for this PR.

james-s-willis (approved):

PR #915, opened 2021-01-13
Created HFB truncate stage

james-s-willis:
Also added a small optimisation when loading the error vector.


lib/stages/HFBTruncate.cpp
```diff
@@ -0,0 +1,141 @@
...
+    w_prec = config.get<float>(unique_name, "weight_fixed_precision");
+    if (w_prec < 0)
+        throw std::invalid_argument("HFBTruncate: config: "
+                                    "weight_fixed_precision should be positive (is "
+                                    + std::to_string(w_prec) + ").");
+    hfb_prec = config.get<float>(unique_name, "data_fixed_precision");
+    if (hfb_prec < 0)
+        throw std::invalid_argument("HFBTruncate: config: "
+                                    "data_fixed_precision should be positive (is "
+                                    + std::to_string(hfb_prec) + ").");
```
jrs65, line 58: I'm not sure what the best way of exiting in this circumstance is. Should these be FATAL_ERROR's instead?

jrs65 (changes requested):

jrs65 (approved):

PR #920, opened 2021-01-26
Jl/bass

luojing1211:
Added one stage for merging the beam data to a bigger frame.

### DIFF
--- a/config/beam_packet_merge.yaml
+++ b/config/beam_packet_merge.yaml
@@ -1,0 +1,52 @@
+##########################################
+#
+# raw_file_read.yaml
+#
+# Configuration for tracking extra beam 
+# receivers' raw data file reading
+#
+# Author: Jing Luo
+#
+##########################################
+---
+type: config
+log_level: info
+cpu_affinity: [0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15]
+samples_per_data_set: 49152
+num_pol: 2
+buffer_depth: 105
+
+telescope:
+  name: CHIMETelescope
+  require_gps: true
+  query_gps: true
+  require_frequency_map: true
+  query_frequency_map: true
+  allow_default_frequency_map: false
+
+dataset_manager:
+  use_dataset_broker: false
+
+beam_metadata_pool:
+  kotekan_metadata_pool: BeamMetadata
+  num_metadata_objects: 4096
+
+beam_recv_buffer:
+  kotekan_buffer: standard
+  num_frames: 2048
+  metadata_pool: beam_metadata_pool
+  frame_size: samples_per_data_set * num_pol
+
+raw_file_read:
+  base_dir: /drives/scratch/test_beam_raw
+  file_name: bass_beam_data
+  read0:
+    kotekan_stage: rawFileRead
+    file_ext: raw
+    buf: beam_recv_buffer
+
+beam_merge:
+  kotekan_stage: mergeRawBuffer
+  in_buf: beam_recv_buffer
+  out_buf: beam_recv_buffer
+
--- a/config/beam_packet_merge.yaml
+++ b/config/beam_packet_merge.yaml
@@ -10,11 +10,12 @@
 ##########################################
 ---
 type: config
-log_level: info
+log_level: debug2
 cpu_affinity: [0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15]
 samples_per_data_set: 49152
 num_pol: 2
 buffer_depth: 105
+raw_frames_per_merged_frame: 2
 
 telescope:
   name: CHIMETelescope
@@ -33,9 +34,16 @@ beam_metadata_pool:
 
 beam_recv_buffer:
   kotekan_buffer: standard
-  num_frames: 2048
+  num_frames: 10
   metadata_pool: beam_metadata_pool
   frame_size: samples_per_data_set * num_pol
+
+merged_buffer:
+  kotekan_buffer: standard
+  num_frames: 10 
+  metadata_pool: beam_metadata_pool
+  # the size of the beam metadata is 64
+  frame_size: (samples_per_data_set * num_pol + 64) * raw_frames_per_merged_frame 
 
 raw_file_read:
   base_dir: /drives/scratch/test_beam_raw
@@ -48,5 +56,12 @@ raw_file_read:
 beam_merge:
   kotekan_stage: mergeRawBuffer
   in_buf: beam_recv_buffer
-  out_buf: beam_recv_buffer
+  out_buf: merged_buffer
 
+#raw_file_write:
+#  kotekan_stage: rawFileWrite
+#  in_buf: merged_buffer
+#  base_dir: /drives/scratch/test_beam_raw
+#  file_name: merged_beam_data
+#  file_ext: raw
+#  exit_after_n_files: buffer_depth
--- a/config/beam_packet_merge.yaml
+++ b/config/beam_packet_merge.yaml
@@ -15,7 +15,9 @@ cpu_affinity: [0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15]
 samples_per_data_set: 49152
 num_pol: 2
 buffer_depth: 105
-raw_frames_per_merged_frame: 2
+raw_frames_per_merged_frame: 6
+# The size of metadata with freq ID
+freq_id_metadata_size: 72
 
 telescope:
   name: CHIMETelescope
@@ -40,10 +42,10 @@ beam_recv_buffer:
 
 merged_buffer:
   kotekan_buffer: standard
-  num_frames: 10 
+  num_frames: 10
   metadata_pool: beam_metadata_pool
-  # the size of the beam metadata is 64
-  frame_size: (samples_per_data_set * num_pol + 64) * raw_frames_per_merged_frame 
+  # the size of the beam metadata with frequency is 64
+  frame_size: (samples_per_data_set * num_pol + freq_id_metadata_size) * raw_frames_per_merged_frame 
 
 raw_file_read:
   base_dir: /drives/scratch/test_beam_raw
@@ -58,10 +60,10 @@ beam_merge:
   in_buf: beam_recv_buffer
   out_buf: merged_buffer
 
-#raw_file_write:
-#  kotekan_stage: rawFileWrite
-#  in_buf: merged_buffer
-#  base_dir: /drives/scratch/test_beam_raw
-#  file_name: merged_beam_data
-#  file_ext: raw
-#  exit_after_n_files: buffer_depth
+raw_file_write:
+  kotekan_stage: rawFileWrite
+  in_buf: merged_buffer
+  base_dir: /drives/scratch/test_beam_raw
+  file_name: merged_beam_data
+  file_ext: raw
+  exit_after_n_files: buffer_depth
--- a/config/beam_recv_raw_record.yaml
+++ b/config/beam_recv_raw_record.yaml
@@ -1,0 +1,55 @@
+##########################################
+#
+# beam_recv_template.yaml
+#
+# Starter configuration for extra beam receivers
+#
+# Author: Andre Renard
+#
+##########################################
+---
+type: config
+log_level: info
+cpu_affinity: [0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15]
+samples_per_data_set: 49152
+num_pol: 2
+buffer_depth: 105
+
+telescope:
+  name: CHIMETelescope
+  require_gps: true
+  query_gps: true
+  require_frequency_map: true
+  query_frequency_map: true
+  allow_default_frequency_map: false
+
+dataset_manager:
+  use_dataset_broker: false
+
+beam_metadata_pool:
+  kotekan_metadata_pool: BeamMetadata
+  num_metadata_objects: 4096
+
+beam_recv_buffer:
+  kotekan_buffer: standard
+  num_frames: 2048
+  metadata_pool: beam_metadata_pool
+  frame_size: samples_per_data_set * num_pol
+
+recv:
+  kotekan_stage: bufferRecv
+  buf: beam_recv_buffer
+  listen_port: 12020
+  num_threads: 2
+
+beam_inspect:
+  kotekan_stage: BeamInspect
+  in_buf: beam_recv_buffer
+
+raw_file_write:
+  kotekan_stage: rawFileWrite
+  in_buf: beam_recv_buffer
+  base_dir: /drives/scratch/test_beam_raw
+  file_name: beam_data
+  file_ext: raw
+  exit_after_n_files: buffer_depth
--- a/config/raw_file_read.yaml
+++ b/config/raw_file_read.yaml
@@ -1,0 +1,51 @@
+##########################################
+#
+# raw_file_read.yaml
+#
+# Configuration for tracking extra beam 
+# receivers' raw data file reading
+#
+# Author: Jing Luo
+#
+##########################################
+---
+type: config
+log_level: info
+cpu_affinity: [0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15]
+samples_per_data_set: 49152
+num_pol: 2
+buffer_depth: 105
+
+telescope:
+  name: CHIMETelescope
+  require_gps: true
+  query_gps: true
+  require_frequency_map: true
+  query_frequency_map: true
+  allow_default_frequency_map: false
+
+dataset_manager:
+  use_dataset_broker: false
+
+beam_metadata_pool:
+  kotekan_metadata_pool: BeamMetadata
+  num_metadata_objects: 4096
+
+beam_recv_buffer:
+  kotekan_buffer: standard
+  num_frames: 2048
+  metadata_pool: beam_metadata_pool
+  frame_size: samples_per_data_set * num_pol
+
+raw_file_read:
+  base_dir: /drives/scratch/test_beam_raw
+  file_name: bass_beam_data
+  read0:
+    kotekan_stage: rawFileRead
+    file_ext: raw
+    buf: beam_recv_buffer
+
+beam_inspect:
+  kotekan_stage: BeamInspect
+  in_buf: beam_recv_buffer
+
--- a/lib/metadata/BeamMetadata.hpp
+++ b/lib/metadata/BeamMetadata.hpp
@@ -26,4 +26,8 @@ struct BeamMetadata {
     uint32_t scaling;
 };
 
+struct FreqIDBeamMetadata:BeamMetadata {
+    uint32_t frequency_bin; 	
+};
+
 #endif // BEAMMETADATA_HPP
--- a/lib/stages/CMakeLists.txt
+++ b/lib/stages/CMakeLists.txt
@@ -47,6 +47,7 @@ add_library(
     bufferSwitch.cpp
     bufferDelay.cpp
     InputSubset.cpp
+    mergeRawBuffer.cpp
     # RFI Pipeline Processes
     rfiVDIF.cpp
     rfiAVXVDIF.cpp
--- a/lib/stages/mergeRawBuffer.cpp
+++ b/lib/stages/mergeRawBuffer.cpp
@@ -29,62 +29,80 @@ mergeRawBuffer::mergeRawBuffer(Config& config_, const std::string& unique_name,
 
     // Apply config
     _samples_per_data_set = config.get<uint32_t>(unique_name, "samples_per_data_set");
-
+    _num_pol = config.get<uint32_t>(unique_name, "num_pol");
+    _raw_frames_per_merged_frame = config.get<uint32_t>(unique_name, "raw_frames_per_merged_frame"); 
     in_buf = get_buffer("in_buf");
     register_consumer(in_buf, unique_name.c_str());
 
     out_buf = get_buffer("out_buf");
+
     register_producer(out_buf, unique_name.c_str());
     
-    INFO("Samples_per_data11 {:d} frame_size {:d}", _samples_per_data_set, (uint32_t)in_buf->frame_size);
-
-//    if (_samples_per_data_set != (uint32_t)in_buf->frame_size) {
-//        throw std::runtime_error("compressLostSamples in_frame has the wrong size.");
-//    } 
-
 }
 
 mergeRawBuffer::~mergeRawBuffer() {}
 
 
 void mergeRawBuffer::main_thread() {
-    // frameID in_buffer_ID(in_buf);
-    frameID frame_id(in_buf);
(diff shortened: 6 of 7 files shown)


andrerenard:
I'm going through the pull request, and there are a lot of config files, but I don't see one that receives from `bufferRecv` and writes to `nDiskFileWrite`.   Can you update one of them to the state you'd like to run on the full cluster?

lib/stages/mergeBeamBuffer.cpp
```diff
@@ -0,0 +1,130 @@
...
+    // on the incoming frame number
+    uint32_t in_frame_counter = 0;
+    
+    while (!stop_thread) {
+        // Get an input buffer, This call is blocking!
+        in_frame = wait_for_full_frame(in_buf, unique_name.c_str(), in_buffer_ID);
+        if (in_frame == nullptr)
+            break;
+
+	BeamMetadata* in_metadata = (BeamMetadata*)get_metadata(in_buf, in_buffer_ID);
```
andrerenard, line 61: The indentation in this function makes it a bit hard to read.  You can use `make clang-format` to clean up the indentation if your editor doesn't have a way to do it for you.

lib/stages/mergeBeamBuffer.cpp
```diff
@@ -0,0 +1,130 @@
...
+	if (sub_frame_index == 0){
+            out_frame = (uint8_t*)wait_for_empty_frame(out_buf, unique_name.c_str(), out_buffer_ID);
+            if (out_frame == nullptr)
+                break;
+	    allocate_new_metadata_object(out_buf, out_buffer_ID);
+            MergedBeamMetadata* out_metadata = (MergedBeamMetadata*)get_metadata(out_buf, out_buffer_ID);
+	    // populate the merged_buffer/out_buffer frame header.
+	    out_metadata -> sub_frame_pre_frame = _sub_frames_per_merged_frame;
+            out_metadata -> sub_frame_metadata_size = FreqIDBeamMeta_size;
+            out_metadata -> sub_frame_data_size = in_buf -> frame_size;
```
andrerenard, line 79: Our coding standard doesn't use spaces here.
```suggestion
	    out_metadata->sub_frame_pre_frame = _sub_frames_per_merged_frame;
            out_metadata->sub_frame_metadata_size = FreqIDBeamMeta_size;
            out_metadata->sub_frame_data_size = in_buf->frame_size;
```

lib/stages/mergeBeamBuffer.cpp
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	// Get the start position of a sub_out_frame's meatadata
+	// Make sure out_frame is pointed to the right place.
+        if (out_frame == nullptr)
+            break;
```
andrerenard, line 85: This can likely just be an `assert(out_frame != nullptr)` since provided the logic above is working correctly, you shouldn't be able to get to a state where out_frame isn't set.

lib/stages/mergeBeamBuffer.cpp
```diff
@@ -0,0 +1,130 @@
...
+	// Get the start position of a sub_out_frame's data
+	uint32_t out_frame_data_pos = out_frame_metadata_pos + sizeof(FreqIDBeamMetadata);
+	// Get the memory block for the data of of a subframe.
+	uint8_t* sub_frame_data = &out_frame[out_frame_data_pos];
+	// copy data from in frame to sub_out_frame
+	memcpy(sub_frame_data, in_frame, in_buf -> frame_size);
+        
+	// Debug code. Check if the subframe matedata is the same. 
+   	FreqIDBeamMetadata* debud_frame_metadata = (FreqIDBeamMetadata *)&out_frame[sub_frame_index * sub_frame_size];
+	DEBUG2("Sub frame Beam RA: {:f}, Dec: {:f}, scaling: {:d}, beam_num: {:d}, freq_id {:d}\n", debud_frame_metadata -> ra, debud_frame_metadata -> dec, debud_frame_metadata -> scaling, debud_frame_metadata-> beam_number,  debud_frame_metadata-> frequency_bin);
```
andrerenard, line 119: Variables only used in a debug statement need to be only included if we are doing a Debug type build, since otherwise it will generate an unused variable warning in Release builds. 
```suggestion
        #ifdef DEBUGGING
        FreqIDBeamMetadata* debug_frame_metadata = (FreqIDBeamMetadata *)&out_frame[sub_frame_index * sub_frame_size];
        DEBUG2("Sub frame Beam RA: {:f}, Dec: {:f}, scaling: {:d}, beam_num: {:d}, freq_id {:d}\n", debug_frame_metadata->ra, debug_frame_metadata->dec, debug_frame_metadata->scaling, debug_frame_metadata->beam_number,  debug_frame_metadata->frequency_bin);
        #endif
```

lib/stages/mergeBeamBuffer.cpp
```diff
@@ -0,0 +1,130 @@
...
+	    out_metadata -> sub_frame_pre_frame = _sub_frames_per_merged_frame;
+            out_metadata -> sub_frame_metadata_size = FreqIDBeamMeta_size;
+            out_metadata -> sub_frame_data_size = in_buf -> frame_size;
+	}
+
+	// Get the start position of a sub_out_frame's meatadata
+	// Make sure out_frame is pointed to the right place.
+        if (out_frame == nullptr)
+            break;
+	uint32_t out_frame_metadata_pos = sub_frame_size * sub_frame_index;
```
andrerenard, line 86: When you are doing complex indexing like this it's generally a good idea to add an `assert()` to check your computed index doesn't go past the end of the array.  e.g. `assert(out_frame_data_post < out_buf->frame_size)`

lib/stages/mergeBeamBuffer.cpp
```diff
@@ -0,0 +1,130 @@
...
+	sub_frame_metadata -> ra = in_metadata -> ra;
+	sub_frame_metadata -> dec = in_metadata -> dec;
+	sub_frame_metadata -> scaling = in_metadata -> scaling;
+        
+	// Add the frequency id into the metadata, for chime one stream has on frequency.
+	// TODO Have the output buffer handle the stream with mulitple frequency
+	const uint32_t num_freq_per_stream = Telescope::instance().num_freq_per_stream();
+
+        for (uint32_t f = 0; f < num_freq_per_stream; ++f) {
+	    // TODO this does not handling the stream with mulitple frequencies. since at right now,
```
andrerenard, line 105: This is generally best handled by encoding the variable `num_freq_per_stream` into your metadata, and then having a variable length section after the metadata, but before the voltage data, for storing each bin.  This does make the indexing more complex, but its really the only way to account for a variable length in the metadata.

lib/stages/nDiskFileWrite.hpp
```diff
@@ -34,11 +34,13 @@
  * @conf num_disks      Int , the number of drives to read from
  * @conf disk_base      String, the path to the mounted drives
  * @conf disk_set       String, the disk name.
+ * @conf file_ext       String, the extenstion of the output file.
```
andrerenard, line 37: Mention the default in the comment.

lib/stages/splitMergedBeamBuffer.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+splitMergedBeamBuffer::splitMergedBeamBuffer(Config& config_, const std::string& unique_name,
+                                             bufferContainer& buffer_container) :
+    Stage(config_, unique_name, buffer_container,
+          std::bind(&splitMergedBeamBuffer::main_thread, this)) {
+    
+    // Apply config
+    //_samples_per_data_set = config.get<uint32_t>(unique_name, "samples_per_data_set");
+    //_num_pol = config.get<uint32_t>(unique_name, "num_pol");
+    //_raw_frames_per_merged_frame = config.get<uint32_t>(unique_name, "raw_frames_per_merged_frame");
```
andrerenard, line 33: If you don't need these lines just remove them.

lib/stages/mergeBeamBuffer.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+
+/**
+ * @class mergeBeamBuffer
+ * @brief A stage to merge n single beam frame to one single 
+ *        beam frame.
+ *
+ * @par Buffers
+ * @buffer in_buf Kotekan buffer of raw packets.
+ *     @buffer_format Array of @c chars
```
andrerenard, line 30: Include a line for each buffer the metadata type.
```suggestion
 *     @buffer_format Array of @c chars
 *     @buffer_metadata MergedBeamMetadata
```

lib/stages/splitMergedBeamBuffer.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+    /// Primary loop to wait for buffers, split the merged beam buffer.
+    void main_thread() override;
+
+private:
+    /// Merged buffer for the merged
+    struct Buffer* in_buf;
+
+    /// Frame merged buffer 
+    struct Buffer* out_buf;
```
andrerenard, line 57: The comments here appear to be copied from the other stage, and need to be updated.

andrerenard (changes requested):
Sorry for the delay with this review.   Overall it looks good and appears to be doing everything correctly.   There are just some comments mostly related to code style and commenting.   You might want to have a look at our style guide here: http://lwlab.dunlap.utoronto.ca/kotekan/dev/dev_style_guide.html   And also running `make clang-format` should automatically clean up a large number of the problems.  

I would like to see the configuration you intend to run with the full array (e.g. with `bufferRecv` and `nDiskFileWrite` just so I can double check all components). 

Also given the number of configuration files you have here, we might want to move them into a sub folder in `config/` so our main config folder doesn't get to large.

andrerenard:
Also can you rename the PR to something more descriptive? e.g. `Tracking beam receiver with raw file format writers.`

config/beam_packet_merge.yaml
```diff
@@ -0,0 +1,69 @@
...
+# beam receivers' raw packet to a more 
+# packed buffer format and write them to 
+# files. 
+#
+# Author: Jing Luo
+#
+##########################################
+---
+type: config
+log_level: debug2
```
andrerenard, line 15: FYI: You can set the log level for any stage individually.  So if you don't want to see debug messages from `bufferRecv` you can set this to `info` or `warn` and add a `log_level: debug2` line into the block for the stage you want to see the extra debugging from.

config/tracking_beam_test/tracking_beam_frame_to_ndisk.yaml
```diff
@@ -0,0 +1,77 @@
...
+beam_recv_buffer:
+  kotekan_buffer: standard
+  num_frames: 2048
+  metadata_pool: beam_metadata_pool
+  frame_size: samples_per_data_set * num_pol
+
+merged_buffer:
+  kotekan_buffer: standard
+  num_frames: 10
+  metadata_pool: beam_metadata_pool
```
andrerenard, line 46: Doesn't this pool need to be the `merged_beam_metadata_pool` pool?

config/tracking_beam_test/tracking_beam_frame_to_ndisk.yaml
```diff
@@ -0,0 +1,77 @@
...
+  query_frequency_map: true
+  allow_default_frequency_map: false
+
+dataset_manager:
+  use_dataset_broker: false
+
+beam_metadata_pool:
+  kotekan_metadata_pool: FreqIDBeamMetadata
+  num_metadata_objects: 4096
+
```
andrerenard, line 36: Missing `merged_beam_metadata_pool`

config/tracking_beam_test/tracking_beam_frame_to_ndisk.yaml
```diff
@@ -0,0 +1,77 @@
...
+#
+##########################################
+---
+type: config
+log_level: debug2
+cpu_affinity: [0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15]
+samples_per_data_set: 49152
+num_pol: 2
+buffer_depth: 105
+sub_frames_per_merged_frame: 11
```
andrerenard, line 18: Is there a reason this is `11`?   Those are still going to be small files.  I would go for something like `500` which should give ~50MB files, which will be more efficient to write.

config/tracking_beam_test/tracking_beam_frame_to_ndisk.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+beam_recv_buffer:
+  kotekan_buffer: standard
+  num_frames: 2048
+  metadata_pool: beam_metadata_pool
+  frame_size: samples_per_data_set * num_pol
+
+merged_buffer:
+  kotekan_buffer: standard
+  num_frames: 10
```
andrerenard, line 45: I would make this at least 30, you will always have 8 frames busy with writes, and you want enough buffer to handle temporary slow downs in the disks.   Make sure you increase the pool size as well when you do this. 
```suggestion
  num_frames: 30
```

andrerenard (changes requested):
I think we should make this changes to the configuration before we run.

config/tracking_beam_test/tracking_beam_frame_to_ndisk.yaml
luojing1211: OK, it was just for the init test. Will change it

config/tracking_beam_test/tracking_beam_frame_to_ndisk.yaml
luojing1211: Will add it

lib/stages/mergeBeamBuffer.cpp
```diff
@@ -0,0 +1,134 @@
...
+               debud_frame_metadata->ra, debud_frame_metadata->dec, debud_frame_metadata->scaling,
+               debud_frame_metadata->beam_number, debud_frame_metadata->frequency_bin);
+
+        mark_frame_empty(in_buf, unique_name.c_str(), in_buffer_ID);
+        in_buffer_ID++;
+        in_frame_counter++;
+
+        if (sub_frame_index == _sub_frames_per_merged_frame - 1) {
+            mark_frame_full(out_buf, unique_name.c_str(), out_buffer_ID);
+            out_buffer_ID++;
```
andrerenard, line 131: Minor comment: you might want to set `out_frame == nullptr;` here, just in case the logic above didn't work for some reason it would seq fault instead of writing to an old pointer.

config/tracking_beam_test/tracking_beam_frame_to_ndisk.yaml
```diff
@@ -0,0 +1,81 @@
...
+  query_gps: true
+  require_frequency_map: true
+  query_frequency_map: true
+  allow_default_frequency_map: false
+
+dataset_manager:
+  use_dataset_broker: false
+
+beam_metadata_pool:
+  kotekan_metadata_pool: FreqIDBeamMetadata
```
andrerenard, line 34:
```suggestion
  kotekan_metadata_pool: MergedBeamMetadata
```

lib/stages/splitMergedBeamBuffer.hpp
```diff
@@ -0,0 +1,60 @@
+
+/**
+ * @file
+ * @Split the merged beam buffer to single framed buffer.
```
andrerenard, line 4: This isn't a valid doxygen string.  Use `@brief` instead

lib/stages/mergeBeamBuffer.hpp
```diff
@@ -0,0 +1,67 @@
...
+ *     @buffer_metadata MergedBeamMetadata
+ * @conf   samples_per_data_set  Int.    No. of samples.
+ * @conf   sub_frames_per_merged_frame  Int.   Number of frames to merge.
+ *
+ * @author Jing
+ *
+ *
+ */
+
+
```
andrerenard, line 45: There are a lot of extra lines in this file, and some of the other files.   `clang-format` doesn't remove them, but generally it's generally nicer not to have more than one line between sections, and no blank lines between comments and the code they describe (since that can stop the auto docs from picking them correctly).
```suggestion
 * @author Jing
 *
 */
```

andrerenard (commented):
Can you reduce the number of config files in this PR to just ones we will use in production modes?  I'd like to keep the configurations in the repo to just configurations we will use to record/read final data products.

lib/stages/BeamBufferSort.cpp
```diff
@@ -0,0 +1,295 @@
...
+	out_buf_frame0.push_back(0);
+	out_buf_start_chan.push_back(buf_start_f);
+	buf_start_f += buf_nchan;
+    }
+    
+}
+
+BeamBufferSort::~BeamBufferSort() {}
+
+void BeamBufferSort::make_freq_meta(FreqIDBeamMetadata* out_meta, uint32_t freq_id,\
```
andrerenard, line 84: Minor point, I'd generally call this `fill_` function, since we aren't actually making/allocating new memory. 
```suggestion
void BeamBufferSort::fill_freq_meta(FreqIDBeamMetadata* out_meta, uint32_t freq_id,\ 
```

lib/stages/splitMergedBeamBuffer.hpp
```diff
@@ -0,0 +1,65 @@
...
+ * @buffer out_buf Kotekan buffer of single frame beam packets.
+ *     @buffer_format Array of @c uint32_t
+ *
+ * @author Jing Santiago Luo
+ *
+ *
+ */
+
+
+class splitMergedBeamBuffer : public kotekan::Stage {
```
andrerenard, line 40: There are a lot of extra blank lines here and in other places.   Generally it's best to only use blank lines where it adds in readability, and normally no more than one at a time. 
```suggestion
 * @author Jing Santiago Luo
 */
class splitMergedBeamBuffer : public kotekan::Stage {
```

lib/stages/BeamBufferSort.hpp
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * @class BeamBufferSort
+ * @brief This stage sort the beam buffer by frequency and time and split buffer by frequencies.
+ *
+ * @par Buffers
+ * @buffer in_buf Kotekan single frame tracking beam buffer.
+ *     @buffer_format Array of @c chars
+ * @buffer out_buf Kotekan sorted tracking beam buffer.
+ *     @buffer_format Array of @c uint32_t
+ *
```
andrerenard, line 32: The `@config` section is missing here.

lib/stages/BeamBufferSort.cpp
```diff
@@ -0,0 +1,295 @@
...
+    }
+    
+    while (!stop_thread) {
+	// Receive a new frame
+        uint8_t* input = wait_for_full_frame(in_buf, unique_name.c_str(), input_frame_id);
+	if (input == nullptr)
+            break;
+        
+	// Get the metadata for each frame.
+        if (has_freq_bin) {
```
andrerenard, line 161: It's likely more flexible to setup this variable as a string like `metadata_type`, and then you can check `metadata_type == "BeamMetadata"`, etc.  It's a little more self-descriptive than having a boolean, and more flexible for other types.  Ideally the buffer would tell you what metadata it has and you would select on that, but it currently doesn't support that.

lib/stages/BeamBufferSort.cpp
```diff
@@ -0,0 +1,295 @@
...
+        else {
+	    // Frequency ID is not in the metadata. will add it to the metadata.
+            non_freq_metadata = (BeamMetadata*)get_metadata(in_buf, input_frame_id);
+            fpga_seq_start = non_freq_metadata -> fpga_seq_start;
+            const uint32_t num_freq_per_stream = Telescope::instance().num_freq_per_stream();
+            for (uint32_t f = 0; f < num_freq_per_stream; ++f) {
+                freq_bin = Telescope::instance().to_freq_id(non_freq_metadata -> stream_id, f);
+            }
+	    // Build the full Frequency IDed metadata
+	    in_metadata = (FreqIDBeamMetadata*)malloc(FreqIDBeamMeta_size);
```
andrerenard, line 176: `in_metadata` gets created for each new frame, and never gets freed.  This will create a memory leak that will eventually crash the program.   Ideally you should just create the space for this once at startup, and then free it in the destructor (if you needed it).   Since the same variable is used for both type of metadata you have to be careful to only free in the case you created it, and not in the case where you change this pointer to what is returned by `get_metadata()`

lib/stages/BeamBufferSort.hpp
```diff
@@ -0,0 +1,108 @@
...
+        uint64_t fpga_seq_start, timespec ctime, uint64_t stream_id,        
+        dset_id_t dataset_id, uint32_t beam_number, float ra, float dec,
+	uint32_t scaling);
+    /// Transfer a non-frequency id'ed metadata to frequency id'ed metadata
+    void nonfreq_meta_2_freq_meta(BeamMetadata* nonfreq_meta,  
+        FreqIDBeamMetadata* freq_meta, uint32_t freq_id);
+    ///Fill up the empty frame.
+    void fill_empty_frame(uint32_t time_idx, uint32_t freq_idx, 
+        uint64_t fpga_seq_start0, timespec ctime0, uint32_t beam_number, 
+	double ra, double dec, double scaling);
```
andrerenard, line 59: These should be private, nothing outside of this stage should be calling these functions.

andrerenard (changes requested):
I need to review the logic in a bit more depth still, but I've commented on some things you can  start addressing.   Also could you run `clang-format` on this code, address the comment regarding the extra configuration files, and also fix the CI compile errors?   That will make it easier for me to read/understand what the code is doing.

lib/stages/BeamBufferSort.cpp
```diff
@@ -0,0 +1,295 @@
...
+            fpga_seq_start = in_metadata -> fpga_seq_start;
+        }
+        else {
+	    // Frequency ID is not in the metadata. will add it to the metadata.
+            non_freq_metadata = (BeamMetadata*)get_metadata(in_buf, input_frame_id);
+            fpga_seq_start = non_freq_metadata -> fpga_seq_start;
+            const uint32_t num_freq_per_stream = Telescope::instance().num_freq_per_stream();
+            for (uint32_t f = 0; f < num_freq_per_stream; ++f) {
+                freq_bin = Telescope::instance().to_freq_id(non_freq_metadata -> stream_id, f);
+            }
```
andrerenard, line 174: This only works in the one frequency case.  For more than one frequency, you are just recording the last frequency in the set.

lib/stages/mergeBeamBuffer.hpp
```diff
@@ -0,0 +1,68 @@
...
+
```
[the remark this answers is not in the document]
luojing1211: The Stage of mergeBeamBuffer is not used in the current beam sorting work. Should we keep it?

PR #925, opened 2021-02-05
Debugging upchan kernel.

james-s-willis:


andrerenard:
I think most of the changes in this PR can be dropped, but it might be worth extracting the square wave parts and merging that in case we want to use it later.

config/tests/chime_frb_verification.yaml
```diff
@@ -16,7 +16,7 @@ log_level: info
 num_elements: 2048
 num_local_freq: 1
 num_data_sets: 1
-samples_per_data_set: 49152
+samples_per_data_set: 4608
```
andrerenard, line 19: Should we just leave this as it was?

config/tests/chime_verify_upchan.yaml
```diff
@@ -0,0 +1,330 @@
+##########################################
+#
+# FRB / Tracking beamformer verification
+#
+# Runs both GPU and CPU FRB beamformer pipelines, include re-ordering, hybrid beamformer, transposer, upchannelizer.
+# using 3840 samples for a quick run, change it to 49152 for the standard operation setting
+# Choose either testDataGen to make fake data or read from a previous baseband dump. DPDK should work too but i just didn't put it here
```
andrerenard, line 7: I think this comment needs updating.

config/tests/chime_verify_upchan.yaml
```diff
@@ -0,0 +1,330 @@
...
+    kotekan_stage: testDataGen
+    out_buf: upchan_in
+
+write:
+  kotekan_stage: rawFileWrite
+  #in_buf: upchan_in
+  in_buf: network_buffer_0
+  file_name: 'upchan_in'
+  file_ext: 'dump'
+  base_dir: '.'
```
andrerenard, line 201: Do we need to write this data out?

config/tests/chime_verify_upchan.yaml
```diff
@@ -0,0 +1,330 @@
...
+    instrument_name: chimeHFB
+
+read_gain:
+    kotekan_stage: ReadGain
+    updatable_config:
+      gain_frb: /updatable_config/frb_gain
+      gain_tracking: /updatable_config/tracking_gain
+    in_buf: network_buffer_0
+    gain_frb_buf: gain_frb_buffer
+    gain_tracking_buf: gain_tracking_buffer
```
andrerenard, line 294: Indentation is inconsistent.   We seem to mostly use 2 spaces in yaml, so let's stick to that.

config/tests/chime_verify_upchan.yaml
```diff
@@ -0,0 +1,330 @@
...
+
+write:
+  kotekan_stage: rawFileWrite
+  #in_buf: upchan_in
+  in_buf: network_buffer_0
+  file_name: 'upchan_in'
+  file_ext: 'dump'
+  base_dir: '.'
+
+cpu_simulate:
```
andrerenard, line 203: Do we actually need to CPU simulate in this config, aren't you just dumping the results of the square wave data run through the GPU?   There isn't a comparison stage, so I think we are just doing unneeded work here?

andrerenard (changes requested):

config/tests/chime_frb_verification.yaml
james-s-willis: So originally this was set to `3840` to improve the speed. I changed it to `49152` as I added a check in `hsaBeamformHFBSum.cpp` to make sure it was a multiple of 12. But I should have just set it to `4608`.

andrerenard (approved):

PR #926, opened 2021-02-08
fix(visAccumulate): check first frame for null dataset_id

nritsche:


jrs65:
@nritsche I think this could be cleaner using `std::optional`, just use that for `dset_id_in`, and modify the conditional on L278 to include the case the value isn't set, then I don't think you need any of the first frame logic.

andrerenard (changes requested):
This looks good, and works for my tests, but could you implement @jrs65's suggestion above, and then I can retest and merge.

nritsche:
> @nritsche I think this could be cleaner using `std::optional`, just use that for `dset_id_in`, and modify the conditional on L278 to include the case the value isn't set, then I don't think you need any of the first frame logic.

Can you explain how this would work?

> @nritsche I think this could be cleaner using `std::optional`, just use that for `dset_id_in`, and modify the conditional on L278 to include the case the value isn't set, then I don't think you need any of the first frame logic.

I don't see how std::optional with no value adds information to `dset_id_t::null`. In either case how would you do what you propose without entering the branch for every frame?

jrs65:
If I'm understanding the underlying issue correctly, at the moment it doesn't enter the branch at L278 (for develop branch) on the first run because it can't tell the difference between the incoming data having a null dataset ID and that it's on the first run and it hasn't seen a dataset ID before (an unfortunate consequence of previously deciding to use dset_id_t::null as a sentinel meaning "unset" in this task). Using a `std::optional<dset_id_t>` means that `dset_id_t::null` no longer needs to represent the not previously set state as the `std::optional` does that for us.

Obviously the conditional would need modifying too, to something like `if (!dset_id_in || *dset_id_in != dset_id_new)`

andrerenard (approved):
I've tested this and it appears to work as intended at the Bleien site.

Issue #464, opened 2019-07-31
Fix stages/systems which don't shutdown correctly.


andrerenard:
Currently a number of stages (for various reasons) cannot correctly shutdown:

- [ ] Baseband dumps
- [ ] DPDK
- [ ] HSA

The reasons for each of these stages is different.  However because of this setup, we cannot recover from some badly formated configs ( see for example #450 ) and must exit the system under the current design.  It also means we cannot just remove a kotekan configuration and return the system to idle - and then later provide a new config - without restarting the binary. 

The big roadblock here is the HSA and DPDK stages which don't fully support a clean exit due to API limitations.  However it should be possible with some limited retention of state information, to allow the user part of the system to start and stop dynamically.  That would allow us to return to the original model of being able to run different configs without restarting the kotekan binary.

PR #927, opened 2021-02-11
Send baseband to a receiver node instead of writing files directly

cubranic:
This change tears out the logic that existed to write baseband into files (i.e., over NFS) from the basebandReadout stage, and makes it truly just a "readout": when a baseband request arrives, the stage now copies out the requested data into an output Kotekan buffer, with event details recorded in the output frame metadata. The idea is that the output buffer will simply be forwarded to a baseband receiver node using a `bufferSend` stage.  Also adds multi-frequency support.


cubranic:
I'm converting the PR to a draft, since it hasn't yet been tested on the site cluster, and the receiver still needs to be written. However, the Python tests have been improved to where we verify exactly what the dumped baseband data is, so I'm reasonably confident that this side works correctly.

There are still a few things that could be done to improve and tidy up the code:
- it should handle output buffer using a different frame size (smaller, for efficiency and tuned to the network) that the input buffer. The code can already handle this in principle, I just need to modify it to use the output size instead of `_samples_per_data_set` when initializing `out_remaining` in line 419.
- the whole BasebandReadoutManager becomes unnecessary: the readout stage can keep the request queue within itself and register itself directly with BasebandAPIManager, as only the readout queue accesses the requests. (Once this is done, I anticipate implementing baseband part of #464 will be pretty simple.)

cubranic:
Adding @leungcalvin and @kiyo-masui so we can consider how this will affect multi-frequency streams (#689). I think the basic outline will remain the same there:
- the readout stage registers itself with the API manager for each frequency that it handles
- when a request comes in, starting with the lowest frequency, output that frequency's samples to the output buffer
- the receiver will be oblivious of the multi-frequency mode, as it will receive each frequency in separate frames, just as in the basic case.

leungcalvin:
> Adding @leungcalvin and @kiyo-masui so we can consider how this will affect multi-frequency streams (#689). I think the basic outline will remain the same there:
> 
> * the readout stage registers itself with the API manager for each frequency that it handles
> * when a request comes in, starting with the lowest frequency, output that frequency's samples to the output buffer
> * the receiver will be oblivious of the multi-frequency mode, as it will receive each frequency in separate frames, just as in the basic case.

I agree that things seem to me like they might be simpler this time around. If I understand the logic correctly, what happens is:

1) request arrives, asks for a certain span (`trigger_start_fpga`, `trigger_len_fpga`) for its `freq_id`.
2) every readout stage call `wait_for_data`, which locks `manager_lock`, waits for the data to arrive, then calls `lock_range` to lock some number of frames for that frequency. then unlocks `manager_lock` so that `add_replace_frame` can keep going. Presumably `manager_lock` does not stay locked for too long when this happens!
3) call `extract_data` on the passed `BasebandDumpData` object. `extract_data` copies the baseband data, which lives in `dump_start_frame` and `dump_end_frame` for that data's `freq_id` to an output buffer.
4) `unlock_range(dump_start_frame,dump_end_frame)` is called to release the frames holding data back into the ring buffer.

In the past we simply looped over a call to `get_data` in a loop over different `freq_idx`'s, which essentially put 2),3), and 4) in a for loop over different frequencies in the right ordering (high freq to low freq / low freq_id to high freq_id), and that seemed to work fine. If we are fine with that approach I would say we could just call `wait_for_data`, `extract_data`, and `unlock_range` in a for loop...

It seeems inelegant to lock `manager_lock` many times in a for loop. Maybe the right way to deal with this seems to be(?) to change `frames_lock` to hold recursive mutexes, which keep track of how many times they've been locked and unlocked. Then they can be locked once for each freq_id in the stream. We can then loop over 3) and 4) instead of 2),3), and 4). When the last `unlock` has been called on each frame it can be released back into the buffer.

It has been a while since I've thought about this so I'd be happy to discuss this further...

jrs65:
@cubranic I was just going to have this a quick look through, but does it need rebasing or something? There's masses of totally unrelated changes. I skimmed through about 10 of the 53 changes files before giving up!

cubranic:
That's odd @jrs65 . It must be because I had merged develop -- I then reverted it, but I think at this point I really need to rebase and clean up the commits. Hang on.

cubranic:
@jrs65 I cleaned up the commit history so that only the files I truly changed show up in the diff list.

lib/stages/basebandReadout.cpp
```diff
@@ -417,218 +339,165 @@ basebandDumpData basebandReadout::get_data(uint64_t event_id, int64_t trigger_st
...
+                out_metadata = (BasebandMetadata*)get_metadata(out_buf, out_frame_id);
 
-    char temp[256];
-    std::string username = (getlogin_r(temp, 256) == 0) ? temp : "unknown";
-    file.createAttribute<std::string>("system_user", HighFive::DataSpace::From(username))
-        .write(username);
+                out_metadata->event_id = event_id;
+                out_metadata->freq_id = freq_id;
+                out_metadata->event_start_fpga = data.trigger_start_fpga;
+                out_metadata->event_end_fpga = data.trigger_start_fpga + data.trigger_length_fpga;
```
james-s-willis, line 441: Maybe create a `set_metadata` for `BasebandMetadata`?

james-s-willis (commented):
@andrerenard this is a big PR. So the idea is we want store the baseband data in a kotekan buffer to send it to a receiver node to be written out to a file? It's similar to HFB backend.

andrerenard (approved):

PR #933, opened 2021-02-25
Add acquisition type to file writer

james-s-willis:


lib/stages/BaseWriter.hpp
```diff
@@ -142,6 +142,9 @@ class BaseWriter : public kotekan::Stage {
     // Parameters saved from the config files
     std::string instrument_name;
 
+    // Acquisition type
+    std::string acq_type;
```
jrs65, line 146:
```suggestion
    const std::string acq_type;
```

lib/stages/HFBWriter.cpp
```diff
@@ -44,7 +44,9 @@ REGISTER_KOTEKAN_STAGE(HFBWriter);
 
 HFBWriter::HFBWriter(kotekan::Config& config, const std::string& unique_name,
                      kotekan::bufferContainer& buffer_container) :
-    BaseWriter(config, unique_name, buffer_container){};
+    BaseWriter(config, unique_name, buffer_container) {
+    acq_type = "hfb";
```
jrs65, line 48:
```suggestion
    BaseWriter(config, unique_name, buffer_container),
    acq_type("hfb") {
```

lib/stages/VisWriter.cpp
```diff
@@ -43,7 +43,9 @@ REGISTER_KOTEKAN_STAGE(VisWriter);
 
 VisWriter::VisWriter(kotekan::Config& config, const std::string& unique_name,
                      kotekan::bufferContainer& buffer_container) :
-    BaseWriter(config, unique_name, buffer_container){};
+    BaseWriter(config, unique_name, buffer_container) {
+    acq_type = "corr";
```
jrs65, line 47:
```suggestion
    BaseWriter(config, unique_name, buffer_container),
    acq_type("corr") {
```

lib/utils/visFile.cpp
```diff
@@ -106,7 +107,12 @@ void visFileBundle::add_file(time_ctype first_time) {
 
     // Construct the name of the new file
     uint32_t time_since_start = (uint32_t)(first_time.ctime - acq_start_time);
-    std::string file_name = fmt::format(fmt("{:08d}_{:04d}"), time_since_start, freq_chunk);
+    std::string file_name;
```
jrs65, line 110: You might consider just leaving the chunk in the name here for uniformity. It is a bit daft as we don't use it, but maybe better than the "if hfb" special case.

jrs65: Particularly as your test code above still adds the chunk digits.

jrs65 (changes requested):

lib/utils/visFile.cpp
```diff
@@ -106,7 +107,12 @@ void visFileBundle::add_file(time_ctype first_time) {
...
+    if (acq_type == "hfb") {
```
jrs65, line 111: I'm not sure about this part. It's taken some code that has been carefully kept generic and added an if statement that dispatches on acq type. It feels a bit out of place.

I've been trying to think of ways around this, and I can see two. The first is to let the file type implementation decide what it wants to be called, and have it change the name in there. That mostly works, but requires some breaking apart of the path that the file constructors get.

What I think is a much better solution is to move the policy of the file naming up into the writer task. This both simplifies some existing things, and makes this possible. I think you do something like:
- Assign `acqname_fmt` and `filename_fmt` params into visFileBundle. These replace the hardcoded fmt strings used to currently create those names and should accept a short set of named params to create the format (those named params are provided using `fmt::arg`).
- This also allows removing the instrument name and freq chunk arguments from visFileBundle as they are only used to create the acq name.
- Define the acq and file name templates in the Writer classes. The vis and hfb writers should use the same template for now (but in both cases they will directly insert the instrument name into it), but they can diverge on the file name template.

jrs65 (changes requested):

lib/utils/visFile.cpp
james-s-willis: So have something like:
```
    // Start the acq and create the directory if required
    if (acq_name.empty()) {
        // Format the time (annoyingly you still have to use streams for this)
        acq_name =
            fmt::format("{:%Y%m%dT%H%M%SZ}_{:s}", *std::gmtime(&t), acqname_fmt);
        // Set the acq fields on the instance                                                                                                                                                                      
        acq_start_time = first_time.ctime;

        // Create acquisition directory. Don't bother checking if it already exists, just let it
        // transparently fail
        mkdir((root_path + "/" + acq_name).c_str(), 0775);
    }
```
for the `acq_name`?

What about the filename though:
```
   uint32_t time_since_start = (uint32_t)(first_time.ctime - acq_start_time);
    std::string file_name;
    if (acq_type == "hfb") {
        file_name = fmt::format(fmt("hfb_{:08d}_{:04d}"), time_since_start, freq_chunk);
    } else {
        file_name = fmt::format(fmt("{:08d}_{:04d}"), time_since_start, freq_chunk);
    }
```
Would you add the `time_since_start` to HFB and Vis Writer also?

jrs65 (changes requested):
This is looking good.

I would suggest delegating the setting the full format to the writer class, and using named parameters to allow them to be much more flexible. I think at the moment you're not really passing on a format string, just a string which adds a prefix or suffix.

Something like (for HFBWriter): `acq_fmt = "{acq_start:%Y%m%dT%H%M%SZ}_chime_hfb"` and `file_fmt = "hfb_{seconds_since_start:08d}.h5"`, which are passed directly into visFileBundle and then use a format command using named parameters against any relevant variable that only visFileBundle knows about, which is something like: `file_name = fmt::format(file_fmt, fmt::arg("acq_start", *std::gmtime(&t)), fmt::arg("seconds_since_start", time_since_start))`.

lib/stages/BaseWriter.hpp
```diff
@@ -98,7 +98,7 @@
 class BaseWriter : public kotekan::Stage {
 public:
     BaseWriter(kotekan::Config& config, const std::string& unique_name,
-               kotekan::bufferContainer& buffer_container);
```
jrs65, line 101: Why have `acq_fmt` be assigned directly in the base classes, but have `file_fmt` be passed to the constructor? It would probably be more obvious if they both behaved the same.

jrs65 (changes requested):
This looks great. One minor change requested, but nothing major.

In the long run we may want to expand the list of parameters fed into the formatting operation to allow more flexibility in the naming scheme, but I don't think anything for now.

lib/stages/BaseWriter.hpp
james-s-willis: I've made them both the same now, @jrs65

jrs65 (approved):

PR #934, opened 2021-03-01
Remove dummy gatingState from frame when dataset ID changes.

### LINKED ISSUE #919: visAccumulate sometimes but not always inserts a gating state into output stream
I think this happens because the initial dataset_id generated does not include a gatingState, but when new IDs are generated (based on a change in the incoming state) they get a dummy gatingState inserted. Fortunately the critical states list for visWriter doesn't contain gating and so things mostly work out fine.

For the file `/mnt/gong/staging/20201223T011000Z_chimestack_corr/01873247_0000.h5`, this shows the frames that have a gating state inserted:
Unknown-36

james-s-willis:
Fixes #919

### DIFF
--- a/lib/stages/visAccumulate.cpp
+++ b/lib/stages/visAccumulate.cpp
@@ -132,6 +132,9 @@ visAccumulate::visAccumulate(Config& config, const std::string& unique_name,
         }
     }
 
+    // register base dataset states to prepare for getting dataset IDs for out frames
+    register_base_dataset_states(instrument_name, freqs, inputs, prods);
+
     in_buf = get_buffer("in_buf");
     register_consumer(in_buf, unique_name.c_str());
 
@@ -154,9 +157,6 @@ visAccumulate::visAccumulate(Config& config, const std::string& unique_name,
                     "gating config={:s}].",
                     num_freq_in_frame, gating_conf.dump());
     }
-
-    // register base dataset states to prepare for getting dataset IDs for out frames
-    register_base_dataset_states(instrument_name, freqs, inputs, prods);
 
     // Register gating update callbacks
     std::map<std::string, std::function<bool(nlohmann::json&)>> callbacks;
@@ -231,8 +231,6 @@ void visAccumulate::register_base_dataset_states(
     base_dataset_states.push_back(dm.create_state<inputState>(inputs).first);
     base_dataset_states.push_back(dm.create_state<prodState>(prods).first);
     base_dataset_states.push_back(dm.create_state<eigenvalueState>(0).first);
-    base_dataset_states.push_back(
-        dm.create_state<gatingState>(*gated_datasets.at(0).spec.get()).first);
     base_dataset_states.push_back(
         dm.create_state<metadataState>(weight_type, instrument_name, git_tag).first);
 }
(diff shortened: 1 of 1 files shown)


jrs65 (changes requested):
I'm pretty sure this removes the gatingState entirely even when we're actually doing a gated acquisition and need it.

james-s-willis:
@jrs65 I have registered the gatingState on startup. Is this what you meant?

lib/stages/visAccumulate.cpp
```diff
@@ -231,6 +231,8 @@ void visAccumulate::register_base_dataset_states(
     base_dataset_states.push_back(dm.create_state<inputState>(inputs).first);
     base_dataset_states.push_back(dm.create_state<prodState>(prods).first);
     base_dataset_states.push_back(dm.create_state<eigenvalueState>(0).first);
+    base_dataset_states.push_back(
```
jrs65, line 234: If you add it to the base dataset states I think it means that when you actually have a gated stream it will have a new gatingState in its ancestors as well as this one. That sort of suggests that the input has been gated twice (even though it hasn't), so I think you probably want to add the uniform gating state elsewhere.

jrs65 (changes requested):

lib/stages/visAccumulate.cpp
james-s-willis: @jrs65 I'm unsure where to put it then. Should it be in the main loop:

```
    while (!stop_thread) {

        // Fetch a new frame and get its sequence id
        uint8_t* in_frame = wait_for_full_frame(in_buf, unique_name.c_str(), in_frame_id);
        if (in_frame == nullptr)
            break;

        // Check if dataset ID changed
        dset_id_t ds_id_in_new = get_dataset_id(in_buf, in_frame_id);
        if (!ds_id_in || ds_id_in_new != *ds_id_in) {
            ds_id_in = ds_id_in_new;

            // Register base dataset. If no dataset ID was was set in the incoming frame,
            // ds_id_in will be dset_id_t::null and thus cause a root dataset to
            // be registered.
            base_dataset_id = dm.add_dataset(base_dataset_states, *ds_id_in);
            DEBUG("Registered base dataset: {}", base_dataset_id)

            // Set the output dataset ID for the main visibility accumulation
            gated_datasets.at(0).output_dataset_id = base_dataset_id;
        }   

        int32_t* input = (int32_t*)in_frame;
        uint64_t frame_count = (get_fpga_seq_num(in_buf, in_frame_id) / samples_per_data_set);

        // Start and end times of this frame
        timespec t_s = ((chimeMetadata*)in_buf->metadata[in_frame_id]->metadata)->gps_time;
        timespec t_e = add_nsec(t_s, samples_per_data_set * tel.seq_length_nsec());
```

jrs65: Okay. Yeah, I think the reason this isn't super obvious is because of the bug that is being fixed in here.

I think in the if statement in the code above, when it detects that the dataset ID has changed, it needs to update all the gated_datasets with new dataset IDs.

Essentially the line starting with `gated_datasets.at(0).output_dataset_id ...` needs to get replaced with a loop over gated_datasets and each one needs to have its `output_dataset_id` replaced with the return value of `register_gate_dataset(spec)`.

That should take care of it all, and you can remove the gate dataset registration from the base dataset states.

jrs65 (approved):

PR #949, opened 2021-04-15
Use a `std::unordered_set` to keep track of frame ids to drop

cubranic:


james-s-willis (approved):

andrerenard:
@james-s-willis just to confirm you checked this PR works now that you've updated the unit tests?

james-s-willis:
@andrerenard yes, this works with my update to the unit tests.

PR #950, opened 2021-04-21
Kotekan Debug Tool

james-s-willis:


andrerenard:
Closing so that @Jintao-Xu can open a new PR on this branch with @james-s-willis as reviewer.

tools/debug_tool/templates/test.html
```diff
@@ -0,0 +1,250 @@
...
+    //console.log("Got response: " + message.network_buffer_0.frames);
+    //drawTree(buffers);
+  }
+
+  // Call poll() again to get the next message
+  setTimeout(() => { poll(); }, POLL_WAIT_TIME_MS);
+}
+
+// Start polling kotekan for metrics
+//poll();
```
james-s-willis, line 95: @Jintao-Xu this is the function I was talking about. It should poll the kotekan endpoint every POLL_WAIT_TIME_MS.

james-s-willis: And you would want to call your draw function where I call `drawTree(buffers)`.

PR #951, opened 2021-04-21
Updates to pulsar gating script

tristpinsm:


nritsche:
@tristpinsm could you give a short description what these changes are about. Is there an issue for this?
It looks like you have to run black (again?). Maybe you need to first update black to `19.10b0`.

tristpinsm:
These are changes that were developed based on feedback from the 26m operators when we were regularly observing pulsars, but never got merged. There's also some stuff to keep it up to date with changes in coco. I have black version 20.8b1, so I guess I have to downgrade.

python/kotekan/scripts/polyco_tools.py
```diff
@@ -32,6 +38,39 @@ def parse_parfile(fname):
...
+                line = [l.strip() for l in line.split()]
+                if not len(line) == 71:
+                    raise ValueError(
+                        "File {} contains line with unexpected number of entries.".format(
+                            fname
+                        )
+                    )
+                # skip index
+                _ = line.pop[0]
+                name = re.match("^OBS(.+)$", line.pop(0))
```
nritsche, line 48: Should we add some error handling in case this doesn't match as expected?

python/kotekan/scripts/polyco_tools.py
```diff
@@ -271,12 +310,67 @@ def update_polyco(
...
+    "--parfile-dir",
+    type=str,
+    default=PARFILE_DIR,
+    help="Directory containing pulsar parfiles.",
+)
+@click.option(
+    "--reference",
+    type=float,
+    default=None,
+    help="Reference time for observations.",
```
nritsche: `show_default="Current time."`

python/kotekan/scripts/polyco_tools.py
```diff
@@ -88,6 +88,16 @@ def mjd(unixtime):
     click.echo(unix2mjd(ts))
 
 
+@click.command()
+@click.argument('unixtime', required=False, default=None, type=float)
+def lst(unixtime):
+    """ Convert unix time to LST (in hours).
+        Will print LST now if no time is provided. """
```
nritsche: `show_default="Current time/now."`

python/kotekan/scripts/polyco_tools.py
```diff
@@ -343,6 +377,7 @@ def update_polyco(
 )
 @click.pass_context
 def import_schedule(ctx, fname, url, tempo_dir, parfile_dir, reference):
+    """NOT IMPLEMENTED YET"""
```
nritsche, line 350: What does this mean? We should remove this or prevent people from using it if it doesn't work at all.

nritsche (changes requested):
Thanks! the commit history made it much easier to understand what's going on when reviewing. However some of the commits seem to fix each other and we should squash those.
Can you either rewrite the commit history into a minimal set or write a longer commit message that we can use for a squash commit merge of this PR?

I commented on the `list.pop[0]` in the function that parses the scheduling files. I don't understand how that would not error. Did you test it?

python/kotekan/scripts/polyco_tools.py
```diff
@@ -58,7 +58,7 @@ def mjd(unixtime):
     "--end-time",
     type=float,
     default=None,
-    help="Specify an end time to generate polyco / match segment with maximum overlap",
+    help="Specify an end time to generate polyco / match segment with maximum overlap. Defaults to start time + 1.",
```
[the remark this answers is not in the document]
tristpinsm: It's specified by the `time-spec` argument, and the default is MJD.

tristpinsm:
> Can you either rewrite the commit history into a minimal set or write a longer commit message that we can use for a squash commit merge of this PR?
I had to do a pretty extensive rebase to get this up to date with `develop`, so the commit history might be messed up, and I intended it to be squashed merged.

> I commented on the `list.pop[0]` in the function that parses the scheduling files. I don't understand how that would not error. Did you test it?
No, the schedule importing functionality is half-baked, but I don't think that should hold things back. This code has already been languishing for way too long, and getting that working is pretty low-priority.

I will fix the docstrings etc. Thanks!

python/kotekan/scripts/polyco_tools.py
```diff
@@ -265,7 +262,7 @@ def update_polyco(
                 check_call(["at", end_str], stdin=tmpf)
         else:
             print("Sending update to {}...".format(url))
-            r = requests.post(url, json=update)
+            r = requests.post(url + "/update-pulsar-gating", json=update)
```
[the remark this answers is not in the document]
tristpinsm: Now you specify the `coco` URL, but the endpoints are fixed. It's used for two different endpoints, so you would need to make them separately configurable, and that doesn't seem useful.

python/kotekan/scripts/polyco_tools.py
tristpinsm: I could make it just do nothing and print that out instead.

python/kotekan/scripts/polyco_tools.py
```diff
@@ -144,22 +144,30 @@ def update_polyco(
     url,
     tempo_dir,
 ):
-    """Generate a gating polyco update from a parfile and send to kotekan."""
+    """Generate a gating polyco update from a parfile and send to kotekan.
+    Required arguments are the path to the parfile and the start time for the polyco
+    (enter 'now' to use current time minus 0.2 days).
```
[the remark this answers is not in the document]
tristpinsm: Click doesn't accept a `help` parameter for arguments.

python/kotekan/scripts/polyco_tools.py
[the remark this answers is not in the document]
tristpinsm: this doesn't seem to work... In the docs they only mention its use for user input options, but I don't know if that implies it's not meant to work for other types. In any case I don't see it show up in the help.

python/kotekan/scripts/polyco_tools.py
nritsche: You can `raise NotImplementedError`

tristpinsm:
yes, I think so.

anjakefala (changes requested):
Please remove commits for the untested code!

PR #953, opened 2021-04-23
Fix problem with tracking beam gain updates

### LINKED ISSUE #935: majority of channels for CHIME/Pulsar beamformed voltages are zero-weighted
On 12 March 2021, a few of us noticed that the majority of data being transmitted to CHIME/Pulsar showed signs of faulty frequency data. An example of what we are currently seeing is shown in the histogram attached to this issue, which shows a histogram of time-averaged intensity values across the 1,024 channels we receiver at CHIME/Pulsar for `B0950+08`; the left-hand ("bad") panel was recorded on 12 March, and the right-hand ("good") panel was recorded on 10 March (at ~1am ET).

While still under investigation, we have so far determined the following:

1. the first data scan to show these issues was recorded on 10 March at ~3:30pm ET;
2. this issue is present for beams `[1,2,3,6,7,8]`, but not for `[0,4,5,9]`;
3. @emmanuelfonseca reverted gain solutions to those recorded on 9 March, to check and see if the solutions are related to this issue; however, the issue persisted despite switching to older gains that produced normal/good data products;
4. @emmanuelfonseca hardbooted at least one of the nodes receiver the affected beams, to check and see if this was somehow related to drop packets not being reflected in Grafana metrics; this restart did not alleviate the issue in any clear way.

B0950_59283_59285_channel_average_hist

andrerenard:
Fixes a logical error introduced by #811.  Basically the code implicitly required all the beams to be updated every time, but #811 changed it so only one beam was updated and the values of the other beams were just whatever was currently in that frame.  Because of the fact there are 10 beams, the buffer depth for the gains was 5, and all gains are loaded at once on startup, this caused the issue to get mostly hidden - the system just ended up using old gains for some beams. 

The hotfix corrects this by keeping a local set of gains and copying them into each new output frame.  

Closes #935

### DIFF
--- a/lib/stages/ReadGain.cpp
+++ b/lib/stages/ReadGain.cpp
@@ -78,6 +78,13 @@ ReadGain::ReadGain(Config& config, const std::string& unique_name,
     register_producer(gain_tracking_buf, unique_name.c_str());
     update_gains_tracking = false;
 
+    // Create a space to store gains for all the tracking beams
+    // Initial values will be set by the call back functions below on startup
+    tracking_beam_gains = (float*)malloc(gain_tracking_buf->frame_size);
+    if (tracking_beam_gains == nullptr) {
+        throw std::runtime_error("Could not allocate memory in ReadGain for tracking gains");
+    }
+
     using namespace std::placeholders;
 
     // listen for gain updates FRB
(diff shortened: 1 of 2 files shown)


james-s-willis (approved):
Looks good to me

PR #954, opened 2021-04-29
Count all frames in gated datasets

tristpinsm:
Should we also copy over `rfi_total`?


jrs65:
> Should we also copy over `rfi_total`?

Yeah, I think so.

jrs65 (approved):

jrs65 (approved):

PR #955, opened 2021-04-29
Timeout waiting for full frame in writer

tristpinsm:


lib/stages/BaseWriter.cpp
```diff
@@ -95,19 +96,24 @@ void BaseWriter::main_thread() {
 
     frameID frame_id(in_buf);
 
+    const int tsec = (int)acq_timeout;
+    const timespec timeout = {tsec, (long)((acq_timeout - tsec) * 1e9)};
```
jrs65, line 100: It looks like you could replace this with a call to `double_to_ts`

lib/stages/BaseWriter.cpp
```diff
@@ -95,19 +96,24 @@ void BaseWriter::main_thread() {
...
+    int status;
```
jrs65, line 101: This should be defined within the loop to limit the scope, I would just do `auto status = wait_....`

jrs65 (changes requested):

jrs65 (approved):

PR #958, opened 2021-05-05
Add a statistic tracker for min/max/avg/std

Jintao-Xu:
- Deque is used for min and max with O(1) processing time
- Advanced calculation formulas are used for avg and std_dev with O(1) processing time
- Unit tests have been added to test min, max, avg and std_dev

### DIFF
--- a/lib/utils/visUtil.cpp
+++ b/lib/utils/visUtil.cpp
@@ -8,6 +8,7 @@
 #include <regex>     // for sregex_token_iterator, match_results<>::_Base_type, _NFA, regex
 #include <sstream>   // for basic_stringbuf<>::int_type, basic_stringbuf<>::pos_type, basic_st...
 #include <stdexcept> // for runtime_error, invalid_argument
+#include <limits>    // for std::numeric_limits<int>::max() and std::numeric_limits<int>::min()
 
 using nlohmann::json;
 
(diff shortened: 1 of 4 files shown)


james-s-willis (approved):
This looks really good, @Jintao-Xu! It might be useful to have your unit tests included in the kotekan repo and for them to run as CI tests for regression purposes. What do you think @andrerenard ?

lib/utils/visUtil.cpp
```diff
@@ -238,6 +238,97 @@ double movingAverage::average() {
...
+    }
+
+    mean = this->get_avg();
+
+    for (size_t i = 0; i < count; i++) {
+        index = (front + i) % buf_size;
+        standardDeviation += pow(rbuf[index] - mean, 2);
+    }
+
+    return sqrt(standardDeviation / count);
```
andrerenard, line 329: For most of what we are measuring using the sample standard deviation will make more sense.
```suggestion
    return sqrt(standardDeviation / (count - 1));
```

andrerenard (changes requested):
This is looking great, thanks!  I've added some notes inline. 

While I know I said we'd just compute the mean/std values over the array when asked, that was mostly to get started, and there is a better way to do these.  If you keep track of partial statistics as you add (and remove) samples, you can convert your mean/std functions to run in O(1) time, at only the expense of some O(1) operations in the `add_sample()` function.   For example for the average, you can just track the sum of all values in the array, and add/subtract from the sum when you add a new value.  Then in the avg function just return `sum/count`.  A similar approach can be done for the standard deviation:  https://jonisalonen.com/2014/efficient-and-accurate-rolling-standard-deviation/

You can do something similar with the min/max functions, but that's a more complex exercise, since it involves more complex data structures.  The algorithms you would most likely want to use are `ascending minima` and `descending maxima` using a heap or deque.  It's not obvious for our relatively simple case with small values of N that this is worth it, but it's an interesting problem to think about.

andrerenard:
And I agree with @james-s-willis we should add some `boost` unit tests for the class: http://lwlab.dunlap.utoronto.ca/kotekan/dev/dev_unittests.html#testing-a-c-function   See for example: https://github.com/kotekan/kotekan/blob/develop/tests/boost/test_truncate.cpp

Jintao-Xu:
@andrerenard @andrerenard Thanks for your suggestions! I'll work on them and produce better code. Also will add unit tests as well.

Jintao-Xu:
- Used advanced formula provided by Andre to compute std_dev.
- Added deque to keep tracking min and max values

lib/utils/visUtil.hpp
```diff
@@ -527,6 +528,110 @@ class movingAverage {
     bool initialised = false;
 };
 
+/**
+ * @class SlidingWindowMinMax
```
andrerenard, line 533: I think class is is based off this [work](https://www.nayuki.io/page/sliding-window-minimum-maximum-algorithm) correct?  If so, you should include a reference to the original blog post/code where the original ideas/code come from.

lib/utils/visUtil.hpp
```diff
@@ -527,6 +528,110 @@ class movingAverage {
...
+ **/
+class SlidingWindowMinMax {
+
+public:
+    /**
+     * @brief Get the current minimum value from the front of minDeque.
+     *
+     * @return The current minimum value.
+     **/
+    double getMinimum();
```
andrerenard, line 545: Since this isn't being included as an external library but instead being modified to become a part of our main source code, it should be formatted to match our coding standards. 
```suggestion
    double get_minimum();
```

andrerenard (changes requested):
Thanks for the updates!  I need to still convince myself the sliding min/max works in all cases, but overall this looks really good.   Some comments inline.

lib/utils/visUtil.hpp
```diff
@@ -527,6 +529,112 @@ class movingAverage {
     bool initialised = false;
 };
 
+// The following class is modified based on
+// https://www.nayuki.io/page/sliding-window-minimum-maximum-algorithm
+/**
+ * @class SlidingWindowMinMax
+ *
+ * @brief Use two deques to keep tracking minimum and maximum values.
+ **/
```
andrerenard, line 538: Keep all the comments in the same block.  Doxygen allows for extended description details in the main body of the comment block. 
```suggestion
/**
 * @class SlidingWindowMinMax
 *
 * @brief Use two deques to keep tracking minimum and maximum values.
 *
 * This class is based on a modified version of:
 * https://www.nayuki.io/page/sliding-window-minimum-maximum-algorithm
 **/
```

andrerenard (approved):
This all looks good thanks!  I will get this merged in now.  There is actually an interesting aside question here of what's more efficient for the min/max in different use cases.  The worse case insertion cost here is O(n), but if you are reading the currently min/max after each sample this is more much more optimal than doing a brute force array read.   However, if you only check the min/max periodically, it might be more efficient to just do the brute force approach and save on insertion costs.   I don't expect the difference will matter practically, but as an extension someday it might be worth having two modes for min/max as an option.

PR #962, opened 2021-05-27
Pipeline Viewer

Jintao-Xu:
The latest update:

- fixed mismatched nodes and wrong links
- fixed overlap issue
- added zooming feature
- printed utilization for every buffer
- updated utilization dynamically


tools/debug_tool/templates/pipeline_tree.html
```diff
@@ -0,0 +1,275 @@
+<!DOCTYPE html>
```
james-s-willis, line 1: This file needs tidying up. I would removed the unused commented out regions and add more documentation to the code.

tools/debug_tool/templates/pipeline_tree.html
```diff
@@ -0,0 +1,275 @@
...
+    // Replace "&#39;" chars as single quote "'"
+    var rawData =  "{{data.json()}}".split('&#39;').join('"');
+    // Add double quotes to None elements ('None') then parse json into javascript object
+    var buffers = jQuery.parseJSON(rawData.split('None').join('"None"'));
+    // Get buffer names
+    var bufNames = Object.keys(buffers);
+        
+    console.log(buffers);
+
+    d3.json("{{url_for('static', filename='chris.json')}}", function (error, graph) {
```
james-s-willis, line 68: So we don't actually use the `chris.json` file here. So i think that part can be removed

james-s-willis (changes requested):
This is looking good. Just need to tidy files up and remove ones which are not needed.

james-s-willis (approved):
This looks good to me. Thanks for making the changes @Jintao-Xu! @andrerenard, do you want to review it before it gets merged?

james-s-willis:
Oh I just missed that some of the CI tests still need fixing, @Jintao-Xu.

Jintao-Xu:
> Oh I just missed that some of the CI tests still need fixing, @Jintao-Xu.

I'm not sure what is wrong in that test, and it does not provide much information. @james-s-willis 

> ##[debug]Loading env
Run OPTS=(--rm --mount type=bind,src=$(pwd),target=/code/kotekan -w /code/kotekan/ ${IMG_CORE}:develop)
##[debug]/usr/bin/bash -e /home/runner/work/_temp/24ee463c-1c93-4769-85e0-c8bf885abee4.sh
would reformat /code/kotekan/tools/debug_tool/debug_server.py
Oh no! 💥 💔 💥
1 file would be reformatted, 73 files would be left unchanged.
Error: Process completed with exit code 1.
##[debug]Finishing: Run black

james-s-willis:
Oh that is an error from a python formatting tool called, `black`. Try running that on your code.

nritsche:
If you expand the first line, you can see that the command that failed is `black --exclude docs --check .`:
```
Run OPTS=(--rm --mount type=bind,src=$(pwd),target=/code/kotekan -w /code/kotekan/ ${IMG_CORE}:develop)
  OPTS=(--rm --mount type=bind,src=$(pwd),target=/code/kotekan -w /code/kotekan/ ${IMG_CORE}:develop)
  docker run "${OPTS[@]}" black --exclude docs --check .
  shell: /usr/bin/bash -e {0}
  env:
    IMG_CORE: docker.pkg.github.com/kotekan/kotekan/kotekan-core
    IMG_IWYU: docker.pkg.github.com/kotekan/kotekan/kotekan-iwyu
    PYTEST_TIMEOUT: 60
    GIT_DIFF: 
    GIT_DIFF_FILTERED: 
    MATCHED_FILES: 
##[debug]/usr/bin/bash -e /home/runner/work/_temp/24ee463c-1c93-4769-85e0-c8bf885abee4.sh
would reformat /code/kotekan/tools/debug_tool/debug_server.py
Oh no! 💥 💔 💥
1 file would be reformatted, 73 files would be left unchanged.
```

You can let `black` format the python code in this repository by running `black --exclude-docs .` (see http://lwlab.dunlap.utoronto.ca/kotekan/dev/dev_style_guide_code.html#automatic-code-formatting).

Jintao-Xu:
@james-s-willis @nritsche That works! Thanks for your help.👍

tools/debug_tool/templates/pipeline_tree.html
```diff
@@ -0,0 +1,242 @@
...
+
+.label {
+  fill: rgb(0, 0, 0);
+  font-family: sans-serif;
+  font-size: 15;
+}
+
+</style>
+</head>
+<body>
```
andrerenard, line 30: Generally I put scripts in the `<head>` not the `<body>`

tools/debug_tool/debug_server.py
```diff
@@ -0,0 +1,70 @@
...
+
+    # Show directory contents
+    files = os.listdir(abs_path)
+    return render_template("files.html", files=files)
+
+
+# Update kotekan metrics
+@app.route("/", defaults={"path": ""})
+@app.route("/update", methods=["GET", "POST"])
+def update():
```
andrerenard, line 54: Is there a way to make this more generic. We'll want to have lots of endpoints passed through to this server, so having to add them all manually will be annoying.   Can we set this up so we have something like `/kotekan_api/<endpoint>` gets mapped to read and return `{KOTEKAN_ADDRESS}/<endpoint>`

tools/debug_tool/debug_server.py
```diff
@@ -0,0 +1,70 @@
...
+
+app = Flask(__name__)
+app.config["CORS_ORIGINS"] = ["*"]
+CORS(app, support_credentials=True)
+
+KOTEKAN_ADDRESS = "http://localhost:12048"
+
+# Display kotekan pipeline
+@app.route("/", defaults={"path": ""})
+@app.route("/debug_tool/pipeline_tree.html", methods=["GET", "POST"])
```
andrerenard, line 14: Can we just serve this file via the `dir_listing()` function below?   I'd like this webserver to be fairly generic so we can just add more html/css/javascript files to a directory and not have to add custom functions to this pipeline.

andrerenard (changes requested):
All the logic here is looking good, but before we get too far into development, I'd just like to move somethings around and generalize the code a bit.   See inline comments.

andrerenard:
One other thing I'd like to add quickly to this PR would be a new page in the "User's Guide" section of the documentation to explain how to start this server and use it.  e.g. `docs/sphinx/user/dev_pipeline_viewer.rst`

tools/debug_tool/static/pipeline_viewer.js
```diff
@@ -0,0 +1,232 @@
...
+
+    init_svg() {
+        var width = 960, height = 500;
+
+        this.#d3cola = cola.d3adaptor(d3)
+            .linkDistance(80)
+            .size([width, height]);
+
+        // Add a svg section and employ zooming
+        this.#svg = d3.select("body").append("svg")
```
andrerenard, line 85: Can you make the `"body"` argument a variable that gets passed to `init_svg()` from the main html file.   e.g. I would like to be able to append this svg more generically to other html elements.

tools/debug_tool/templates/pipeline_tree.html
```diff
@@ -0,0 +1,28 @@
...
+    <a href="../index.html">cola.js home</a>
+<h1>Kotekan Pipeline</h1>
+<script>
+    get_data().then(function (buffers) {
+        var pipeline = new PipelineViewer(buffers);
+        pipeline.init_svg();
+        pipeline.parse_data();
+        pipeline.create_objs();
+        pipeline.start_buff_ult();
```
andrerenard, line 23: To simplify this for the user of the `PipelineViewer` object, I would provide just one function call that initializes everything by calling each of these functions internally.

tools/debug_tool/static/pipeline_viewer.js
```diff
@@ -0,0 +1,232 @@
...
+    #node;
+    #label;
+
+    constructor(buffers) {
+        this.buffers = buffers;
+        this.bufNames = Object.keys(this.buffers);
+    }
+
+    init_svg() {
+        var width = 960, height = 500;
```
andrerenard, line 78: I would make these optional arguments to the function, so they can be adjusted by each html page (if this class gets used in more than one context).

tools/debug_tool/debug_server.py
```diff
@@ -0,0 +1,56 @@
+from flask import Flask, render_template, request, send_file, abort
+from flask_cors import CORS
+from requests import get
+import os
+
+app = Flask(__name__)
+app.config["CORS_ORIGINS"] = ["*"]
+CORS(app, support_credentials=True)
+
+KOTEKAN_ADDRESS = "http://localhost:12048"
```
andrerenard, line 10: This should be an optional argument to the python script, with `http://localhost:12048` as the default.  It will likely be the case that the user will want to run the server locally, but debug a kotekan instance on a remote server.

andrerenard (changes requested):
The structure of this is looking much better!  I just have a few minor details I'd like to adjust, and then it should be done.

tools/debug_tool/static/pipeline_viewer.js
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    start_viewer() {
+        this.init_svg();
+        this.parse_data();
+        this.create_objs();
+        this.start_buff_ult();
+    }
+
+    init_svg(width = 960, height = 500) {;
```
andrerenard, line 83: These optional parameters should be at the level of `start_viewer()` so that the user on the HTML side can adjust them if needed, same with `margin` on `create_objs()`

andrerenard (changes requested):
One small comment to address, but otherwise this looks good.

andrerenard (approved):

PR #964, opened 2021-05-31
Jl/beamsort

luojing1211:
This is the new beamSort stage, with controllable dump size. Feel free to comment on it. An operation diagram is included. 
BeamBufferSort

Question: Since we have the beam sorted stage, do we need the mergeBeamBuffer stage?


lib/stages/BeamBufferSort.cpp
```diff
@@ -0,0 +1,414 @@
...
+            non_freq_metadata = (BeamMetadata*)get_metadata(in_buf, input_frame_id);
+            fpga_seq_start = non_freq_metadata -> fpga_seq_start;
+            const uint32_t num_freq_per_stream = Telescope::instance().num_freq_per_stream();
+            for (uint32_t f = 0; f < num_freq_per_stream; ++f) {
+                freq_bin = Telescope::instance().to_freq_id(non_freq_metadata -> stream_id, f);
+            }
+	    // Build the full Frequency IDed metadata
+	    in_metadata = temp_freq_metadata;
+	    nonfreq_meta_2_freq_meta(non_freq_metadata, in_metadata, freq_bin);
+        }
```
james-s-willis, line 216: So sometimes the incoming beam metadata doesn't contain a frequency bin?

lib/stages/BeamBufferSort.cpp
```diff
@@ -0,0 +1,414 @@
...
+            freq_bin = in_metadata -> frequency_bin;
+            fpga_seq_start = in_metadata -> fpga_seq_start;
+        }
+        else {
+	    // Frequency ID is not in the metadata. will add it to the metadata.
+            non_freq_metadata = (BeamMetadata*)get_metadata(in_buf, input_frame_id);
+            fpga_seq_start = non_freq_metadata -> fpga_seq_start;
+            const uint32_t num_freq_per_stream = Telescope::instance().num_freq_per_stream();
+            for (uint32_t f = 0; f < num_freq_per_stream; ++f) {
+                freq_bin = Telescope::instance().to_freq_id(non_freq_metadata -> stream_id, f);
```
james-s-willis, line 211: What happens when `num_freq_per_stream` is `> 1`? The value of `freq_bin` will get overwritten after each iteration of the loop.

lib/stages/BeamBufferSort.cpp
```diff
@@ -0,0 +1,414 @@
...
+        }
+	// Copy in frame to the queue.
+        //INFO("In queue {:d} {:d}", time_offset, freq_bin);
+	FreqIDBeamMetadata* sub_frame_metadata = (FreqIDBeamMetadata*)&sort_queue[time_offset][freq_bin][0];
+	memcpy(sub_frame_metadata, in_metadata, FreqIDBeamMeta_size);
+	// copy data
+	uint8_t* sub_frame_data = &sort_queue[time_offset][freq_bin][FreqIDBeamMeta_size];
+	memcpy(sub_frame_data, input, in_buf->frame_size);
+        // Debug line
+	FreqIDBeamMetadata* debug_meta = (FreqIDBeamMetadata*)&sort_queue[time_offset][freq_bin][0];
```
james-s-willis, line 392: Isn't `debug_meta` equal to `sub_frame_metadata`? Also, should that `INFO` not be `DEBUG`?

james-s-willis (commented):
First round of comments.

lib/stages/BeamBufferSort.cpp
```diff
@@ -0,0 +1,414 @@
...
+    }
+    // Free memory
+    // Free queue
+    for (int i = 0; i < wait_nframes; ++i){
+        for (uint32_t j = 0; j < total_freq_chan; j++){
+            free(sort_queue[i][j]);
+        }
+    // Free temp freqid metadata
+    free(temp_freq_metadata);
+    }
```
james-s-willis, line 412: Can this not be done in the destructor?

lib/stages/BeamBufferSort.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+
+/**
+ * @class BeamBufferSort
+ * @brief This stage sort the beam buffer by frequency and time and split buffer by frequencies.
+ *
+ * @par Buffers
+ * @buffer in_buf Kotekan single frame tracking beam buffer.
+ *     @buffer_format Array of @c chars
+ * @buffer out_buf Kotekan sorted tracking beam buffer.
```
james-s-willis, line 30: Can describe how there can be multiple output buffers here.

lib/stages/mergeBeamBuffer.cpp
```diff
@@ -0,0 +1,138 @@
...
+        uint32_t out_frame_data_pos = out_frame_metadata_pos + sizeof(FreqIDBeamMetadata);
+        // Make sure the position does not pass the frame size.
+        assert(out_frame_data_pos < (uint32_t)out_buf->frame_size);
+        // Get the memory block for the data of of a subframe.
+        uint8_t* sub_frame_data = &out_frame[out_frame_data_pos];
+        // copy data from in frame to sub_out_frame
+        memcpy(sub_frame_data, in_frame, in_buf->frame_size);
+
+// Debug code. Check if the subframe matedata is the same.
+#ifdef DEBUGGING
```
james-s-willis, line 119: Is the `DEBUGGING` macro defined by `kotekan`? Or do you define it yourself?

lib/stages/nDiskFileWrite.hpp
```diff
@@ -34,11 +34,13 @@
  * @conf num_disks      Int , the number of drives to read from
  * @conf disk_base      String, the path to the mounted drives
  * @conf disk_set       String, the disk name.
+ * @conf file_ext       String, Default 'vdif', the extenstion of the output file.
```
james-s-willis, line 37:
```suggestion
 * @conf file_ext       String, Default 'vdif', the extention of the output file.
```

lib/stages/splitMergedBeamBuffer.cpp
```diff
@@ -0,0 +1,128 @@
...
+	    sub_frame_pre_frame = in_metadata->sub_frame_pre_frame;
+            sub_frame_metadata_size = in_metadata->sub_frame_metadata_size;
+	    sub_frame_data_size = in_metadata->sub_frame_data_size;
+	    DEBUG2("Number of subframes {:d}, subframe metasize {:d}, subframe data size {:d}:\n",
+                   in_metadata->sub_frame_pre_frame, in_metadata->sub_frame_metadata_size,
+                   in_metadata->sub_frame_data_size);
+	}
+	else{
+	    if (sub_frame_pre_frame == 0){
+	        ERROR("'sub_frame_pre_frame' is required in the configuration when the merged buffer metadata is not provided.");
```
james-s-willis, line 68: I think you can just use `FATAL_ERROR` here without the need for `exit(-1)`.

lib/stages/splitMergedBeamBuffer.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+            // Allocated the new metadata for the output buffer
+            allocate_new_metadata_object(out_buf, out_buffer_ID);
+            FreqIDBeamMetadata* out_metadata = (FreqIDBeamMetadata*)get_metadata(out_buf, out_buffer_ID);
+            // Get pointer to the beam metadata with frequency
+            FreqIDBeamMetadata* sub_frame_metadata =
+                (FreqIDBeamMetadata*)&in_frame[sub_frame_start];
+	    DEBUG2("Frame start {:d}", sub_frame_start);
+            // Fill up the out put buffer metadata.
+            out_metadata->fpga_seq_start = sub_frame_metadata->fpga_seq_start;
```
james-s-willis, line 105: Will a `memcpy` not work here?

james-s-willis (changes requested):
Do we have any unit tests in place for these changes? The logic in `BeamBufferSort` is quite complicated and could go wrong if not tested for edge cases. For `HFBAccumulate` I created a test that had missing frames to check that the integration could handle it. Could you do something similar? For example you could have the initial frame and miss `wait_nframes` frames and then add a frame with `fpga_seq_start` advanced in time. Another would be to receive only the same frequency over multiple time steps.

Also, are all config files needed? Or are some of them for testing and debugging purposes?

config/beam_record_sorted.yaml
```diff
@@ -0,0 +1,124 @@
...
+    # sort_buffer_5:
+          # kotekan_buffer: standard
+    # sort_buffer_6:
+          # kotekan_buffer: standard
+    # sort_buffer_7:
+          # kotekan_buffer: standard
+    # sort_buffer_8:
+          # kotekan_buffer: standard
+    # sort_buffer_9:
+          # kotekan_buffer: standard
```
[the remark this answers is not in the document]
luojing1211: Those are for splitting frequencies to different disks. The number of buffers depends on the number of disks to use.

lib/stages/BeamBufferSort.cpp
```diff
@@ -0,0 +1,414 @@
...
+    subframe_time_nsec = time_resolution * samples_per_data_set * 1e9;
+    // Queue frame time resolution
+    // TODO this will be changed. 
+    queue_frame_resolution_nsec = subframe_time_nsec;
+    time_resolution_nsec = time_resolution * 1e9;
+    // Number of data samples a queue frame represents.
+    // TODO this will be changed.
+    queue_frame_represent_size = samples_per_data_set; 
+    // Time delay vector for each channel
+    std::vector<double> time_delay(total_freq_chan, 0.0); // need to change here
```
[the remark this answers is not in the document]
luojing1211: This was put there for the gated pulsar. But I think I can remove it now

lib/stages/BeamBufferSort.cpp
luojing1211: It needs a frequency mapping, but some other stage may add a frequency bin to it. For example the mergebeambuffer stage. So it would be better to have a check.

lib/stages/BeamBufferSort.cpp
luojing1211: That is something that needs to fix. If the num_freq_per_stream > 1. I don't know how the data is arranged. But for CHIME, this should be ok

luojing1211:
Most of the config files are for debugging purposes.  I will remove them before merging. Can we focus our review on the code first. After the code changes, I will have to test the new changes by using those config files.

lib/stages/BeamBufferSort.cpp
luojing1211: temp_freq_metadata is defined in the `main_thread()`. I don't think it can be done in the destructor.

luojing1211:
Quick question, do we still want to keep the BeamBufferMerge stage? Since we are using the BeamBufferSort stage.

andrerenard:
@luojing1211 Could you bring this branch upto data with develop, and fix the CI build and lint issues?  Let me know if you run into any issues with the linting (`iwyu` can be a tricky one to get working).

andrerenard (changes requested):
Can you reduce the number of config files to just the ones used in production?  I'd like to keep the active configuration files just to production and/or test cases with a pass/fall result in `config/tests/`.   Also which configuration file are you currently using for CHIME data capture?  T

lib/utils/util.c
```diff
@@ -50,6 +50,27 @@ void make_rfi_dirs(int streamID, const char* write_to, const char* time_dir) {
...
+
+int make_dir(const char* dir_name) {
+    // Make the data dir from the absolute path
+    int err = 0;
+    err = mkdir(dir_name, 0777);
+
+    if (err != -1) {
+        return err;
+    }
+    if (errno == EEXIST) {
```
andrerenard, line 62: We should adjust the logic here.  Do we care if the directory already exists?  Perhaps we could pass in a flag to indicate if we should fail if the directory exists or not?   At the very least if we want to exit the program regardless of if the failure was EEXIST or something else, we should log both cases.

Also it looks like a lot of the code in this file uses the old style logging.  We should use the `(INFO|WARN|ERROR)_F`, and `FATAL_ERROR_F()` macros instead.  At least in your new functions.

lib/utils/util.h
```diff
@@ -40,6 +40,7 @@ extern "C" {
 #endif
 
 void make_rfi_dirs(int streamID, const char* write_to, const char* time_dir);
+int make_dir(const char* dir_name);
```
andrerenard, line 43: I'm not sure it makes sense to return the error code, if when we encounter an error we exit immediately?  If you want the caller to make this judgement (which is reasonable), then don't exit the program in the function.

lib/stages/rawFileRead.cpp
```diff
@@ -67,8 +67,13 @@ void rawFileRead::main_thread() {
             snprintf(full_path, full_path_len, "%s/%s_%s_%07d.%s", base_dir.c_str(), hostname,
                      file_name.c_str(), file_num, file_ext.c_str());
         } else {
-            snprintf(full_path, full_path_len, "%s/%s_%07d.%s", base_dir.c_str(), file_name.c_str(),
-                     file_num, file_ext.c_str());
+            if (ndisk_wrote)
```
andrerenard, line 70: Why don't you loop over the num_disks?  If you are reading nDisk written files, don't you want to read from each disk?   I would also change this variable name to `ndisk_mode`.

lib/stages/CMakeLists.txt
```diff
@@ -48,6 +51,8 @@ add_library(
     bufferDelay.cpp
     BufferSplit.cpp
     InputSubset.cpp
+    mergeBeamBuffer.cpp
+    BeamBufferSort.cpp
```
andrerenard, line 57: I'm also really bad at this, but we should try to be consistent with naming new stages with a starting capital letter.   (It will make it easier when we go back in and rename all the old lower case stages).

lib/stages/oneDiskVDIFWrite.hpp
```diff
@@ -0,0 +1,107 @@
...
+ * @conf   ref_year                 Double. The reference year for VDIF header. It only accepts
+ *                                  the integer year or the half year (e.g. 2020 or 2020.5)
+ *
+ * @author Jing Santiago Luo
+ *
+ *
+ */
+
+
+class oneDiskVDIFWrite : public kotekan::Stage {
```
andrerenard, line 65: Remove extra spaces
```suggestion
 * @author Jing Santiago Luo
 */
class oneDiskVDIFWrite : public kotekan::Stage {
```

lib/stages/oneDiskVDIFWrite.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+#include <stdint.h> // for uint32_t
+#include <string>   // for string
+#include <vector>   // for vector
+
+
+using std::vector;
+
+
+/**
```
andrerenard, line 26: Remove extra spaces
```suggestion
 */

#ifndef ONE_DISK_VDIF_WRITE
#define ONE_DISK_VDIF_WRITE

#include "BeamMetadata.hpp" // for BeamMetadata
#include "Config.hpp"       // for Config
#include "Stage.hpp"
#include "bufferContainer.hpp" // for bufferContainer
#include "vdif_functions.h"   // for VDIFHeader

#include <stdint.h> // for uint32_t
#include <string>   // for string
#include <vector>   // for vector

using std::vector;

/**
```

lib/stages/oneDiskVDIFWrite.cpp
```diff
@@ -0,0 +1,338 @@
...
+	snprintf(data_log_file, full_path_len, "%s/%s.txt", data_dir_name, "data_log");
+	j_file = fopen(data_log_file, "w");
+    }
+    
+        
+    // allocate the vdif buffer 
+    vdif_frame_data_size = vdif_samples_per_frame * vdif_freq_per_frame * num_pol;
+    vdif_frame_size = vdif_frame_header_size + vdif_frame_data_size;
+    INFO("VDIF size {:d} {:d}", vdif_frame_data_size, vdif_frame_size);
+    vdif_frame = (uint8_t*) malloc(vdif_frame_size);
```
andrerenard, line 152: Check that `malloc()` returned valid memory, e.g. `!= nullptr`

lib/stages/oneDiskVDIFWrite.cpp
```diff
@@ -0,0 +1,338 @@
...
+	    for (uint32_t ii = 0; ii < vdif_freq_per_frame; ii++){
+                // Copy data to payload
+	        // The input buffer is order as freq0(n_sample * n_pol) freq1(n_sample* n_pol) ...
+	        // But the VDIF is order as freq0 (1_sample * n_pol) freq1(1_sample * n_pol) ...
+                src_start = (freq_offset + ii) * in_buf_frame_size + (time_offset + tt) * num_pol  + in_metadata -> sub_frame_metadata_size;
+                target_start = (ii * num_pol) + tt * num_pol * vdif_freq_per_frame;
+	        for (uint32_t jj = 0; jj < num_pol; jj++){
+	            //if (ii >= 0 && ii < 20){		
+	            //    if (tt > 620 && tt < 630 && time_offset > 49000 && time_offset < 51000)
+	   	    //        INFO("target_index {:d} src_index {:d} tt {:d} ii {:d} freq offset {:d} time offset {:d} frameszie {:d}", target_start +jj, src_start + jj, tt, ii, freq_offset, time_offset, in_buf_frame_size);
```
andrerenard, line 276: You should either make these commented out `INFO` statements into `DEBUG` or `DEBUG2` statements, or remove them entirely.

lib/stages/oneDiskVDIFWrite.cpp
```diff
@@ -0,0 +1,338 @@
...
+    vdif_header -> frame_len = (vdif_frame_data_size + vdif_frame_header_size) / 8;
+    vdif_header -> vdif_version = 0;
+    vdif_header -> data_type = 1;
+    vdif_header -> bits_depth = 4 - 1;  //  bits per sample minus 1
+    vdif_header -> station_id = 9999;
+    vdif_header -> edv = 0;   // for now
+    vdif_header -> eud1 = 0; 
+    vdif_header -> eud2 = 0;
+    vdif_header -> eud3 = 0;
+    vdif_header -> eud4 = 0;
```
andrerenard, line 174: The formatting here doesn't follow our design guide, e.g. no spaces in member access `vdif_header->invalid`.  I think once you run clang-format, that should automatically fix this.

lib/stages/oneDiskVDIFWrite.cpp
```diff
@@ -0,0 +1,338 @@
...
+    }
+    
+        
+    // allocate the vdif buffer 
+    vdif_frame_data_size = vdif_samples_per_frame * vdif_freq_per_frame * num_pol;
+    vdif_frame_size = vdif_frame_header_size + vdif_frame_data_size;
+    INFO("VDIF size {:d} {:d}", vdif_frame_data_size, vdif_frame_size);
+    vdif_frame = (uint8_t*) malloc(vdif_frame_size);
+    // Cast vdif header
+    vdif_header = (VDIFHeader*) &vdif_frame[0];
```
andrerenard: Redundant array access and dereference.
```suggestion
    vdif_header = (VDIFHeader*) vdif_frame;
```

lib/stages/oneDiskVDIFWrite.cpp
```diff
@@ -0,0 +1,338 @@
...
+    // cast vdif payload
+    payload = &vdif_frame[vdif_frame_header_size];
```
andrerenard, line 156: You could also use pointer math.  Although one could argue it's clearer to read with an array access.
```suggestion
    payload = vdif_frame + vdif_frame_header_size;
```

lib/stages/oneDiskVDIFWrite.cpp
```diff
@@ -0,0 +1,338 @@
...
+
+oneDiskVDIFWrite::~oneDiskVDIFWrite(){}
+
+
+void oneDiskVDIFWrite::main_thread(){
+    int fd;
+    int payload_ctr;
+    uint8_t* input = nullptr;
+    frameID in_frame_id(in_buf);
+    uint8_t* vdif_frame;
```
andrerenard, line 88: It would be good practice to set all your pointers to `nullptr` when you declare them.

andrerenard (changes requested):
Not done with the review, but here are some comments to get started.

andrerenard:
Also take a look at the LGTM warnings, a few of those alerts look like they should be fixed.

lib/stages/CMakeLists.txt
luojing1211: So it would be "MergeBeamBuffer.cpp" instead of "mergeBeamBuffer.cpp"?

lib/utils/util.c
luojing1211: For the case of n disk vdif recording, it is always a new data directory that each time Kotekan starts. So we need to exit and throw an error at this place. Can you send me some examples of the new style of logging?

lib/utils/util.h
luojing1211: Do you mean return an error code in the function, but the user can make a judgment in the program?

lib/stages/rawFileRead.cpp
luojing1211: This function was designed for different name conventions, during the development. I did not mean to change this stage. But somehow I have to commit it. Do you want me to change the stage?

PR #966, opened 2021-06-10
Support old HartRAO hardware

andrerenard:


config/hartrao_reduced_links.yaml
```diff
@@ -0,0 +1,333 @@
...
+  gpu_output_split_buffer_3:
+    kotekan_buffer: standard
+  gpu_output_split_buffer_4:
+    kotekan_buffer: standard
+  gpu_output_split_buffer_5:
+    kotekan_buffer: standard
+#  gpu_output_split_buffer_6:
+#    kotekan_buffer: standard
+#  gpu_output_split_buffer_7:
+#    kotekan_buffer: standard
```
james-s-willis, line 111: Same here and all following commented out regions for GPUs 6 and 7,

config/hartrao_reduced_links.yaml
```diff
@@ -0,0 +1,333 @@
...
+    - gpu_output_split_buffer_1
+    - gpu_output_split_buffer_2
+    - gpu_output_split_buffer_3
+    - gpu_output_split_buffer_4
+    - gpu_output_split_buffer_5
+#    - gpu_output_split_buffer_6
+#    - gpu_output_split_buffer_7
+
+metadata:
+  #kotekan_stage: chimeMetadataDump
```
james-s-willis, line 243: Is this debug?

lib/opencl/kernels/kv_corr_sm_legacy.cl
```diff
@@ -0,0 +1,113 @@
+//#define SAMPLES_PER_DATA_SET
+//#define NUM_ELEMENTS
+//#define NUM_FREQS
+//#define BLOCK_SIZE
+//#define COARSE_BLOCK_SIZE
```
james-s-willis, line 5: These are commented out here but are actually used in the code.

james-s-willis (changes requested):
So there are 6 GPU buffers which get merged into 1 buffer and are only processed by 1 GPU?

andrerenard:
> So there are 6 GPU buffers which get merged into 1 buffer and are only processed by 1 GPU?

Yes that's correct.  With the CHIME configurations we always have one stream of data with 1 GPU, but with most other configurations we take a few streams of data and process them with the same GPU.

config/hartrao_reduced_links.yaml
```diff
@@ -0,0 +1,333 @@
...
+  lost_samples_buffer_3:
+    kotekan_buffer: standard
+  lost_samples_buffer_4:
+    kotekan_buffer: standard
+  lost_samples_buffer_5:
+    kotekan_buffer: standard
+#  lost_samples_buffer_6:
+#    kotekan_buffer: standard
+#  lost_samples_buffer_7:
+#    kotekan_buffer: standard
```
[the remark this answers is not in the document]
andrerenard: The system at HartRAO currently doesn't have enough working sticks of memory in the system to process all 8 streams.  I've left those in to un-comment once that's fixed.

config/hartrao_reduced_links.yaml
andrerenard: Oh yeah I didn't even notice that line was there.

lib/opencl/kernels/kv_corr_sm_legacy.cl
andrerenard: Ah those are passed in at run-time.  I'll add a note to that effect.

james-s-willis (approved):
Looks good to me now!

dcrichton:
Looks good to me too with the cinttypes includes. Do you think it's worth adding the `subset_active` stage parts of the config that we're currently running to the repo? Given that the number of active elements will vary at HartRAO over the next few months I don't think so.

andrerenard:
Yeah I don't think we need the `subset_active` part if that's actively changing.  I'm going to merge this in now.

PR #968, opened 2021-06-10
Create a new object to track CPU usage per stage

Jintao-Xu:
New object: CpuMonitor
- Run in a new thread to keep tracking stage CPU usage periodically
- Add a new endpoint /cpu_ult to get the CPU usage message
- Enable CPU monitor by config file
- Only support Linux
- Start CPU monitor from kotekanMode
- Read thread lists from Stages passed by kotekanMode

### DIFF
--- a/lib/core/Stage.cpp
+++ b/lib/core/Stage.cpp
@@ -17,7 +17,7 @@
 #include <regex>         // for match_results<>::_Base_type
 #include <sched.h>       // for cpu_set_t, CPU_SET, CPU_ZERO
 #include <stdexcept>     // for runtime_error
-#include <sys/syscall.h> // for SYS_gettid
+#include <sys/syscall.h> // for SYS_gettid // IWYU pragma: keep
 // IWYU pragma: no_include <syscall.h>
 #include <system_error> // for system_error
 #include <thread>       // for thread
--- a/lib/core/Stage.cpp
+++ b/lib/core/Stage.cpp
@@ -7,20 +7,20 @@
 
 #include "fmt.hpp" // for format
 
-#include <algorithm>    // for copy, max, find
-#include <chrono>       // for seconds
-#include <cstdlib>      // for abort
-#include <cxxabi.h>     // for __forced_unwind
-#include <exception>    // for exception
-#include <future>       // for async, future, future_status, future_status::timeout, launch
-#include <pthread.h>    // for pthread_setaffinity_np, pthread_setname_np
-#include <regex>        // for match_results<>::_Base_type
-#include <sched.h>      // for cpu_set_t, CPU_SET, CPU_ZERO
-#include <stdexcept>    // for runtime_error
-#include <syscall.h>    // for SYS_gettid
-#include <system_error> // for system_error
-#include <thread>       // for thread
-#include <unistd.h>     // for syscall
+#include <algorithm>     // for copy, max, find
+#include <chrono>        // for seconds
+#include <cstdlib>       // for abort
+#include <cxxabi.h>      // for __forced_unwind
+#include <exception>     // for exception
+#include <future>        // for async, future, future_status, future_status::timeout, launch
+#include <pthread.h>     // for pthread_setaffinity_np, pthread_setname_np
+#include <regex>         // for match_results<>::_Base_type
+#include <sched.h>       // for cpu_set_t, CPU_SET, CPU_ZERO
+#include <stdexcept>     // for runtime_error
+#include <sys/syscall.h> // for SYS_gettid
+#include <system_error>  // for system_error
+#include <thread>        // for thread
+#include <unistd.h>      // for syscall
 
 namespace kotekan {
 
--- a/lib/core/Stage.cpp
+++ b/lib/core/Stage.cpp
@@ -17,7 +17,7 @@
 #include <regex>         // for match_results<>::_Base_type
 #include <sched.h>       // for cpu_set_t, CPU_SET, CPU_ZERO
 #include <stdexcept>     // for runtime_error
-#include <sys/syscall.h> // for SYS_gettid
+#include <sys/syscall.h> // for SYS_gettid // IWYU pragma: keep
 #include <system_error>  // for system_error
 #include <thread>        // for thread
 #include <unistd.h>      // for syscall
@@ -163,7 +163,7 @@ void Stage::unregister_tid(pid_t tid) {
     }
 }
 
-std::vector<pid_t> Stage::get_tids() {
+const std::vector<pid_t>& Stage::get_tids() {
     return thread_list;
 }
 
--- a/lib/core/Stage.cpp
+++ b/lib/core/Stage.cpp
@@ -17,7 +17,8 @@
 #include <regex>         // for match_results<>::_Base_type
 #include <sched.h>       // for cpu_set_t, CPU_SET, CPU_ZERO
 #include <stdexcept>     // for runtime_error
-#include <sys/syscall.h> // for SYS_gettid // IWYU pragma: no_include
+#include <sys/syscall.h> // for SYS_gettid 
+// IWYU pragma: no_include <syscall.h>
 #include <system_error>  // for system_error
 #include <thread>        // for thread
 #include <unistd.h>      // for syscall
--- a/lib/core/Stage.cpp
+++ b/lib/core/Stage.cpp
@@ -7,7 +7,7 @@
 
 #include "fmt.hpp" // for format
 
-#include <algorithm>    // for copy, max
+#include <algorithm>    // for copy, max, find
 #include <chrono>       // for seconds
 #include <cstdlib>      // for abort
 #include <cxxabi.h>     // for __forced_unwind
@@ -17,10 +17,10 @@
 #include <regex>        // for match_results<>::_Base_type
 #include <sched.h>      // for cpu_set_t, CPU_SET, CPU_ZERO
 #include <stdexcept>    // for runtime_error
+#include <syscall.h>    // for SYS_gettid
 #include <system_error> // for system_error
 #include <thread>       // for thread
-#include <unistd.h>
-#include <sys/syscall.h>
+#include <unistd.h>     // for syscall
 
 namespace kotekan {
 
@@ -102,7 +102,7 @@ void Stage::set_cpu_affinity(const std::vector<int>& cpu_affinity_) {
 }
 
 void Stage::start() {
-    this_thread = std::thread([&](){
+    this_thread = std::thread([&]() {
 #if !defined(MAC_OSX)
         pid_t tid = syscall(SYS_gettid);
         register_tid(tid);
--- a/lib/core/Stage.cpp
+++ b/lib/core/Stage.cpp
@@ -17,11 +17,11 @@
 #include <regex>         // for match_results<>::_Base_type
 #include <sched.h>       // for cpu_set_t, CPU_SET, CPU_ZERO
 #include <stdexcept>     // for runtime_error
-#include <sys/syscall.h> // for SYS_gettid 
+#include <sys/syscall.h> // for SYS_gettid
 // IWYU pragma: no_include <syscall.h>
-#include <system_error>  // for system_error
-#include <thread>        // for thread
-#include <unistd.h>      // for syscall
+#include <system_error> // for system_error
+#include <thread>       // for thread
+#include <unistd.h>     // for syscall
 
 namespace kotekan {
 
--- a/lib/core/Stage.cpp
+++ b/lib/core/Stage.cpp
@@ -128,6 +128,7 @@ void Stage::join() {
 
 void Stage::stop() {
     stop_thread = true;
+    unregister_tid(this_thread.native_handle());
 }
 
 void Stage::main_thread() {}
@@ -136,7 +137,6 @@ Stage::~Stage() {
     stop_thread = true;
     if (this_thread.joinable())
         this_thread.join();
-    unregister_tid(this_thread.native_handle());
 }
 
 std::string Stage::dot_string(const std::string& prefix) const {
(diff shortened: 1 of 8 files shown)


lib/core/Stage.cpp
```diff
@@ -22,6 +22,8 @@
 
 namespace kotekan {
 
+std::map<std::string, pid_t> Stage::thread_list;
```
andrerenard, line 25: I'd prefer this not be global (in general we avoid using global if at all possible).  I would make this a member variable which just stores the PIDs for each object.   It is possible for a stage to have more than one PID associated with it's unique name (e.g. a stage with more than one thread is allowed).   Then you can just make this a vector, and return a vector.   The `CpuMonitor` class will then need to be passed a reference to the `stages` variable from `KotekanMode`, and iterate that get the list of PID(s) for each stage.

lib/core/Stage.cpp
```diff
@@ -135,10 +138,34 @@ Stage::~Stage() {
...
+    pid_t tid;
+    void* others;
+};
+
+void Stage::register_tid(pthread_t ptr) {
+    pid_t tid = ((pthread_fake*)ptr)->tid;
+    thread_list[unique_name] = tid;
+}
+
+void Stage::unregister_tid() {
```
andrerenard, line 160: As a follow up to the comment above, this should likely take a `ptr` convert to a `tid` and remove it from the vector.

lib/core/kotekanMode.cpp
```diff
@@ -120,9 +120,15 @@ void kotekanMode::start_stages() {
         INFO_NON_OO("Starting kotekan_stage: {:s}...", stage.first);
         stage.second->start();
     }
+#if defined(CPU_MONITOR) && !defined(MAC_OSX)
```
andrerenard, line 123: I'm going to suggest that we enable/disable the `CpuMonitor` via the configuration file instead of a `cmake` option.  There is no performance penalty to just always compiling `CpuMonitor` into the code, making it easier to enable for an existing binary later.   I'm going to propose the following config structure:
```
cpu_monitor:
    enabled: [bool]  # default false
    cpu_affinity: [array] # use the configuration scope to allow this to default to the "global" `cpu_affinity` value set in the config. 
```
See the `restServer::set_server_affinity()` for an example of how this might be done. 
The `#if !defined(MAC_OSX)` for MacOS builds can stay as is.

lib/core/cpuMonitor.hpp
```diff
@@ -0,0 +1,62 @@
...
+     * @brief Compute stage CPU usage periodically (thread entry function).
+     * Get CPU stat from /proc/stat and stage stat from /proc/self/tid/stat.
+     * Thread list maintained and passed by Stage class.
+     **/
+    void track_cpu();
+
+private:
+    std::thread this_thread;
+    bool stop_thread;
+    std::map<std::string, CpuStat> ult_list;
```
andrerenard, line 56: This might need to be slightly more complex, since each stage can have more than one thread perhaps: `std::map<std::string, std::vector<CpuStat>>`

lib/core/cpuMonitor.cpp
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        // Read each thread stats based on tid
+        std::map<std::string, pid_t> thread_list = Stage::get_thread_list();
+        for (auto element : thread_list) {
+            char fname[40];
+            snprintf(fname, sizeof(fname), "/proc/self/task/%d/stat", element.second);
+            FILE* thread_fp = fopen(fname, "r");
+
+            if (thread_fp) {
```
andrerenard, line 67: If this fails you should log the reason in an `else` clause.

lib/core/cpuMonitor.cpp
```diff
@@ -0,0 +1,114 @@
...
+                // Get the 14th (utime) and the 15th (stime) stats
+                uint32_t utime = 0, stime = 0;
+                ret = fscanf(thread_fp, "%*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %u %u",
```
andrerenard, line 70: I would check the `ret` value matches what you want separately from the other logic, and log failures to read the file.

lib/core/cpuMonitor.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+void CpuMonitor::track_cpu() {
+    while (!stop_thread) {
+        uint32_t cpu_times[10];
+        uint32_t cpu_time = 0;
+
+        // Read CPU stats from /proc/stat first line
+        std::string stat;
+        FILE* cpu_fp = fopen("/proc/stat", "r");
```
andrerenard, line 47: Check return value for error conditions.

lib/core/cpuMonitor.cpp
```diff
@@ -0,0 +1,114 @@
...
+        int ret = fscanf(cpu_fp, "%*s %u %u %u %u %u %u %u %u %u %u", &cpu_times[0], &cpu_times[1],
+                         &cpu_times[2], &cpu_times[3], &cpu_times[4], &cpu_times[5], &cpu_times[6],
+                         &cpu_times[7], &cpu_times[8], &cpu_times[9]);
+        fclose(cpu_fp);
+
+        // Compute total cpu time
+        if (ret == 10) {
```
andrerenard, line 54: Also log if ret isn't what you expect, or if there was an error in `fscanf`

andrerenard (changes requested):
This is great functionality to have!  I have a few comments inline, mostly just design considerations.

Jintao-Xu:
@andrerenard Not sure why there is a segfault after receiving SIGINT. Maybe the way I detach or finish the thread is wrong?

> ^CGot SIGINT, shutting down kotekan...
Attempting to stop and join kotekan_stages...
Sending shutdown signal to buffer: cpu_hfb_output_buffer_0
Sending shutdown signal to buffer: cpu_output_buffer_0
Sending shutdown signal to buffer: lost_samples_buffer
Sending shutdown signal to buffer: network_buffer_0
Joining kotekan_stage: /cpu_simulate/cpu_data_sim_0...
/gen_data/network_buffer_0: Generated a random test data set in network_buffer_0[3]
/cpu_simulate/cpu_data_sim_0: Simulating GPU beamform processing done for network_buffer_0[0] result is in cpu_output_buffer_0[0]
No consumers are registered on cpu_output_buffer_0 dropping data in frame 0...
No consumers are registered on cpu_hfb_output_buffer_0 dropping data in frame 0...
Joining kotekan_stage: /gen_data/network_buffer_0...
Joining kotekan_stage: /gen_data/test_data_gen_lost_samples...
Joining kotekan_stage: /screen_dump...
Segmentation fault (core dumped)

Jintao-Xu:
> @andrerenard Not sure why there is a segfault after receiving SIGINT. Maybe the way I detach or finish the thread is wrong?
> 
> > ^CGot SIGINT, shutting down kotekan...
> > Attempting to stop and join kotekan_stages...
> > Sending shutdown signal to buffer: cpu_hfb_output_buffer_0
> > Sending shutdown signal to buffer: cpu_output_buffer_0
> > Sending shutdown signal to buffer: lost_samples_buffer
> > Sending shutdown signal to buffer: network_buffer_0
> > Joining kotekan_stage: /cpu_simulate/cpu_data_sim_0...
> > /gen_data/network_buffer_0: Generated a random test data set in network_buffer_0[3]
> > /cpu_simulate/cpu_data_sim_0: Simulating GPU beamform processing done for network_buffer_0[0] result is in cpu_output_buffer_0[0]
> > No consumers are registered on cpu_output_buffer_0 dropping data in frame 0...
> > No consumers are registered on cpu_hfb_output_buffer_0 dropping data in frame 0...
> > Joining kotekan_stage: /gen_data/network_buffer_0...
> > Joining kotekan_stage: /gen_data/test_data_gen_lost_samples...
> > Joining kotekan_stage: /screen_dump...
> > Segmentation fault (core dumped)

It has been fixed.

lib/core/Stage.cpp
```diff
@@ -141,4 +143,28 @@ std::string Stage::dot_string(const std::string& prefix) const {
     return fmt::format("{:s}\"{:s}\" [shape=box, color=darkgreen];\n", prefix, get_unique_name());
 }
 
+// Used to get tid from pthread_t
+struct pthread_fake {
+    char offset[720];
+    pid_t tid;
+    void* others;
+};
```
andrerenard, line 151: I'm not very comfortable with this as a method of getting the `tid`, it's far too dependent on the `pthread` library implementation on a given system.   As far as I can tell this might not even be portable between Linux distributions. 

I think the correct solution is to use `gettid()`, of course this requires it be called from within the thread, but I think this is possible without editing every `main_thread()` function.  Basically in the `start()` function use a lambda (or create a new function and function pointer) to start the thread on a new function - and then from that function call `gettid()`, register the thread_id with `register_tid` followed by `main_thread()`, followed by an register call.   This gives us a space to work directly within the thread, without making changes to the thread itself.  Note that you'll have to be careful about how the `*this` pointer is handled, and you'll have to avoid using this method for MacOS, since `gettid()` is a Linux only system call.

andrerenard (changes requested):
This is looking very good, but I'd like to revisit the way we get the `tid` since the present implementation is dependent on pthread library version, and not memory safe should the library use a different `pthread_t` structure.

Jintao-Xu:
The abort issue is about std::thread and pthread_detach(). Pthread_detach() cannot work on std::thread object, so when std::thread is destructed it is still associated. Also, thread.detach() has to be after pthread_setaffinity_np().

andrerenard:
> The abort issue is about std::thread and pthread_detach(). Pthread_detach() cannot work on std::thread object, so when std::thread is destructed it is still associated. Also, thread.detach() has to be after pthread_setaffinity_np().

Why do you need to detach the thread?  I think you'd be better off just joining the thread in the destructor.

Jintao-Xu:
> > The abort issue is about std::thread and pthread_detach(). Pthread_detach() cannot work on std::thread object, so when std::thread is destructed it is still associated. Also, thread.detach() has to be after pthread_setaffinity_np().
> 
> Why do you need to detach the thread? I think you'd be better off just joining the thread in the destructor.

I saw somewhere they are using detach and others using join. I'm not sure when we should use one of them?

lib/core/cpuMonitor.cpp
```diff
@@ -0,0 +1,172 @@
...
+    // Register CPU usage callback
+    restServer::instance().register_get_callback(
+        "/cpu_ult", std::bind(&CpuMonitor::cpu_ult_call_back, this, _1));
+}
+
+CpuMonitor::~CpuMonitor() {
+    restServer::instance().remove_get_callback("/cpu_ult");
+
+    try {
+        this_thread.join();
```
andrerenard, line 33: You should also check if we ever started the CPU Monitor, no need to have a warning if we never intended this thread to be joinable.    And it might be a good idea to call `stop()` inside this function just in case it isn't called externally.

andrerenard:
> I saw somewhere they are using detach and others using join. I'm not sure when we should use one of them?

Generally you can detach if the thread exits on its own in a fixed amount of time, and we don't need to confirm clean exit before terminating.  In this case since it's a "forever" thread, it's good to make sure we've cleanly shut it down.  As a general rule for our code, use joinable threads unless there is a good reason not to.

lib/core/Stage.hpp
```diff
@@ -51,6 +52,21 @@ class Stage : public kotekanLogging {
...
+
+    /**
+     * @brief Remove a tid from thread_list.
+     */
+    void unregister_tid(pid_t ptr);
+
+    /**
+     * @brief Get tids from the current stage.
+     */
+    std::vector<pid_t> get_tids();
```
andrerenard, line 68: I think we should return this as a const reference to save on the copy constructor time.  Since normally we don't need to do any modifications to the array once it's returned other than read it.

lib/core/Stage.cpp
```diff
@@ -7,18 +7,20 @@
...
+#include <chrono>        // for seconds
+#include <cstdlib>       // for abort
+#include <cxxabi.h>      // for __forced_unwind
+#include <exception>     // for exception
+#include <future>        // for async, future, future_status, future_status::timeout, launch
+#include <pthread.h>     // for pthread_setaffinity_np, pthread_setname_np
+#include <regex>         // for match_results<>::_Base_type
+#include <sched.h>       // for cpu_set_t, CPU_SET, CPU_ZERO
+#include <stdexcept>     // for runtime_error
+#include <sys/syscall.h> // for SYS_gettid
```
andrerenard, line 20:
```suggestion
#include <syscall.h> // for SYS_gettid
```
I think this will fix the iwyu error.

lib/core/cpuMonitor.cpp
andrerenard: This was marked as resolved, but you aren't checking the if `fopen` was successful.  This could cause a problem at the `fscanf` lines below if the file handle returned isn't valid.

andrerenard (changes requested):
This is almost ready to go!  Just a few small comments.

lib/core/Stage.cpp
Jintao-Xu: This will cause another test failure unfortunately. it cannot find this file without sys/

andrerenard: Ah that's annoying.  We should just tell IWYU to ignore this include problem then: https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md

Jintao-Xu:
Both IWYU pragma "keep" and "no_include" should be used to avoid test failure.

andrerenard (approved):
I think this is ready to merge, but please see slack regarding the return of the endpoint when CPU utilization isn't enabled.   It might be nice to clean up the behavior there before merging.

Jintao-Xu:
> I think this is ready to merge, but please see slack regarding the return of the endpoint when CPU utilization isn't enabled. It might be nice to clean up the behavior there before merging.

That sounds pretty reasonable. I'll create a separate PR to do that after this one is merged. Thanks!

PR #971, opened 2021-06-29
Create endpoint timer metrics

Jintao-Xu:
- Create a group of endpoint metrics in the REST server with the label "endpoint"
- Add StatTracker to get avg and max reply time
- Use a threshold of 1.0 to max reply time


Jintao-Xu:
This PR is refactored, and a new one will be created.

PR #975, opened 2021-07-13
Improve stat tracker

Jintao-Xu:
- add tracker name
- add tracker unit
- add timestamp
- add brute force min/max
- update unit test

### DIFF
--- a/lib/utils/visUtil.cpp
+++ b/lib/utils/visUtil.cpp
@@ -314,10 +314,9 @@ double StatTracker::get_max() {
     } else {
         // brute force way to get max
         double max = std::numeric_limits<double>::lowest();
-        int start = (end - count) % buf_size;
-        for (size_t i = 0; i < count; i++) {
-            max = std::max(max, rbuf[(start + i) % buf_size].value);
-        }
+        size_t size = std::min(count, buf_size);
+        for (size_t i = 0; i < size; i++) max = std::max(max, rbuf[i].value);
+
         return max;
     }
 }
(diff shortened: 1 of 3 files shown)


james-s-willis (approved):
Looks good to me

lib/utils/visUtil.hpp
```diff
@@ -583,9 +584,13 @@ class StatTracker {
     /**
      * @brief Create a ring buffer.
      *
+     * @param name Buffer name.
```
andrerenard, line 587:
```suggestion
     * @param name The statistic's name.
```
Buffer gets used a lot to reference a very specific thing in the pipeline context, best to use another term when possible.

lib/utils/visUtil.cpp
```diff
@@ -301,14 +308,36 @@ double StatTracker::get_max() {
...
+
+    if (is_optimized) {
+        return min_max.get_max();
+    } else {
+        // brute force way to get max
+        double max = std::numeric_limits<double>::lowest();
+        int start = (end - count) % buf_size;
+        for (size_t i = 0; i < count; i++) {
+            max = std::max(max, rbuf[(start + i) % buf_size].value);
+        }
```
andrerenard, line 320: The logic here is more complex than it needs to be.   In the case that `count < buf_size` since we always start at `0` you can just start with `i = 0`, and once `count == buf_size` then we only need to find the max of the entire array - it doesn't matter where we start.   So a simple `rbuf[i].value` will do the trick.  Same for `min`.

lib/utils/visUtil.hpp
```diff
@@ -583,9 +584,13 @@ class StatTracker {
...
+     * @param unit Sample unit.
      * @param size The size of the ring buffer.
+     * @param is_optimized Flag of min/max optimization.
```
andrerenard, line 590: I'd like slight more description of what this flag does.  e.g. what does this flag do, and why might you choose to use it (or not).   You could add this to the main body of the comment block after the `@brief` line, since there isn't a lot of space in the `@param` and it's an important part of how this class works.

andrerenard (changes requested):
Looks good, just a few small comments inline.

andrerenard (approved):

PR #976, opened 2021-07-13
Exclude 'select' as a backend API for libevent

andrerenard:
* Fixes an issue where if libevent picked `select` as its backend API, we would be limited 1024 file handles.


james-s-willis (changes requested):
Looks good to me. Just need to fix IWYU errors.

james-s-willis:
You still need to remove this include from `restClient.cpp`:
```
2021-07-22T21:12:53.4933875Z /code/kotekan/lib/utils/restClient.cpp should remove these lines:
2021-07-22T21:12:53.4935071Z - #include <event2/event_struct.h>  // lines 13-13
```

james-s-willis:
Once that has been fixed it's good to go.

james-s-willis (approved):

PR #977, opened 2021-07-13
Fix phase center for CHIME tracking beamformer

andrerenard:
* Moves the instrument latitude and longitude from define constants into the configuration file. 
* Fixes an issue were a `float` was read as an `int`
* Updates the CHIME phase center to the correct location (appears to have been set to the Pathfinder location)

### DIFF
--- a/config/chime_science_run_gpu.yaml
+++ b/config/chime_science_run_gpu.yaml
@@ -70,6 +70,8 @@ frb_scaling: 0.05 #1.0
 # Tracking beams global options
 feed_sep_NS: 0.3048
 feed_sep_EW: 22.0
+# Latitude and longitude converted from final geocentric (XYZ) position as calculated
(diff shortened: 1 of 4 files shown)


bwmeyers:
Can we also note in the code where these coordinates originate? Something like 
```
Converted from final geocentric (XYZ) position calculated in https://bao.chimenet.ca/doc/documents/1327 via NGS NOAA website.
```

cherryng (approved):
Thanks for making the changes. I think they all look fine. Having lat and Long as double should be fine. the lines where they are being used in hsaTrackingUpdatePhase.cpp are all casting to double, anyways. And sorry about feed_seq_EW being int32. it should be a float!

andrerenard:
I've added the comment on how the lat/long are computed to the config files, will merge this once it passes CI.

PR #978, opened 2021-07-19
Create kotekan trackers

Jintao-Xu:
- create kotekan trackers cpp/hpp
- set up trackers from config
- register /trackers endpoint
- print tracker content in json format


Jintao-Xu:
config format:
> screen_dump:
  kotekan_stage: hexDump
  in_buf: example_buffer
  trackers: [{name: test_0, unit: ms, size: 9, is_optimized: false}, {name: test_1, unit: km}]

endpoint print:
> {
> "/screen_dump/test_0": {
> "is_optimized": false,
> "samples": [
> {
> "timestamp": 1626789723054,
> "value": 72.0
> },
> {
> "timestamp": 1626789722382,
> "value": 64.0
> },
> {
> "timestamp": 1626789722466,
> "value": 65.0
> },
> {
> "timestamp": 1626789722550,
> "value": 66.0
> },
> {
> "timestamp": 1626789722634,
> "value": 67.0
> },
> {
> "timestamp": 1626789722718,
> "value": 68.0
> },
> {
> "timestamp": 1626789722802,
> "value": 69.0
> },
> {
> "timestamp": 1626789722886,
> "value": 70.0
> },
> {
> "timestamp": 1626789722970,
> "value": 71.0
> }
> ],
> "size": 9,
> "unit": "ms"
> },
> "/screen_dump/test_1": {
> "is_optimized": true,
> "size": 100,
> "unit": "km"
> }
> }

lib/core/kotekanTrackers.cpp
```diff
@@ -0,0 +1,68 @@
...
+    nlohmann::json trackers_json = {};
+
+    for (auto& it : trackers) {
+        trackers_json[it.first] = it.second->get_json();
+    }
+
+    conn.send_json_reply(trackers_json);
+}
+
+std::shared_ptr<StatTracker> KotekanTrackers::add_tracker(std::string name, std::string unit,
```
andrerenard, line 42: I think you should ask for the unique name, and make the map key an `std::pair<std::string, std::string>` with values, `(unique_name, tracker_name)`.  This way you won't run into a problem where you have two stages generating the same key for their stat tracker, and that will make it possible to group stat trackers by stage object names in the returned json.

lib/core/Stage.hpp
```diff
@@ -85,6 +97,9 @@ class Stage : public kotekanLogging {
 
     bufferContainer& buffer_container;
 
+    // A map to store all stat trackers in this stage.
+    std::map<std::string, std::shared_ptr<StatTracker>> stat_trackers;
```
andrerenard, line 101: Likewise, I don't think we need to store this in the base class for every stage, just let each subclass store the copies of it's stat trackers as it sees fit.

lib/utils/visUtil.cpp
```diff
@@ -341,19 +347,42 @@ double StatTracker::get_min() {
...
     }
     return std_dev;
 }
 
+nlohmann::json StatTracker::get_json() {
+    std::lock_guard<std::mutex> lock(tracker_lock);
+
+    nlohmann::json tracker_json = {};
+    tracker_json["unit"] = unit;
+    tracker_json["is_optimized"] = is_optimized;
```
andrerenard, line 372: I don't think there is any reason to return this value.

lib/utils/visUtil.cpp
```diff
@@ -341,19 +347,42 @@ double StatTracker::get_min() {
...
+    tracker_json["size"] = buf_size;
```
andrerenard, line 373: I'm also not sure we need this either.   The length of the array below gives most of this information implicitly.  Knowing the max length of the tracker doesn't add much for the viewer.

lib/core/kotekanTrackers.hpp
```diff
@@ -0,0 +1,68 @@
...
+     * @brief Adds a new tracker
+     *
+     * @param name The name of the tracker.
+     * @param unit The unit of the tracker.
+     * @param size The size of the tracker with default of 100.
+     * @param is_optimized The switch of min/max methods of the tracker.
+     * @return a shared pointer to the newly created tracker
+     * @throw std::runtime_error if the tracker with that name is already registered.
+     */
+    std::shared_ptr<StatTracker> add_tracker(std::string name, std::string unit, size_t size = 100,
```
andrerenard, line 52: We should also have a call to remove the tracker.  And see my comment in the function regarding changing the required arguments here.

lib/core/kotekanTrackers.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+#include "fmt.hpp"  // for format, fmt
+#include "json.hpp" // for json, basic_json<>::value_type
+
+#include <functional> // for _Bind_helper<>::type, _Placeholder, bind, _1, placeholders
+#include <stdexcept>  // for runtime_error
+#include <utility>    // for pair
+
+namespace kotekan {
+namespace trackers {
```
andrerenard, line 13: I don't think we need this namespace.

lib/core/kotekanTrackers.cpp
```diff
@@ -0,0 +1,68 @@
...
+}
+
+KotekanTrackers& KotekanTrackers::instance() {
+    static KotekanTrackers _instance;
+    return _instance;
+}
+
+void KotekanTrackers::register_with_server(restServer* rest_server) {
+    using namespace std::placeholders;
+    rest_server->register_get_callback("/trackers",
```
andrerenard, line 28: I would like to have two end points here.  `/trackers_current` and `/trackers`.   This current endpoint can stay the same, but we should add a `trackers_current` which returns just the current min/max/avg/std/etc. for each tracker.   I can see the JSON returned from `/trackers` getting very large, and JSON processing in Javascript is somewhat slow, so for realtime watching of stats we really just want the current values.

lib/core/kotekanTrackers.cpp
```diff
@@ -0,0 +1,68 @@
...
+                                       std::bind(&KotekanTrackers::trackers_callback, this, _1));
+}
+
+void KotekanTrackers::trackers_callback(connectionInstance& conn) {
+    nlohmann::json trackers_json = {};
+
+    for (auto& it : trackers) {
+        trackers_json[it.first] = it.second->get_json();
```
andrerenard, line 36: hmm, we might want to think about the structure of this a little more.  We kind of want to group these clearly by the stages `unique_name` so that we can associate them with the pipeline graph easily.

andrerenard (changes requested):
This looks really good.  However I want to adjust some aspects of the design, please see the notes inline.   Also let's add a tracker to one of the example `stages` so that we have something to test against.

andrerenard:
We should also add a function to dump the JSON with all the values to a file in the case that kotekan exits with an error - and also add an option to the config file to control if and where that dump file will be written.

lib/core/kotekanTrackers.hpp
Jintao-Xu: The remove() function should remove individual tracker or all trackers in a given stage?

andrerenard: I'd give the option for both.

lib/core/kotekanMode.cpp
```diff
@@ -53,6 +54,9 @@ kotekanMode::~kotekanMode() {
     restServer::instance().remove_get_callback("/pipeline_dot");
     restServer::instance().remove_all_aliases();
 
+    // Dump all trackers before cleanup.
+    KotekanTrackers::instance().dump_trackers();
```
andrerenard, line 58: I would move this to `kotekan.cpp` and only dump in the case that we exited with an error: see line 495.   I don't think this dump makes sense outside of an error triggered shutdown.

lib/core/kotekanTrackers.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+    nlohmann::json return_json = {};
+
+    for (auto& stage_itr : trackers) {
+        for (auto& tracker_itr : trackers[stage_itr.first]) {
+            return_json[stage_itr.first][tracker_itr.first] = tracker_itr.second->get_json();
+        }
+    }
+
+    std::ofstream dump_file(dump_path + "/trackers_dump.json");
```
andrerenard, line 125: We are going to want to make this file name based on the current time.  Otherwise it'll keep getting overwritten with each crash.   Also it might be worth including the host name of the system as well, since in the case of the CHIME cluster, we have many nodes sharing the same file system (NFS), so it won't be clear which node a file belongs too.

lib/core/kotekanTrackers.cpp
```diff
@@ -0,0 +1,135 @@
...
+    if (dump_file.is_open()) {
+        dump_file << return_json;
```
andrerenard, line 127: You should explicitly close this file when you are done with it.

lib/stages/hexDump.cpp
```diff
@@ -55,6 +61,11 @@ void hexDump::main_thread() {
         // Prints the hex data to screen
         hex_dump(16, (void*)&frame[_offset], _len);
 
+        // Add sample to trackers
+        static int cnt = 0;
```
andrerenard, line 65: I would avoid using a static here.  With more than one stage, this will have possibly unexpected results.   If you just move this declaration outside the `while`, then it will achieve the same effect.

lib/stages/hexDump.cpp
```diff
@@ -55,6 +61,11 @@ void hexDump::main_thread() {
...
+        tracker_0->add_sample(cnt++);
+        tracker_1->add_sample(cnt << 1);
```
andrerenard, line 67: It's likely better to make the stat tracker example based on something that's theoretically useful.  This is more of a unit test style example.   Perhaps you could track the time needed in testDataGen to populate a new frame with values, or something like that.

andrerenard (changes requested):
I still need to look over that data structure changes in a little more detail, but here are some comments from my first read over the code.

lib/core/kotekanTrackers.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * @brief Remove all trackers in the given stage
+     *
+     * @param stage_name The given stage name.
+     */
+    void remove_tracker(std::string stage_name);
+
+    /**
+     * @brief Remove a simgle tracker.
```
Jintao-Xu, line 77:
```suggestion
     * @brief Remove a single tracker.
```

config/examples/basic_pipeline.yaml
```diff
@@ -3,6 +3,10 @@ type: config
 log_level: DEBUG
 cpu_affinity: [0,1,2,3]
 
+trackers_dump:
```
andrerenard, line 6: I would rename this slightly.   Call `trackers_dump:` just `trackers:` and then `path:` -> `dump_path:`, `enabled:` -> `enable_crash_dump`   That way if we want to add additional parameters, we have a logical part of the configuration to add them.

lib/core/kotekanMode.cpp
```diff
@@ -88,6 +89,13 @@ void kotekanMode::initalize_stages() {
     // Apply config for Telescope class
     Telescope::instance(config);
 
+    // Create and register kotekan trackers before stages created
+    KotekanTrackers::instance().register_with_server(&restServer::instance());
+    if (config.get_default<bool>("/trackers_dump", "enabled", false)) {
```
andrerenard, line 94: I would put this logic in the constructor, and pass `config` to the constructor like the telescope and datasetManager objects.

lib/core/kotekanTrackers.cpp
```diff
@@ -0,0 +1,153 @@
...
+                                                          bool is_optimized) {
+    if (stage_name.empty()) {
+        ERROR_NON_OO("Empty stage name. Exiting.");
+        throw std::runtime_error("Empty stage name.");
+    }
+    if (tracker_name.empty()) {
+        ERROR_NON_OO("Empty tracker name. Exiting.");
+        throw std::runtime_error("Empty tracker name.");
+    }
+    if (unit.empty()) {
```
andrerenard, line 76: I'm not sure we actually need this check.  It is very possible to have a dimensionless quantity which is still meaningful as a statistic.  In which case I would just have a null string.

lib/core/kotekanTrackers.cpp
```diff
@@ -0,0 +1,153 @@
...
+        ERROR_NON_OO("Empty unit for tracker {:s}:{:s}. Exiting.", stage_name, tracker_name);
+        throw std::runtime_error(
+            fmt::format(fmt("Empty unit name: {:s}:{:s}"), stage_name, tracker_name));
+    }
+
+    std::shared_ptr<StatTracker> tracker_ptr =
+        std::make_shared<StatTracker>(tracker_name, unit, size, is_optimized);
+
+    std::lock_guard<std::mutex> lock(trackers_lock);
```
andrerenard, line 85: I would also lock when removing and reading them as well.  Very unlikely to be a contention there, but it is theoretically possible.

lib/core/kotekanTrackers.cpp
```diff
@@ -0,0 +1,153 @@
...
+    char* host_name = new char[20];
+    int ret = gethostname(host_name, sizeof(host_name));
+    if (ret == -1) {
+        ERROR_NON_OO("Error from gethostname()");
+        exit(errno);
+    }
+
+    std::chrono::system_clock::time_point timestamp = std::chrono::system_clock::now();
+    std::string time = std::to_string(
+        std::chrono::duration_cast<std::chrono::milliseconds>(timestamp.time_since_epoch())
```
andrerenard, line 139: I would format this as an `%Y%m%dT%H%M%SZ` string, not just epoch time stamp.

lib/core/kotekanTrackers.cpp
```diff
@@ -0,0 +1,153 @@
...
+            .count());
+
+    std::ofstream dump_file(dump_path + "/" + std::string(host_name) + "_" + time + ".json");
```
andrerenard, line 142: Add something like `crash_data` or `crash_stats` to the file name, so it's clear if someone comes across this file what it is.

lib/core/kotekanTrackers.hpp
```diff
@@ -0,0 +1,110 @@
...
+#include <memory>   // for shared_ptr
+#include <mutex>    // for mutex
+#include <stddef.h> // for size_t
+#include <string>   // for string
+
+namespace kotekan {
+
+typedef std::map<std::string, std::shared_ptr<StatTracker>> stage_trackers_t;
+
+class KotekanTrackers {
```
andrerenard, line 17: This is missing a class doc string/comments.   See for example `metrics` (although even that is maybe a little sparse on details).  It's good to have something that gives an overview of what the class is trying to achieve and how to use it.

lib/core/kotekanTrackers.hpp
```diff
@@ -0,0 +1,110 @@
...
+    void trackers_current_callback(connectionInstance& conn);
+
+    /**
+     * @brief Adds a new tracker
+     *
+     * @param stage_name The name of the stage.
+     * @param tracker_name The name of the tracker.
+     * @param unit The unit of the tracker.
+     * @param size The size of the tracker with default of 100.
+     * @param is_optimized The switch of min/max methods of the tracker.
```
andrerenard, line 61: Maybe add something like "See `statTracker` for details."  Since otherwise it might be unclear where to get information on this flag.

lib/stages/testDataGen.cpp
```diff
@@ -68,6 +69,10 @@ testDataGen::testDataGen(Config& config, const std::string& unique_name,
     using namespace std::placeholders;
     restServer::instance().register_post_callback(
         endpoint, std::bind(&testDataGen::rest_callback, this, _1, _2));
+
+    // Create stat tracker
+    kotekan::KotekanTrackers& KT = kotekan::KotekanTrackers::instance();
```
andrerenard, line 74: I think you should already be using the `kotekan` namespace, so the first part of this is likely not needed.

lib/stages/testDataGen.cpp
```diff
@@ -68,6 +69,10 @@ testDataGen::testDataGen(Config& config, const std::string& unique_name,
...
+    timer = KT.add_tracker(unique_name, "test_data_gen_timer", "sec");
```
andrerenard, line 75:
```suggestion
    timer = KT.add_tracker(unique_name, "frame_fill_time", "sec");
```
The unique_name should provide context for what this stage is doing, so you just need to name what metric is tracking.  In this case how long it takes to fill a frame.

andrerenard (changes requested):
This is looking really good, and I think all the structure is solid now.  Just some minor inline comments.

lib/stages/testDataGen.cpp
Jintao-Xu: There will be error of unknown type name, so I just leave it like this.

Jintao-Xu:
Not sure what is wrong with those CI failures, but seems not related to code.

andrerenard:
> Not sure what is wrong with those CI failures, but seems not related to code.

Let's see if they resolve themselves, github has been moving around it's URLs for docker containers and that might be causing unexpected issues.  Hopefully it's a short term issue, if not we might have to update our docker/github actions files.

lib/stages/testDataGen.cpp
andrerenard: Yeah you are right, this is needed.

andrerenard:
I'm going to update this branch with develop to see if that resolves the CI issue.

lib/utils/visUtil.cpp
```diff
@@ -341,19 +347,52 @@ double StatTracker::get_min() {
...
+    }
+
+    return tracker_json;
+}
+
+nlohmann::json StatTracker::get_current_json() {
+    nlohmann::json tracker_json = {};
+
+    tracker_json["unit"] = unit;
+    tracker_json["min"] = get_min();
```
andrerenard, line 388: We should add the current value, and timestamp of the current value to this as well.

andrerenard (changes requested):
This is almost ready to go, just see the one inline comment.

andrerenard (approved):

PR #980, opened 2021-07-30
Add cpu monitor not-started message

Jintao-Xu:


james-s-willis (approved):
Looks good to me

lib/core/cpuMonitor.cpp
```diff
@@ -143,6 +143,12 @@ void CpuMonitor::track_cpu() {
 }
 
 void CpuMonitor::cpu_ult_call_back(connectionInstance& conn) {
+
+    if (!this_thread.joinable()) {
+        conn.send_text_reply("CPU monitor has not been started.");
```
andrerenard, line 148: I would sent this as a JSON formatted object. That way you don't have to check if the data is parsable or not, just parse as normal, and then look for an error string.

andrerenard (changes requested):

andrerenard (approved):

PR #981, opened 2021-08-04
Add cpu stats to web interface

Jintao-Xu:
- Read cpu stats from cpu_ult
- Print total cpu usage
- Show usr and sys usages in tooltip

### DIFF
--- a/tools/debug_tool/static/pipeline_viewer.js
+++ b/tools/debug_tool/static/pipeline_viewer.js
@@ -3,27 +3,63 @@ function isIE() { return ((navigator.appName == 'Microsoft Internet Explorer') |
 // Time between updating kotekan metrics
 const POLL_WAIT_TIME_MS = 1000;
 // Poll kotekan via web server every POLL_WAIT_TIME_MS and update page with new metrics
-function poll(label, index) {
-    get_data().then(function (new_buffers) {
-        update_buf_utl(new_buffers, label, index);
+function poll(buffer_labels, stage_labels) {
+    get_data("/buffers").then(function (new_buffers) {
+        update_buf_utl(new_buffers, buffer_labels);
     });
 
+    get_data("/cpu_ult").then(function (cpu_stats) {
+        update_cpu_utl(cpu_stats, stage_labels);
+    })
+
     // Call poll() again to get the next message
-    setTimeout(() => { poll(label, index); }, POLL_WAIT_TIME_MS);
+    setTimeout(() => { poll(buffer_labels, stage_labels); }, POLL_WAIT_TIME_MS);
 }
 
 // Update buffer utilization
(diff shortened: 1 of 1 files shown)


Jintao-Xu:
image

Jintao-Xu:
The test failure is irrelevant. Seems like github cache has something wrong.

tools/debug_tool/static/pipeline_viewer.js
```diff
@@ -3,27 +3,63 @@ function isIE() { return ((navigator.appName == 'Microsoft Internet Explorer') |
...
+                tot_ult += usr + sys;
+            }
+        });
+        // Limit to two decimals
+        usr_ult = usr_ult.toFixed(2);
+        sys_ult = sys_ult.toFixed(2);
+        tot_ult = tot_ult.toFixed(2);
+
+        el.select("title").text("usr: " + usr_ult + "%; sys: " + sys_ult + "%");
+        el.select("#utl").text(tot_ult + "%");
```
andrerenard, line 56:
```suggestion
        el.select("#utl").text("CPU: " + tot_ult + "%");
```
Just add something to indicate what this value is.

tools/debug_tool/static/pipeline_viewer.js
```diff
@@ -3,27 +3,63 @@ function isIE() { return ((navigator.appName == 'Microsoft Internet Explorer') |
...
+
+        d3.select(cpu_stats[name]).forEach((stage) => {
+            var threads = Object.keys(stage[0]);
+            for (tid of threads) {
+                var usr = stage[0][tid]["usr_cpu_ult"];
+                var sys = stage[0][tid]["sys_cpu_ult"];
+                usr_ult += usr;
+                sys_ult += sys;
+                tot_ult += usr + sys;
+            }
```
andrerenard, line 48: For now this works fine.  However, if we have time it might be nice to break out each thread for multi-threaded stages. And also indicate the which CPUs the threads are allowed to run on (e.g. thread CPU affinity), however that requires more changes on the C++ side, so I think that can be another PR later when we have time.

andrerenard (changes requested):
Just one small request, and otherwise this looks good to merge.

tools/debug_tool/static/pipeline_viewer.js
Jintao-Xu: That's a good idea!

andrerenard (approved):

andrerenard:
Since there is no CI run on this file anyway, I'm going to ahead and merge it without the github actions, since it seems they may be stuck in a broken state on this PR.    (Somewhat related, we should consider a `JSLint` action in our CI, now that we are including some web code into the project)

PR #982, opened 2021-08-06
Change metrics from references to smart pointers.

### LINKED ISSUE #979: Return smart pointers in prometheus metrics
A problem showed up when working on PR #972 were references to individual Prometheus metrics can deleted before the user of the metric is done with it.   This only happens in the case of non-stage objects which use prometheus metrics (e.g. the HTTP/REST server) stay alive past the life of the metrics object on shutdown. 

I think the correct solution here is return the shared pointer directly.   At the moment it uses a smart pointer in some places, but then dereferences it and returns a reference, instead of a return by value of the smart pointer itself.   This results in the reference becoming detached from the shared pointer, which defeats the point of using a shared pointer in the first place.

andrerenard:
Fix an issue where the prometheus metrics singleton could be deleted before the objects referencing the individual metrics.   Resolves this by returning smart pointers instead of references. 

Closes #979

### DIFF
--- a/kotekan/kotekan.cpp
+++ b/kotekan/kotekan.cpp
@@ -462,8 +462,8 @@ int main(int argc, char** argv) {
 
     auto& metrics = prometheus::Metrics::instance();
     metrics.register_with_server(&rest_server);
-    auto& kotekan_running_metric = metrics.add_gauge("kotekan_running", "main");
-    kotekan_running_metric.set(running);
+    auto kotekan_running_metric = metrics.add_gauge("kotekan_running", "main", {});
+    kotekan_running_metric->labels({}).set(running);
 
     basebandApiManager& baseband = basebandApiManager::instance();
     baseband.register_with_server(&rest_server);
@@ -473,7 +473,7 @@ int main(int argc, char** argv) {
         // Update running state
         {
             std::lock_guard<std::mutex> lock(kotekan_state_lock);
-            kotekan_running_metric.set(running);
+            kotekan_running_metric->labels({}).set(running);
         }
 
         if (sig_value == SIGINT) {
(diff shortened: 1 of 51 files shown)


andrerenard:
The key changes in this PR are in `lib/core/prometheusMetrics.[c|h]pp`, the rest of the changes is just updating the code to work with them.

james-s-willis (approved):
Looks good to me :+1:

kotekan/kotekan.cpp
```diff
@@ -463,8 +463,8 @@ int main(int argc, char** argv) {
 
     auto& metrics = prometheus::Metrics::instance();
     metrics.register_with_server(&rest_server);
-    auto& kotekan_running_metric = metrics.add_gauge("kotekan_running", "main");
-    kotekan_running_metric.set(running);
+    auto kotekan_running_metric = metrics.add_gauge("kotekan_running", "main", {});
+    kotekan_running_metric->labels({}).set(running);
```
jrs65, line 467: I've been trying to figure out what the point of this is (i.e. why call labels every time), and I think it's because you `.labels(...)` itself is leaky (i.e. it returns references to the `Metric` objects that are in theory owned by the `MetricsFamily` instance), and you're trying to avoid storing references that may be long lived.

This seems like it's also a fragile point (although probably much less fragile) and so if you're going to fix it you should do it properly. I think someone could still in theory bind a reference to a raw `Metric` object in an intialization list and that may outlive the `MetricsFamily` objects it belonged to, or the `MetricsFamily` could delete the instance from its `deque` (though no deletions are done at the moment, that's not to say they will never be done).

Potential resolutions to this I can imagine:
- labels should return a `std::shared_ptr<Metric>` and `MetricsFamily` should store shared pointers.
- labels should return an uncopyable proxy object that holds a `shared_ptr` to metrics family (this won't deal with the issue of deletion of the Metric from the MetricFamily).

Not sure if there's any other options. I wondered about trying to remove the move assignment operator to prevent binding a reference to the temporary returned by labels, but I don't think that works at all.

jrs65 (changes requested):
I'm not sure exactly where it should go, but I think this could do with some clear documentation/comments about the point of the changes. It took me a while to remember the point.

PR #984, opened 2021-08-13
Add sidebar to show trackers_current info

Jintao-Xu:
Add a sidebar to display tracker current info

### DIFF
--- a/tools/debug_tool/static/pipeline_viewer.js
+++ b/tools/debug_tool/static/pipeline_viewer.js
@@ -3,13 +3,17 @@ function isIE() { return ((navigator.appName == 'Microsoft Internet Explorer') |
 // Time between updating kotekan metrics
 const POLL_WAIT_TIME_MS = 1000;
 // Poll kotekan via web server every POLL_WAIT_TIME_MS and update page with new metrics
-function poll(buffer_labels, stage_labels) {
+function poll(buffer_labels, stage_labels, sidebar) {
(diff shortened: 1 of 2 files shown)


Jintao-Xu:
image
image

Jintao-Xu:
Since I cannot trigger redraw properly after using many online solutions, I added this new sidebar to display trackers. It works fine, and interaction goes smoothly.

It is possible if we also want each stage node as a button to control tracker display on the right.

andrerenard:
I think it would be good to make clicking the node in the graph open the item in the side bar.   I suggest we also exclude the stages which haven't been clicked on.  In some of our pipelines we have 50+ stages, and it could get hard to find the item you want in the side bar. 

In terms of the sidebar format.   I would like to see the current value as well, and also display the units.    It might also be worth creating a standard format for all the numbers that doesn't require a line for each number.   e.g.

```
fill_frame_time:  <cur_value> <unit> (<avg>, <std>, <min>, <max>).
```
And then include a legend to indicate the order of values.   Although option would be to generate a table with a header. 

Also I would like to see maybe the first 2 stats trackers in the node on the graph itself, just showing the name, current value, and unit.

Jintao-Xu:
image
image

andrerenard (approved):
This all looks great!  Long term it looks like it will make sense to use the KotekanTrackers for CPU utilization, and that might simplify some of the code here, but that's for another PR.   For now we can merge this, and just focus on displaying the historical data.

PR #940, opened 2021-04-01
Support for next gen hardware.

andrerenard:
This is an unpleasantly large PR.   The main items include:

- Adding initial support for AMD HIP
- Fixes issues with OpenCL PCIe transfer speeds (memory registration)
- Updates to the CUDA wrapper (fixes transfer speed issue, improves profiling, and fixes issues with recent cmake changes)
- Numerous test configuration files for pushing the new hardware to its limits.  (Needs clean up before merging)
- New custom profiling API for all GPU libraries.

PR #986, opened 2021-08-17
Improvements to the GPU patient objects

andrerenard:
This PR is the first a few PRs designed to breakup #940 into more manageable PRs. 

* Adds a memory registration function for CUDA and OpenCL to improve host<->device async copy speeds
* Switch to using the new `KotekanTrackers` object to generate more stats from the command run times
* Add new profiling output functions for command line based kernel performance logging (useful for rapid testing)

### DIFF
--- a/lib/gpu/gpuProcess.hpp
+++ b/lib/gpu/gpuProcess.hpp
@@ -23,6 +23,14 @@ public:
                kotekan::bufferContainer& buffer_container);
     virtual ~gpuProcess();
     void main_thread() override;
+
+    /**
+     * @brief Reset server call back function for the profiling information
+     *
+     * @param conn The connection object required for the callback
+     * @todo This function will be deprecated by the KotekanTrackers stats once the
+     *       python script for monitoring GPU utilization is updated to that system.
+     */
     void profile_callback(kotekan::connectionInstance& conn);
 
     /// Returns the dot string formatted graph for the GPU pipeline
@@ -33,6 +41,7 @@ protected:
                                        const std::string& unique_name) = 0;
     virtual gpuEventContainer* create_signal() = 0;
     virtual void queue_commands(int gpu_frame_id) = 0;
+    virtual void register_host_memory(struct Buffer * host_buffer) = 0;
     void results_thread();
     void init(void);
 
@@ -50,6 +59,7 @@ protected:
     // Config variables
     uint32_t _gpu_buffer_depth;
     uint32_t gpu_id;
+    uint32_t gpu_thread_id;
 };
 
 #endif // GPU_PROCESS_H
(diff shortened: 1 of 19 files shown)


lib/gpu/gpuProcess.hpp
```diff
@@ -50,6 +59,7 @@ class gpuProcess : public kotekan::Stage {
     // Config variables
     uint32_t _gpu_buffer_depth;
     uint32_t gpu_id;
+    uint32_t gpu_thread_id;
```
Jintao-Xu, line 62: This variable is not used.

lib/opencl/clProcess.cpp
```diff
@@ -46,5 +55,35 @@ void clProcess::queue_commands(int gpu_frame_id) {
         signal = ((clCommand*)command)->execute(gpu_frame_id, signal);
     }
     final_signals[gpu_frame_id]->set_signal(signal);
-    INFO("Commands executed.");
+    DEBUG("Commands executed.");
 }
+
+void clProcess::register_host_memory(struct Buffer* host_buffer) {
+    // Register the host memory in in_buf with the OpenCL run time.
+    ERROR("Called register_host_memory !");
```
Jintao-Xu, line 63: Might need to change the log level.

Jintao-Xu (commented):
They look good, but I'm not sure if we should use the same execute tracker for all of them. If something goes wrong, it will be hard to tell where the problem is.

andrerenard:
> They look good, but I'm not sure if we should use the same execute tracker for all of them. If something goes wrong, it will be hard to tell where the problem is.

I'm not too sure what you mean by the same tracker for all of them.  Do you mean the fact that each subclass uses the same tracker from the patient class?   In this case I don't see that as a problem, since we only ever run with one type of subclass based on which GPU API we are targeting.

Jintao-Xu:
> > They look good, but I'm not sure if we should use the same execute tracker for all of them. If something goes wrong, it will be hard to tell where the problem is.
> 
> I'm not too sure what you mean by the same tracker for all of them. Do you mean the fact that each subclass uses the same tracker from the patient class? In this case I don't see that as a problem, since we only ever run with one type of subclass based on which GPU API we are targeting.

Sorry for the confusion. Yup, that is exactly what I meant. Then everything looks great!

lib/utils/visUtil.cpp
```diff
@@ -364,6 +364,13 @@ double StatTracker::get_std_dev() {
     return std_dev;
 }
 
+double StatTracker::get_current() {
+    std::lock_guard<std::mutex> lock(tracker_lock);
```
james-s-willis, line 368: That's a handy class, you don't even have to unlock the mutex explicitly like in C.

james-s-willis (changes requested):
Looks okay to me barring the `ERROR` statements.

lib/utils/visUtil.cpp
Jintao-Xu: The mutex has been changed to recursive_mutex.
```suggestion
    std::lock_guard<std::recursive_mutex> lock(tracker_lock);
```

PR #987, opened 2021-08-18
Add dumped files to web interface

Jintao-Xu:


Jintao-Xu:
image

andrerenard:
Looks good! I think the `time` and `name` labels are backwards, and also likely better to put the `time` after the `name`, `cur` and `unit` columns.

Jintao-Xu:
> Looks good! I think the `time` and `name` labels are backwards, and also likely better to put the `time` after the `name`, `cur` and `unit` columns.

Oops, my bad.

Jintao-Xu:
image

PR #989, opened 2021-08-20
Add stats in dump files

Jintao-Xu:


Jintao-Xu:
I think it might be a problem if someone tries to call get_xxx() in run time. To ensure get_json and get_current_json get the right stats and samples, the lock should be inside of them.

andrerenard:
> I think it might be a problem if someone tries to call get_xxx() in run time. To ensure get_json and get_current_json get the right stats and samples, the lock should be inside of them.

I think you can use `std::recursive_mutex` to get around this problem, and still have locks in the `get_xxx()` functions.  See: https://en.cppreference.com/w/cpp/thread/recursive_mutex

andrerenard (approved):

PR #998, opened 2021-10-06
Baseband archiver script

chitrangpatel:
Conversion script for CHIME/FRB from raw binary files to hdf5.


andrerenard:
Can you rebase this on `baseband-remove-bip-buffer` and have this PR directed into `baseband-remove-bip-buffer` instead of `develop`?   It's hard to see your changes against Davor's original, and as I make changes to Davor's code in his old branch, it will be hard to keep this in sync.

chitrangpatel:
Sure!

python/kotekan/scripts/baseband_archiver.py
```diff
@@ -213,13 +229,18 @@ def process_raw_file(file_name: str, config: Dict[str, int], dry_run: bool = Fal
...
         )
-        sample_present[
-            frame_start_idx : (frame_start_idx + frame_metadata.valid_to)
-        ] = True
+        # sample_present[
+        #    frame_start_idx : (frame_start_idx + frame_metadata.valid_to)
+        # ] = True
+        clip_after += frame_metadata.valid_to
+
+    baseband = baseband[:clip_after,]
```
leungcalvin, line 237: Looks good to me!

leungcalvin (commented):
Everything looks good to me. The gold standard would be to incorporate `convert()` into `test_baseband_readout.py`. From there on the only regime that we will not have tested is on the sky.

python/kotekan/scripts/baseband_archiver.py
```diff
@@ -99,9 +99,22 @@ def create_baseband_archive(frame_metadata: baseband_buffer.BasebandMetadata):
         frame_metadata.freq_id,
         frame_metadata.frame_fpga_seq,
     )
-    file_name = f"baseband_{ event_id }_{ freq_id }.h5"
+    date = datetime.utcfromtimestamp(frame_metadata.time0_ctime).date()
+    y = str(date.year).zfill(4)
+    m = str(date.month).zfill(2)
+    d = str(date.day).zfill(2)
+    # confirm root path on the recv node.
+    root = f"/data/chime/baseband/raw"
```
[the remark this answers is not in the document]
chitrangpatel: agreed. Its now a cli parameter.

config/chime_science_run_gpu.yaml
```diff
@@ -32,7 +32,7 @@ log_level: info
 num_elements: 2048
 num_local_freq: 1
 num_data_sets: 1
-samples_per_data_set: 49152
+samples_per_data_set: 512
```
andrerenard, line 35: @chitrangpatel can you revert this?  I'm not sure how it got into this PR, but it's clearly wrong.

config/baseband_archiver.yaml
```diff
@@ -0,0 +1,2063 @@
+##########################################
+#
+# baseband_archiver.yaml
+#
+##########################################
+num_elements: 2048
+sampling_rate: 2
+fft_length: 2048
+nyquist_zone: 2
+samples_per_data_set: 49152
```
andrerenard, line 10: Do we actually need this value?  Or is it incorrect?  The frames coming from the x-engine are 512 samples long because that's what the transfer buffer uses.  (The primary buffer is 49152, but that doesn't have any impact on the baseband system down stream).

andrerenard (changes requested):

config/baseband_archiver.yaml
chitrangpatel: Actually, we don't need this file at all. I had added it because we had imagined not using the realtime Kotekan config for this system at all. Since we are currently still using the realtime config, I will remove this and we can revisit this later.

config/chime_science_run_gpu.yaml
chitrangpatel: I think Davor had set it to 49152 while he was converting and we changed it to 512 since thats what we are currently using. Should I revert or leave it as is?

andrerenard: Revert to develop, it should be `49152`

config/baseband_archiver.yaml
andrerenard: Ah perfect, yes let's just remove this file then.

config/chime_science_run_gpu.yaml
andrerenard: Sorry revert to the parent branch, develop will have other changes as well.

andrerenard (approved):

PR #1002, opened 2021-11-10
Fallback to system time if GPS time is not set

andrerenard:
Also fixes an issue where the ICETelescope object didn't correctly return the `gps_time_enabled()` value correctly.

### DIFF
--- a/lib/stages/visAccumulate.cpp
+++ b/lib/stages/visAccumulate.cpp
@@ -304,6 +304,13 @@ void visAccumulate::main_thread() {
 
         // Start and end times of this frame
         timespec t_s = ((chimeMetadata*)in_buf->metadata[in_frame_id]->metadata)->gps_time;
+        // GPS time not set
+        if (t_s.tv_sec == 0) {
+            TIMEVAL_TO_TIMESPEC(
+                &((chimeMetadata*)in_buf->metadata[in_frame_id]->metadata)->first_packet_recv_time,
+                &t_s);
+            WARN("GPS time not set, using much less accurate system time instead.");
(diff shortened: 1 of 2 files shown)


lib/stages/visAccumulate.cpp
```diff
@@ -304,6 +305,13 @@ void visAccumulate::main_thread() {
 
         // Start and end times of this frame
         timespec t_s = ((chimeMetadata*)in_buf->metadata[in_frame_id]->metadata)->gps_time;
+        // If GPS time is not set, fall back to the system's best guess at the time.
+        if (t_s.tv_sec == 0) {
```
jrs65, line 309: This is probably cleaner to test `Telescope::instance().gps_time_enabled()`, plus then you only need to do the test once and then just save the result.

lib/stages/visAccumulate.cpp
```diff
@@ -304,6 +305,13 @@ void visAccumulate::main_thread() {
...
+            TIMEVAL_TO_TIMESPEC(
+                &((chimeMetadata*)in_buf->metadata[in_frame_id]->metadata)->first_packet_recv_time,
+                &t_s);
+            WARN("GPS time not set, using much less accurate system time instead.");
```
jrs65, line 313: If anyone expects that they might want to run in this state, this is going to be a very verbose message. Maybe only warn the first time (or when `gps_time_enabled()` is evaluated if doing the above suggestion).

jrs65 (changes requested):

jrs65 (approved):

PR #1003, opened 2021-11-10
Add high cadence (~80ms) config for single iceboard systems

andrerenard:


andrerenard:
I'm going to close this, on second thought, I don't think it's worth having two configuration files for what is a relatively small change.

PR #1004, opened 2021-11-19
fix(ICETelescope) Fix mapping for shuffle128

andrerenard:


lib/utils/ICETelescope.cpp
```diff
@@ -129,8 +129,7 @@ freq_id_t ICETelescope::to_freq_id(stream_t stream, uint32_t ind) const {
         case 8: // 16 ICEBoards (256 elements) e.g. Pathfinder/HIRAX-128
             return stream_id.slot_id + stream_id.link_id * 16 + ind * 128;
         case 16: // 8 ICEBoards (128 elements) e.g. Allenby
-            // TODO: Check this mapping
```
aelanman, line 132: Since you removed the TODO to check the mapping, I assume this has been done?

lib/utils/ICETelescope.cpp
```diff
@@ -129,8 +129,7 @@ freq_id_t ICETelescope::to_freq_id(stream_t stream, uint32_t ind) const {
         case 8: // 16 ICEBoards (256 elements) e.g. Pathfinder/HIRAX-128
```
aelanman, line 129: This also applies to GBO and Hat Creek, right?

aelanman (approved):
Looks good to me. Just a couple comments to confirm some details.

lib/utils/ICETelescope.cpp
andrerenard: Yes, I checked the mapping with JF.

lib/utils/ICETelescope.cpp
andrerenard: Yes, I'll add those just for completeness.

PR #1005, opened 2021-11-19
fix(logging) Fix logging for OpenCL API errors

andrerenard:


aelanman (approved):
Looks fine to me. It's just replacing char arrays with standard library strings?

andrerenard:
> Looks fine to me. It's just replacing char arrays with standard library strings?

The main change is `internal_logging()`  -> `FATAL_ERROR_NON_OO()` which is the new logging API.  That works better with `std::string`, so I also changed the function above.

PR #1006, opened 2021-11-19
Add HIP support

andrerenard:
Adds AMD ROCm HIP language support. 

These files have been around for a while, but never merged because #940 was too large to easily review.


andrerenard:
TODO: ~The Input/Output data command objects need to be made generic (see recent changes with the CUDA commands).~  Done now.

lib/hip/hipProcess.cpp
```diff
@@ -0,0 +1,65 @@
...
+REGISTER_KOTEKAN_STAGE(hipProcess);
+
+// TODO Remove the GPU_ID from this constructor
+hipProcess::hipProcess(Config& config_, const string& unique_name,
+                       bufferContainer& buffer_container) :
+    gpuProcess(config_, unique_name, buffer_container) {
+    device = new hipDeviceInterface(config_, gpu_id, _gpu_buffer_depth);
+    dev = device;
+    device->prepareStreams();
+    // CHECK_HIP_ERROR(hipProfilerStart());
```
kvand: This should probably be a config option rather than a hard-uncomment, but I don't think worth holding up the merge on account of.

lib/hip/hipUtils.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+#include <assert.h>
+
+// Convenience function for checking HIP runtime API results
+// can be wrapped around any runtime API call. No-op in release builds.
+#define CHECK_HIP_ERROR(result)                                                                    \
+    if (result != hipSuccess) {                                                                    \
+        FATAL_ERROR_NON_OO("Error at {:s}:{:d}; Error type: {:s}", __FILE__, __LINE__,             \
+                           hipGetErrorString(result));                                             \
+    }
```
kvand, line 22: Should I expect this macro to be identical to the NON_OO one?

kvand (approved):
A couple of tiny things, but this looks good enough to me to merge in.

lib/hip/hipUtils.hpp
andrerenard: It's almost the same.  The key difference is the `NON_OO` uses the global config log level (instead of the per-stage overridable log_level).  However, in cases like this where it might be called from a function or class that doesn't inherit from `kotekanLogging` we have to use this version.

lib/hip/hipProcess.cpp
andrerenard: We already have a `profiling` configuration option in the base class, so that's an easy change.  However now, I'm not sure why this was commented out in the first place.  I was running profiling tests on this branch, so either this is a commit error, or we don't actually need this line.   I'll check this before merging.

andrerenard: Turns out this deals with the external profiling application, not the internal timers.  So it's simply not needed, will remove.

PR #1008, opened 2021-11-30
cleanup and delete files after conversion

chitrangpatel:


python/kotekan/scripts/convert_baseband.py
```diff
@@ -191,6 +191,7 @@ def convert_data(sqlite, conn, e, num_threads):
                     f"UPDATE conversion SET status = 'FINISHED' WHERE event_no = {e[0]}"
                 )
                 conn.commit()
+                os.system(f"rm -rf /data/baseband_raw/baseband_raw_{e[0]}")
```
andrerenard, line 194: Should we also delete the lock in the case that we've decided to process a file which has an expired lock?

chitrangpatel: The above command will only run if all the files have been converted to hdf5 and the conversion process starts only when there are no more lock files for an event so it should not trigger if there are lock files in the directory.

andrerenard: What about the case that the lock files are very old, e.g. the writer crashed and never cleared the files.   I thought there was a check to allow the system to still process locked files if they exceeded some age (e.g. 24 hours)?

chitrangpatel: Yes, I have it set to 6 hours (can increase it to 24 if we like). This part will only run when hdf5 files for all the raw data files exist on the baseband archiver. Those hdf5 files can be created from the data files or the lock files. So whenever this part runs, we first ensure that the hdf5 files exist.

andrerenard: 6 hours should be fine.   I'm not entirely sure I follow the logic.  What happens if the baseband recv and/or the entire corrector crashes mid dump (e.g. power failure) and doesn't remove the lock files.  And then later when your script checks their won't be hdf5 files, but there will be lock files.  Is that case also handled?  And if so I think we want to also delete the lock files once we finish?

chitrangpatel: right so if that happens, the conversion script will wait for 6 hours. In `/data/baseband_raw/baseband_raw_<event_no>` it will find `....data` and `....data.lock` files. It will check if the current time is > 6 hours plus the event time. If it is, it will use the lock files and convert to hdf5. When it confirms that all the `lock` files have a corresponding hdf5 file, it will call this statement.

python/kotekan/scripts/convert_baseband.py
```diff
@@ -154,8 +154,13 @@ def convert_data(sqlite, conn, e, num_threads):
     if unlocked is True or datetime.datetime.utcnow() > datetime.datetime.strptime(
         e[1], "%Y-%m-%d %H:%M:%S.%f"
     ) + datetime.timedelta(hours=3):
```
andrerenard, line 156: Why is this 3 hours?  Didn't we say this should be 6 hours below?

andrerenard (changes requested):

python/kotekan/scripts/convert_baseband.py
chitrangpatel: yes. Fixed it.

andrerenard (approved):

PR #1012, opened 2022-01-10
config(chime_keep_gpu_warm) Fix issues resulting from API changes

andrerenard:
* Adds the instrument coordinates
* Removes the baseband dump system config block (not longer needed with the new version).


anjakefala (approved):

anjakefala:
Can you link to the API changes?

andrerenard:
They are in: #977 and #927 (which is huge, but as far as this PR is concerned just removes the need to have a baseband section for the keep warm config).

PR #1013, opened 2022-01-25
Add the Astron Tensor-Core N2 kernels

### LINKED ISSUE #946: CUDA Kernels for HIRAX
This is a catch all ticket for the various CUDA related development work for HIRAX:

- [ ] Transpose kernels to format data for the N^2 Tensor Core kernel
- [X] N^2 Tensor-Core kernels
- [ ] Full array beamforming kernels for FRB searches.   This includes an FFT channelization stage. 
- [ ] Tracking beamformer kernel
- [ ] RFI kernels? 
- [X] Stages for post processing N^2.  
- [ ]  Beamforming post processing and network stages for full array and tracking beams.

andrerenard:
* Adds John Romein's CUDA based Tensor-Core kernel for correlation, and command objects to support it
* Adds support for generic input/output command objects
* Support run-time compiled kernels
* Adds new kernel/command object to convert from the ICEBoard formatted data into a shape that's efficient for the TC correlator. 

This is based on the code that @kvand wrote in `kv/cuda_corr`, but updated to the latest version of the Astron kernel  pulled forward to the current develop branch, and includes CUDA framework related changes from @andrerenard to generalize some functions.


lib/cuda/CMakeLists.txt
```diff
@@ -2,21 +2,23 @@ project(kotekan_cuda)
...
-    cudaCorrelatorKernel.cu)
+    # Kernels
+    cudaCorrelatorAstron.cpp
+    cudaShuffleAstron.cu)
 
 target_link_libraries(
     kotekan_cuda
     PRIVATE libexternal kotekan_libs -lstdc++
-    INTERFACE ${CUDA_LIBRARIES})
+    INTERFACE ${CUDA_LIBRARIES} cuda nvrtc)
```
kvand, line 21: Is it weird that these aren't already in ${CUDA_LIBRARIES}? Might be worth taking a quick look to make sure we're not overly complicating things...

lib/cuda/cudaCorrelatorAstron.cpp
```diff
@@ -0,0 +1,85 @@
...
+    _num_elements = config.get<int>(unique_name, "num_elements");
+    _num_local_freq = config.get<int>(unique_name, "num_local_freq");
+    _samples_per_data_set = config.get<int>(unique_name, "samples_per_data_set");
+    _num_data_sets = config.get<int>(unique_name, "num_data_sets");
+    _block_size = config.get_default<int>(unique_name, "block_size", 2);
+    _elements_per_thread_block =
+        config.get_default<int>(unique_name, "elements_per_thread_block", 128);
+    _num_blocks = config.get<int>(unique_name, "num_blocks");
+    _buffer_depth = config.get<int>(unique_name, "buffer_depth");
+    _gpu_mem_input = config.get_default<std::string>(unique_name, "in_gpu_mem", "voltage");
```
kvand: _gpu_mem_input/output, and in/out_gpu_mem are pretty stupid names, sorry for using those. Probably best to change those to match your much-improve "voltage" config param and something like "n2_output" (or maybe "corrleation_matrix"?)

lib/cuda/cudaInputData.hpp
```diff
@@ -14,15 +14,8 @@
...
  *
  * This is a cudaCommand that async copies a buffer from CPU to GPU.
- * This code also passes metadata along.
- *
- * @par GPU Memory
- * @gpu_mem in_buf           Input buffer, arbitrary size
- *     @gpu_mem_type         staging
- *     @gpu_mem_format       Any
  *
  * @author Keith Vanderlinde
```
kvand: Add Andre as author, or add a revision line here tagged to him?

lib/cuda/cudaOutputData.cpp
```diff
@@ -30,6 +30,8 @@ cudaOutputData::cudaOutputData(Config& config, const std::string& unique_name,
         }
     }
 
+    _gpu_mem = config.get_default<std::string>(unique_name, "gpu_mem", "n2_output");
```
kvand: this shouldn't be just hardcoded to the n2 data, should it?

kvand (changes requested):
Looks good overall! Only one substantive comment (a hardcode in the gpu_memory_out that I think should be config-based), plus a couple of variable name cleanups in the n^2 kernel wrapper.

lib/cuda/CMakeLists.txt
andrerenard: It turns out we do...  We need both for low level memory operations and run-time compiling respectively.   Annoyingly `cmake` is also in the middle of transitioning how CUDA libs are included, but it requires a newer version of cmake than on a lot of our systems.  The new system just needs to enable the language and it auto links the core libraries. These two items move to `CUDA::nvrtc` and `CUDA::cuda_driver` which get included with a new FindCUDAToolKit() macro, but would still need to be included.    This code seems to be maximally supported, if a bit clunky since it doesn't go through `find()` or another search macro.

lib/cuda/cudaOutputData.cpp
andrerenard: It's just a default value, but you are right, it really shouldn't have any default, that's just going to lead to bugs in config files.  Removed (in InputData as well).

lib/cuda/cudaCorrelatorAstron.cpp
andrerenard: I'm going to change them to `gpu_mem_voltage`, `gpu_mem_correlation_matrix`, etc.  These names end up getting repeated a lot the config file, which makes them somewhat hard to read/search without some difference.  And I want to have some designation for GPU memory names.

andrerenard: I'm also going to remove all default values for these, thinking about this carefully, it's just opening us up to future configuration file errors to have defaults here.

kvand (approved):
Looks good to me now.

PR #1015, opened 2022-04-06
Add support for Jinja config files

andrerenard:
This PR adds an optional config file type with the extension `.j2`.    Files with this type are written like normal YAML but with Jinja commands processed by the Jinja template engine converting them into pure YAML.   The process looks like "Jinga -> YAML -> JSON".   

This added two new kotekan command line options:
`--variables (-e) [json]` Pass extra variable to the jinja engine on the command line.
`--print-yaml (-y)` Print the YAML being used (works for both .yaml and .j2 files.

In addition to the above mode, the error handling around YAML files has been cleaned up, and no longer core dumps.   For example an error in a YAML configuration file will now report the error type and line number:
```
Error parsing yaml: 
while scanning a simple key
  in "../../config/single_iceboard_n2.yaml", line 246, column 1
could not find expected ':'
  in "../../config/single_iceboard_n2.yaml", line 248, column 1
Unable to load config from ../../config/single_iceboard_n2.yaml
```


kvand:
This makes jinja2 a required dependency, doesn't it? Presumably there should be a cmake check for it as well? Or throw an error if the import fails?

andrerenard:
It's only required if you want to run a `.j2` file, so existing installs should be ok without adding it, but I will add it as a requirement in `python/requirements.txt`, and a note in the documentation.  (cmake doesn't deal with python related things, at least not at the moment.)   For the error case where it's not installed, it should pass through the python import error message to kotekan, but since this is a new requirement, I add an exception handler to make the error message a bit cleaner.

andrerenard:
@kvand I've addressed the comment regarding the `jinja2` requirement.   Added it to `requirements.txt`, and updated docs and error handling.

kotekan/kotekan.cpp
```diff
@@ -356,15 +463,39 @@ int main(int argc, char** argv) {
...
+            exec_format_str += "-e '{:s}' ";
+        }
+        exec_format_str += "{:s}";
+
+        std::string exec_command = "";
+        if (jinja_variables == "")
+            exec_command = fmt::format(exec_format_str, yaml_to_json, config_file_name);
+        else
+            exec_command =
+                fmt::format(exec_format_str, yaml_to_json, jinja_variables, config_file_name);
```
ketiltrout: There is a potential security vulnerability here.

If I run kotekan with a malicious value for `-e` along the lines of:
```
kotekan -e "' \`rm -rvf *\` '" [...]
```
then I can execute arbitrary shell commands as the UID of kotekan.  This becomes an escalation attack whenever kotekan is installed setuid. (I use a backtick subshell to ensure my malicious payload is executed before the command I'm co-opting.)

The root problem is the use of `popen`, which invokes `sh(1)` to parse the string it's given.  Given that it's kotekan itself that built the string, this is completely unnecessary.  It would be best to replace `popen` with the standard `fork/execvp` idiom which would remove the need to compose a sh-style command line because you'd be passing the arguments to `python` explicitly, bypassing the shell parser.

It *might* be enough to run the user-supplied string through a JSON validator to ensure valid JSON, but I suspect JSON is expressive enough that I can escape the sh-single quotes within valid JSON, so I think the only other safe option is to properly `sh`-escape the user-supplied string before creating the command-line.

ketiltrout (changes requested):
I suspect I have way too much experience with IPC in C.  It's made me paranoid, but if you'd like me to implement a fork-and-exec replacement for the `popen` call, it should be fairly straightforward for me to do.

kotekan/kotekan.cpp
andrerenard: Good catch!   I don't think we have any situations where the privilege of `kotekan` is higher than the user in command line mode, but this should be fixed nonetheless.   If you want to write fork-and-exec replacement, that would be appreciated!  But if you don't have time I can do that as well, just let me know?

ketiltrout: sure, I'll give it a stab. I'll try to get something done today.

andrerenard: Thanks!

python/scripts/config_to_json.py
```diff
@@ -0,0 +1,128 @@
...
+
+Extra variables can be passed in with the -e flag, in json/dict format:
+ -e '{"my_val": 20, "my_val2": 30, "my_array": [0,2,1]}'
+   
+The JSON is returned in STDOUT, and error messages are returned in STDERR
+"""
+try:
+    import yaml, json, sys, os, subprocess, errno, argparse
+except ImportError as err:
+    sys.stderr.write(
```
ketiltrout, line 23: There's no guarantee that `sys` is in scope here.

Realistically, the only import that's going to fail for which installing the additional requirements will help is `yaml`.

I'd move the import of everything else (all of which are bundled with Python as part of its standard library) out of the try block.  If the import of any of them fail, then the user's Python is broken in some weird way and it's probably best for them if we just let Python explode (which will happen here anyways if importing `sys` failed.)

Also, I don't know in which order imports are processed.  A missing `yaml` may cause the exception handler to fire before the interpreter has had a chance to consider importing `sys`.

ketiltrout (approved):
An unlikely occurrence.  Not sure what would actually happen.

python/scripts/config_to_json.py
andrerenard: Good point, this way I can also improve the error message around this block.

PR #1018, opened 2022-04-06
Replace popen() call with fork+exec

ketiltrout:
Changes the signature of `exec` to take a `std::vector<std::string>&` instead of a simple `std::string&`.

Not sure if you want to tweak the error messages or not.

### DIFF
--- a/kotekan/kotekan.cpp
+++ b/kotekan/kotekan.cpp
@@ -30,8 +30,8 @@
 #include <string.h>    // for strdup
 #include <string>      // for string, basic_string, operator!=, operator<<, operator==
 #include <strings.h>   // for strcasecmp
-#include <syslog.h>    // for closelog, openlog, LOG_CONS, LOG_LOCAL1, LOG_NDELAY
 #include <sys/wait.h>  // for waitpid
+#include <syslog.h>    // for closelog, openlog, LOG_CONS, LOG_LOCAL1, LOG_NDELAY
 #include <type_traits> // for underlying_type, underlying_type<>::type
 #include <unistd.h>    // for optarg, sleep, pipe, fork, execvp, dup2
 #include <utility>     // for pair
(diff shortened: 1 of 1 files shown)


kotekan/kotekan.cpp
```diff
@@ -282,25 +283,71 @@ void print_json_version() {
...
+      for (i = 0; i < cmd.size(); ++i) {
+          args[i] = strdup(cmd[i].c_str());
+      }
+      args[i] = NULL;
+
+      // exec to subprocess.  On success, this does not return.
+      execvp(args[0], args);
+
+      // exec-ing failed
+      std::cerr << "exec to " << cmd[0] << " failed!" << std::endl;
```
andrerenard: For future reference, we use the built-in logging macros.   In this case `ERROR_NON_OO(format_string, values)`, the format string uses `fmt` style notation.  See: https://fmt.dev/latest/syntax.html

kotekan/kotekan.cpp
```diff
@@ -282,25 +283,71 @@ void print_json_version() {
...
 
-std::string exec(const std::string& cmd) {
-    std::array<char, 256> buffer;
-    std::string result;
-    FILE* pipe = popen(cmd.c_str(), "r");
+std::string exec(std::vector<std::string>& cmd) {
+    // create a pipe for interprocess communication
+    int pipefds[2];
+    if (pipe(pipefds))
+      throw std::runtime_error("Could not create a pipe");
```
andrerenard: For reference, we use 4 space indentation.   If you run `make clang-format` from `build/` it will automatically apply the style guide.

kotekan/kotekan.cpp
```diff
@@ -466,24 +513,22 @@ int main(int argc, char** argv) {
         std::lock_guard<std::mutex> lock(kotekan_state_lock);
         INFO_NON_OO("Opening config file {:s}", config_file_name);
 
-        // Create the format string, adding the yaml dump, and extra vars if needed
-        std::string exec_format_str = "python -c '{:s}' ";
+        // Create the command line, adding the yaml dump, and extra vars if needed
+        std::vector<std::string> exec_command;
```
andrerenard, line 517: I like this way of constructing the command string a lot better!

andrerenard (approved):
This looks great, thanks for doing this!   A few small comments, but I can easily address them in my branch.  Will merge in first thing tomorrow.

PR #1019, opened 2022-04-20
Remove HCC support

andrerenard:
AMD has deprecated HCC support in favour of HIP.   We never really had functional HCC support, so this just removes all the early code related to HCC.


kvand (approved):
Looks good to me.

PR #1022, opened 2022-04-26
Add FRB beamformer

eschnett:


eschnett:
@andrerenard This code should compile, but doesn't do anything useful yet.

Currently, the CI tests `lint` and `iwyu` are fighting, I can't make them both happy. They disagree about the order of `#include` statements.

lib/stages/FRBBeamformer.hpp
```diff
@@ -0,0 +1,20 @@
+#ifndef FRBBEAMFORMER_H
+#define FRBBEAMFORMER_H
+
+#include <string>
+
+#include "bufferContainer.hpp"
+#include "Config.hpp"
+#include "Stage.hpp"
+
+class FRBBeamformer : public kotekan::Stage {
```
andrerenard, line 10: Haven't looked over this in detail yet, but you will want to use a `commandObject` (`cudaCommand` in this case) not a stage for invoking the kernel.  See for example: `lib/cuda/cudaCorrelatorAstron.cpp` or `lib/cuda/cudaShuffleAstron.cu`

andrerenard:
How much of this process is could be automated?   It feels like there is a risk as we make changes to the Julia code, there will be a lot of work to update all these intermediate files, without getting any readable code level viability to what's actually changed.  Also with the primary code living elsewhere, this is essentially equivalent to including a library with a precompiled binary object. 

If feels like a better approach would be to include the Julia code directly into kotekan, and use cmake scripts to build all these intermediate and final generated source files at compile time (and never commit them directly into the repo).  This would make Julia an (optional) dependency of kotekan when using these kernels, but I think that's a reasonable trade off for enhanced maintainability and git history clarity.  Although maybe I'm not appreciating the complexity of that? 

Alternatively if you want your Julia source code to be external to kotekan, then perhaps we could make it into a library with standard header files and link it into kotekan?

eschnett:
At the moment I'm experimenting with which format is best to transfer PTX code. There are currently two ways, one uses `asm` statements with PTX code, the other uses CUDA tools to generate a file.

Both are automated, or can be. The first requires piping Julia into a perl script (to add quotes around the PTX code to turn it into C), the second requires running two CUDA commands.

PR #1030, opened 2022-06-16
jinja support as module

aelanman:
Replaces the `config_to_json.py` script with a new python submodule with the same functionality.


andrerenard (changes requested):
I would like to keep this as one file.  You can do that by using an `if __name__ == "__main__":` block to encase the parts of the code currently in `config_to_json.py`.   Doing it this way will allow kotekan to call the script directly (or embed it) without having to first install the `kotekan.config` python package in `config.py`, but also allow someone to install/import that same file without triggering the script.

andrerenard:
You can remove the `Update the call in kotekan.cpp` part from this PR, I can take care of the `cmake` scripts to do that once this is merged.

aelanman:
@andrerenard Thanks for the review -- I've addressed your comments and the checks are all passing.

python/kotekan/config.py
```diff
@@ -0,0 +1,163 @@
...
+
+        # Convert to yaml
+        config_yaml_raw = template.render(options_dict)
+
+        # Dump the rendered yaml file if requested
+        if dump:
+            sys.stderr.write(config_yaml_raw + "\n")
+
+        # TODO Should we also lint the output of the template?
+        config_yaml = yaml.safe_load(config_yaml_raw)
```
andrerenard, line 128: This should be in a `try:` block to catch `yaml.YAMLError` and output the parsing error as above and in the original script.

andrerenard (changes requested):

python/kotekan/config.py
aelanman: Why do we want to parse the error and print to stdout instead of just allowing the YAMLError to be raised normally?

andrerenard: Because when called within `kotekan.cpp` the error message doesn't make it to command line of the person running the code.  I think this was because kotekan assumes the config will come via STDOUT, and the python error message from the exception also went to STDOUT.   This resulted in us not seeing what line number/syntax error was generating the problem - just a JSON parse error from kotekan.  Given we've moved this to a library + script design, perhaps the correct way to deal with this is to allow it raise the exception here, but in the script part use a `try:` block to catch the exception and write out the err string details to STDERR and exit the script without writing anything to STDOUT.

aelanman: That makes sense. I just pushed some changes that add an option to the script mode of `config.py`, which will write out exceptions to stderr instead of stdout.

python/kotekan/config.py
```diff
@@ -0,0 +1,172 @@
...
+    parser.add_argument(
+        "-s", "--stderr", help="Print exceptions to stderr", action="store_true"
+    )
+    args = parser.parse_args()
+
+    options = args.variables
+
+    try:
+        print(load_config_file(args.name, dump=args.dump, jinja_options=options))
+    except (IOError, yaml.YAMLError, ImportError, jinja2.TemplateNotFound) as err:
```
andrerenard: We should add `jinja2.exceptions.TemplateSyntaxError` to the list of exceptions.

python/kotekan/config.py
```diff
@@ -0,0 +1,172 @@
...
+        if args.stderr:
+            sys.stderr.write(err.message + "\n")
```
andrerenard: I think this should be:
```suggestion
            sys.stderr.write("Failure loading config: " + str(err) + "\n")
```
The `err.message` property doesn't exist for all the exceptions, but "to string" seems to work for all the ones I tried.

python/kotekan/config.py
```diff
@@ -0,0 +1,172 @@
...
+    All files without a ".j2" extension are treated as YMAL, files with a ".j2"
+    extension are processed by the Jinja template render.
+
+    Extra variables can be passed in with the -e flag, in json/dict format:
+     -e '{"my_val": 20, "my_val2": 30, "my_array": [0,2,1]}'
+
+    The JSON is returned in STDOUT, and error messages are returned in STDERR
+    """
+
+    parser = argparse.ArgumentParser(
```
andrerenard, line 151: We need to `import argparse`, otherwise (at least with my system) this script won't run.

andrerenard (changes requested):
This is looking very good, but opon testing I found a few small issues we should fix before merging.

andrerenard (approved):
Looks good now, thanks for all these updates!   Once it passes CI you can merge.   Please remember to use the github squash merge with the kotekan repo.  I generally also try to include the PR description in the squashed commit message.

PR #1031, opened 2022-07-07
Added the new compute_phases function

VishwangiShah312:


lib/stages/ReadGain.cpp
```diff
@@ -133,6 +140,26 @@ bool ReadGain::update_gains_frb_callback(nlohmann::json& json) {
     return true;
 }
 
+bool ReadGain::update_UT1_UTC_offset(nlohmann::json& json) {
```
andrerenard: This function should be moved to the `BeamformingPhaseUpdate` base class files.   Since that's where this value is needed.  I don't think anything in ReadGains needs the value.

lib/stages/FloatPhaseUpdate.cpp
```diff
@@ -11,13 +11,87 @@ FloatPhaseUpdate::FloatPhaseUpdate(kotekan::Config& config, const std::string& u
...
+    if (month < 3) {
+        month = month + 12;
+        year = year - 1;
+    }
+    uint day = utc_time->tm_mday;
+    uint hour = utc_time->tm_hour;
+    float JD = 2 - int(year / 100.) + int(int(year / 100.) / 4.) + int(365.25 * year) //See this conversion for Skyfield
+               + int(30.6001 * (month + 1)) + day + 1720994.5; 
+    double UT = (hour) + (utc_time->tm_min / 60.)
+               + (utc_time->tm_sec + _beam_coord.UT1_UTC + gps_time.tv_nsec / 1.e9) / 3600.;                //TO DO:UT1_UTC should be a double value in struct _beam_coord containing UT1-UTC in seconds
```
andrerenard, line 30: Once the UTC-UT1 call back function is added to the base class, then the value should be used here.

config/chime_science_run_gpu.yaml
```diff
@@ -1017,6 +1019,7 @@ read_gain:
   updatable_config:
     gain_frb: /updatable_config/frb_gain
     gain_tracking: /updatable_config/tracking_gain
+    UT1_UTC: /updatable_config/UT1_UTC_offset
```
andrerenard: We'll need to add the `FloatPhaseUpdate` stage as a new entry, then add to that entry the block:
```
updatable_config:
  UT1_UTC: /updatable_config/UT1_UTC_offset
```

config/chime_science_run_gpu.yaml
```diff
@@ -110,6 +110,8 @@ updatable_config:
   frb_gain:
     kotekan_update_endpoint: json
     frb_gain_dir: /mnt/frb-archiver/daily_gain_solutions/Latest_FRB
+  UT1_UTC_offset:
+    UT1_UTC_val: 0.00
```
andrerenard: Should we make this a recent value, so we have at least a rough estimate when we start the system?    We might also want to have some system in the scheduler to check if the value is old?   e.g. in the case we fully restart the array, and it's still some number of hours before the next update?   Alternatively maybe the TBS should update every 30-60 minutes?

config/chime_science_run_gpu.yaml
```diff
@@ -3966,3 +3969,1029 @@ input_reorder:
...
+ [-33.03944463862491, -31.810682490811796],
+ [-33.03906693550881, -31.505882724833615],
+ [-33.038689232392706, -31.20108295885543],
+ [-33.038311529276605, -30.896283192877245],
+ [-33.037933826160504, -30.59148342689906],
+ [-33.0375561230444, -30.286683660920875],
+ [-33.0371784199283, -29.98188389494269],
+ [-33.0368007168122, -29.677084128964506],
+ [-33.0364230136961, -29.37228436298632],
+ [0, 0],
```
andrerenard, line 4010: We might want to have a check in the `compute_phases` function to set feeds without a position to have [0,0] for their phase to null those values in the beamformer.   Since I'm not sure if we can always expect that feeds without positions are also nulled in the gains (although most of the time I'd expect that to be the case).

andrerenard (changes requested):
This is looking good, but we need to fix a few small details before it's ready to test.  See inline comments.

config/chime_science_run_gpu.yaml
VishwangiShah312: What should the name of this stage be? `float_phase_update` like the `read_gain` stage?

lib/stages/ReadGain.cpp
VishwangiShah312: Done!

lib/stages/FloatPhaseUpdate.cpp
VishwangiShah312: Done!

config/chime_science_run_gpu.yaml
andrerenard: I think `beamformer_phase_generation`.  Since that's basically what this stage is doing.

lib/stages/ReadGain.hpp
```diff
@@ -97,6 +97,8 @@ class ReadGain : public kotekan::Stage {
 
     /// Directory path where gain files are
     std::string _gain_dir_frb;
+    /// UT1-UTC value
+    std::float _UT1_UTC_val;
```
andrerenard: Can you remove these extra lines that aren't used in ReadGain.hpp, and .cpp.

config/chime_science_run_gpu.yaml
VishwangiShah312: Done!

config/chime_science_run_gpu.yaml
VishwangiShah312: Set the value to recent value.

lib/stages/ReadGain.hpp
VishwangiShah312: Done!

config/chime_science_run_gpu.yaml
VishwangiShah312: Done!

andrerenard:
Closing this PR as we've merged the changes in https://github.com/kotekan/kotekan/pull/1261

PR #1033, opened 2022-07-14
Better error message when metadata pool is too small

andrerenard:
Changes the error message from the confusing:
```
kotekan: /code/build/kotekan/lib/core/metadata.c:131: request_metadata_object: Assertion `container != NULL' failed.
```
To the more helpful:
```
The metadata pool `/main_pool` is out of metadata objects, try increasing `num_metadata_objects:` in the config.
```


aelanman (approved):
Looks good to me! Thanks for making the update

PR #1053, opened 2022-09-19
Optimized copy_packet_vdif loop

ninaneiman:
Me and @mhvk moved a few things from the inner-most loop of the copy_packet_vdif to make it faster. It works, data seems to be still intacked. Can check it with the real source to be sure. Could you check if new loop makes sense?

### DIFF
--- a/lib/dpdk/iceBoardVDIF.hpp
+++ b/lib/dpdk/iceBoardVDIF.hpp
@@ -291,7 +291,13 @@ void iceBoardVDIF::copy_packet_vdif(struct rte_mbuf* mbuf) {
     int from_idx = header_offset + offset;
     int mbuf_len = mbuf->data_len;
     for (uint32_t time_step = 0; time_step < samples_per_packet; ++time_step) {
+        int output_idx_base = (vdif_frame_location * frame_size +           // Frame location in output buffer.
+			       vdif_header_len +        // Offset for the vdif header.
+			       tel.to_freq_id(encoded_id, 0) + // Offset for ARO as in tel.to_freq_id for _num_freq_per_stream = 128
+			       vdif_packet_len * num_elements * time_step); // Time step in output frame.
         for (int freq = 0; freq < 128; ++freq) {
+	    int output_idx = (output_idx_base +
+			      freq * 8); // Location in the VDIF packet is just frequency.
             for (int elem = 0; elem < num_elements; ++elem) {
 
                 // Advance to the next mbuf in the chain.
@@ -301,19 +307,11 @@ void iceBoardVDIF::copy_packet_vdif(struct rte_mbuf* mbuf) {
                     from_idx -= mbuf_len; // Subtract the last mbuf_len from the current idx.
                     mbuf_len = mbuf->data_len;
                 }
-
-                int output_idx =
-                    vdif_frame_location * frame_size +           // Frame location in output buffer.
-                    vdif_packet_len * num_elements * time_step + // Time step in output frame.
-                    vdif_packet_len * elem + // VDIF pack for the correct element (ThreadID).
-                    vdif_header_len +        // Offset for the vdif header.
-                    tel.to_freq_id(encoded_id,
-                                   freq); // Location in the VDIF packet is just frequency.
-
                 // After all that indexing copy one byte :)
                 out_buf_frame[output_idx] = *(rte_pktmbuf_mtod(mbuf, char*) + from_idx);
 
                 from_idx += 1;
+		output_idx += vdif_packet_len; // VDIF packet for the next element (ThreadID).
             }
             // If we only take 2 elements, then we have to skip 14
             from_idx += total_num_elements - num_elements;


andrerenard (changes requested):
The logic all checks out to me (at least assuming I didn't miss something, it's been a while since I've thought about this code).   It looks like you just need to run `make clang-format` and commit the format changes to make it pass the CI, and then once we double check with a source transit we can merge.

lib/dpdk/iceBoardVDIF.hpp
```diff
@@ -291,7 +291,13 @@ void iceBoardVDIF::copy_packet_vdif(struct rte_mbuf* mbuf) {
     int from_idx = header_offset + offset;
     int mbuf_len = mbuf->data_len;
     for (uint32_t time_step = 0; time_step < samples_per_packet; ++time_step) {
+        int output_idx_base = (vdif_frame_location * frame_size +           // Frame location in output buffer.
+			       vdif_header_len +        // Offset for the vdif header.
+			       tel.to_freq_id(encoded_id, 0) + // Offset for ARO as in tel.to_freq_id for _num_freq_per_stream = 128
```
mhvk, line 296: @andrerenard - this call to a method seemed to be the dominant reason the compiler couldn't make this optimization itself. In principle, one could calculate something like
```
freq0_offset = tel.to_freq_id(encoded_id, 0)
freq_step = 1024 // tel.num_freq_per_stream
```
and then loop to `freq < tel.num_freq_per_stream`. I think this covers the contents of the switch statement inside `to_freq_id` -- but is this worth it?

lib/dpdk/iceBoardVDIF.hpp
```diff
@@ -301,19 +307,11 @@ void iceBoardVDIF::copy_packet_vdif(struct rte_mbuf* mbuf) {
...
-                int output_idx =
-                    vdif_frame_location * frame_size +           // Frame location in output buffer.
-                    vdif_packet_len * num_elements * time_step + // Time step in output frame.
-                    vdif_packet_len * elem + // VDIF pack for the correct element (ThreadID).
-                    vdif_header_len +        // Offset for the vdif header.
-                    tel.to_freq_id(encoded_id,
-                                   freq); // Location in the VDIF packet is just frequency.
-
                 // After all that indexing copy one byte :)
                 out_buf_frame[output_idx] = *(rte_pktmbuf_mtod(mbuf, char*) + from_idx);
```
mhvk, line 311: I think this could be made faster still with pointer arithmetic, but perhaps the compiler optimizes it already. With it, possibly more than 2 inputs could be saved. Anyway, that definitely seems for later. Next step is probably to ensure we can pick any two arbitrary input channels.

lib/dpdk/iceBoardVDIF.hpp
andrerenard: Yeah we refactored the handling of the `tel.to_freq_id()` function after this code was written, so that's likely where this performance regression showed up.  In this mode `encoded_id` doesn't change at all for a given instance of `iceBoardVDIF`, and in fact if it does, I believe other parts of the system will reject those packets.   So you could just cache the output of this function.  There is also no way for this code to run with anything other than the FPGA in `shuffle16` mode, so hard coding the frequency generation, isn't really a problem here.   That might well be worth doing, since this function is called 390625 times/second even after moving it up here.

lib/dpdk/iceBoardVDIF.hpp
andrerenard: The compiler should optimize out `rte_pktmbuf_mtod`, since I believe it's a header level inline function.   If we are getting the performance we need, not obvious it needs to be changed, but might be worth doing if we need to extract more than two inputs, or different patterns of inputs.

andrerenard: The thing we have to be careful about is `mbuf` is linked list, so as we make changes, we have to make sure we are still iterating to the next part of that linked list correctly.

lib/dpdk/iceBoardVDIF.hpp
```diff
@@ -291,7 +291,13 @@ void iceBoardVDIF::copy_packet_vdif(struct rte_mbuf* mbuf) {
...
+        int output_idx_base = (vdif_frame_location * frame_size +           // Frame location in output buffer.
+			       vdif_header_len +        // Offset for the vdif header.
+			       tel.to_freq_id(encoded_id, 0) + // Offset for ARO as in tel.to_freq_id for _num_freq_per_stream = 128
+			       vdif_packet_len * num_elements * time_step); // Time step in output frame.
         for (int freq = 0; freq < 128; ++freq) {
+	    int output_idx = (output_idx_base +
+			      freq * 8); // Location in the VDIF packet is just frequency.
             for (int elem = 0; elem < num_elements; ++elem) {
 
                 // Advance to the next mbuf in the chain.
```
mhvk, line 303: @andrerenard - I don't quite know the structure of these data buffers. Can one be sure that all 16 channels are in one buffer. I.e., can one move this test outside of the element loop? (I recall trying to move it outside of the frequency loop and that seemed not OK, but not really sure any more.)

andrerenard: With the ICEBoards all the ADC channels (elements) are in every buffer.   What changes is the number of elements and number of frequency channels (inversely).  However, this particular code was only every intended to be run with the ICEBoard in `shuffle16` mode.  Any other mode would require more computers, and so couldn't assemble full 1024 frequency channel VDIF packets.   It should be fine to make the assumption that num_elements = 16.

Issue #943, opened 2021-04-14
New C++ Buffers


andrerenard:
Currently the `Buffer` is the oldest part of the framework dating all the way back to the days we used C for everything.  This needs to be updated with a C++ version, and additional features need to be added.  

Given how deeply integrated the current buffers are into the framework, it will require some careful thought to developing a new one.   Either we will need to provide a compatibility layer to the old API, and/or allow for using a mixed configuration of new buffers along with the old ones. 

Some of the key new features we would like to support:

- [ ] Modern easy to use C++17 interface.  e.g. remove the frame_id value and introduce a `next_frame()` style function.
- [ ] Better control over memory allocation type to meet different system needs (GPU/NIC transfer frames with driver registration, NUMA alloc, huge page allocations, external memory device allocations (e.g. for CXL/PCIe 5.0 NICs/GPUs, etc.)
- [ ] Return typed frames using the visFrameView model and move details about sub-frame dimensions to the buffer itself.
- [ ] All statistics features from #942 
- [ ] Different types of queues.  e.g. more than just ring-buffers.
- [ ] Dynamic queue/ring size.  There would still be a fixed number of frames allocated at start-up - since this is an expensive process, but we would allow removing select frames from the ring for slow speed processing. e.g. this could be used for baseband readout buffers, periodic statistics checks on high-speed data products, etc.
- [ ] Customizable back-pressure behavior.  e.g. no-drop, drop oldest, drop low priority, drop newest.  Also replaces the `valve` stage.  
- [ ] Keeping support for all the zero copy and generic buffer manipulation stages

PR #1054, opened 2022-09-19
new stage: BeamAssemble, new metadata registry, and more

mnchord:
There is a problem with get_metadata somewhere, which ends the run with an assertion.

### DIFF
--- a/lib/stages/BeamAssemble.cpp
+++ b/lib/stages/BeamAssemble.cpp
@@ -95,8 +95,8 @@ void BeamAssemble::main_thread() {
             }
         }
 
-        // check if the timestamp corresponds to the past time frame which no longer considered
-        if ( out_frame_map.size() && current_frame_timestamp - arriving_data_timeout < out_frame_map.begin()->first)
+        // check if the timestamp corresponds to the past time frame which is no longer considered/accepted
+        if ( out_frame_map.size() && current_frame_timestamp + arriving_data_timeout < out_frame_map.begin()->first)
         {
             // log the receiving outdated frame frequency
             for (uint32_t f = 0; f < num_freq_per_stream; ++f) {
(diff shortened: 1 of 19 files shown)


lib/stages/BeamAssemble.cpp
```diff
@@ -0,0 +1,173 @@
...
+                out_frame_map.erase(it++);
+            }
+            else {
+                it++;
+            }
+        }
+
+        // determine which output frame to write on based on the timestamp: if this is a new time
+        // stamp, make a new entry in the map and acquire an empty frame from the output buffer
+        if (out_frame_map.find(current_frame_time_stamp) == out_frame_map.end()) {
```
andrerenard: This needs to check if the current timestamp is outside the recv window in the past, otherwise this will get a new output frame for the current input frame - even if it's very old, and then release it on the next input frame.   You should likely also log if you are dropping a frame because it is too old, since ideally we want the tune the system to not drop frames in normal operation.

andrerenard: The logic here also doesn't sort output frames into time order.  I think that's ok, as long as we are ok with doing the sorting in the conversion from raw file to time-transposed HDF5.   But it would be worth thinking about if it makes more sense to just time sort here.

lib/stages/BeamAssemble.cpp
```diff
@@ -0,0 +1,173 @@
...
+            // wait for the first available empty frame
+            out_frame = wait_for_empty_frame(out_buf, unique_name.c_str(), out_frame_id);
+            if (out_frame == nullptr) {
+                // error/warning message: no output frame buffer available
```
andrerenard: No need for a warning here, getting `nullptr` means we are in shutdown mode.

lib/utils/visUtil.hpp
```diff
@@ -261,6 +261,14 @@ inline double ts_to_double(const timespec& ts) {
     return (ts.tv_sec + 1e-9 * ts.tv_nsec);
 }
 
+/**
+ * @brief Converts a timespec type into UNIX time in milliseconds as a uint64_t.
+ * @param  ts Time as timespec.
+ * @return    Time in milliseconds as a uint64_t.
+ */
+constexpr uint64_t timespec_to_milliseconds(const timespec& ts) {
```
andrerenard: It's a bit odd to use milliseconds, our base unit of time in all our measurements is nano seconds.   Generally we timestamp data to resolution of -/+ 10 nanoseconds.   This is dropping a lot of precision and risks introducing rounding errors.

lib/stages/BeamAssemble.cpp
```diff
@@ -0,0 +1,173 @@
...
+        BeamMetadata* metadata = (BeamMetadata*)get_metadata(in_buf, in_frame_id);
+
+        // get the current time stamp in milliseconds from the current input frame metadata
+        current_frame_time_stamp = timespec_to_milliseconds(metadata->gps_time);
+
+        // loop over the open output frames and release the output frames if the timeout has reached
+        for (auto it = out_frame_map.begin(); it != out_frame_map.end();) {
+            // check if the time has passed more than arriving_data_timeout since this output frame
+            // has been opened
+            if (current_frame_time_stamp - it->first > arriving_data_timeout) {
```
andrerenard: I don't think this should use the timestamp from the current input frame.  That value could be older than what's in the map, e.g. a delayed input frame received out of order.  In that case, then `current_frame_time_stamp - it->first` could wrap around since it's an unsigned value, and the newer frame would get expelled.

andrerenard (changes requested):
I haven't done a full review, and I don't see any obvious reason for the metadata assert problem.   However, I think the logic around the map for the active output frames has some issues.  See inline comments.

lib/stages/BeamAssemble.cpp
mnchord: I added another check before those lines to get the out-of-window input frames and print them out.

Well, std::map is a sorted associative container that contains key-value pairs with unique keys, here the time stamp. Keys are sorted by using the default std::less comparison function. When creating the HDF5 files, one can check the metadata for gps_time and check if any sorting required.

lib/stages/BeamAssemble.cpp
mnchord: Yes, you're right, though I saw the same check in other stages. Anyway, I removed "error/warning message:" comment.

lib/utils/visUtil.hpp
mnchord: Well, since the timescale for the arriving data window timeout is in milliseconds, it is reasonable to avoid extra divisions for converting from nano- to milliseconds and having such a helper function.

lib/stages/BeamAssemble.cpp
mnchord: Well, it exactly reflects the need for timeout tweak. Also, there are several approaches to do this: based on the out-of-window input frames gps_time, mean of some input frames gps_time, etc. A simple method could be: getting the window wide enough and only check the latest input frame gps_time, which I did here.

andrerenard: I think the window should be based on the highest time value currently in the map, not the value you just received.  That value could be an old value coming from a delayed node, in that case you would get unexpected behavior, since the frames newer than that would get removed, because it's subtracting a smaller unsigned number by a larger one - resulting in a very large unsigned number. 

I think using this model without sorting in the output frame, the algorithm should look something like:
- Check if the new timestamp is older than the largest value in the map minus the timeout (the other comment).  Drop that frame if it's too old. 
- Add the new value to map if it doesn't exist in the map.
- If there was a change to the map, check if the map has any values smaller/older that the highest time value minus the timeout and remove those entries.

mnchord: I had implemented exactly what you mentioned in my last commit. I fixed something there, and resubmitted. The first value of the map has the largest timestamp.

Again, the map is sorted, meaning that the frame with the lowest timestamp will be marked as full first (to be used in the next stage) and so on. Therefore, the frames will be release in ascending order in time and the next stage is responsible to process them in the same order to preserve it, if any ordering required there.

andrerenard: The buffers act as rings and are processed in frame ID order, not the order in which they are released.    Releasing the frames in this way would sort if the buffer was a FIFO queue, but not with the current design.  Having said that, I think we can leave the reordering to the transpose/conversion from raw files to HDF5. 

However, I still think there is a problem on this line.  There is no guarantee that current_frame_time_stamp will be the latest timestamp.    If `current_frame_time_stamp < it->first` (e.g. if the timestamp value we just got is less than any value currently in the map), then the LHS of the inequality in the `if (current_frame_time_stamp - it->first > arriving_data_timeout)` statement will become a very large number (since it is unsigned), that will be greater than the timeout, and then inadvertently remove newer entries in the map.

mnchord: Well, adding another feature to the ring buffer is easier and more rational than carrying this sorting need along to every stage or any postprocessing down the road.

That's a very good point; and the solution to simply accompany another condition `(current_frame_timestamp > it->first)` with boolean `&&` operator or removing the unsigned from those types. I fixed and submitted.

andrerenard: We can talk about it more in the meeting today, but changing the ring buffer is something we have to be very careful about doing.  We currently have many telescopes using it with dozens of stages.  Making changes there  requires substantial testing at this point.  It would also be sunk cost, since my current thinking is that we will add a new C++ version of the buffer which supports different operating modes (rings, FIFO, priority queues, and more), and then slowly replace the use of the older version in the existing systems, see #943  That is a project I want to do, but it requires a careful design proposal, a lot of modification to the core framework components, and likely several months of work.  From a project priority point of view, we should not start building a new buffer until we have at least completed the gated beams project and the core Outriggers functionality.

PR #1055, opened 2022-09-22
 new stage: BeamAssemble, new metadata registry, and more

mnchord:
It seems working:

```
/raw_capture: Data file write done for /drives/CHF/7/20220922T111406Z_chime_assemble_test_vdif/0000000063.vdif, lost_packets -975031214
/raw_capture: Data file write done for /drives/CHF/0/20220922T111406Z_chime_assemble_test_vdif/0000000064.vdif, lost_packets -975031214
/raw_capture: Data file write done for /drives/CHF/1/20220922T111406Z_chime_assemble_test_vdif/0000000065.vdif, lost_packets -975031214
/raw_capture: Data file write done for /drives/CHF/2/20220922T111406Z_chime_assemble_test_vdif/0000000066.vdif, lost_packets -975031214
^CGot SIGINT, shutting down kotekan...
Attempting to stop and join kotekan_stages...
Sending shutdown signal to buffer: asm_beam_recv_buffer
Sending shutdown signal to buffer: beam_recv_buffer
Joining kotekan_stage: /beam_assemble...
/beam_assemble: Beam Assembly statistics for 1008ms timeout window: 
    number of incoming frames: 57762
    number of accepted frames: 57762
    number of missed frames: 0

Joining kotekan_stage: /raw_capture...
Joining kotekan_stage: /recv...
kotekan shutdown with status: CLEAN_EXIT
```
:


lib/metadata/FrequencyAssembledMetadata.hpp
```diff
@@ -0,0 +1,65 @@
...
+ * @par lower_band_frequency, higher_band_frequency
+ *      The frequency id range: [low to high]
+ *
+ * @author Mehdi Najafi
+ */
+struct FrequencyBin {
+    /// The lower frequency range
+    uint32_t lower_band_frequency  = std::numeric_limits<uint32_t>::max();
+    /// The higher frequency range
+    uint32_t higher_band_frequency = std::numeric_limits<uint32_t>::min();
```
andrerenard: Because the metadata structs on an overlay on memory, and the full struct gets zeroed between use, these numeric_limits values won't actually be set when you get a new metadata object in the pipeline.

Also I'm not sure how useful the lower/upper values will be.  The way the correlator works, most X-engine nodes get frequency bins spread out between the lower and upper half of the band.  e.g. one node might have bins, 1, 250, 500, and 700.  So having the edges of the band doesn't say much.  The more useful value to store is the `num_frequencies` value below, because that gives how much of the band was captured.   I think we could just remove these.

lib/metadata/FrequencyAssembledMetadata.hpp
```diff
@@ -0,0 +1,65 @@
...
+    /// The number of frequencies
+    uint32_t num_frequencies;
```
andrerenard: We should likely change the name here to something like `num_received_frequencies` or `num_valid_frequencies`.  The variable name `num_frequencies` is commonly used in the code to refer to the total number of frequencies in the system - which in the chime case is always 1024 for example.

lib/metadata/assembledChimeMetadata.hpp
```diff
@@ -0,0 +1,18 @@
...
+* @author Mehdi Najafi
+* @date   28 AUG 2022
+*****************************************************/
+
+#include "chimeMetadata.hpp"
+#include "FrequencyAssembledMetadata.hpp"
+typedef FrequencyAssembledMetadata<chimeMetadata> assembledChimeMetadata;
+
+// register this new metadata
+REGISTER_KOTEKAN_METADATA(assembledChimeMetadata)
```
andrerenard: We don't need an assembled chime metadata object, I'd prefer not to create new classes which aren't used.   If we need it down the road we can add it.

lib/metadata/CMakeLists.txt
```diff
@@ -2,5 +2,13 @@
 # symbols are ever added it will need to be turned back into a standard CMake library
 project(kotekan_metadata)
 
+find_package(Python COMPONENTS Interpreter REQUIRED)
+execute_process(COMMAND ${Python_EXECUTABLE} enum_metadatas.py
```
andrerenard: I think this is adding more complexity than we need here, and it appears to be breaking the CI checks in various ways.   I don't think it's all that big a problem if we get repeated registrations of the same type at startup, we can just remove the warnings around that for metadata.  I'd prefer that to the added complexity of needing python scripts and auto generated files to make everything work.

lib/stages/BeamAssemble.cpp
```diff
@@ -0,0 +1,207 @@
...
+        // get the current timestamp in milliseconds from the current input frame metadata
+        current_frame_timestamp = timespec_to_milliseconds(metadata->gps_time);
+
+        // loop over the open output frames and release the output frames if the timeout has reached
+        for (auto it = out_frame_map.begin(); it != out_frame_map.end();) {
+            // check if the time has passed more than arriving_data_timeout since this output frame
+            // has been opened
+            if (current_frame_timestamp - it->first > arriving_data_timeout) {
+                // release the output frame in the map
+                mark_frame_full(out_buf, unique_name.c_str(), it->second.second);
```
andrerenard: It would be interesting for debug to print out the number of frequencies in the frame at the time we release it.   I assume that will be fairly stable, based on how many X-engine nodes are on line, but it would be nice to confirm how many frequencies we are putting into the output frame.

lib/stages/BeamAssemble.cpp
```diff
@@ -0,0 +1,207 @@
...
+        received_beam_frames_count++;
+
+        // retrieve the metadata and create a new metadata based on it
+        BeamMetadata* metadata = (BeamMetadata*)get_metadata(in_buf, in_frame_id);
+
+        // get the current timestamp in milliseconds from the current input frame metadata
+        current_frame_timestamp = timespec_to_milliseconds(metadata->gps_time);
+
+        // loop over the open output frames and release the output frames if the timeout has reached
+        for (auto it = out_frame_map.begin(); it != out_frame_map.end();) {
```
andrerenard: I think you only need to run this loop if you are adding a new entry to the map, e.g. in the `if` condition at line 127.

lib/stages/BeamAssemble.cpp
```diff
@@ -0,0 +1,207 @@
...
+                auto freq_id = Telescope::instance().to_freq_id(metadata->stream_id, f);
+                if (late_beam_printout) {
+                    // print info about the late beam
+                    INFO( "LATE Beam RA: {:f}, Dec: {:f}, scaling: {:d}, freq: {:d}, first value: {:d}+{:d}i\n",
+                            metadata->ra, metadata->dec, metadata->scaling, freq_id, 
+                            in_frame[0] & 0x0F, (in_frame[0] & 0xF0) >> 4 );
+                }
+            }
+        }
+        else {
```
andrerenard: There is some style is inconsistency with the bracket placement around for/if/else statements.  You can just run `make clang-format` in the build directory to automatically apply the `clang-format` rules, which are also checked by the CI build system.

lib/stages/BeamAssemble.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+                // determine the offset based on the frequency id
+                auto out_frame_offset = freq_id * (out_buf->frame_size - num_freq_per_output_frame) / num_freq_per_output_frame;
+                // is that the same as: freq_id * in_frame_freq_size ?
+                // assert(out_frame_offset == freq_id * in_frame_freq_size);
+
+                // copy the data from in_frame to the out_frame + some offset with no change
+                // Note that the first num_freq_per_output_frame is the frequency id indicator,
+                // which is 0 or 1 indicating the frequency data copied to the output
+                memcpy(out_frame + num_freq_per_output_frame + out_frame_offset, in_frame, in_frame_freq_size);
```
andrerenard: For `num_freq_per_stream > 1` the in_frame layout does not follow this pattern.  The pattern for multi-frequency beams is (time, frequency, pol), where each byte is still a complex 4+4-bit [Re,Im] offset encoded pair.  And `size(pol) = 2`, `size(frequency) = num_freq_per_stream`, and `size(time) = samples_per_data_set`.

lib/stages/BeamAssemble.hpp
```diff
@@ -0,0 +1,83 @@
...
+ * @par received_beam_frames_count
+ *      a counter to keep track of the number of beam frames received
+ * 
+ * @par accepted_beam_frames_count
+ *      a counter to keep track of the number of beam frames received
+ *      and copied to the output buffer frames
+ *
+ * @par missed_beam_frames_count
+ *      a counter to keep track of the number of beam frames received
+ *      late and ignored (not copied to any output buffer frames)
```
andrerenard: The config options should use the `@conf` custom doxygen tag.  The idea being we should be able to generate nice autodocs off these headers once we get time to work on that.

andrerenard (changes requested):
I've gone over the code in a bit more detail.  I think the core algorithm is looking good now, I don't see any logic errors at this point (aside from some multi-frequency indexing).  There are a few smaller issues in the inline comments.

lib/stages/BeamAssemble.cpp
andrerenard: To be clear, `(time, frequency, pol)` is slowest to fastest varying index.

andrerenard:
The assembler also needs to check for the case when the tracking beam pointing (Ra/Dec coordinates) change. Because the nodes might not all get updates at the same time, it is possible for the same time step to have different `ra`, `dec`, and `scaling` values.  We will not want to mix pointings into the output frames, since doing so safely would require keeping per-frequency metadata.  

There are two main ways to address this:
1. Create more than one output frame.  e.g. one frame per set of Ra/Dec/Scaling.  
2. When adding a new output frame, record the Ra/Dec/Scaling values, then check each incoming frame with the same time stamp against that Ra/Dec/Scaling, if the values don't match, discard the incoming frame. 

The issue with 1. is it will momentarily double the data rate to disk, which can only be sustained for very short periods of time.   The optimal solution there is to create an output frame for every Ra/Dec/Scaling value, but then only write to disk the frame with the highest number of received frequencies.  However, that's a somewhat tricky thing to implement in a streaming pipeline.   Option 2. is far easier to implement, and in practice should cover the vast majority of cases.  I think at this stage we should just implement option 2, but keep track of how many frames we are rejecting on pointing changes.

lib/stages/BeamAssemble.cpp
mnchord: I added that. The output looks like this:
```
/raw_capture: Data file write done for /drives/CHF/7/20220928T214023Z_chime_assemble_test_vdif/0000000335.vdif, lost_packets 1400382879
/beam_assemble: Released output frame RA: 83.975502, Dec: 22.028116, scaling: 128, freq bin: [1,1023], num_freqs: 804

/raw_capture: Data file write done for /drives/CHF/0/20220928T214023Z_chime_assemble_test_vdif/0000000336.vdif, lost_packets 1400382879
/beam_assemble: Released output frame RA: 83.975502, Dec: 22.028116, scaling: 128, freq bin: [1,1023], num_freqs: 804

/raw_capture: Data file write done for /drives/CHF/1/20220928T214023Z_chime_assemble_test_vdif/0000000337.vdif, lost_packets 1400382879
/beam_assemble: Released output frame RA: 83.975502, Dec: 22.028116, scaling: 128, freq bin: [1,1023], num_freqs: 804

/raw_capture: Data file write done for /drives/CHF/2/20220928T214023Z_chime_assemble_test_vdif/0000000338.vdif, lost_packets 1400382879
/beam_assemble: Released output frame RA: 83.975502, Dec: 22.028116, scaling: 128, freq bin: [1,1023], num_freqs: 804

/raw_capture: Data file write done for /drives/CHF/3/20220928T214023Z_chime_assemble_test_vdif/0000000339.vdif, lost_packets 1400382879
/beam_assemble: Released output frame RA: 83.975502, Dec: 22.028116, scaling: 128, freq bin: [1,1023], num_freqs: 804

```

lib/metadata/CMakeLists.txt
mnchord: Don't worry; Removed.

lib/metadata/assembledChimeMetadata.hpp
mnchord: Removed.

lib/metadata/FrequencyAssembledMetadata.hpp
mnchord: Very good point; I changed it.

lib/stages/BeamAssemble.cpp
mnchord: The loop releases the outdated frames, also.

lib/stages/BeamAssemble.cpp
mnchord: Fixed.

lib/stages/BeamAssemble.cpp
mnchord: Thanks, I fixed it.

lib/stages/BeamAssemble.hpp
mnchord: Fixed.

lib/metadata/FrequencyAssembledMetadata.hpp
mnchord: Thanks, I fixed that. The lower/upper is just for a quick check; or debugging purposes like this:
The new result says: `freq bin: [1,1023], num_freqs: 804` does it mean it did not receive `220` freq ids?

Also, it says when I pressed Ctrl+C:
```
/beam_assemble: Released output frame RA: 83.975502, Dec: 22.028116, scaling: 128, freq bin: [1,1022], num_freqs: 264

/beam_assemble: Released output frame RA: 83.975502, Dec: 22.028116, scaling: 128, freq bin: [1,1023], num_freqs: 652

/beam_assemble: Released output frame RA: 83.975502, Dec: 22.028116, scaling: 128, freq bin: [1,1023], num_freqs: 794

/beam_assemble: Released output frame RA: 83.975502, Dec: 22.028116, scaling: 128, freq bin: [1,1023], num_freqs: 804

/beam_assemble: Released output frame RA: 83.975502, Dec: 22.028116, scaling: 128, freq bin: [1,1023], num_freqs: 804

```

lib/stages/BeamAssemble.cpp
andrerenard: Right, but you'll only have outdated frames if you've added a new entry to the map.  This loop doesn't use the system time, it uses the latest time it received.  So it only changes which entries are old, when new entries are added.

lib/stages/BeamAssemble.hpp
```diff
@@ -0,0 +1,86 @@
...
+ * @par received_beam_frames_count
+ *      a counter to keep track of the number of beam frames received
+ *
+ * @par accepted_beam_frames_count
+ *      a counter to keep track of the number of beam frames received
+ *      and copied to the output buffer frames
+ *
+ * @par missed_beam_frames_count
+ *      a counter to keep track of the number of beam frames received
+ *      late and ignored (not copied to any output buffer frames)
```
andrerenard, line 61: Since these are private members, and not exposed via the metrics API, I would just put these comments over the variables themselves in the class below.  That also will annotate them directly in Doxygen.

lib/metadata/FrequencyAssembledMetadata.hpp
```diff
@@ -0,0 +1,73 @@
...
+ *      The frequency id range: [low to high]
+ *
+ * @author Mehdi Najafi
+ */
+struct FrequencyBin {
+    /// The lower frequency range
+    uint32_t lower_band_received_frequency = std::numeric_limits<uint32_t>::max();
+    /// The upper frequency range
+    uint32_t upper_band_received_frequency = std::numeric_limits<uint32_t>::min();
+    /// The number of frequency ids in the frame attached to this metadata
```
andrerenard, line 38:
```suggestion
    /// The lower frequency range
    uint32_t lower_band_received_frequency;
    /// The upper frequency range
    uint32_t upper_band_received_frequency;
    /// The number of frequency ids in the frame attached to this metadata
```
Since this code structure is never directly instantiated, there isn't much point in setting the values in the header, and it might just be confusing.  Better to just add a commend above that this structure should be initialized with the function below before use.

andrerenard (changes requested):
This is looking good, and I like the simplified metadata factory.  Just a few small inline comments.  I'll do another full review once we get the new file writer and/or the Ra/Dec/Scaling logic changes.

lib/metadata/FrequencyAssembledMetadata.hpp
mnchord: Thanks; done.

lib/stages/BeamAssemble.hpp
mnchord: Lets' talk about that loop; I think it should be moved to the end of the thread loop.


I added the comments above their definition. Shall I keep the above comments as well in the class header notes/comments?

andrerenard: I'd remove the comments them from the class header.

lib/stages/nDiskMultiFormatWriter.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+std::string nDiskMultiFormatWriter::get_file_extension() {
+    // go through a nested if-else to find the requested file_format
+    if (file_format == "raw") {
+        return "raw";
+    } else {
+        if (file_format == "HDF5") {
+            return "h5";
+        }
+    }
```
andrerenard: We are not using this to output hdf5, so I don't think we should include code here unless we actually support that mode, let's just remove the hdf5 bit for now. 

Separately, we should still have backwards compatibility with the other nDiskFileWrite code, so we can remove that older version, that means we should support the `.vdif` file extension.

lib/stages/nDiskMultiFormatWriter.cpp
```diff
@@ -0,0 +1,392 @@
...
+    // retrieve the path and file naming convention, like /drives/CHF/{n}/{DATE}T{TIME}Z_asm
+    disk_path = config.get<std::string>(unique_name, "disk_path");
+    // retrieve the requested output file format
+    file_format = config.get<std::string>(unique_name, "file_format");
+    // retrieve the expected number of frequencies in the output
+    num_freq_per_output_frame = config.get<uint32_t>(unique_name, "num_freqs");
+    // retrieve the maximum number of frames to be written to each file
+    max_frames_per_file = config.get<int64_t>(unique_name, "max_frames_per_file");
+    // write to disk boolean as a on/off switch for any real disk activity
+    write_to_disk = config.get<bool>(unique_name, "write_to_disk");
```
andrerenard: I think we should remove this `write_to_disk` option.  It was originally there for some ARO modes were we wanted to run the pipeline and get the logs from the file writer, but not write files.  The better approach would be to create a separate stage that outputs the metadata for debugging, and just remove this stage from the config if we don't want to have file writing.

lib/stages/nDiskMultiFormatWriter.cpp
```diff
@@ -0,0 +1,392 @@
...
+    if (file_format == "raw") {
+        return "raw";
+    } else {
+        if (file_format == "HDF5") {
+            return "h5";
+        }
+    }
+
+    ERROR("Unknown requested file format: {:s}\n", file_format);
+    exit(-1);
```
andrerenard: I think we already talked about this, but replace this with the `FATAL_ERROR()` macro.

lib/stages/nDiskMultiFormatWriter.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+std::string nDiskMultiFormatWriter::get_dataset_folder_name(const int n) {
+    std::string path(disk_path);
+    string_replace_all(path, "{DATE}", fmt::format("{:%Y%m%d}", invocation_time));
+    string_replace_all(path, "{TIME}", fmt::format("{:%H%M%S}", invocation_time));
+    string_replace_all(path, "{n}", fmt::format("{}", n));
+    return path;
+}
+
+void nDiskMultiFormatWriter::save_metadata() {
```
andrerenard: We can remove this entire function.

lib/stages/nDiskMultiFormatWriter.cpp
```diff
@@ -0,0 +1,392 @@
...
+    // retrieve the requested output file format
+    file_format = config.get<std::string>(unique_name, "file_format");
+    // retrieve the expected number of frequencies in the output
+    num_freq_per_output_frame = config.get<uint32_t>(unique_name, "num_freqs");
+    // retrieve the maximum number of frames to be written to each file
+    max_frames_per_file = config.get<int64_t>(unique_name, "max_frames_per_file");
+    // write to disk boolean as a on/off switch for any real disk activity
+    write_to_disk = config.get<bool>(unique_name, "write_to_disk");
+    // write to disk boolean as a on/off switch for any real disk activity
+    write_metadata_and_gains =
```
andrerenard: We can remove this option.

lib/stages/nDiskMultiFormatWriter.cpp
```diff
@@ -0,0 +1,392 @@
...
+    for (uint32_t i = 0; i < num_disks; i++) {
+        std::string folder = get_dataset_folder_name(i);
+        int err = mkdir(folder.c_str(), 0777);
+
+        if (err != -1) {
+            continue;
+        }
+
+        if (errno == EEXIST) {
+            printf("The folder: %s, already exists.\nPlease delete the data set, or use another "
```
andrerenard: This should be converted to the modern `ERROR()` macro.  The `\n` tags won't be needed once it's converted.

lib/stages/nDiskMultiFormatWriter.cpp
```diff
@@ -0,0 +1,392 @@
...
+                   "name.\n",
+                   folder.c_str());
+        } else {
+            perror("Error creating data set directory.\n");
+            printf("The directory was: %s \n", folder.c_str());
```
andrerenard: Convert to the modern `ERROR()` macro.

lib/stages/nDiskMultiFormatWriter.cpp
```diff
@@ -0,0 +1,392 @@
...
+        }
+        exit(errno);
```
andrerenard: use `exit_kotekan()` to trigger clean shutdown.

lib/stages/nDiskMultiFormatWriter.cpp
```diff
@@ -0,0 +1,392 @@
...
+                    ERROR("Could not copy {:s} to {:s}\n", gain_files[j], dest);
+                    exit(-1);
+                } else {
+                    INFO("Copied gains from {:s} to {:s}\n", gain_files[j], dest);
+                }
+            }
+        }
+        // save settings
+        save_metadata();
+    }
```
andrerenard: We can also remove all of this.

lib/stages/nDiskMultiFormatWriter.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+    // thread infinite loop
+    while (!stop_thread) {
+
+        // This call is blocking.
+        frame = wait_for_full_frame(buf, unique_name.c_str(), frame_id);
+        if (frame == nullptr)
+            break;
+
+        // INFO("Got buffer id: {:d}, disk id {:d}", frame_id, disk_id);
```
andrerenard: We generally shouldn't leave commented code like this.  Ether set this to `DEBUG` or `DEBUG2` or remove it entirely.

lib/stages/nDiskMultiFormatWriter.cpp
```diff
@@ -0,0 +1,392 @@
...
+                // write the number of frames in this file
+                bytes_written = write(fd, &num_frames, sizeof(num_frames));
+                // reset the num_frames
+                num_frames = 0;
+            }
+
+            if (fd == -1) {
+                ERROR("Cannot open file");
+                ERROR("File name was: {:s}", file_name);
+                exit(errno);
```
andrerenard: Update exit handling here and below.

lib/stages/nDiskMultiFormatWriter.cpp
```diff
@@ -0,0 +1,392 @@
...
+            if (mc != nullptr) {
+                if (write(fd, mc->metadata, mc->metadata_size) != (int32_t)mc->metadata_size) {
+                    ERROR("Failed to write metadata_size to disk for file {:s}", file_name);
+                    exit(-1);
+                }
+            }
+
+            // write the number of frequencies
+            bytes_written =
+                write(fd, &num_freq_per_output_frame, sizeof(num_freq_per_output_frame));
```
andrerenard: I think this `num_freq_per_output_frame` should be written into the output frame in the previous step (or put into the metadata).   If you do that, then this code just writes the frame size and frame without needing to know the type of data it is writing, which makes it a lot more flexible.

lib/stages/nDiskMultiFormatWriter.cpp
```diff
@@ -0,0 +1,392 @@
...
+            bytes_written = write(fd, &frame_byte_size, sizeof(size_t));
+
+            // write the frame data
+            bytes_written = write(fd, frame, buf->frame_size);
+
+            if (bytes_written != buf->frame_size) {
+                ERROR("Failed to write buffer to disk!!!  Abort, Panic, etc.");
+                exit(-1);
+            } else {
+                // INFO("Data written to file!");
```
andrerenard: Remove or turn into a `DEBUG2()` type statement.

lib/stages/nDiskMultiFormatWriter.cpp
```diff
@@ -0,0 +1,392 @@
...
+                }
+            }
+
+            // write the number of frequencies
+            bytes_written =
+                write(fd, &num_freq_per_output_frame, sizeof(num_freq_per_output_frame));
+
+            // write the frame size in bytes
+            size_t frame_byte_size = (size_t)buf->frame_size;
+            bytes_written = write(fd, &frame_byte_size, sizeof(size_t));
```
andrerenard: Check if the write call was successful.

lib/stages/nDiskMultiFormatWriter.hpp
```diff
@@ -0,0 +1,163 @@
...
+ * @class nDiskMultiFormatWriter
+ * @brief Consumer ``kotekan::Stage`` which writes input data on multiple drives in various formats.
+ *
+ * This is a consumer which initiates n threads to write to ``n`` disks. Each drive will receive
+ * data from every ``n``th buffer, stored within a common-named subfolder. Within each folder
+ * the data files will be numbered incrementally across the disks.
+ *
+ * @par Buffers
+ *  @buffer in_buf The kotkean buffer with the data to be written
+ *  @buffer_metadata metadata The kotkean metadata to be written
```
andrerenard: These comments should indicate the type of frame/metadata the stage expects, in this case it's a `standard` buffer with `any` metadata.

andrerenard (changes requested):
A few comments on the `nDiskMultiFormatWriter`.  I think most of this we already covered on the call, but there are a few other small things to adjust.

lib/stages/nDiskMultiFormatWriter.cpp
mnchord: Fixed with another wrapper for all write calls.

lib/stages/nDiskMultiFormatWriter.cpp
mnchord: It seems unnecessary, so I removed it.

lib/stages/nDiskMultiFormatWriter.cpp
mnchord: I fixed it, please check the new one.

lib/stages/nDiskMultiFormatWriter.cpp
mnchord: I removed it.

lib/stages/nDiskMultiFormatWriter.cpp
mnchord: I removed it completely.

lib/stages/nDiskMultiFormatWriter.cpp
mnchord: Yep, I replaced it with FATAL_ERROR.

lib/stages/nDiskMultiFormatWriter.cpp
mnchord: I was thinking it may be useful to keep like previous writer. I removed it.

lib/stages/nDiskMultiFormatWriter.cpp
mnchord: Oh, yes! I put "vdif" there,

lib/stages/nDiskMultiFormatWriter.hpp
mnchord: Yes, agree. I removed the second one which was about the metadata.

lib/stages/nDiskMultiFormatWriter.cpp
mnchord: It is define as an input config parameter for BeamAssemble stage. If I add it as the leading uint64_t in each frame, we need to change the buffer size by adding 4 as well, right?

lib/stages/nDiskMultiFormatWriter.cpp
mnchord: Yes, I replaced them with a better msg macros.

lib/stages/nDiskMultiFormatWriter.cpp
mnchord: Yes, agree; I replaced them with FATAL_ERROR instead, which would exit the kotekan safely.

lib/stages/nDiskMultiFormatWriter.cpp
mnchord: Fixed.

lib/stages/nDiskMultiFormatWriter.cpp
mnchord: Removed them all.

lib/stages/nDiskMultiFormatWriter.cpp
mnchord: Yes, I agree, I used DEBUG.

PR #1060, opened 2022-10-17
Set GPU device clocks from config file

ShrutiBhatporia:
Set user-defined clocks to graphics and memory. Function `cudaDeviceInterface::set_device_clocks() `set the core and memory clock values from the the config file `config/examples/basic_pipeline.yaml`. The patch also lists available clock configurations for the available device.


config/examples/basic_pipeline.yaml
```diff
@@ -2,6 +2,8 @@
 type: config
 log_level: DEBUG
 cpu_affinity: [0,1,2,3]
+gpu_core_clock: 2000   #in MHz
+gpu_mem_clock: 2000    #in MHz
```
andrerenard: We should move this to `config/tests/verify_cuda_n2_astron.yaml`, since this config doesn't contain a `cudaProcess` and so won't use these variables. Put them in the `gpu_0:` block of that config. 

As for values, I think we should use `1005` as the core clock, and `7251` as the memory clock.   The A40 we are running this on only supports `7251` as the memory clock.

lib/cuda/cudaDeviceInterface.cpp
```diff
@@ -24,6 +42,22 @@ cudaDeviceInterface::~cudaDeviceInterface() {
...
+
+    // Set default clocks to zero
+    uint32_t gpu_mem_clock = config.get_default<uint32_t>(unique_name, "gpu_core_clock", 1000);
+    uint32_t gpu_core_clock = config.get_default<uint32_t>(unique_name, "gpu_core_clock", 1000);
+
+    // Get and update the GPU clocks from the config file
+    gpu_mem_clock = config.get<uint32_t>(unique_name, "gpu_mem_clock");
+    gpu_core_clock = config.get<uint32_t>(unique_name, "gpu_core_clock");
+
+    CHECK_CUDA_ERROR(nvmlDeviceSetApplicationsClocks (gpu_id_, gpu_mem_clock, gpu_core_clock))
```
andrerenard: From what I've read on the API, the GPU will only accept clocks exactly which match options in the list given by the `nvmlDeviceGetSupportedMemoryClocks` and `nvmlDeviceGetSupportedGraphicsClocks` functions.   If the clocks in the config don't match one of the options from the list, then the most robust logic would be to select the closest valid clock and use that.  With a log message to tell the user the actual clock used.

lib/cuda/cudaDeviceInterface.cpp
```diff
@@ -24,6 +42,22 @@ cudaDeviceInterface::~cudaDeviceInterface() {
     cleanup_memory();
 }
 
+void cudaDeviceInterface::set_device_clocks(const string& unique_name, int32_t gpu_id_) {
+
+    // Set default clocks to zero
+    uint32_t gpu_mem_clock = config.get_default<uint32_t>(unique_name, "gpu_core_clock", 1000);
+    uint32_t gpu_core_clock = config.get_default<uint32_t>(unique_name, "gpu_core_clock", 1000);
```
andrerenard: I would set the default to `0`, and then check below these lines if the value is zero or not.   If one or both is, then don't set the clocks, otherwise apply the clocks.   That way if someone doesn't provide a set of clocks in the config, it won't try to set anything and just use the GPU's default.

lib/cuda/cudaDeviceInterface.hpp
```diff
@@ -53,6 +53,13 @@ class cudaDeviceInterface final : public gpuDeviceInterface {
     void async_copy_host_to_gpu(void* dst, void* src, size_t len, cudaEvent_t pre_event,
                                 cudaEvent_t& copy_pre_event, cudaEvent_t& copy_post_event);
 
+    /**
+    * @brief Print avialable colcks for given GPU and set sevice clock
+    * @param unique_name Base path from CudaProcess
+    * @param gpu_id_  Device ID
+    */
+    void cudaDeviceInterface::set_device_clocks(const string& unique_name, int32_t gpu_id_);
```
andrerenard, line 62: Since this is member function you can use the `gpu_id` variable from the base class directly and don't need to pass it into this function. 

Similarly I think we should make `unique_name` a member variable of this class, and then we won't need to pass that around either.

andrerenard (changes requested):
This looks good, but we'll need a few changes, see the inline comments.   

For the CI failure, you just need to run `make clang-format` from the `build/` directory, that should fix any formatting issues, and then you can push the changes.

config/examples/basic_pipeline.yaml
ShrutiBhatporia: Shifted the definition of `gpu_core_clock` and `gpu_mem_clock` to the given header file.

lib/cuda/cudaDeviceInterface.cpp
ShrutiBhatporia: Values from `nvmlDeviceGetSupportedMemoryClocks` and `nvmlDeviceGetSupportedGraphicsClocks` are given to `nvmlDeviceSetApplicationsClocks` now in order to keep the values restricted allowed GPU clock values.

lib/cuda/cudaDeviceInterface.cpp
ShrutiBhatporia: Set the clock values to `0` and applied the conditions for keeping the values zero if the clock values from the config files are not matching the allowed GPU clocks.

lib/cuda/cudaDeviceInterface.hpp
ShrutiBhatporia: Done

lib/cuda/cudaDeviceInterface.cpp
```diff
@@ -24,6 +44,47 @@ cudaDeviceInterface::~cudaDeviceInterface() {
...
+    // Set default clocks to zero
+    uint32_t gpu_mem_clock = config.get_default<uint32_t>(unique_name, "gpu_mem_clock", 0);
+    uint32_t gpu_core_clock = config.get_default<uint32_t>(unique_name, "gpu_core_clock", 0);
+
+    uint32_t get_gpu_mem_clock, get_gpu_core_clock;
+
+    nvmlDeviceGetMaxClockInfo (gpu_id_, gpu_mem_clock, get_gpu_mem_clock);
+    nvmlDeviceGetMaxClockInfo (gpu_id_, gpu_core_clock, get_gpu_core_clock);
+
+    if (get_gpu_mem_clock != 0 & get_gpu_core_clock != 0) {
```
andrerenard:
```suggestion
    if (gpu_mem_clock != 0 && gpu_core_clock != 0) {
```
Use logical `&&` not bitwise `&` here.   Also I think you should be checking the clocks from the config, not the max clocks from above (which I don't think are needed).

lib/cuda/cudaDeviceInterface.cpp
```diff
@@ -24,6 +44,47 @@ cudaDeviceInterface::~cudaDeviceInterface() {
...
+
+        // Get and update the GPU clocks from the config file
+        gpu_mem_clock = config.get<uint32_t>(unique_name, "gpu_mem_clock");
+        gpu_core_clock = config.get<uint32_t>(unique_name, "gpu_core_clock");
```
andrerenard: The config file won't change while this function is being called, so I don't think it makes sense to call the function again.   To make sure both clocks are set if one is set, use if statements and throw an `std::runtime_error` if only one value is set.

lib/cuda/cudaDeviceInterface.cpp
```diff
@@ -24,6 +44,47 @@ cudaDeviceInterface::~cudaDeviceInterface() {
...
+
+        for (int i = 0; i < mem_count; ++i)  {
+             mem_clock[i] == get_gpu_mem_clock;
+             break;
+        }
+
+        if (i == mem_count)
+            gpu_mem_clock = config.get_default<uint32_t>(unique_name, "gpu_mem_clock", 0)
```
andrerenard: This function won't return anything different if the clocks don't match one of the clocks in the list.

The optimal thing to do here is find the closest clock (rounding up) from the supported clocks list and use that one.   e.g. if the allowed clocks are 900, 1000, and 1200 and someone sets the clock to 1100, then round up to 1200.  Or if the user sets 1300, then round down to the highest clock option - again 1200 in this example.

lib/cuda/cudaDeviceInterface.cpp
```diff
@@ -15,6 +17,24 @@ cudaDeviceInterface::cudaDeviceInterface(Config& config_, int32_t gpu_id_, int g
...
+        INFO("{:d}  ", mem_clock[i]);
+    }
+
+    INFO("Allowed GPU graphics clocks(MHz): ");
+    for (int i = 0; i < core_count; ++i)  {
+        INFO("{:d}  ", core_clock[i]);
+    }
+
+    CHECK_CUDA_ERROR(set_device_clocks(mem_clock, core_clock,
+                                       mem_count, core_count));
```
andrerenard:
```suggestion
    set_device_clocks(mem_clock, core_clock, mem_count, core_count));
```
The set clocks function is `void`, so it won't return anything for `CHECK_CUDA_ERROR` to check.

lib/cuda/cudaDeviceInterface.cpp
```diff
@@ -24,6 +44,47 @@ cudaDeviceInterface::~cudaDeviceInterface() {
...
+                                            unsigned int mem_count, unsigned int core_count) {
+
+    // Set default clocks to zero
+    uint32_t gpu_mem_clock = config.get_default<uint32_t>(unique_name, "gpu_mem_clock", 0);
+    uint32_t gpu_core_clock = config.get_default<uint32_t>(unique_name, "gpu_core_clock", 0);
+
+    uint32_t get_gpu_mem_clock, get_gpu_core_clock;
+
+    nvmlDeviceGetMaxClockInfo (gpu_id_, gpu_mem_clock, get_gpu_mem_clock);
+    nvmlDeviceGetMaxClockInfo (gpu_id_, gpu_core_clock, get_gpu_core_clock);
```
andrerenard, line 56: I don't think these are being called correctly.  The second argument should be `nvmlClockType_t`, e.g. `NVML_CLOCK_MEM`.   However, I don't think you need to call these at all.  Since the list of clocks from `nvmlDeviceGetSupported*Clocks` will have the max clock it it.

lib/cuda/cudaDeviceInterface.cpp
```diff
@@ -15,6 +17,24 @@ cudaDeviceInterface::cudaDeviceInterface(Config& config_, int32_t gpu_id_, int g
     INFO("Number of CUDA GPUs: {:d}", max_num_gpus);
 
     cudaSetDevice(gpu_id);
+
+    // Find out how many GPUs clocks are allowed.
+    unsigned int mem_clock, core_clock, mem_count, core_count;
```
andrerenard: `mem_clock` and `core_clock` should be array pointers, otherwise the code below won't compile.

lib/cuda/cudaDeviceInterface.cpp
```diff
@@ -24,6 +44,47 @@ cudaDeviceInterface::~cudaDeviceInterface() {
     cleanup_memory();
 }
 
+void cudaDeviceInterface::set_device_clocks(unsigned int mem_clock, unsigned int core_clock,
+                                            unsigned int mem_count, unsigned int core_count) {
```
andrerenard: Here `mem_clock` and `core_clock` also need to be array pointers.

andrerenard (changes requested):
Thanks for the updates.  I've left a few inline comments, I think a few things need to be adjusted.

lib/cuda/cudaDeviceInterface.cpp
ShrutiBhatporia: Included the change.

lib/cuda/cudaDeviceInterface.cpp
ShrutiBhatporia: mem_clock and core_clock are pointers now, will make other changes accordingly.

lib/cuda/cudaDeviceInterface.cpp
ShrutiBhatporia: My initial intent was to get the maximum GPU clock in formation and, check it against the current set device clock, thank you for the information regarding  `nvmlDeviceGetSupported*Clocks`. That solves the query.

lib/cuda/cudaDeviceInterface.cpp
ShrutiBhatporia: The cuda statements to acquire clocks throw the error and are shifted outside the for loop.

lib/cuda/cudaDeviceInterface.cpp
ShrutiBhatporia: Added a patch for finding the clock closest to the list and setting the clock. The comparison is made on the integer values of the frequency clocks.

PR #1062, opened 2022-10-24
feat: added config for testing the drone

jrs65:


andrerenard:
@mondana @ljgray Was this ever used/will be be ever used, should we merge it in, or just abandon this PR?

mondana:
This is potentially the one used during Will & Laura & Richard's drone campaign and Will's paper .... Please keep.

andrerenard:
If it's useful should we just merge it in?  Otherwise we run the risk of not keeping it updated when there are renaming changes, etc.

ljgray:
Even if we never use it again, we should probably just merge it since it was used for the paper

andrerenard (approved):
Merging this in as it was used in a paper, and we may want to use it in the future.  Bypassing CI errors, as the develop branch is waiting on changes from the `chord` branch to fix the CI.

PR #1063, opened 2022-10-28
clear cache memory in nDiskFileWrite

ninaneiman:
Added `syncfs` and `posix_fadvise` to clear cache memory taken by each file in `nDiskFileWrite.cpp`. It works and it solved the cache memory build-up and consequent packet loss while running kotekan with `nDiskFileWrite` stage. Could you check if this is the correct sync function?


andrerenard (approved):
Looks good to me.

PR #1067, opened 2022-11-10
Ng/baseband reader aro

mhvk:
@andrerenard - I think the below would dump data correctly for us but thought I'd at least point you to it so you can have a look (last commit only; 5c58bb212329c3b139513ba6e107c85e36e180e7). We'll test it later today...

p.s. Would it be possible to be added to kotekan so I can push branches directly?

### DIFF
--- a/lib/stages/basebandReadout.cpp
+++ b/lib/stages/basebandReadout.cpp
@@ -79,9 +79,10 @@ basebandReadout::basebandReadout(Config& config, const std::string& unique_name,
         _inputs[order_inds[i]] = inputs_copy[i];
     }
 
-    // Memcopy byte alignments assume the following.
-    if (_num_elements % 128) {
-        throw std::runtime_error("num_elements must be multiple of 128");
+    // Memcopy byte alignments assume data has shape (num_elements, n_time, num_local_freq),
+    // while the straight copy assumes (num_elements=16, num_local_freq, n_time).
+    if (!(_num_elements % 128 == 0 || _num_elements == 16)) {
+        throw std::runtime_error("num_elements must be multiple of 128 or equal to 16");
     }
 
     register_consumer(in_buf, unique_name.c_str());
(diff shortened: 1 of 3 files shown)


lib/stages/basebandReadout.cpp
```diff
@@ -509,10 +510,32 @@ basebandDumpData::Status basebandReadout::extract_data(basebandDumpData data) {
...
+                    // frequency we have to skip over the others.
+                    assert(_num_elements == 16);
+                    // Make pointers to 16-byte unsigned integers to be able to copy all
+                    // 16 elements in one go, leaving it to the compiler to optimize this.
+                    __int128* out_ptr = (__int128*)out_frame;
+                    __int128* in_ptr = (__int128*)in_buf_data;
+
+                    out_ptr += out_start;
+                    in_ptr += in_start;
+                    for (int i = 0; i < copy_len; i++) {
```
andrerenard: I think you need to limit the copy length with `copy_len = std::min(in_end - in_start, out_remaining);` as is done in the `_num_freq_per_stream == 1` case to avoid running over the input frame end.

lib/stages/basebandReadout.cpp
```diff
@@ -509,10 +510,32 @@ basebandDumpData::Status basebandReadout::extract_data(basebandDumpData data) {
...
+                    // different frequencies, so to get a time stream for a given
+                    // frequency we have to skip over the others.
+                    assert(_num_elements == 16);
+                    // Make pointers to 16-byte unsigned integers to be able to copy all
+                    // 16 elements in one go, leaving it to the compiler to optimize this.
+                    __int128* out_ptr = (__int128*)out_frame;
+                    __int128* in_ptr = (__int128*)in_buf_data;
+
+                    out_ptr += out_start;
+                    in_ptr += in_start;
```
andrerenard: You need to offset this by `stream_freq_idx` as well, since this loop is run once per frequency in the stream, but the index into the input buffer always starts at the first frequency.

andrerenard (changes requested):
I've added you to the repo, note that you still cannot push directly to main or develop, since those require a PR.

I've also added some inline comments.

lib/stages/basebandReadout.cpp
mhvk: That's done at the start of the `else` clause.

lib/stages/basebandReadout.cpp
mhvk: Ah, how silly of me to forget to do that! Thanks!

lib/stages/basebandReadout.cpp
mhvk: Actually, now I see that there is a minimum of `1` there -- I had overlooked that. That also means that branch works differently from what I thought -- was it correct all the time?

lib/stages/basebandReadout.cpp
```diff
@@ -509,10 +510,32 @@ basebandDumpData::Status basebandReadout::extract_data(basebandDumpData data) {
...
+                    for (int i = 0; i < copy_len; i++) {
+                        *out_ptr = *in_ptr;
```
andrerenard: There might be an advantage to using a `_mm_stream_load_si128` followed by an `_mm_stream_si128`.  Since we aren't going to be reading the output until later in another thread, using the NT hint for store reduces cache pressure and avoids a line read.  And because the data read per frequency loop is likely too large for the cache, the prefetcher is likely hurting not helping.  The compiler is unlikely to add the NT hint on its own.

andrerenard: The arrays are all allocated using an aligned memory allocator, so the alignment conditions for using those instructions should be met.

lib/stages/basebandReadout.cpp
andrerenard: Yes the branch is correct with that minimum of `1`.   At least for the outriggers case num_elements=128, frequencies_per_stream=16 that has been working (we even have B0329 pulses).

mhvk: Yes, we're a bit jealous of your pulses...  Hopefully we get some too today, and then can check whether they correlate with the 46 m...

andrerenard:
Is this ready?  Sorry I've lost the thread.   If you want me to review/merge, can you take it out of "draft" mode?

mhvk:
@andrerenard - probably best to close this and focus on #1097 now.

PR #1068, opened 2022-11-30
Cuda GPU integration via loading compiled PTX files

dstndstn:
This adds the ability to load compiled PTX kernel for NVIDIA/CUDA GPUs.

Adds the baseband beamforming PTX kernel for the CHORD project. 

As a major side-effect, we also changed the buffer size type from int to size_t, so this produced a cascade of type changes throughout!


config/tests/onehot_cuda_baseband_beamformer_voltage.yaml
```diff
@@ -0,0 +1,177 @@
+##########################################
+#
+# Config to test Kendrick & Erik's cuda baseband beamformer kernel
+# using one-hot testing: setting one element in the voltage array
+# to a non-zero value, with the phase array all one constant value.
+#
+##########################################
```
andrerenard, line 7: I'd remove all the onehot configs until they generate a pass-fail result.

config/tests/verify_cuda_baseband_beamformer.yaml
```diff
@@ -0,0 +1,140 @@
...
+    log_level: debug
+
+gen_shift:
+    type: const
+    value: 0x0
+    kotekan_stage: testDataGen
+    out_buf: host_shift_buffer
+    log_level: debug
+
+#data_format: "cuda_wmma"
```
andrerenard: Can likely remove this line.

lib/cuda/cudaBasebandBeamformer.cpp
```diff
@@ -0,0 +1,141 @@
...
+using kotekan::Config;
+
+#define CHECK_CU_ERROR(result)                                                                     \
+    if (result != CUDA_SUCCESS) {                                                                  \
+        const char* errstr = NULL;                                                                 \
+        cuGetErrorString(result, &errstr);                                                         \
+        internal_logging(LOG_ERR, __log_prefix, "Error at {:s}:{:d}; Error type: {:s}", __FILE__,  \
+                         __LINE__, errstr);                                                        \
+        std::abort();                                                                              \
+    }
```
andrerenard: Why is this not not using `CHECK_CUDA_ERROR` from `cudaUtil.hpp`?

lib/cuda/cudaBasebandBeamformer.cpp
```diff
@@ -0,0 +1,141 @@
...
+    int64_t maxsize;
+    int64_t dims[N];
+    int64_t len;
+};
+typedef CuDeviceArray<int32_t, 1> kernel_arg;
+
+cudaEvent_t cudaBasebandBeamformer::execute(int gpu_frame_id, cudaEvent_t pre_event) {
+    pre_execute(gpu_frame_id);
+
+    size_t voltage_len = (size_t)_num_elements * _num_local_freq * _samples_per_data_set;
```
andrerenard: It feels like a really minor optimization, but we could make these `*_len` variables member values, and compute them once in the constructor.  Which might also make the code a little easier to read.

lib/gpu/gpuDeviceInterface.hpp
```diff
@@ -37,7 +37,7 @@ class gpuDeviceInterface : public kotekan::kotekanLogging {
      * NOTE: if accessing an existing named region then len must match the existing
      * length or the system will throw an assert.
      */
-    void* get_gpu_memory_array(const std::string& name, const uint32_t index, const uint32_t len);
+    void* get_gpu_memory_array(const std::string& name, const uint32_t index, const size_t len);
```
andrerenard, line 40: Oh this is potentially even more painful... The `hsaDeviceInterface`, and `hipDeviceInterface` also need to be updated, which will likely cascade into more areas in those two frameworks.   It's a bit tricky to test that, because one only really works on the old ROCm drivers code on CHIME, and the other just isn't used outside of testing.  But at the very least we should update the signatures in those so we don't immediately fail in those builds.

lib/metadata/oneHotMetadata.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+#include <sys/time.h>
+#include <vector>
+
+#pragma pack()
+
+struct oneHotMetadata {
+    struct chimeMetadata chime;
+    int frame_counter;
+    std::vector<int> hotIndices;
```
andrerenard, line 19: We should change this to a fixed size array, or if we plan to do that in the next PR, at least put a big TODO here to not use vector.

lib/stages/printSparse.hpp
```diff
@@ -0,0 +1,114 @@
...
+#include <functional> // for bind
+#include <regex>      // for match_results<>::_Base_type
+#include <stdexcept>  // for invalid_argument, runtime_error
+#include <stdlib.h>   // for size_t
+#include <string>     // for allocator, string, operator+, to_string, char_traits
+#include <vector>     // for vector
+
+
+template<typename A_Type>
+class printSparse : public kotekan::Stage {
```
andrerenard, line 38: This needs the standard stage comments.

lib/stages/printSparse.hpp
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }
+
+        INFO("printSparse: {:s}[{:d}] has {:d} elements set.", buf->buffer_name, buf_id, nset);
+        if (nset == 0) {
+            INFO("PY sparse[\"{:s}\"][{:d}] = (None, 0)", buf->buffer_name, frame_counter);
+        }
+
+        mark_frame_empty(buf, unique_name.c_str(), buf_id);
+        buf_id = (buf_id + 1) % buf->num_frames;
```
andrerenard: You could use the frameID type to simply the logic here to `buf_id++`

lib/stages/testDataGen.cpp
```diff
@@ -123,6 +141,9 @@ void testDataGen::main_thread() {
     double frame_length =
         samples_per_data_set * ts_to_double(Telescope::instance().seq_length()) / num_links;
 
+    if (((type == "random") || (type == "onehot")) && _seed)
```
andrerenard: We should likely add a comment to the header docs to indicate a seed of 0 is basically non-seeded.

lib/stages/testDataGen.cpp
```diff
@@ -214,7 +279,7 @@ void testDataGen::main_thread() {
         frame_id_abs += 1;
         if (num_frames >= 0 && frame_id_abs >= num_frames) {
             INFO("Frame ID greater than the no. of frames");
-            exit_kotekan(ReturnCode::CLEAN_EXIT);
+            // exit_kotekan(ReturnCode::CLEAN_EXIT);
```
andrerenard: I think we can just remove this entirely.  And also the `INFO` statement above.  Or at least change it to something like "Reached maximum number of frames, stopping data generation".

lib/testing/gpuSimulateCudaBasebandBeamformer.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include "Config.hpp"          // for Config
+#include "Stage.hpp"           // for Stage
+#include "bufferContainer.hpp" // for bufferContainer
+
+#include <stdint.h> // for int32_t
+#include <string>   // for string
+
+/**
+ * @class gpuSimulateCudaBasebandBeamformer
+ * @brief Stage for faking CUDA baseband beamforming.
```
andrerenard, line 18: Doc string should include the expected configuration and buffer parameters.

tests/bb_bf_onehot_post.py
```diff
@@ -0,0 +1,16 @@
+print("VOLTAGE    CPU(re,im)       CPU(re,im)sh     CPU       GPU")
```
andrerenard, line 4: It would be good to have some comments on what this script does, and where it gets used.

tests/bb_bf_onehot_pre.py
```diff
@@ -0,0 +1,5 @@
+bb = {}
```
andrerenard, line 21: Same here.

lib/stages/CMakeLists.txt
```diff
@@ -75,6 +75,8 @@ add_library(
     DPDKShuffleSimulate.cpp
     hexDump.cpp
     BadInputFlag.cpp
+    hexDump.cpp
```
andrerenard, line 77:
```suggestion
```
`hexDump.cpp` seems to be listed twice.

andrerenard (changes requested):

lib/cuda/cudaBasebandBeamformer.cpp
dstndstn: I moved this to cudaUtil.hpp -- it's for API functions starting "cu" vs those starting "cuda".  "driver" vs "runtime" API levels.

lib/cuda/cudaBasebandBeamformer.cpp
dstndstn: Done!

lib/gpu/gpuDeviceInterface.hpp
dstndstn: Okay, so I tried propagating this all through cl/hsa/hip directories at the level of grepping for things.

lib/metadata/oneHotMetadata.hpp
dstndstn: I just added a big ol' TODO

lib/stages/printSparse.hpp
dstndstn: done!

lib/stages/printSparse.hpp
dstndstn: Thanks for the tip!

lib/stages/testDataGen.cpp
dstndstn: Done

lib/stages/testDataGen.cpp
dstndstn: Done

lib/testing/gpuSimulateCudaBasebandBeamformer.hpp
dstndstn: Done!

tests/bb_bf_onehot_post.py
dstndstn: done; it aint pretty

lib/stages/CMakeLists.txt
dstndstn: good catch, thanks

lib/cuda/cudaBasebandBeamformer.cpp
```diff
@@ -0,0 +1,145 @@
...
+                                 + std::to_string(cuda_nelements)
+                                 + " for the CUDA baseband beamformer");
+    if (_num_local_freq != cuda_nfreq)
+        throw std::runtime_error("The num_local_freq config setting must be "
+                                 + std::to_string(cuda_nfreq)
+                                 + " for the CUDA baseband beamformer");
+    if (_samples_per_data_set != cuda_nsamples)
+        throw std::runtime_error("The samples_per_data_set config setting must be "
+                                 + std::to_string(cuda_nsamples)
+                                 + " for the CUDA baseband beamformer");
```
andrerenard, line 35: Not something to fix, but if you prefer format strings you can use `fmt:format("some text {:d}", cuda_nsamples)`.

config/tests/onehot_cuda_baseband_beamformer.yaml
```diff
@@ -0,0 +1,149 @@
+##########################################
+#
+# verify_cuda_baseband_beamformer.yaml
```
andrerenard: Delete this

andrerenard (changes requested):

andrerenard (approved):

PR #1069, opened 2022-11-30
Import Kendrick's "N2k" N-squared correlator kernel for Cuda

dstndstn:
N2k is a small stand-alone library that provides a C++ interface to the Cuda kernel.
N2k in turn uses the "gputils" library.

Also included:
- pass C++17 flags to the Cuda compiler
- a kotekan stage to zero out the lower-triangle of a (sub-integrated) correlation matrix, to allow comparison between the GPU results and a CPU implementation
- a kotekan stage performing the CPU equivalent of the GPU kernel
- we added the "random_signed" type to the testDataGen stage, generating 4+4 integers that are in the range -7 to +7 (no -8)

The `config/tests/verify_cuda_n2k.yaml` file implements a quick unit test comparing the CPU and GPU implementations.


lib/testing/gpuSimulateN2k.hpp
```diff
@@ -0,0 +1,29 @@
...
+#define GPU_SIMULATE_N2K_HPP
+
+#include "Config.hpp"          // for Config
+#include "Stage.hpp"           // for Stage
+#include "bufferContainer.hpp" // for bufferContainer
+
+#include <stdint.h> // for int32_t
+#include <string>   // for string
+
+class gpuSimulateN2k : public kotekan::Stage {
```
andrerenard, line 34: This needs documentation with a description, and list of input/output buffers, and config options.   The header descriptions are very useful for constructing the config files.

lib/testing/zeroLowerTriangle.hpp
```diff
@@ -0,0 +1,30 @@
...
+#define ZERO_LOWER_TRIANGLE_HPP
+
+#include "Config.hpp"          // for Config
+#include "Stage.hpp"           // for Stage
+#include "bufferContainer.hpp" // for bufferContainer
+
+#include <stdint.h> // for int32_t
+#include <string>   // for string
+
+class zeroLowerTriangle : public kotekan::Stage {
```
andrerenard, line 43: Same here, we need stage documentation.

lib/testing/zeroLowerTriangle.cpp
```diff
@@ -0,0 +1,100 @@
...
+                for (int x = 0; x < _num_elements; ++x) {
+                    for (int y = 0; y < _num_elements; ++y) {
+                        int real = 0;
+                        int imag = 0;
+                        if (x <= y) {
+                            real = input[(((tout * _num_local_freq + f) * _num_elements + x)
+                                              * _num_elements
+                                          + y)
+                                             * 2
+                                         + 0];
```
andrerenard: I don't love how clang-format handled this, it's quite hard to read...  We might want layout these by hand, and then include the clang-format ignore comment `// clang-format off` and `// clang-format on`

lib/cuda/cudaCorrelator.hpp
```diff
@@ -0,0 +1,45 @@
...
+#define KOTEKAN_CUDA_CORRELATOR_CUH
+
+#include "cudaCommand.hpp"
+#include "cudaDeviceInterface.hpp"
+#include "n2k.hpp"
+
+/**
+ * @class cudaCorrelator
+ * @brief cudaCommand for doing an N2 correlation, with Cuda code from Kendrick.
+ *
```
andrerenard, line 11: Should include comments on the config parameters, and gpu memory expectations.

config/tests/verify_cuda_n2k.yaml
```diff
@@ -0,0 +1,109 @@
...
+    kotekan_buffer: standard
+    num_frames: buffer_depth
+    frame_size: num_local_freq * (samples_per_data_set / sub_integration_ntime) * num_elements * num_elements * 2 * sizeof_int
+    metadata_pool: main_pool
+
+gen_data:
+    type: random_signed
+    value: 1532
+    #type: const
+    #value: 0x11
```
andrerenard:
```suggestion
```
Generally I don't like to keep commented out code/config.

lib/testing/gpuSimulateN2k.cpp
```diff
@@ -0,0 +1,107 @@
...
+                            }
+
+                        output[(((tout * _num_local_freq + f) * _num_elements + x) * _num_elements
+                                + y)
+                                   * 2
+                               + 0] = +real;
+                        output[(((tout * _num_local_freq + f) * _num_elements + x) * _num_elements
+                                + y)
+                                   * 2
+                               + 1] = -imag;
```
andrerenard: Same strange formatting here.

andrerenard (changes requested):
This overall looks really good, we should just make sure it has good documentation in the headers, and then we can merge.

lib/testing/gpuSimulateN2k.cpp
dstndstn: done

dstndstn:
The IWYU test failure seems to be a false positive -- it's complaining about ExampleProducer.cpp etc, which do not exist on this branch!

dstndstn:
I'm going to try merging in `develop` and see if that fixes it.

dstndstn:
The IWYU failure reports problems with:
```
>>> Fixing #includes in '/code/kotekan/lib/stages/ExampleProducer.hpp'
>>> Fixing #includes in '/code/kotekan/lib/stages/ExampleProducer.cpp'
>>> Fixing #includes in '/code/kotekan/lib/stages/ExampleConsumer.hpp'
>>> Fixing #includes in '/code/kotekan/lib/stages/ExampleConsumer.cpp'
>>> Fixing #includes in '/code/kotekan/lib/stages/ExampleDotProduct.hpp'
>>> Fixing #includes in '/code/kotekan/lib/stages/ExampleDotProduct.cpp'
```
which are not relevant to this branch.

lib/cuda/cudaCorrelator.hpp
```diff
@@ -0,0 +1,68 @@
...
+ * - sub_integration_ntime -- number of time samples that will be summed
+ *     into the correlation matrix
+ *
+ * Input buffer (gpu):
+ * - gpu_mem_voltage -- size per frame: samples_per_data_set * num_elements * num_local_freq
+ *   -4+4-bit complex voltage inputs
+ *
+ * Output buffer (gpu):
+ * - gpu_mem_correlation_triangle -- size per frame: (samples_per_data_set / sub_integration_ntimes)
+ * * num_elements^2 * 2 * sizeof(int32) -Complex correlation values
```
andrerenard: For the headers, please use doxygen tags for the buffers and config parameters.  We haven't made a ton of use of them yet, but the idea is to be able to auto generate documents for each `CommandObject/Stage`, and if nothing else it creates a standardized way of listing these.   See for example: https://github.com/kotekan/kotekan/blob/develop/lib/hsa/hsaBeamformHFBSum.hpp for a command object, and https://github.com/kotekan/kotekan/blob/develop/lib/stages/eigenVis.hpp for a stage.

andrerenard (changes requested):
Just one comment about the header comment formatting, but otherwise this looks good to merge.

lib/cuda/cudaCorrelator.hpp
dstndstn: Done!

config/tests/verify_cuda_n2k.yaml
```diff
@@ -0,0 +1,107 @@
...
+    frame_size: num_local_freq * (samples_per_data_set / sub_integration_ntime) * num_elements * num_elements * 2 * sizeof_int
+    metadata_pool: main_pool
+
+gen_data:
+    type: random_signed
+    value: 1532
+    kotekan_stage: testDataGen
+    out_buf: host_voltage_buffer
+
+data_format: "cuda_wmma"
```
andrerenard: I don't think this line is needed.

lib/cuda/CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
 project(kotekan_cuda)
 
+# NVIDIA Ampere
+set(CMAKE_CUDA_ARCHITECTURES 86)
```
andrerenard: I feel like we should provide the option to override this with a `-DCUDA_ARCHITECTURE=` command line option, and have the default as 86.  Something like:
```
if(${CUDA_ARCHITECTURE})
    set(CMAKE_CUDA_ARCHITECTURES ${CUDA_ARCHITECTURES})
else()
    set(CMAKE_CUDA_ARCHITECTURES 86)
endif()
```

andrerenard (approved):

PR #1075, opened 2023-02-14
Make clang-format work with version 8 and 14

andrerenard:
Ideally we'd bump the clang-format version to 14 or higher, but due to the current CI system being locked to ubuntu 18.04, we cannot do that without also making lots of other changes.   We should bump the ubuntu version in the CI system, but this will be done in another ticket.


dstndstn (approved):
LGTM!

PR #1076, opened 2023-02-15
Add support for more than 3 cuda streams and custom synchronization between them

andrerenard:
Currently there are 3 cuda streams, two for copies, and one for compute.  This change allows for an arbitrary number of streams set by the `num_streams` option.   When using this option it becomes necessary to  provide a `cuda_stream` for each commandObject. 

To synchronize between streams, a new command object called `cudaSyncStream` can be used to place cuda wait events which depend on other streams. 

All configurations moving forward will need to use this object to sync between the kernel and copy streams.


andrerenard:
@dstndstn I still need to do a few more correctness tests, but this seems to be passing the initial tests, and it's ready for review.

dstndstn:
(no rush, just putting this here while I think of it --)
@andrerenard , I'd appreciate your comment on [0744ecb](https://github.com/kotekan/kotekan/pull/1076/commits/0744ecbaa0fc4c3903479b4b1e9e1ef646cb857e) -- I added an additional cudaSyncStream constructor that doesn't run the initialization steps whose arguments come from the config file, so I can do something different in the cudaSyncInput constructor.  Maybe you want to recommend a better way to do this.

dstndstn (approved):
This seems reasonable to me!

dstndstn:
By the way, it looks like the one failing CI test, the parallel python tests, are consistently failing, not just intermittently from a floating-point issue, but related to setting memory policy:
```
Creating visBuffer named dumpvis_buf0 with 4 frames, frame size of 176 and metadata pool vis_pool on numa_node 0
Failed to set memory policy: Operation not permitted (1)
Failed to start kotekan with config file /tmp/tmpdd2cyatj, error message: Could not create the buffer: dumpvis_buf0
Exiting...
```

dstndstn:
@andrerenard From my perspective, I think this is ready, but you should review my changes.  (I don't seem to be able to request a review from you, since it's your PR?)  Thanks!

lib/cuda/cudaBasebandBeamformer.cpp
```diff
@@ -79,11 +80,10 @@ cudaEvent_t cudaBasebandBeamformer::execute(int gpu_frame_id, cudaEvent_t pre_ev
...
         (int32_t*)device.get_gpu_memory_array(_gpu_mem_info, gpu_frame_id, info_len);
 
-    if (pre_event)
-        CHECK_CUDA_ERROR(cudaStreamWaitEvent(device.getStream(CUDA_COMPUTE_STREAM), pre_event, 0));
-    CHECK_CUDA_ERROR(cudaEventCreate(&pre_events[gpu_frame_id]));
-    CHECK_CUDA_ERROR(
-        cudaEventRecord(pre_events[gpu_frame_id], device.getStream(CUDA_COMPUTE_STREAM)));
+    if (pre_events[cuda_stream_id])
+        CHECK_CUDA_ERROR(
+            cudaStreamWaitEvent(device.getStream(cuda_stream_id), pre_events[cuda_stream_id], 0));
```
andrerenard: I think we can actually remove the wait event on the pre_event.  Since the pre_event is now on the same stream streams, this doesn't provide any meaningful synchronization past what the stream ordering guarantees already.   Before this was used to sync streams, but that's now provided by `syncStream`.

lib/cuda/cudaSyncStream.cpp
```diff
@@ -0,0 +1,52 @@
...
+    cudaCommand(config, unique_name, host_buffers, device, "sync", "") {
+    set_source_cuda_streams(config.get<std::vector<int32_t>>(unique_name, "source_cuda_streams"));
+    set_command_type(gpuCommandType::BARRIER);
+}
+
+cudaSyncStream::cudaSyncStream(kotekan::Config& config, const std::string& unique_name,
+                               kotekan::bufferContainer& host_buffers, cudaDeviceInterface& device,
+                               bool called_by_subclasser) :
+    cudaCommand(config, unique_name, host_buffers, device, "sync", "") {
+    (void)called_by_subclasser;
```
andrerenard, line 19: This still needs a command type set for the internal profiling.  
```suggestion
    (void)called_by_subclasser;
    command_type = gpuCommandType::BARRIER;
```

lib/cuda/cudaSyncStream.hpp
```diff
@@ -0,0 +1,43 @@
...
+    cudaSyncStream(kotekan::Config& config, const std::string& unique_name,
+                   kotekan::bufferContainer& host_buffers, cudaDeviceInterface& device);
+    ~cudaSyncStream();
+    int wait_on_precondition(int gpu_frame_id) override;
+    cudaEvent_t execute(int gpu_frame_id, const std::vector<cudaEvent_t>& pre_events) override;
+    void finalize_frame(int frame_id) override;
+
+protected:
+    /// A constructor meant to be called by subclassers who need to
+    /// override the constructor behavior
```
andrerenard: I cannot really think of a better way to do this.  However, I might add a little more to this comment to explain what's different.  e.g. doesn't get streams from the configuration file.

andrerenard (commented):
Thanks for the changes!   It looks much cleaner now for the default case.   I cannot "request changes" on my own PR, but there are a few small changes in this comment we should do before merging.

lib/cuda/cudaBasebandBeamformer.cpp
dstndstn: Does this go for all cudaCommand compute subclasses?  Eg this same snippet also appears in `cudaCorrelatorAstron.cpp`, `cudaCorrelator.cpp` and `cudaOutputDataZero.cpp`.

lib/cuda/cudaSyncStream.cpp
dstndstn: Done!

lib/cuda/cudaSyncStream.hpp
dstndstn: done

lib/cuda/cudaBasebandBeamformer.cpp
andrerenard: Yes I think we can remove this from all subclasses (other than the sync of course).

dstndstn: Right, I see what you mean now, these were just waiting on the previous event in their stream, which is... already how a stream works :)  Done!

dstndstn:
Just the usual CI test failing (python test failing to set memory policy)!  Good to merge?

andrerenard:
Hmm, that's a new failure mode... I'm going to rerun it and see if it's consistent.

dstndstn:
FWIW I also see it in https://github.com/kotekan/kotekan/pull/1078

andrerenard:
Ah, looks like I broke that with another PR, but didn't catch it because it was also failing with the unrelated float issue.  It doesn't impact this PR though, so I'll merge this in, and then fix that separately.

PR #1074, opened 2023-02-13
Add huge page support to buffers and improve NUMA handling

andrerenard:
* Add support for 2MB huge pages
* Allocate all buffer structure memory to the requested NUMA domain
* Fix a problem with the bufferFactory error handling
* Fixed some typos

PR #1079, opened 2023-02-22
Fix a memory lock issue with restricted docker instances.

andrerenard:
In #1074 we introduced an issue for restricted docker containers which don't allow memory locking or adjusting the memory policy.  This was preventing some github CI tests from running.   To fix this we add back the `NO_MEMLOCK` compiler option which removes the binary's ability to make memory policy changes.

This change also introduces a "zero_new_frames" config option to override the default behavior of zeroing new frames.   This is useful in situations where a stage doesn't assume anything about the frame contents (true for almost all stages), and so can reduce the startup time by just not zeroing the memory.   The option is `true` by default to preserve backwards compatibility with any configs that assume it's true.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -319,8 +319,10 @@ if(${COMPILE_DOCS})
     add_subdirectory(docs EXCLUDE_FROM_ALL)
 endif()
 
-# Turn off memlocking of buffers. This is useful when running in restricted environments e.g. Docker
-# containers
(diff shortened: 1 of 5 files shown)


andrerenard:
@dstndstn finally got this to pass the CI.  It's ready to review anytime, and should fix the new full-build test issues.

dstndstn (approved):
Looks good!

PR #1080, opened 2023-02-23
Fix Config.hpp build warning for boolean types

andrerenard:
This stops the compiler from trying to compile a `configEval<bool>`, which was unsupported by the program logic already but getting compiled by the template logic and generating a warning.

### DIFF
--- a/lib/core/Config.hpp
+++ b/lib/core/Config.hpp
@@ -86,7 +86,7 @@ public:
     }
 
     /**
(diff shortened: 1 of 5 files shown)


andrerenard:
@dstndstn no rush on this one; I got annoyed enough at that warning that I finally fixed it...

dstndstn (approved):
Hallelujah!

PR #1084, opened 2023-03-07
Support newer versions of DPDK

andrerenard:
* Adds support for DPDK 21.11 included in Ubuntu 22.04
* Removes the `-DUSE_DPDK=ON` cmake flag, and auto detects DPDK with `pkg-config`
* Keeps support for older versions of DPDK via the new `-DUSE_OLD_DPDK=ON` cmake flag
* Changes config file parameter:  `master_lcore_cpu` -> `main_lcore_cpu`
* Removes an old Jenkinsfile
* Adds check for UDP checksum failure in supporting hardware
* Tested on the Intel E810 NIC.


dstndstn (approved):
Looks good to me!

PR #1085, opened 2023-03-09
Disable warnings from CUDA library headers

andrerenard:
* Prevent warnings form CUDA headers from being included in cmake output. 
* Fix a few unused variable warnings.


dstndstn:
Hallelujah!

dstndstn (approved):

PR #1086, opened 2023-03-22
Try implementing a re-chunking GPU command

dstndstn:
This is my attempt at implementing a GPU command that assembles multiple input frames into one output frame.

The awkward thing is that you then only want to run subsequent GPU commands when a full frame has been assembled.  So I added a "quit" from to the command->execute() call.  If a command requests to quit, then all subsequent commands get command->skipped_execute() called, instead of command->execute().  This lets those commands do any cleanup they might have to do.

One ugly thing about this implementation is that even though GPU commands following the rechunker don't get called every time, the gpu_frame_id still increases in lockstep.  So if we're rechunking by a factor of 5, then the command following rechunking will get called on gpu_frame_id 4, 9, 14, 19, etc, mod the gpu buffer depth.

One side effect of this is that the cudaRechunk command assembles its output frame into a temporary buffer, because it doesn't know until the output is assembled which gpu_frame_id it is supposed to put the output into!  (Yes, it could look ahead, but I didn't want to have to think about whether that frame was already in use.)

It does work, though!

In the testing YAML included here, I generate a series of input buffers with values 1,2,3,...20, and it rechunks it by a factor of 5, and the results are
/hexDump: hexDump: Got buffer host_output_buffer[0]
  0000  01 02 03 04 05
/hexDump: hexDump: Got buffer host_output_buffer[1]
  0000  06 07 08 09 0a
/hexDump: hexDump: Got buffer host_output_buffer[2]
  0000  0b 0c 0d 0e 0f
/hexDump: hexDump: Got buffer host_output_buffer[3]
  0000  10 11 12 13 14


dstndstn:
The only files you'll want to look at are `cudaProcess.cpp` and `cudaOutputData.cpp`  (and possibly `cudaRechunk.cpp`).

dstndstn:
The issue that the gpu_frame_id is not uniform throughout the pipeline could be avoided by, eg, having each gpuCommand keep its own gpu_frame_id counter that would increment on each execute() call.

PR #1093, opened 2023-05-23
Kko extending bbdump

shionandrew:
Changed the kko gpu.j2.yaml configuration to:

```
baseband_buffer_depth: 222
baseband_output_buffers:
    num_frames: 165 * num_local_freq
max_dump_samples: 548864
```

to enable 1.4s long baseband dumps. The staging buffer can hold up to 1.7s of data, but max_dump_samples is configured to prevent dumps longer than 1.4s: the extra 300ms in the staging buffer is meant to hold data for an FRB in the event that one is triggered during the 30-40 minutes it takes for a 1.4s long dump to be written to disk.


config/chime_kko_gpu.j2
```diff
@@ -225,7 +225,7 @@ zero_samples:
 
 ### Start Baseband Section ###
 baseband:
-    max_dump_samples: 1024000 # 1024ms
+    max_dump_samples: 548864 #1024000 # 1024ms
```
andrerenard: Could we just update the comment here to `# 1.4 seconds`? 
```suggestion
    max_dump_samples: 548864 # 1.4 seconds.
```

andrerenard (changes requested):

PR #1097, opened 2023-07-02
Allow ARO to write threads with input and frequency ranges

mhvk:
@andrerenard - this is an edit of your old `iceBoardVDIF.hpp` to allow more control over what is recorded at ARO (mostly in preparation for getting a CHORD feed there, which will need more data to be written). cc @ninaneiman 

Note that I rebased on `develop` but am not sure that is what I should do -- a quick try at compiling this on cloyd got me something without `dpdkCore`, so I'm probably doing something wrong (maybe just omitting the `CMake`?)

But I thought I would make the PR anyway as the code seems in good enough shape to ask for comments (and it does work properly on top of the branch we were using at ARO). I'll happily squash commits if needed.

The last commit is the main one. It contains also the main idea:
```
What to save is controlled by the config file, as follows:
Writing is done in num_thread chunks of num_elements * num_freq samples,
picking some of the 16 elements and some of the 1024 frequencies,
as given by offsets and frequencies.

Examples: reproduce old-style of all frequencies of 2 inputs in 2 separate threads:

num_threads: 2
num_elements: 1
num_freq: 1024
offsets: [8, 9]
frequencies: [0, 0]

CHORD-like 300-800, supposing 6&7 contain 0-400, and 8&9 400-800.

num_threads: 5
num_elements: 2
num_freq: 256
offsets: [6, 8, 8, 8, 8]
frequencies: [0, 0, 256, 512, 768]

CHORD-like 800-1500, supposing 10&11 contain 800-1200, and 8&9 1200-1500.

num_threads: 7
num_elements: 2
num_freq: 256
offsets: [10, 10, 10, 10, 8, 8, 8]
frequencies: [0, 256, 512, 768, 256, 512, 768]

Save all frequencies of four consecutive inputs

num_threads: 1
num_elements: 4
num_freq: 1024
offsets: [8]
frequencies: [0]

Save half the frequencies of eight consecutive inputs
num_threads: 1
num_elements: 8
num_freq: 512
offsets: [8]
frequencies: [0]
```

### DIFF
--- a/lib/dpdk/iceBoardVDIF.hpp
+++ b/lib/dpdk/iceBoardVDIF.hpp
@@ -388,6 +388,8 @@ void iceBoardVDIF::copy_packet_vdif(struct rte_mbuf* mbuf) {
 
     // Offset in output frame for first frequency of this receiver buffer thread (0-7).
     uint32_t freq_offset = tel.to_freq_id(encoded_id, 0);
+    // Sanity check: there should be an integer number of frames in the input buffer.
+    assert((mbuf_data_max - mbuf_data_ptr) % (128 * total_num_elements) == 0);
 
     // Times are in separate frames, so time stride equals the size of the VDIF framesets.
     for (uint8_t* out_t_f0 = out_t0_f0;
@@ -395,6 +397,7 @@ void iceBoardVDIF::copy_packet_vdif(struct rte_mbuf* mbuf) {
          out_t_f0 += vdif_frameset_size) {
         // Pointer to start of thread in VDIF frame (updated in loop).
         uint8_t* out_t_p = out_t_f0;
+        //
         for (uint32_t i_thread = 0; i_thread < num_threads; i_thread++) {
             // Pointer to starting input and frequency in input buffer for this thread.
             char* in = mbuf_data_ptr + buffer_offsets[i_thread];
@@ -440,6 +443,8 @@ void iceBoardVDIF::copy_packet_vdif(struct rte_mbuf* mbuf) {
             mbuf_data_base = rte_pktmbuf_mtod(mbuf, char*);
             mbuf_data_max = mbuf_data_base + mbuf->data_len;
             mbuf_data_ptr = mbuf_data_base + mbuf_offset;
+            // Sanity check: there should be an integer number of frames in the input buffer.
+            assert((mbuf_data_max - mbuf_data_ptr) % (128 * total_num_elements) == 0);
         }
     }
 }
--- a/lib/dpdk/iceBoardVDIF.hpp
+++ b/lib/dpdk/iceBoardVDIF.hpp
@@ -377,76 +377,76 @@ void iceBoardVDIF::copy_packet_vdif(struct rte_mbuf* mbuf) {
     auto& tel = Telescope::instance();
     stream_t encoded_id = ice_encode_stream_id(port_stream_id);
 
-    // Create the parts of the VDIF frame that are in this packet.
-    char* mbuf_data_base = rte_pktmbuf_mtod(mbuf, char*);
-    char* mbuf_data_max = mbuf_data_base + mbuf->data_len;
-    char* mbuf_data_ptr = mbuf_data_base + header_offset;
+    // Pointer to where we should start storing our data.
     uint8_t* out_t0_f0 =
         (out_buf_frame                              // Start of output buffer.
          + vdif_frame_location * vdif_frameset_size // Frame location in output buffer.
          + vdif_header_len);                        // Offset for the vdif header.
-
     // Offset in output frame for first frequency of this receiver buffer thread (0-7).
     uint32_t freq_offset = tel.to_freq_id(encoded_id, 0);
-    // Sanity check: there should be an integer number of frames in the input buffer.
-    assert((mbuf_data_max - mbuf_data_ptr) % (128 * total_num_elements) == 0);
+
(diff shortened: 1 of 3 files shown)


mhvk:
@andrerenard - I would love some overall review of this PR, but probably also need help with the branch I should base this on. As I mentioned on top, if I try compiling this version, which I rebased on `develop`, on our ARO machines, no dpdk is produced so nothing runs. Should I rebase on another branch/tag? (Right now, the version that works is based on 4dd7f66a4304855a31d53656a4f4758361fa920c)

andrerenard:
Try adding `-DUSE_OLD_DPDK=ON` to the `cmake` options.  I've been updating the code to work with the much newer DPDK versions and auto detect them, which creates a lot of odd backwards compatibility issues.   I think that should work, but if it doesn't, message me on slack with the cmake output and any compile errors your are getting.

mhvk:
@andrerenard - thanks, adding `-DUSE_OLD_DPDK=ON` worked. I also fixed the style problems (not understanding the `iwyu` ones that seem to add includes that are not needed; let me know if those should go somewhere else...). So, this is ready for review!

lib/dpdk/iceBoardVDIF.hpp
```diff
@@ -267,56 +359,87 @@ inline void iceBoardVDIF::handle_lost_samples(int64_t lost_samples) {
...
+    char* mbuf_data_base = rte_pktmbuf_mtod(mbuf, char*);
+    char* mbuf_data_max = mbuf_data_base + mbuf->data_len;
+    char* mbuf_data_ptr = mbuf_data_base + header_offset;
+    uint8_t* out_t0_f0 =
+        (out_buf_frame                              // Start of output buffer.
+         + vdif_frame_location * vdif_frameset_size // Frame location in output buffer.
+         + vdif_header_len);                        // Offset for the vdif header.
+
+    // Offset in output frame for first frequency of this receiver buffer thread (0-7).
+    uint32_t freq_offset = tel.to_freq_id(encoded_id, 0);
```
andrerenard: Since we know this function can be a bit expensive to call, a small optimization might be to cache this value.   The value `port_stream_id` is only set once on the first packet.  So you could have all the logic to set `freq_offset` happen only if say `freq_offset` is `-1` (making that a member variable).  That could bypass the two related lines above as well.

lib/dpdk/iceBoardVDIF.hpp
```diff
@@ -267,56 +359,87 @@ inline void iceBoardVDIF::handle_lost_samples(int64_t lost_samples) {
...
+
+    // Times are in separate frames, so time stride equals the size of the VDIF framesets.
+    for (uint8_t* out_t_f0 = out_t0_f0;
+         out_t_f0 < out_t0_f0 + vdif_frameset_size * samples_per_packet;
+         out_t_f0 += vdif_frameset_size) {
+        // Pointer to start of thread in VDIF frame (updated in loop).
+        uint8_t* out_t_p = out_t_f0;
+        for (uint32_t i_thread = 0; i_thread < num_threads; i_thread++) {
+            // Pointer to starting input and frequency in input buffer for this thread.
+            char* in = mbuf_data_ptr + buffer_offsets[i_thread];
```
andrerenard: I'm worried that `in` might exceed `mbuf_data_max`.   Have you checked in all cases below that the `in` pointer never reads past the end of the current `mbuf` (for each of the sizes it can be cast to)?   It may actually be possible that it just happens to work out because of how our packet size, the mbuf sizes, and headers line up, but generically I'm not sure this should work, and would like to double check.

andrerenard (changes requested):
Thanks for making these changes.  Overall it looks good, but I haven't been able to convince myself it will always avoid reading past the end of an `mbuf`.   Is this something you've tested? 

Also what's the performance like on this?  Is this able to work in all the key ARO cases without large packet losses?

lib/dpdk/iceBoardVDIF.hpp
mhvk: Yikes, yes, it does, and some debug statements show the buffer size is 2048, of which 58 is header, so 1990 left, which is indeed not even a multiple of 16... I think I have an idea how to do it, using that it won't be split over more than 2 frames, but clearly this is not as simple as I thought...

lib/dpdk/iceBoardVDIF.hpp
mhvk: OK, might as well, indeed. Done in last commit.

lib/dpdk/iceBoardVDIF.hpp
mhvk: OK, now completely rewrote. As I realized I could not count on things being properly aligned, I tested just using `std::copy_n` and it goes just as fast without trying to be clever and have fixed `num_elements`. So, the whole is much simpler now, even with the partial samples.

mhvk:
@andrerenard - on the larger question: I tested that this still works properly for recording standard ARO data with the CHIME receiver. It is substantially faster so one can now also record 4 elements at a go (if combined in one thread), and it can /almost/ keep up with the 1000 MHz band that we need to record for the CHORD receiver that we hope to mount in a few weeks (300-800 + 1000-1500 in 10 threads of 100 MHz and 2 polarizations each). I hope that with the slightly faster other computer we have there (squirrel), it will "just work".

Anyway, thanks for pointing out the big blunder; I hope you don't find further ones!

PR #1078, opened 2023-02-17
[WIP] Add CUDA FRB beamformer

dstndstn:

PR #1107, opened 2023-08-10
Basic changes from Dustin's MVP work

dstndstn:
This PR is trying to extract the changes to some function signatures and structures from Dustin's MVP development branch (dstn/frb-bf).  This was done by generating a giant "git diff" and then selectively applying changes with "git commit -p", trying to group commits in a sensible way.

Important changes are:
- Cuda: Create a pipeline state object (cudaPipelineState) that is passed to each stage's execute() function, as a way of passing information between stages
- silly, for debugging: GPU: pass a GPU frame counter that increments each time queue_commands() is called.
- graphviz: record GPU buffers used by each stage, and add them to the graphviz call graph
- Cuda stages: add the option to require a flag to be set in the cudaPipelineState object in order for a stage to run
- GPU memory buffers: add the ability to create GPU memory views, allowing one stage to see only a subset of a memory buffer created by a different stage.  Used in a few places in the MVP to handle rechunking sorts of operations.
- add some float16_t data type handling

And other changes that were hard to extricate from these diffs include:
- cudaBasebandBeamformer: make the 'info' buffer a local buffer (not known to kotekan)


dstndstn:
btw the Cuda build works

dstndstn:
This plus #1108 make #1078 obsolete

dstndstn:
After switching the macos build from OS version 10.15 to 11, the h5py compile fails;
https://github.com/kotekan/kotekan/actions/runs/5856283980/job/15875669797?pr=1107

dstndstn:
After merging in Jim's CI PR, I dropped the macos-11 build.  Can reinstate it if anyone cares...

dstndstn:
@jbmertens After merging in `develop` to pick up your CI changes, it looks like the 1804 test is failing, but in a way that makes me think this could be an "unlucky" numerical thing -- some of the tests are not deterministic, unfortunately....

jbmertens (approved):

jbmertens:
I've approved pending lint changes. IWYU is passing, but might not actually be running, which I'll look into in #1117.

jbmertens (approved):

dstndstn:
Note that all the github-actions commits are irrelevant -- we deleted the macos-11 build while merging in Jim's CI changes on the `develop` branch.

lib/cuda/cudaCommand.hpp
```diff
@@ -25,6 +25,23 @@
 #include <string>
 #include <vector>
 
+class cudaPipelineState : public kotekan::kotekanLogging {
```
andrerenard, line 28: Normally I would say this needs to be documented...  However I think can remove this in the medium term.

lib/gpu/gpuProcess.cpp
```diff
@@ -166,7 +167,7 @@ void gpuProcess::main_thread() {
         DEBUG("Waiting for free slot for GPU[{:d}][{:d}]", gpu_id, gpu_frame_id);
         // We make sure we aren't using a gpu frame that's currently in-flight.
         final_signals[gpu_frame_id]->wait_for_free_slot();
-        queue_commands(gpu_frame_id);
+        queue_commands(gpu_frame_id, gpu_frame_counter);
```
andrerenard, line 171: Do we want to replace this with a mod counter like `FrameID` now, or address that in another PR?

lib/stages/printSparse.hpp
```diff
@@ -73,6 +75,29 @@ printSparse<A_Type>::printSparse(kotekan::Config& config, const std::string& uni
...
+}
+static std::string format_python_string(uint32_t x) {
+    return fmt::format("{}", x);
+}
+#if KOTEKAN_FLOAT16
+static std::string format_python_string(float16_t x) {
+    return fmt::format("{}", (float)x);
+}
+#endif
+
```
andrerenard: This feels like it should be in a `utils/` file, as I don't think this is the only place we are likely to use it.   Also maybe add some simple documentation to explain what the idea of each of these functions are?

lib/testing/testDataCheck.hpp
```diff
@@ -65,6 +70,24 @@ almost_equal(A_Type x, A_Type y, double epsilon) {
...
+}
+#if KOTEKAN_FLOAT16
+static std::string format_nice_string(float16_t x) {
+    return fmt::format("{}", (float)x);
+}
+#endif
+static std::string format_nice_string(float x) {
+    return fmt::format("{}", x);
+}
+
```
andrerenard: And it is reused here, so should be in a utils file.

tests/boost/CMakeLists.txt
```diff
@@ -65,6 +65,20 @@ target_link_libraries(dataset_broker_producer2 PRIVATE libexternal kotekan_utils
 target_link_libraries(dataset_broker_producer PRIVATE libexternal kotekan_utils kotekan_core)
 target_link_libraries(dataset_broker_consumer PRIVATE libexternal kotekan_utils kotekan_core)
 
+# Many tests now need the Cuda include directory for float16_t via visUtil.hpp
```
andrerenard, line 68: Will this build if `cuda` isn't enabled/installed, but boost tests are?   Or are we making cuda a requirement of the boost tests?

andrerenard (changes requested):
This looks good!  I think there are some larger restructuring things to deal with, but not worth holding up this PR.

lib/gpu/gpuProcess.cpp
dstndstn: I'd like to hold off on this one, since I'm hoping for more sweeping changes to this in the near future!

lib/testing/testDataCheck.hpp
dstndstn: Done!

lib/stages/printSparse.hpp
dstndstn: Done!

tests/boost/CMakeLists.txt
dstndstn: I just tested with `cmake -DUSE_CUDA=OFF ..` and these tests are still built and they run.  So I think it's fine

lib/cuda/cudaCommand.hpp
dstndstn: hah!, sweet, I got away with it!

dstndstn:
Hallelujah!!
Thanks!!

PR #1109, opened 2023-08-10
Documentation updates and starting to update CI some

jbmertens:
This PR is meant to accomplish a few goals, especially developing CI/infrastructure. Specific changes include:
1) A new 22.04 build has been introduced which includes nvidia tools (@dstndstn), julia (@eschnett), and more. Kotekan appears to build and run on python 3.10 / ubuntu 22.04, although some tests are failing (see below).
2) Documentation builds correctly on 22.04 with doxygen 1.9.7+, and has been updated so it builds on readthedocs.
3) The 18.04 Docker image was no longer building; this has been corrected and it now builds again.
4) The github actions workflows should be able to run locally, especially using act.

The PR is incomplete in that the following issues now exist:
1) Boost tests are not found by pytest on the 22.04 build. This should be investigated.
2) chime-experiment/comet (a dependency for some tests) relies on deprecated functionality, so does not work on 22.04. Some tests are skipped on 22.04 as a result.
3) Some package versions have been fixed for compatibility (esp. blaze) or due to bugs (esp. doxygen). Others have been fixed, but may be ok to relax.


lib/stages/bufferMerge.hpp
```diff
@@ -29,14 +29,17 @@
  *
  * @par Buffers
  * @buffer in_bufs Array of input buffers to merge frames from.
+ *
```
andrerenard: Not sure why there are extra spaces here?   It doesn't seem like they are intended.

lib/utils/LinearAlgebra.hpp
```diff
@@ -43,7 +43,7 @@ template<typename MT, bool SO>
 double rms(const blaze::DenseMatrix<MT, SO>& A) {
     double t = 0.0;
 
-    auto At = ~A;
+    auto At = blaze::trans(A);
```
andrerenard: I just want to flag @jrs65 on this change.  I remember we had tons of problems with the API change - and then the library version change.  I think this is safe as long as we keep the library version the same, but I don't remember all the details on what the issues were.

.github/workflows/main.yml
```diff
@@ -19,11 +20,245 @@ on:
     - master
 
 env:
-  IMG_CORE: docker.pkg.github.com/kotekan/kotekan/kotekan-core
-  IMG_IWYU: docker.pkg.github.com/kotekan/kotekan/kotekan-iwyu
+  REGISTRY: docker.pkg.github.com
+  IMG_CORE: "docker.pkg.github.com/kotekan/kotekan/kotekan-core"
+  IMG_IWYU: "docker.pkg.github.com/kotekan/kotekan/kotekan-iwyu"
+  IMG_CI: "docker.pkg.github.com/${{ github.repository }}/kotekan-ci"
```
andrerenard: We likely need to update the naming here to include the Ubuntu versions.  Otherwise I'm not sure what the difference between `kotekan-core` and `kotekan-ci` is in terms of base Ubuntu version.

.github/workflows/main.yml
```diff
@@ -19,11 +20,245 @@ on:
...
+            -DNO_MEMLOCK=ON \
+            -DUSE_OMP=ON \
+            -DWITH_TESTS=ON \
+            -DCCACHE=ON ..
+        docker run "${OPTS[@]}" make -j 2
+        docker run "${OPTS[@]}" rm -r lib
+        docker run "${OPTS[@]}" ccache -s
+        df -h
+
+    - name: Run parallel python tests
```
andrerenard: If it looks like it's going to be hard to get these to pass on 22.04, then for now, I would disable them, and open a ticket to return to this problem.   There is value in getting this 22.04 image working quickly, and tracking down the unit tests mostly related to CHIME systems might take some time.   As long as we keep running them under the 18.04 image, we shouldn't break anything CHIME needs.

.github/workflows/main.yml
```diff
@@ -19,11 +20,245 @@ on:
...
+            -DCMAKE_LINK_WHAT_YOU_USE=ON \
+            -DCMAKE_BUILD_TYPE=Test \
+            -DARCH=haswell \
+            -DCCACHE=ON \
+            -DWITH_TESTS=ON ..
+        docker run "${OPTS[@]}" make -j 2
+        docker run "${OPTS[@]}" ccache -s
+
+  # Build a full version of kotekan and run the unit tests
+  build-full-tests:
```
andrerenard: I might suggest including the Ubuntu version number here as well.  While we can trace the dependencies, it's nice to get a message that tells you quickly "oh the 18.04 or 22.04 build is breaking".

tools/CI/Dockerfile
```diff
@@ -0,0 +1,135 @@
...
+# So tzdata package doesn't ask for user interaction
+ARG DEBIAN_FRONTEND=noninteractive
+
+# Install any needed packages to run cmake with full CHIME build options
+RUN apt-get -y update && \
+    apt-get install -y software-properties-common=0.99.* && \
+    add-apt-repository ppa:deadsnakes/ppa && \
+    apt-get update && \
+    apt-get upgrade -y && \
+    apt-get install -y python3.11=3.11.* \
```
andrerenard: As we discussed, I think the versions here can be somewhat relaxed for the 22.04 build (although possibly Python will need to stay pinned due to library issues we need to resolve first).   For 18.04 we should keep them locked, since we are keeping support for 18.04 really just to support a few systems which are effectively version locked.

andrerenard (changes requested):
This is looking great.  I think we just need to clean up a few small things, and as long as it's passing all the new CI checks, we can merge.   Note, that generally unless you have a very careful git history, generally we use squash merges into develop, which makes it easier to revert PR if necessary and makes for a cleaner git log.   Generally I copy out description as the squash commit message, so make sure to have an updated (bullet-point is fine) list of changes this PR is doing.

lib/stages/bufferMerge.hpp
jbmertens: Doxygen was complaining about this, but now doesn't! Version changes may have fixed this so I'm reverting.

tools/docker/22.04/Dockerfile
```diff
@@ -0,0 +1,176 @@
...
+
+# Install OpenBLAS and clone Blaze for the eigenvalue processes
+RUN apt-get update && \
+    apt-get -y install libopenblas-dev \
+                       liblapack-dev \
+                       liblapacke-dev \
+                       && \
+    apt-get clean && apt-get autoclean && \
+    git clone https://bitbucket.org/blaze-lib/blaze.git blaze && \
+    cd blaze && git checkout v3.8 && cmake -DCMAKE_INSTALL_PREFIX=/usr/local/ && \
```
andrerenard: Why does blaze change from 3.4 to 3.8?   I think we want to keep it at 3.4 since the version change caused us some problems in the past with a change to the convergence algorithm making its timing less deterministic. At the moment only one system uses blaze for a very specific case.  So it might be better to keep the version the same between Ubuntu versions as well.  Unless of course 3.4 just won't build on 22.04?

jbmertens: 3.4 seems to build as does 3.8, but we can stick with the old version!

.github/workflows/main.yml
jbmertens: I've added a flag for now.

.github/workflows/main.yml
```diff
@@ -1,594 +1,762 @@
...
+        OPTS=(--rm --mount type=bind,src=$(pwd),target=/code/kotekan -w /code/kotekan/tests/ -e PYTHONPATH=/code/kotekan/python/ ${IMG_2204}:${{ needs.build-docker-2204.outputs.branch_name }})
+        docker run "${OPTS[@]}" bash -c "redis-server --daemonize yes; pytest -v -x --timeout=${PYTEST_TIMEOUT} -m serial"
+
+    - name: Run boost tests
+      if: env.RUN_2204_BOOST_TESTS # These aren't working, so for now only run them upon request.
+      run: |
+        OPTS=(--rm --mount type=bind,src=$(pwd),target=/code/kotekan -w /code/kotekan/build-2204/tests/ -e PYTHONPATH=/code/kotekan/python/ ${IMG_2204}:${{ needs.build-docker-2204.outputs.branch_name }})
+        docker run "${OPTS[@]}" pytest -v -x --timeout=${PYTEST_TIMEOUT}
+
+  # Build a full CHIME version of kotekan
```
andrerenard: I don't think we need this.  CHIME for now is 18.04 only.

build/.gitignore
```diff
@@ -1,3 +0,0 @@
-*
-
-!.gitignore
```
andrerenard, line 3: Can you put this back in?

lib/utils/LinearAlgebra.hpp
andrerenard: Let's leave the old syntax.

tests/test_vistestpattern.py
```diff
@@ -272,6 +273,9 @@ def test_pattern_no_noise_freq(tmpdir_factory):
     )
 
 
+@pytest.mark.skipif(
```
andrerenard, line 276: Maybe add a comment about this being related to comet.

tools/docker/22.04/Dockerfile
```diff
@@ -0,0 +1,184 @@
...
+
+RUN git clone https://github.com/kmsmith137/gputils.git \
+    && cd gputils \
+    && make install
+
+RUN git clone https://github.com/kmsmith137/n2k.git \
+  && cd n2k \
+  && make install
+
+RUN apt install -y --no-install-recommends \
```
andrerenard, line 115: Maybe just add a comment we should change this to ROCm if possible.

andrerenard (changes requested):
Looks great, just a few small changes and we can merge.

tools/docker/22.04/Dockerfile
dstndstn: My understanding is that the `rocm/dev-ubuntu-22.04` image is only available with Rocm versions >= 5.3, which don't work with the CHIME code.
https://hub.docker.com/r/rocm/dev-ubuntu-22.04/tags

So we need to use the older `rocm/dev-ubuntu-18.04` image, which supports much older Rocm versions (eg 3.7).

jbmertens: Thanks -- going to leave this as-is in this PR.

andrerenard (approved):
Looks great!  Awesome to be making progress on an improved CI system.   Let's just remember to open a ticket on the comet issues, and also open a ticket to resolve the clang-build issues.  I think a lot of us use clang, so it would be nice to get that build test working.

andrerenard:
Going to override the status checks, because it seems it's expecting the old checks to still pass....

PR #1110, opened 2023-08-10
[IGNORE] add a Dockerfile for building kotekan in the CHORD era

dstndstn:
The dockerfile contains prerequisites (including cuda and dpdk) for building CHORD-era kotekan.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -39,3 +39,12 @@ RUN git clone https://github.com/kmsmith137/n2k.git \
   && make install
 
 RUN apt install -y --no-install-recommends libdpdk-dev && apt clean
+
+RUN apt install -y --no-install-recommends \
+  libopencl-clang-dev opencl-headers ocl-icd-opencl-dev \
+  libopenblas-dev liblapack-dev liblapacke-dev \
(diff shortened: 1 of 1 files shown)


dstndstn:
Pushed to dockerhub as dstndstn/kotekan-docker.  It's a jumbo image (the CUDA layer, which pulls in all the X libraries, is like 8 GB).

dstndstn:
merge #1109 instead, and then I'll pull anything useful out of this PR and add it on top.

I should just close this PR, but I don't want to forget about it entirely!

jbmertens:
Hi @dstndstn - can you tell if the dockerfile in #1109 works for you? I believe I have added in the libraries, etc, from this PR. But, the only GPU-related test I've done is making sure nvidia-smi runs inside docker on Ruby.
(Dockerfile: https://github.com/kotekan/kotekan/blob/5859eda9b16bde93248026dd258a39f5e368b9cc/tools/docker/22.04/Dockerfile )

dstndstn:
Hi @jbmertens , have you built this docker container and pushed it to Docker hub?  Or do I need to build it before testing? :)

PR #1108, opened 2023-08-10
Cuda kernels & Kotekan stages from Dustin's MVP work

dstndstn:
This follows on from #1107 (and includes all commits from #1107 !), adding the Cuda kernels and MVP config files:

- FRB beamformer
- upchannelizer (two sample lengths)
- FRB beam reformer (FRB2 kernel)
- quantizer (FRB3 kernel)
- rechunker (goes between FRB2 and FRB3 kernels)
- chord MVP setup (sets up buffer views to hack our way around some data-length differences)

PR #1111, opened 2023-08-22
Add a CHORD metadata object, and minimally make buffer.[ch] into C++

dstndstn:
This builds on top of #1108 and pulls in changes bringing it equal to the dstn/frb-bf branch.

Adds a (sketch of a) CHORD metadata struct that tracks array shapes and sizes with a simple, statically-sized struct.  Adds plumbing to copy metadata objects into the CUDA GPU world.  (To-do: copy metadata *out* of the CUDA world also!)


lib/core/errors.h
```diff
@@ -132,7 +132,7 @@ void set_error_message_f(const char* format, ...);
     {                                                                                              \
         ERROR_F(m, ##a);                                                                           \
         set_error_message_f(m, ##a);                                                               \
-        exit_kotekan(1);                                                                           \
+        exit_kotekan(ReturnCode::FATAL_ERROR);                                     \
```
andrerenard: Should we also "promote" other `.c/.h` files while we are at it?   I think the buffers were the last thing that needed any pure C code.

config/tests/rechunk.yaml
```diff
@@ -0,0 +1,117 @@
+##########################################
+#
```
andrerenard: Looks like this made it back in.

lib/cuda/cudaCommand.cpp
```diff
@@ -369,3 +369,13 @@ cudaEvent_t cudaCommand::record_end_event(int gpu_frame_id) {
...
+
+struct chordMetadata* cudaCommand::get_chord_metadata(struct metadataContainer* mc) {
+    if (!mc)
+        return nullptr;
+    if (strcmp(mc->parent_pool->type_name, "chordMetadata")) {
+        FATAL_ERROR("cudaCommand: expected metadata to be type \"chordMetadata\", got \"{:s}\".",
+                    mc->parent_pool->type_name);
+    }
+    return static_cast<struct chordMetadata*>(mc->metadata);
+}
```
andrerenard: This should live in another place.  Perhap in the metadata file itself.

lib/cuda/cudaInputData.cpp
```diff
@@ -70,11 +70,20 @@ cudaEvent_t cudaInputData::execute(cudaPipelineState& pipestate,
                                   start_events[pipestate.gpu_frame_id],
                                   end_events[pipestate.gpu_frame_id]);
 
+    // Copy (reference to) metadata also
+    // FIXME -- need to lock here?
```
andrerenard: I believe the answer here is no to needing a lock.  Since you've waited for a full frame in the last function, it already must have (or not have) a metadata associated with the frame. 

As long as you complete your metadata pass/copy before you release the frame, it should be ok.  There is a bit of an issue that metadata isn't globally locked (meaning odd things could happen with multiple consumer pipelines if doing inplace changes, but that's a rare case, and out of scope here).

lib/cuda/cudaOutputData.cpp
```diff
@@ -95,6 +96,25 @@ cudaEvent_t cudaOutputData::execute(cudaPipelineState& pipestate,
...
+            device.get_gpu_memory_array_metadata(_gpu_mem, pipestate.gpu_frame_id);
+        if (meta) {
+            // Attach the metadata to the host buffer frame
+            CHECK_ERROR_F(pthread_mutex_lock(&output_buffer->lock));
+            if (output_buffer->metadata[output_buffer_execute_id] == NULL) {
+                DEBUG("Passing metadata from GPU array {:s}[{:d}] to output buffer {:s}[{:d}]",
+                      _gpu_mem, pipestate.gpu_frame_id, output_buffer->buffer_name,
+                      output_buffer_execute_id);
+                output_buffer->metadata[output_buffer_execute_id] = meta;
+                // .... where is this gonna get freed?
```
andrerenard: When the buffer resets a frame it decrements the reference counter, so the next buffer (or one of the following ones should release it).   As long as you haven't increased the reference counter too much.   But if you do you'll know quickly, since you'll run out of metadata frames to use and the pipeline will crash.

lib/cuda/cudaOutputData.cpp
```diff
@@ -103,17 +123,32 @@ void cudaOutputData::finalize_frame(int frame_id) {
...
+            DEBUG("Passing metadata from input (host) buffer {:s}[{:d}] to {:s}[{:d}]",
+                  in_buffer->buffer_name, in_buffer_id, output_buffer->buffer_name,
+                  output_buffer_id);
+            pass_metadata(in_buffer, in_buffer_id, output_buffer, output_buffer_id);
+            got_metadata = true;
+        }
     } else {
-        DEBUG("Add metadata? from input {:s}[{:d}] to {:s}[{:d}]", in_buffer->buffer_name,
-              in_buffer_id, output_buffer->buffer_name, output_buffer_id);
+        // ?? does this happen?
```
andrerenard: I'm confused by this.  As far as I can tell `got_metadata` will always be `false` at the start of this function call...

andrerenard: I'm also not sure why it's a member variable, I think it's a control variable in just this function...

lib/cuda/cudaUpchannelize.cpp
```diff
@@ -122,6 +123,29 @@ cudaEvent_t cudaUpchannelize::execute(cudaPipelineState& pipestate,
...
+            _gpu_mem_output_voltage, pipestate.gpu_frame_id, mc->parent_pool);
+        struct chordMetadata* meta_out = get_chord_metadata(mc_out);
+        struct chordMetadata* meta_in = get_chord_metadata(mc);
+        chord_metadata_copy(meta_out, meta_in);
+        INFO("cudaUpchannelize: input array shape: {:s}", meta_in->get_dimensions_string());
+        assert(meta_in->get_dimension_name(0) == "T");
+        assert(meta_in->get_dimension_name(2) == "F");
+        meta_out->dim[0] /= _upchan_factor;
+        meta_out->dim[2] *= _upchan_factor;
+        INFO("cudaUpchannelize: output array shape: {:s}", meta_out->get_dimensions_string());
```
andrerenard: Should this be `DEBUG`?   (Above as well)

lib/gpu/gpuDeviceInterface.hpp
```diff
@@ -87,6 +91,38 @@ class gpuDeviceInterface : public kotekan::kotekanLogging {
                                  const std::string& view_name, const size_t view_offset,
                                  const size_t view_len);
 
+    /**
+     * @brief Fetches the metadata (if any) attached to the given GPU
```
andrerenard, line 96: Should likely add the `@param` and `@return` doxygen comments, since it's likely people will be looking up what these do at some point.

lib/gpu/gpuDeviceInterface.cpp
```diff
@@ -108,7 +111,87 @@ void gpuDeviceInterface::create_gpu_memory_array_view(const std::string& source_
...
+    // Make sure the slot is empty
+    if (gpu_memory[name].metadata_pointers[index]) {
+        FATAL_ERROR("claim_gpu_memory_array_metadata for name \"{:s}\"[:d]: slot is not empty.",
+                    name, index);
+    }
+    increment_metadata_ref_count(mc);
+    gpu_memory[name].metadata_pointers[index] = mc;
+}
+
+void gpuDeviceInterface::release_gpu_memory_array_metadata(const std::string& name,
```
andrerenard, line 188: I think we need to add mutex protection to all these functions.  For allocating memory currently we always have just one thread doing the precondition and execute.   However, now some of these functions operate in finalize, which is another thread. 

And of course once we support multiple pipelines, many of the gpuDevice functions will need to become threadsafe.

lib/metadata/chordMetadata.hpp
```diff
@@ -0,0 +1,127 @@
...
+const int CHORD_META_MAX_FREQ = 16;
+
+// Maximum number of dimensions for arrays
+const int CHORD_META_MAX_DIM = 10;
+
+// Maximum length of dimension names for arrays
+const int CHORD_META_MAX_DIMNAME = 16;
+
+struct chordMetadata {
+    struct chimeMetadata chime;
```
andrerenard: We likely want to make this a public inheritance, but I don't think we want to include all of `chimeMetadata` in CHORD, so some refactoring will be necessary.

lib/stages/testDataGen.cpp
```diff
@@ -210,6 +222,52 @@ void testDataGen::main_thread() {
...
+                            // Arbitrarily number the frequency channels...
+                            chordmeta->coarse_freq[i] = i;
+                            chordmeta->freq_upchan_factor[i] = 1;
+                            int64_t fpgacount = frame_id_abs * ntime;
+                            chordmeta->half_fpga_sample0[i] = 2 * fpgacount;
+                            chordmeta->time_downsampling_fpga[i] = 1;
+                        }
+                    }
+
+                    INFO("Chord metadata: array shape {:s}", chordmeta->get_dimensions_string());
```
andrerenard: `DEBUG`?

andrerenard (changes requested):

lib/core/errors.h
dstndstn: I filed #1127 :)

config/tests/rechunk.yaml
dstndstn: Re-killed

lib/cuda/cudaCommand.cpp
dstndstn: Good call, done

lib/metadata/chordMetadata.hpp
andrerenard: This might be something that can become a ticket.

lib/cuda/cudaInputData.cpp
dstndstn: Thanks!  comment removed

lib/cuda/cudaOutputData.cpp
dstndstn: Thanks, and I did check that it doesn't explode in a long run :)

lib/cuda/cudaUpchannelize.cpp
dstndstn: Yes, done!

lib/gpu/gpuDeviceInterface.hpp
dstndstn: Done

lib/gpu/gpuDeviceInterface.cpp
dstndstn: Thanks for the comment, done!

lib/metadata/chordMetadata.hpp
dstndstn: Indeed ;)  https://github.com/kotekan/kotekan/issues/1128

lib/stages/testDataGen.cpp
dstndstn: Done!

lib/cuda/cudaOutputData.cpp
```diff
@@ -103,17 +123,32 @@ void cudaOutputData::finalize_frame(int frame_id) {
...
-        DEBUG("Passing metadata from input {:s}[{:d}] to {:s}[{:d}]", in_buffer->buffer_name,
-              in_buffer_id, output_buffer->buffer_name, output_buffer_id);
-        pass_metadata(in_buffer, in_buffer_id, output_buffer, output_buffer_id);
-        got_metadata = true;
+        if (in_buffer) {
+            DEBUG("Passing metadata from input (host) buffer {:s}[{:d}] to {:s}[{:d}]",
+                  in_buffer->buffer_name, in_buffer_id, output_buffer->buffer_name,
+                  output_buffer_id);
+            pass_metadata(in_buffer, in_buffer_id, output_buffer, output_buffer_id);
+            got_metadata = true;
```
andrerenard: As far as I can tell this variable, `got_metadata`, is not needed.

andrerenard (changes requested):
Thanks for the review changes!  There is just one comment around a variable that I think should be removed.

lib/cuda/cudaOutputData.cpp
dstndstn: Good catch, that was vestigal

andrerenard (approved):
Looks like cmake-lint wants a silly change, but otherwise this is good to merge!

dstndstn:
Hah, so I just finished remove approximately 1,000,000 uses of `struct Buffer` from the codebase... just compiling and checking that, then I'll add that commit to this PR, if that's ok.

dstndstn:
Okay, that's in, it compiles (ship it!).  I'll be in in about 30 minutes.

PR #1114, opened 2023-08-25
Support Julia

eschnett:
Add basic support for Julia


eschnett:
This is a draft. I still need to add Julia to the CI Docker images and test Julia there.

jbmertens:
> This is a draft. I still need to add Julia to the CI Docker images and test Julia there.

@eschnett There is a julia version in the 22.04 docker image now, courtesy @dstndstn,
https://github.com/kotekan/kotekan/blob/develop/tools/docker/22.04/Dockerfile#L154
Will that suffice?

lib/core/juliaManager.hpp
jbmertens, line 1: Could be useful to change this documentation to doxygen-style comments

lib/stages/juliaHelloWorld.cpp
jbmertens, line 1: Could also add some doxygen doc comments here

jbmertens (changes requested):

lib/stages/juliaHelloWorld.cpp
eschnett: Done.

lib/core/juliaManager.hpp
eschnett: Done.

lib/stages/juliaHelloWorld.cpp
eschnett: Done, this time for real.

eschnett:
@jbmertens This is ready for a re-review.

cmake/FindJulia.cmake
```diff
@@ -0,0 +1,193 @@
+# This FindJulia.cmake from
+# <https://github.com/JuliaInterop/libcxxwrap-julia/blob/main/FindJulia.cmake>
+
+# Original FindJulia.cmake from <https://github.com/QuantStack/xtensor-julia-cookiecutter/
```
andrerenard, line 4: Since this is an external file we are pulling in with a different (if somewhat confused license history), we should indicate that in our `LICENSE` file.

andrerenard (changes requested):
LGTM!   Just one inline comment, and the doxygen comment update from Jim, and we can merge.

jbmertens (approved):

cmake/FindJulia.cmake
eschnett: Done.

PR #1118, opened 2023-08-30
possible fix for publish image workflow

jbmertens:


andrerenard (approved):
LGTM!  Let's see if it works.

PR #1119, opened 2023-09-11
Add ASDF library to Docker image

eschnett:


eschnett:
This seems to work. CI fails because it cannot upload the newly generated Docker image.

jbmertens (approved):
There's a permissions issue when pushing the docker image to look into, but it builds fine.

jbmertens:
The docker build seems to be working, but has a permissions issue when trying to publish the image. That's an issue with the test though, not this PR, so merging anyways.

Issue #1034, opened 2022-07-15
Random persistent failures of `test_eigenvis.py` 


andrerenard:
For reasons I don't yet understand `test_eigenvis.py` has been failing the `test_basic()` test with the following error when run in github actions.   It doesn't happen with every PR, but once it happens re-runs of the test don't resolve the problem.  The failure signature always looks like this, always with the same values: 
```
=================================== FAILURES ===================================
__________________________________ test_basic __________________________________
[gw1] linux -- Python 3.7.9 /usr/bin/python3.7
tmpdir_factory = TempdirFactory(_tmppath_factory=TempPathFactory(_given_basetemp=PosixPath('/tmp/pytest-of-root/pytest-0/popen-gw1'), _...pluggy._tracing.TagTracerSub object at 0x7f9ffdc35c10>, _basetemp=PosixPath('/tmp/pytest-of-root/pytest-0/popen-gw1')))
    def test_basic(tmpdir_factory):
        params = default_params
        num_elements = params["num_elements"]
        expected_evec_phase_fact = np.exp(1j * np.arange(num_elements))
        eigen_data = run_eigenvis(tmpdir_factory)
        for frame in eigen_data:
            largest_eval = frame.eval[0]
            largest_evec = frame.evec[:num_elements]
            assert abs(largest_eval - num_elements) / num_elements < 1e-6
            assert np.allclose(largest_evec / largest_evec[0], expected_evec_phase_fact)
            assert np.allclose(abs(largest_evec), 1 / np.sqrt(num_elements))
            zero_eval = frame.eval[1]
            assert zero_eval / num_elements < 1e-6
>           assert frame.erms < 1e-6
E           assert array([1.2018842e-06], dtype=float32) < 1e-06
E            +  where array([1.2018842e-06], dtype=float32) = <kotekan.visbuffer.VisBuffer object at 0x7f9ff1944450>.erms
test_eigenvis.py:68: AssertionError
```
It's been hard to track down, since I cannot reproduce it on my local system.   The test also doesn't seem to involve any random number generation (unless I'm missing some randomization component in the eigenvalue solver).   This makes me think it might be some kind of memory problem, with something not initializing correctly when run in the github actions environment.

jrs65:
Yeah, this is definitely not one of the random statistical tests so it should be entirely deterministic. I wonder if the underlying linear algebra libraries have got changed somehow. Has Github updated its ubuntu-latest image?

andrerenard:
[...]

PR #1129, opened 2023-09-13
make eigenvector rms tolerance limit more lenient

jbmertens:
This should help with the failing eigenvis test on the 18.04 build, #1034 . Accumulated errors may be a bit larger than the 1e-6 tolerance due to adding multiple eigenvectors in the loop here,
https://github.com/kotekan/kotekan/blob/4118a52a2e7b89636274ce4824df0554cb488fa2/lib/stages/eigenVis.cpp#L217
this change is just an attempt to account for that.

### DIFF
--- a/tests/test_eigenvis.py
+++ b/tests/test_eigenvis.py
@@ -49,6 +49,7 @@ def run_eigenvis(tdir_factory, params=None):
     test.run()
     return dump_buffer.load()
 
+
 def test_basic(tmpdir_factory):
 
     params = default_params
--- a/tests/test_eigenvis.py
+++ b/tests/test_eigenvis.py
@@ -7,7 +7,6 @@ from future.builtins.disabled import *  # noqa  pylint: disable=W0401, W0614
 
 import pytest
 import numpy as np
-import sys
 
 from kotekan import runner
 
@@ -57,7 +56,6 @@ def test_basic(tmpdir_factory):
     expected_evec_phase_fact = np.exp(1j * np.arange(num_elements))
 
     eigen_data = run_eigenvis(tmpdir_factory)
-
     for frame in eigen_data:
         largest_eval = frame.eval[0]
         largest_evec = frame.evec[:num_elements]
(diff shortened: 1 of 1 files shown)


andrerenard (approved):
I'm going to merge this now to get the CI passing.  @jrs65 If you think something else should be done, we can revert.

jrs65:
I'm at a conference at the moment so I haven't had time to look at this carefully.

I don't think there is a good explanation for why these failures are happening in some locations and not in others, this just seems to bump the tolerance somewhat arbitrarily without great justification. If it was truly an error accumulation the relaxation to apply would be sqrt(N_eval).

I would suggest changing this to be more honest and just relax the tolerance to 1e-5, *and* add a note to the test to explain that it's to deal with sporadic failures.

jbmertens:
sqrt(N_eval) is related to the variance of the accumulated error, but worst case they could just add, right?

The near-zero eigenvalues seem to be different near-zero values on every system. I'm not sure how much of that is down to hardware, library versions, or something else. But they do at least seem consistently within tolerance.

jrs65:
The reason I think it's hard to apply a lot of this reasoning is that it constructs erms by subtracting the rank-N approximation from the matrix and then summing the squared residuals. In many ways this is not a great way to construct this quantity, as you are subtracting two large numbers. This is likely to be numerically quite noisy, and while it's not biasing the residuals, it probably does boost the squared residuals. However, I don't think there's any clear reasoning why this should depend on N_eval, and not, for instance, the size of the matrix itself.

My proposal is that we just admit we don't really understand it, but that by scaling by that much we find it eliminate the failures on most platforms.

jbmertens:
Ah I see. Yeah maybe there is no ideal solution, just a tradeoff between the test tolerance and actually catching an issue. If you're fine with 1e-5 I'll use that, and add a comment.

jrs65:
Great, thanks! @jbmertens 

FWIW I'm pretty sure I just made up the 1e-6 threshold in the first place, so it's not surprising we would need to tweak it, but I'm still a little concerned about not understanding the lack of consistency on the test. Anyway, that's one for the future.

PR #1131, opened 2023-09-14
hdf5FileWrite: New output method

eschnett:


jbmertens:
I'll am to work through this shortly, @eschnett . One immediate question, did you mean to include the FEngine code in this PR? Fine with me either way, just wanted to check given the PR title.

eschnett:
No, there should not be any F-engine code in there. I'll clean it up, thanks!

lib/stages/hdf5FileWrite.cpp
```diff
@@ -0,0 +1,240 @@
...
+#include "bufferContainer.hpp" // for bufferContainer
+#include "chimeMetadata.hpp"   // for chimeMetadata
+#include "chordMetadata.hpp"   // for chordMetadata
+#include "errors.h"
+#include "kotekanLogging.hpp"    // for ERROR, INFO
+#include "metadata.h"            // for metadataContainer
+#include "prometheusMetrics.hpp" // for Metrics, Gauge
+#include "visUtil.hpp"           // for current_time
+
+#undef NDEBUG
```
jbmertens: Just want to make sure this is here intentionally!

lib/stages/hdf5FileWrite.cpp
```diff
@@ -0,0 +1,240 @@
...
+    _prefix_hostname = config.get_default<bool>(unique_name, "prefix_hostname", true);
+    _exit_after_n_frames = config.get_default<uint32_t>(unique_name, "exit_after_n_frames", 0);
+    _exit_with_n_writers = config.get_default<uint32_t>(unique_name, "exit_with_n_writers", 0);
+}
+
+hdf5FileWrite::~hdf5FileWrite() {}
+
+void hdf5FileWrite::main_thread() {
+    std::string full_path;
+    hid_t fd = -1;
```
jbmertens: Maybe H5I_UNINIT instead of -1 here

lib/stages/hdf5FileWrite.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+        // This call is blocking.
+        frame = wait_for_full_frame(buf, unique_name.c_str(), frame_id);
+        if (frame == nullptr)
+            break;
+
+        // Start timing the write time
+        const double st = current_time();
+
+        if (fd < 0) {
```
jbmertens, line 78: Could check for H5I_UNINIT instead of < 0 here.

lib/stages/hdf5FileWrite.cpp
```diff
@@ -0,0 +1,240 @@
...
+                char hostname[256];
+                gethostname(hostname, sizeof hostname);
+                sbuf << hostname << "_";
+            }
+            sbuf << _file_name << ".h5";
+            full_path = sbuf.str();
+
+            fd = H5Fcreate(full_path.c_str(), H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
+
+            if (fd < 0) {
```
jbmertens, line 91: Could check for H5I_BADID here. (& Similarly, later)

jbmertens (changes requested):
Mostly good to me! Just a couple small items to consider, not really change requests.

lib/stages/hdf5FileWrite.cpp
eschnett: I replaces all the asserts by proper error messages.

lib/stages/hdf5FileWrite.cpp
eschnett: Done.

lib/stages/hdf5FileWrite.cpp
eschnett: I'd rather not, checking for any negative number seems to include more possible error cases.

jbmertens (approved):

jbmertens:
@andrerenard wanted to leave a bit of feedback, just waiting for that before merging.

lib/stages/hdf5FileWrite.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include "Config.hpp"
+#include "Stage.hpp" // for Stage
+#include "bufferContainer.hpp"
+
+#include <stdint.h> // for uint32_t
+#include <string>   // for string
+
+/**
+ * @class hdf5FileWrite
+ * @brief Stream a buffer to disk.
```
andrerenard, line 13: Given this stage in limited to writing just one file, I'm tempted to either move it to `lib/testing/`, and/or make it clear in the comments this stage is used to run tests, and not to be used in a production pipeline (where we'd have to cycle file names).

andrerenard (changes requested):
Thanks for adding this!   Other than the things we need to change with larger system level changes (e.g. having "final" signals), as a testing function this is good to merge.   However, I'd indicate in some way this isn't a stage to use in a production pipeline.

lib/stages/hdf5FileWrite.hpp
eschnett: I added a comment.

jbmertens (approved):

andrerenard (approved):

PR #1133, opened 2023-09-14
Refactor cudaCommands so that each instance is responsible for running a single frame of data at a time

dstndstn:
- add `gpuCommandState` class to hold state between instances of the class
- add `int instance_number` and a (shared ptr to) `gpuCommandState` to the `gpuCommand` constructor.
- for each `gpuCommand` appearance in a pipeline, `_gpu_buffer_depth` instances of the class will be created, with `instance_num` counting up from zero.  A `gpuCommandState` object, if used by the class, will be created and passed to each constructor
- remove the `gpu_frame_id` arg from `gpuCommand` methods `wait_on_precondition`, `execute`, `pre_execute`, and `finalize_frame`
- (set a `gpu_frame_id` member variable in a new `gpuCommand.start_frame(int gpu_frame_id)` call)
- move cuda `build` and `build_ptx` to `cudaDeviceInterface`
- change `gpuDeviceInterface.get_gpu_memory*` so that the `gpu_device_frame` does not have to be mod'ed by the caller


dstndstn:
List of things that could be pulled out into cudaCommandState:
* cudaFRBBeamReformer : cublasHandle_t

To-do things that are still arrays
* "host_info" arrays -- these are going to get refactored anyway in the near future, so not super interested in fixing here
* cudaFRBBeamformer : host_length
* cudaOutputDataZero -- but couldn't we use a cudaMemset instead?

In OpenCL:
* build() --> move to device?  Or move to clCommandState.
* clOutputDataZero, clPresumZero: shared block of zero mem
* clPreseedKernel, clCorrelatorKernel: defineOutputDataMap shared

For the HSA directory, I basically made the minimal adjustment to the existing code to get it building, without actually using this new structure.

No changes in HIP, it's certainly broken.

dstndstn:
I think this might actually be ready for review!

andrerenard (changes requested):
At a high level, for the HSA code, I think we'll need to make a copy of the old `gpu/` objects, rename them something like `gpu*Old` and then have the HSA code inherit from that.  It's a bit ugly, but there is a lot of risk and very little value to upgrading that to the newer data flow.

dstndstn:
Unless I missed something, I think the HSA code should still work as is.  I basically just added in some extra shim functions that adapt from the old gpu/ method signatures to what is in the hsa/ gpu commands.

[narrator: he did indeed miss something!]

lib/gpu/gpuProcess.cpp
```diff
@@ -152,23 +155,27 @@ void gpuProcess::main_thread() {
     bool first_run = true;
 
     while (!stop_thread) {
-        int gpu_frame_id = gpu_frame_counter % _gpu_buffer_depth;
+        int icommand = gpu_frame_counter % _gpu_buffer_depth;
```
andrerenard: I'm not sure this works in the HSA case?   Since there we would only be creating one command per gpu buffer frame.

lib/gpu/gpuProcess.cpp
```diff
@@ -152,23 +155,27 @@ void gpuProcess::main_thread() {
...
+        for (auto& command : commands)
+            command[icommand]->start_frame(gpu_frame_counter);
+
         // Wait for all the required preconditions
         // This is things like waiting for the input buffer to have data
         // and for there to be free space in the output buffers.
         // INFO("Waiting on preconditions for GPU[{:d}][{:d}]", gpu_id, gpu_frame_id);
         for (auto& command : commands) {
-            if (command->wait_on_precondition(gpu_frame_id) != 0) {
+            if (command[icommand]->wait_on_precondition() != 0) {
```
andrerenard: I think the logic here needs to be adjusted to support HSA, since there we only have one command object per-command.  We could copy the same pointer into each command, or we could abstract it to a function like `queue_commands()` which is different for HSA than for CUDA/OpenCL.

lib/gpu/gpuProcess.cpp
```diff
@@ -199,18 +206,19 @@ void gpuProcess::results_thread() {
...
+    int gpu_frame_counter = 0;
 
     while (true) {
         // Wait for a signal to be completed
-        DEBUG2("Waiting for signal for gpu[{:d}], frame {:d}, time: {:f}", gpu_id, gpu_frame_id,
-               e_time());
-        if (final_signals[gpu_frame_id]->wait_for_signal() == -1) {
+        DEBUG2("Waiting for signal for gpu[{:d}], frame {:d}, time: {:f}", gpu_id,
+               gpu_frame_counter, e_time());
+        int icommand = gpu_frame_counter % _gpu_buffer_depth;
```
andrerenard: Same concern here.

andrerenard (changes requested):
I could be wrong, but I don't think the logic in the base class `gpuProcess` reduces to the previous logic in the way it calls the command objects for precondition and finalize for the HSA case.  Although I might also be missing something.  And I agree, it seems like it might work to refactor things such that both HSA and CUDA/OpenCL work under the same base-class.

lib/gpu/gpuProcess.cpp
dstndstn: oh yes, you're right, of course.

Hmm, let's see, I could mod by the actual number of command objects in each vector, since currently it's up to the gpuProcess subclass to create that array of commands in `create_command`, and currently `main` is just assuming that it created `_gpu_buffer_depth` of them...

dstndstn: Or hsaProcess could just store `_gpu_buffer_depth` copies of each command's shared_ptr (oh, except they're not shared_ptrs, darn!) ;)

But maybe I should do something like this,

```
        for (auto& command : commands) {
            int ic = gpu_frame_counter % command.size();
            command[ic]->start_frame(gpu_frame_counter);
        }
```

ie, round-robin the set of objects created for each command.  That probably makes more sense anyway, rather than having gpuProcess assume there are always _gpu_buffer_depth of them!

dstndstn: okay, pushed a commit for that, I think it's better.  I'm going to "resolve" your comments, but feel free to protest!

lib/core/kotekanTrackers.cpp
```diff
@@ -98,16 +98,18 @@ std::shared_ptr<StatTracker> KotekanTrackers::add_tracker(std::string stage_name
...
 
-    if (trackers.count(stage_name) && trackers[stage_name].count(tracker_name)) {
-        ERROR_NON_OO("Duplicate tracker name: {:s}:{:s}. Exiting.", stage_name, tracker_name);
-        throw std::runtime_error(
-            fmt::format(fmt("Duplicate tracker name: {:s}:{:s}"), stage_name, tracker_name));
+    auto val1 = trackers.find(stage_name);
+    if (val1 != trackers.end()) {
+        auto val2 = val1->second.find(tracker_name);
+        if (val2 != val1->second.end())
+            // Exists, return existing tracker.
```
andrerenard, line 107: Update the header description with this change.

lib/cuda/cudaCommand.hpp
```diff
@@ -42,6 +42,15 @@ class cudaPipelineState : public kotekan::kotekanLogging {
...
 };
 
+class cudaCommandState : public gpuCommandState {
+public:
+    cudaCommandState(kotekan::Config&, const std::string&, kotekan::bufferContainer&,
+                     cudaDeviceInterface&) {}
+};
+
+// use this to avoid having to write "std::shared_ptr<cudaCommandState>()"
+extern std::shared_ptr<cudaCommandState> no_cuda_state;
```
andrerenard: I think this should be named `no_cuda_command_state`, I don't like that it's longer, but   `no_cuda_state` makes me think this is for something else.

lib/cuda/cudaCommand.hpp
```diff
@@ -117,15 +116,15 @@ class cudaCommand : public gpuCommand {
...
-    void record_start_event(int gpu_frame_id);
+    void record_start_event();
 
     // For subclassers to call to create & record GPU ending events.
-    cudaEvent_t record_end_event(int gpu_frame_id);
+    cudaEvent_t record_end_event();
 
     /// Events queued after the kernel/copy for synchronization and profiling
-    cudaEvent_t* end_events;
+    cudaEvent_t end_event;
```
andrerenard, line 128: I like this simplification!

lib/gpu/gpuCommand.hpp
```diff
@@ -53,35 +56,41 @@ class gpuCommand : public kotekan::kotekanLogging {
      * @param default_kernel_file_name  (optional) external file (e.g. CL) used by a command
      */
     gpuCommand(kotekan::Config& config, const std::string& unique_name,
-               kotekan::bufferContainer& host_buffers, gpuDeviceInterface& device,
+               kotekan::bufferContainer& host_buffers, gpuDeviceInterface& device, int instance_num,
+               std::shared_ptr<gpuCommandState> = std::shared_ptr<gpuCommandState>(),
```
andrerenard: Missing documentation above for these new parameters.

lib/cuda/cudaCommand.hpp
```diff
@@ -134,16 +133,25 @@ class cudaCommand : public gpuCommand {
...
+CREATE_FACTORY(cudaCommandState, kotekan::Config&, const std::string&, kotekan::bufferContainer&,
                cudaDeviceInterface&);
+
 #define REGISTER_CUDA_COMMAND(newCommand)                                                          \
     REGISTER_NAMED_TYPE_WITH_FACTORY(cudaCommand, newCommand, #newCommand)
 
+#define REGISTER_CUDA_COMMAND_WITH_STATE(newCommand, newCommandState)                              \
+    REGISTER_NAMED_TYPE_WITH_FACTORY_VARIANT(state, cudaCommand, newCommand, #newCommand);         \
+    REGISTER_NAMED_TYPE_WITH_FACTORY(cudaCommandState, newCommandState, #newCommand)
+
```
andrerenard, line 180: This many factory logic feels complex enough we should write some solid documentation about how (and why) it works the way it does.  Perhaps that can go in the base class so we don't repeat for every subclass?

lib/cuda/cudaCommand.hpp
```diff
@@ -108,7 +107,7 @@ class cudaCommand : public gpuCommand {
     /** Releases the memory of the event chain arrays per buffer_id
      * @param gpu_frame_id    The bufferID to release all the memory references for.
```
andrerenard: Update comment.

lib/opencl/clCommand.hpp
```diff
@@ -30,12 +30,20 @@
...
 
 
+class clCommandState : public gpuCommandState {
+public:
+    clCommandState(kotekan::Config&, const std::string&, kotekan::bufferContainer&,
+                   clDeviceInterface&) {}
+};
+
+// use this to avoid having to write "std::shared_ptr<clCommandState>()"
+extern std::shared_ptr<clCommandState> no_cl_state;
```
andrerenard: Same comment here re: possible name confusion.

andrerenard (changes requested):
Not completely done double checking the changes in all the files, but the overall the big picture logic changes look really good!    Some quick comments inline.

lib/core/kotekanTrackers.cpp
dstndstn: Done!

lib/cuda/cudaCommand.hpp
dstndstn: Done!

lib/cuda/cudaCommand.hpp
dstndstn: Me too!

lib/gpu/gpuCommand.hpp
dstndstn: Done!

lib/cuda/cudaCommand.hpp
dstndstn: Done!

lib/cuda/cudaCommand.hpp
dstndstn: Done!

lib/opencl/clCommand.hpp
dstndstn: Done!

lib/gpu/gpuCommand.hpp
```diff
@@ -22,6 +23,8 @@ class gpuDeviceInterface;
 /// Enumeration of known GPU command types.
 enum class gpuCommandType { COPY_IN, BARRIER, KERNEL, COPY_OUT, NOT_SET };
 
+class gpuCommandState : public kotekan::kotekanLogging {};
```
andrerenard, line 36: Might be good to have a comment on what this class is used for here at the base.

lib/cuda/cudaDeviceInterface.cpp
```diff
@@ -87,3 +90,230 @@ void cudaDeviceInterface::async_copy_gpu_to_host(void* dst, void* src, size_t le
     CHECK_CUDA_ERROR(cudaEventCreate(&copy_end_event));
     CHECK_CUDA_ERROR(cudaEventRecord(copy_end_event, getStream(cuda_stream_id)));
 }
+
+void cudaDeviceInterface::build(const std::string& kernel_filename,
+                                const std::vector<std::string>& kernel_names,
```
andrerenard, line 95: Given that it's easier now to accidentally call `build()` multiple times if someone forgets the `inst == 0` check, we should likely check if a kernel with that name already exists and throw an error if it does.    And genetically it's likely good to have that check in case two kernels end up with the same name.

lib/cuda/cudaDeviceInterface.cpp
```diff
@@ -87,3 +90,230 @@ void cudaDeviceInterface::async_copy_gpu_to_host(void* dst, void* src, size_t le
...
+            cuGetErrorString(err, &errStr);
+            FATAL_ERROR("ERROR IN cuModuleGetFunction for correlate: {}", errStr);
+        }
+        if (runtime_kernels[kernel_name] == nullptr) {
+            FATAL_ERROR("Failed to find kernel name \"{}\" in compiled PTX module", kernel_name);
+        }
+    }
+}
+
+void cudaDeviceInterface::build_ptx(const std::string& kernel_filename,
```
andrerenard, line 194: Same for this function.

lib/cuda/cudaOutputData.cpp
```diff
@@ -12,134 +12,123 @@ using kotekan::Config;
...
-                cudaHostRegister(output_buffer->frames[i], output_buffer->frame_size, 0));
+    if (instance_num == 0) {
+        register_producer(output_buffer, unique_name.c_str());
+        for (int i = 0; i < output_buffer->num_frames; i++) {
+            uint flags;
+            // only register the memory if it isn't already...
+            if (cudaErrorInvalidValue == cudaHostGetFlags(&flags, output_buffer->frames[i])) {
+                CHECK_CUDA_ERROR(
+                    cudaHostRegister(output_buffer->frames[i], output_buffer->frame_size, 0));
+            }
```
andrerenard: Not really a part of this PR, so feel free to leave for another time, but I think we can actually get rid of this logic.   It should be handled by gpuProcess.cpp:97 for all buffers.

andrerenard (approved):
A few small comments, but otherwise I think this is ready to merge.  I've gone over the logic a second time to make sure I understand it all, and I don't see any issues.  
My only general comment is the `if (inst == 0)` logic shows up in a few places, but it's maybe not entirely obvious to someone new why it's needed, and I've been trying to decide if we should have helper functions to hide it, or if that's just more confusing?   But for now I think it's sufficient to just add error checking in case someone forgets that, and merge this so we can start seeing how it works in wider testing.

lib/gpu/gpuCommand.hpp
dstndstn: Done!

lib/cuda/cudaDeviceInterface.cpp
dstndstn: Good idea, done!

lib/cuda/cudaDeviceInterface.cpp
dstndstn: Done

lib/cuda/cudaOutputData.cpp
dstndstn: Done!

PR #1134, opened 2023-09-15
Make chordMetadata const-safe

eschnett:
Also use `std::ostringstream` to construct strings.


dstndstn (approved):
This all looks great!

Issue #1117, opened 2023-08-30
CI improvements


jbmertens:
There are a few issues remaining with CI,

- [x] Readthedocs still complains plantuml isn't found
- [x] Boost tests are not found by pytest on the 22.04 build. This should be investigated
- [x] chime-experiment/comet (a dependency for some tests) relies on deprecated functionality, so does not work on 22.04. Some tests are skipped on 22.04 as a result. See: chime-experiment/comet#104
- [ ] IWYU might be broken?
- [ ] Handle PR changes from forks better (include a useful fail message or something?)

PR #1137, opened 2023-10-02
Github actions: add an OpenCL build

dstndstn:


dstndstn:
I just copied the `build-full-tests-2204` target to `build-opencl-2204` and changed the `cmake` command to turn off CUDA and turn on OpenCL.  Since the container has the required OpenCL libraries, this just worked!  Like the `full-tests` run, this runs the python parallel and Boost tests -- I don't know if we need those -- do they test anything that isn't tested in the `full-tests` build?

jbmertens:
This is great! There's now a node with a couple GPUs on it, along with a runner, on sky. I haven't had a chance to really test it yet, but this looks like a good first test. Could you change the runs-on to `runs-on: self-hosted`?

Also just a note, the boost tests haven't been running correctly on 22.04. Still need to investigate that.

.github/workflows/main.yml
```diff
@@ -284,6 +284,78 @@ jobs:
         OPTS=(--rm --mount type=bind,src=$(pwd),target=/code/kotekan -w /code/kotekan/build-2204/tests/ -e PYTHONPATH=/code/kotekan/python/ ${{ needs.build-docker-2204.outputs.docker_img_name }})
         docker run "${OPTS[@]}" pytest -v -x --timeout=${PYTEST_TIMEOUT}
 
+  # Build an OpenCL version of kotekan and run the unit tests
+  build-opencl-2204:
+    runs-on: ubuntu-latest
```
jbmertens, line 289: Change to `runs-on: self-hosted`

.github/workflows/main.yml
```diff
@@ -284,6 +284,78 @@ jobs:
...
+        OPTS=(--rm --mount type=bind,src=$(pwd),target=/code/kotekan -w /code/kotekan/tests/ -e PYTHONPATH=/code/kotekan/python/ ${{ needs.build-docker-2204.outputs.docker_img_name }})
+        docker run "${OPTS[@]}" pytest -v -n auto --dist=loadfile --timeout=${PYTEST_TIMEOUT} -x -m 'not serial'
+
+    - name: Run serial python tests
+      run: |
+        OPTS=(--rm --mount type=bind,src=$(pwd),target=/code/kotekan -w /code/kotekan/tests/ -e PYTHONPATH=/code/kotekan/python/ ${{ needs.build-docker-2204.outputs.docker_img_name }})
+        docker run "${OPTS[@]}" bash -c "redis-server --daemonize yes; pytest -v -x --timeout=${PYTEST_TIMEOUT} -m serial"
+
+    - name: Run boost tests
+      if: env.RUN_2204_BOOST_TESTS # These aren't working, so for now only run them upon request.
```
jbmertens, line 354: Just leaving a note to come back to this (#1117)

jbmertens (changes requested):

dstndstn:
Hi @jbmertens , sorry for not replying sooner!  The OpenCL will actually build on Github's CI infrastructure.  The tests won't do anything, of course :) but I think it would be useful to keep the build to make sure we don't break it (my current ticket was breaking it big time!).

jbmertens:
Ok, thanks @dstndstn! It looks like the build is failing on the local runner, which will take some debugging, but I'll address that in a separate PR. So this one looks good to go!

jbmertens (approved):

dstndstn:
Awesome, thank you!

PR #1140, opened 2023-10-12
(Rebased) Switch to Julia-generated kotekan C++ wrappers for CUDA kernels

dstndstn:
Erik's PR #1088 was to merge into one of my old devel branches that has since been merged to develop, so I created a new branch off of `develop` and then cherry-picked (most of) the commits from #1088.  There were a few conflicts, and in most cases I simply took Erik's changes.  There are couple of places where I wasn't sure whether that was the right thing or not, but I have code-review comments on each of those.


cmake/FindJulia.cmake
dstndstn, line 1: I don't know which of the FindJulia.cmake files we should be using!

config/tests/mvp.yaml
```diff
@@ -154,7 +166,7 @@ gpu:
     # log_level: debug
     profiling: true
     log_profiling: true
-    kernel_path: "../../lib/cuda/kernels/"
+    kernel_path: "lib/cuda/kernels"
```
dstndstn, line 169: We should have a convention for where the kernels are located.  This is relative to the cwd at kotekan runtime, so the `../..` prefix works well if you run from the `build/kotekan` directory, ie, if you run `./kotekan ....` then the `../..` relative paths for the kernels work out.

kotekan/kotekan.cpp
```diff
@@ -408,8 +408,7 @@ void start_new_kotekan_mode(Config& config, bool dump_config) {
 
 int main(int argc, char** argv) {
 
-    std::signal(SIGINT, signal_handler);
-    std::signal(SIGTERM, signal_handler);
+    std::signal(SIGHUP, signal_handler);
```
dstndstn, line 411: I did not really try to follow too carefully the whole SIG* changes, so please confirm that this is what you want!  Should this be both SIGINT and SIGHUP?

lib/cuda/cudaFRBBeamformer.cpp
```diff
@@ -1,280 +1,222 @@
...
+REGISTER_CUDA_COMMAND(cudaFRBBeamformer);
 
 cudaFRBBeamformer::cudaFRBBeamformer(Config& config, const std::string& unique_name,
                                      bufferContainer& host_buffers, cudaDeviceInterface& device) :
-    cudaCommand(config, unique_name, host_buffers, device, "FRB_beamformer", "frb.ptx") {
-
-    // HACK -- at the very beginning of the run, we pretend we have 40 samples of padding,
-    // so that we get 51 outputs (otherwise we would get 50), so that the rechunker produces
-    // an output frame of 256 outputs after every 5 input frames.
-    padded_samples = 40;
```
dstndstn, line 23: I have just taken all of Erik's autogenerated code here, which means we've dropped the padding/overlap handling, but that will get pushed upstream to Erik's IndexSpaces package where this C++ code is auto-generated.

lib/cuda/kernels/BasebandBeamformer.jl
dstndstn, line 1: Was this file supposed to be checked in to this repo?

lib/stages/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ add_library(
     pyPlotN2.cpp
     rawFileRead.cpp
     rawFileWrite.cpp
-    $<$<BOOL:${USE_HDF5}>:hdf5FileWrite.cpp>
+    hdf5FileWrite.cpp
```
dstndstn: ?? Or should we not change this??

lib/stages/CMakeLists.txt
```diff
@@ -80,8 +85,7 @@ add_library(
     ExampleProducer.cpp
     ExampleConsumer.cpp
     ExampleDotProduct.cpp
-    # Julia
-    $<$<BOOL:${Julia_FOUND}>:juliaHelloWorld.cpp>)
```
dstndstn, line 84: ??

lib/stages/hdf5FileWrite.cpp
```diff
@@ -118,120 +125,77 @@ void hdf5FileWrite::main_thread() {
                 const chimeMetadata* const md = (const chimeMetadata*)mc->metadata;
                 {
                     const hid_t space = H5Screate(H5S_SCALAR);
-                    if (space < 0)
-                        ERROR("Could not create data space");
+                    assert(space >= 0);
```
dstndstn, line 128: Please confirm that these changes are the way you want them.

dstndstn (commented):
Erik,  Could you please check my comments and confirm the couple of places where there were merge conflicts with the develop branch.

jbmertens:
Is this PR still relevant? @eschnett @dstndstn

eschnett:
This is very old.

PR #1143, opened 2023-10-30
(Fix for) boost tests fail when dpdk is included in 22.04 builds

jbmertens:


lib/CMakeLists.txt
```diff
@@ -71,25 +71,32 @@ if(${USE_HSA}
 endif()
 
 # Include the old DPDK if the USE_OLD_DPDK flag is given.
-if(${USE_OLD_DPDK})
-    add_definitions(-DOLD_DPDK)
-    add_subdirectory(dpdk)
-    target_link_libraries(kotekan_libs INTERFACE kotekan_dpdk ${DPDK_LIBRARIES})
+if(${WITH_TESTS})
```
andrerenard: This might be a problem.  `WITH_TESTS` currently includes a bunch of stages in `/lib/testing`, which I sometime use with the DPDK networking...   Perhaps create another flag like `WITH_BOOST_TESTS` just for the unit tests and boost include.   Currently the `WITH_TESTS` controls two different build ideas.

andrerenard (changes requested):

andrerenard (approved):
LGTM!

jbmertens:
Fixed a typo, merging in now.

PR #1145, opened 2023-11-06
Add function to create a GPU ring-buffer view

dstndstn:
* add create_gpu_memory_ringbuffer(), to create a large underlying array with an array of views into it.

The motivating use-case is for the GPU upchannelizer, where the inputs are voltages (copied from CPU memory), and the upchannelizer needs to keep a bit of overlap between frames.  This could be implemented using GPU buffer views and having each upchannelizer stage manage its own overlaps, but an alternative is to implement ring-buffer-friendly wrap-around indexing in the upchannelizer kernel, and then have the upchannelizer `execute()` method pass an index into the underlying ring buffer to the kernel.


dstndstn:
the failing test is a red herring

jbmertens (approved):
LGTM!

andrerenard:
This is also now included via the chord->develop rename.

PR #1146, opened 2023-11-15
Rewrite Julia-generated CUDA kernels

eschnett:


julia/bin/make_bb_beams.jl
```diff
@@ -0,0 +1,46 @@
+using Revise
+using HDF5
+
+includet("/Users/eschnett/src/kotekan/lib/stages/FEngine.jl")
```
dstndstn: Please convert to relative path

julia/cxx/bb.cxx
dstndstn, line 1: Not needed?

julia/cxx/frb.cxx
dstndstn, line 1: Not needed?

julia/cxx/upchan.cxx
dstndstn, line 1: Not needed?

lib/stages/FEngine.cpp
```diff
@@ -0,0 +1,545 @@
...
+#include <complex>
+#include <cstdint>
+#include <cstring>
+#include <fstream>
+#include <julia.h>
+#include <juliaManager.hpp>
+#include <string>
+#include <vector>
+
+// Find float16:
```
dstndstn: There is also some Float16 handling in lib/util/visUtil.hpp -- consider moving this code to there and including visUtil.hpp ?

lib/stages/FEngine.cpp
```diff
@@ -0,0 +1,545 @@
...
+    register_producer(S_buffer, unique_name.c_str());
+    register_producer(G_buffer, unique_name.c_str());
+    register_producer(W_buffer, unique_name.c_str());
+
+    INFO("Starting Julia...");
+    kotekan::juliaStartup();
+
+    INFO("Defining Julia code...");
+    {
+        std::ifstream file("lib/stages/FEngine.jl");
```
dstndstn: Could you please add some error checking here, eg if the file doesn't exist (eg, because the user isn't running kotekan from the correct directory)?  From experience, I can say that the resulting error message is not very intuitive!

dstndstn (changes requested):
Looks good!, a few minor comments

julia/bin/make_bb_beams.jl
eschnett: Done.

julia/cxx/bb.cxx
eschnett: I added a comment instead describing the file because it's useful.

julia/cxx/frb.cxx
eschnett: I added a comment instead describing the file because it's useful.

julia/cxx/upchan.cxx
eschnett: I added a comment instead describing the file because it's useful.

lib/stages/FEngine.cpp
eschnett: Did that.

lib/stages/FEngine.cpp
eschnett: Done.

dstndstn:
I'm just testing a Dockerfile update to address the CI build failures with Ubuntu 22.04

dstndstn (approved):

PR #1149, opened 2023-11-23
Enable multiple GPU pipelines with Buffers for signalling

dstndstn:
Enable multiple GPU pipelines that share a gpuDeviceInterface (and therefore named GPU memory), but run at different cadences, (optionally) connected by Buffers that can signal that a piece of GPU memory is available.

This also adds a new kind of Buffer, a RingBuffer, that allows one producer and one consumer to communicate via a ring buffer where the reader and writer can wait for some number of elements to be ready to read, or have space to write; and can then indicate when those items have been read or written.  Currently, this memory is assumed to be external to that class (ie, GPU memory); the RingBuffer itself only manages the signalling between the producer and consumer.

The RingBuffer work involved a major rework of the Buffer class (making it an actual C++ class), and also adding a GenericBuffer from which RingBuffer and (vanilla) Buffer inherit.  This therefore touches most of the files in the project, but mostly in a formulaic manner (I did a lot of Perl search-and-replace!).

I added cudaCopyToRingBuffer.cpp and cudaCopyFromRingBuffer.cpp, but I expect that most of the time, we will want to avoid the extra copy that entails, and instead incorporate the ring-buffering directly into the kernel-specific wrappers.

For highlights, check out:
* config file syntax for ring buffers: config/tests/multi-pipe-ringbuf.yaml
* buffer.hpp :)


lib/core/buffer.hpp
```diff
@@ -41,30 +47,216 @@
...
+    std::vector<bool> is_done;
+};
+
+class GenericBuffer : public kotekan::kotekanLogging {
+public:
+    GenericBuffer(const std::string&, const std::string&, metadataPool* pool, int num_frames);
+    virtual ~GenericBuffer();
+    virtual void print_full_status(){};
+
+    virtual bool is_basic() = 0;
```
andrerenard: Comment on how this is intended to be used.

lib/core/buffer.hpp
```diff
@@ -41,30 +47,216 @@
...
     int last_frame_acquired;
 
     /// Last frame to be released with a call to mark_frame_*
     int last_frame_released;
+
+    // Is this producer/consumer done with each frame?
+    std::vector<bool> is_done;
+};
+
+class GenericBuffer : public kotekan::kotekanLogging {
```
andrerenard, line 91: Should have a doxygen comment for this class.  I like that this uses `kotekanLogging` now.   I could see that being useful for per-buffer log levels.

lib/core/buffer.hpp
```diff
@@ -41,30 +47,216 @@
...
+    std::string buffer_type;
+
+    /// The pool of info objects
+    metadataPool* metadata_pool;
+
+    /// Array of buffer info objects, for tracking information about each buffer.
+    std::vector<metadataContainer*> metadata;
+};
+
+class RingBuffer : public GenericBuffer {
```
andrerenard: Description of how this buffer works as a class doxygen.   And it would be good to add comments for the member functions as well.

lib/core/buffer.hpp
```diff
@@ -41,30 +47,216 @@
...
     /// Last frame to be released with a call to mark_frame_*
     int last_frame_released;
+
+    // Is this producer/consumer done with each frame?
+    std::vector<bool> is_done;
+};
+
+class GenericBuffer : public kotekan::kotekanLogging {
+public:
+    GenericBuffer(const std::string&, const std::string&, metadataPool* pool, int num_frames);
```
andrerenard: While this isn't going to be called directly, I still think it's worth a doc string.  I would also suggest adding the names of the strings to the signature so it gives some hint what they do.

lib/core/buffer.hpp
```diff
@@ -208,6 +497,24 @@ struct Buffer {
 
     /// The NUMA node the frames are allocated in
     int numa_node;
+
+protected:
```
andrerenard:
```suggestion
private:
```

lib/core/buffer.cpp
```diff
@@ -998,31 +754,128 @@ void buffer_free(uint8_t* frame_pointer, size_t size, bool use_hugepages) {
...
+    {
+        buffer_lock lock(mutex);
+        shutdown_signal = true;
+    }
+    empty_cond.notify_all();
+    full_cond.notify_all();
+}
 
-    return buf->metadata[ID];
+/////////////// RING BUFFER
```
andrerenard: I'm tempted to split out ring buffer (both the .hpp and .cpp) into their own files.  These files are getting fairly long.  It may also be worth splitting out the GenericBuffer and "Buffer" into their own files as well.  Long term I'd actually like to rename "Buffer" to something more specific, but this change already has enough wide ranging changes.  So perhaps just break out RingBuffer into their own files for now.

lib/core/buffer.cpp
```diff
@@ -334,532 +244,378 @@ int private_mark_frame_empty(Buffer* buf, const int id) {
...
+            ERROR("The consumer {:s} hasn't been registered, cannot unregister!", name);
+        }
 
-    // Check if removing this consumer would cause any of the frames
-    // which are currently full to become empty.
-    for (int id = 0; id < buf->num_frames; ++id) {
-        if (private_consumers_done(buf, id) == 1) {
-            broadcast |= private_mark_frame_empty(buf, id);
+        /*
+        // Check if removing this consumer would cause any of the frames
```
andrerenard: Why is this commented out?  (I'm redoing the logic of `unregister`, so I can just address it when I make those changes).

lib/core/buffer.cpp
```diff
@@ -34,45 +40,25 @@ struct zero_frames_thread_args {
...
-int private_consumers_done(Buffer* buf, const int ID);
+typedef std::lock_guard<std::recursive_mutex> buffer_lock;
 
-// Returns 1 if all producers are done for the given ID.
-int private_producers_done(Buffer* buf, const int ID);
+GenericBuffer::GenericBuffer(const std::string& _buffer_name, const std::string& _buffer_type,
+                             metadataPool* pool, int _num_frames) :
+    num_frames(_num_frames),
+    shutdown_signal(false), buffer_name(_buffer_name), buffer_type(_buffer_type),
+    metadata_pool(pool), metadata(num_frames, nullptr) {}
```
andrerenard: We should set the log level here as well since we now use `kotekanLogging`, otherwise the buffers will always default to level `INFO`.  It could be passed in via the factory for the buffer (or higher) config scope.   See for example the `Stage` constructor.

andrerenard (changes requested):
First round of comments.  Overall looking really good!    I'm a bit torn on how to name the buffers types here.  At the moment we have "GenericBuffer" as generic, and "Buffer" as a specific type of the generic.  Which maps well with the old naming, but breaks the pattern of having the most generic buffer just called "Buffer" and then named sub-types.  e.g. FrameBuffer (current Buffer), RingBuffer, (ResizableBuffer), etc.  However given the number of changes in this PR, I think it might be worth doing that as a second pass at some point instead of making this more complex now unless your perl scripts make that somewhat easy?

lib/core/buffer.hpp
dstndstn: Yeah, the `is_basic()` is a hack to check whether a `GenericBuffer` instance is a (Frame) `Buffer`, for use in the few places that want data that isn't in `GenericBuffer`.

Ideally I would turn these things into generic functions that subclassers can implement as appropriate, but in practice those are a bit messy -- a good example is `stages/bufferStatus.cpp`, which is creating a `Metrics` object for the number of full frames in the buffers.  `RingBuffer` buffers don't have the concept of "full frames".

I could do a match on the buffer's `type_name`, but there's a type hierarchy, since "standard", "vis",  and "hfb" types all have this.

I could generalize the "get_num_full_frames()" call to something like "count_available_data()" or something like that, with unspecified units, and then `RingBuffer` would have an obvious equivalent thing to return.

lib/core/buffer.hpp
dstndstn: Done; and yes, leveraging kotekanLogging should be very helpful!

lib/core/buffer.hpp
dstndstn: Indeed, done!

lib/core/buffer.cpp
dstndstn: I did set the log prefix and level in `bufferFactory` -- would you prefer to do it in constructors?

dstndstn:
* ~~should `copy_metadata()` be a `friend` function or something like that?  I feel like the buffer's `mutex` should not be a public member variable!~~
* think very carefully about the RingBuffer `wait_and_claim_readable()` and `read()` calls!
* we very strongly need multi-consumers for CHORD!
* what's up with the Buffer() argument zero_new_frames ?  Doesn't seem to be used?

lib/core/buffer.cpp
andrerenard: Oh I missed that somehow!  I don't have a strong preference.  Slightly nicer to have it in the constructor since then the factory doesn't need to remember to do that, and it's a bit cleaner, but realistically that's the only time this will happen.  Eventually it would be nice to move to an abstract factory for the buffers as well, but even then, only the factory needs to remember to include this logic and it's the same for every type.

andrerenard:
> * should `copy_metadata()` be a `friend` function or something like that?  I feel like the buffer's `mutex` should not be a public member variable!

I agree.  It could also become a `src_buf->copy_metadata_to(dest_buf)` type function.   Then it could lock on the src side, and call a function on the other buffer passing the metadata pointer, which would the lock on the dest side before doing the deep copy.   That's a bit more annoying since it involves more perl search and replace, so I'll leave it upto you if you want to do that, or just friend the function.

lib/core/buffer.cpp
```diff
@@ -998,31 +754,128 @@ void buffer_free(uint8_t* frame_pointer, size_t size, bool use_hugepages) {
...
+RingBuffer::RingBuffer(size_t sz, metadataPool* pool, const std::string& _buffer_name,
+                       const std::string& _buffer_type) :
+    GenericBuffer(_buffer_name, _buffer_type, pool, 1),
+    size(sz), elements(0), claimed(0)
+// write_cursor(0),
+// read_cursor(0)
+{}
+
+void RingBuffer::register_producer(const std::string& name) {
+    assert(producers.size() == 0);
```
andrerenard: We should likely throw exception here, since it's quite a strict requirement that we'll want the check in release builds.

lib/core/buffer.hpp
andrerenard: I like the option of `count_available_data()`, or perhaps `get_active_count()`, `get_in_use_count()`?    It's often very useful to be able to see the ratio of used memory (waiting to be processed) to total memory available.   e.g. num_full_frames/num_frames for FrameBuffers.  Since that can be used to measure back-pressure, and do other queue statistics.    Having said that, it might be easier to manage this change in a second PR, and just add a `todo` in the comment for this function, since we have a lot going on in this PR.

lib/core/buffer.hpp
```diff
@@ -41,30 +35,309 @@
...
+     * @return 0 on success, -1 if shutting down.
+     */
+    int wait_and_claim_readable(const std::string& consumer_name, size_t sz);
+
+    /**
+     * @brief Called by a consumer after the given number of elements
+     * has been read.  This number of elements MUST match the number
+     * "reserved" by the @c wait_and_claim_readable() call.
+     */
+    void read(const std::string& consumer_name, size_t sz);
```
andrerenard: English makes this dangerously ambiguous.  Maybe `finish_read()`?  and then `finish_write()` above for consistency.

lib/core/buffer.hpp
```diff
@@ -41,30 +35,309 @@
...
+class RingBuffer : public GenericBuffer {
+public:
+    /**
+     * @brief Build a new RingBuffer for connecting kotekan stages
+     * that want to produce & consume data with different sizes or at
+     * different cadences.
+     *
+     * Currently, only a single producer and single consumer are supported!
+     *
+     * @param ring_size: the number of elements in the ring buffer to be managed
```
andrerenard: Do you see a case where elements != bytes?    If not, I'd be tempted to make this bytes to make it easier to follow.

lib/gpu/gpuProcess.cpp
```diff
@@ -144,7 +142,7 @@ void gpuProcess::main_thread() {
 
     restServer& rest_server = restServer::instance();
     rest_server.register_get_callback(
-        fmt::format(fmt("/gpu_profile/{:d}"), gpu_id),
+        fmt::format(fmt("/gpu_profile/{:s}"), unique_name),
```
andrerenard: This will create some problems for the script: `python/scripts/kernel_utilization.py`...  Might need to do something like grab the output of `/endpoints` to find the names of the gpu profile end points.

config/tests/multi-pipe-ringbuf.yaml
```diff
@@ -0,0 +1,115 @@
...
+gpuA:
+    #profiling: true
+    #log_profiling: true
+    kernel_path: "../../lib/cuda/kernels/"
+    commands: &command_list
+    - name: cudaInputData
+      in_buf: host_in
+      gpu_mem: in
+    - name: cudaSyncInput
+    - name: cudaCopyToRingbuffer
```
andrerenard, line 65: Here it doesn't matter, but generically I think this has to be after a `cudaSyncOutput` and be a copy_out type , since otherwise the `cudaMemcpyAsync` call on the kernel stream can conflict with one of the two copy engines.

lib/cuda/cudaCopyFromRingbuffer.cpp
```diff
@@ -0,0 +1,74 @@
...
+    _output_size = config.get<int>(unique_name, "output_size");
+    _ring_buffer_size = config.get<int>(unique_name, "ring_buffer_size");
+    _gpu_mem_input = config.get<std::string>(unique_name, "gpu_mem_input");
+    _gpu_mem_output = config.get<std::string>(unique_name, "gpu_mem_output");
+    signal_buffer = dynamic_cast<RingBuffer*>(
+        host_buffers.get_generic_buffer(config.get<std::string>(unique_name, "host_signal")));
+    signal_buffer->register_consumer(unique_name);
+
+    input_cursor = 0;
+    set_command_type(gpuCommandType::KERNEL);
```
andrerenard, line 27: I think this needs to be on one of `COPY_IN` or `COPY_OUT` so it gets queued inline with other async copies to avoid stalling the kernel stream waiting for a copy engine.  This comment also applies to CopyToRing.

lib/core/buffer.hpp
```diff
@@ -41,39 +35,351 @@
...
+    std::condition_variable_any empty_cond;
+
+    void private_copy_metadata(int dest_frame_id, GenericBuffer* src, int src_frame_id);
+};
+
+/**
+ * @brief A buffer to manage the signalling between stages when those
+ * stages want to communicate data using a ring buffer.
+ *
+ * This class is a little strange in that it doesn't actually allocate
```
andrerenard: Perhaps we should name it something like `SignalRingBuffer`?  Since I could imagine having a ring buffer in CPU space which actually manages the underlying memory.

lib/core/buffer.hpp
```diff
@@ -41,39 +35,351 @@
...
+ * @conf metadata_pool The name of the metadata pool to associate with the buffer
+ */
+class RingBuffer : public GenericBuffer {
+public:
+    /**
+     * @brief Build a new RingBuffer for connecting kotekan stages
+     * that want to produce & consume data with different sizes or at
+     * different cadences.
+     *
+     * Currently, only a single producer and single consumer are supported!
```
andrerenard: I don't think we need multiple producers (that's a really unlikely case with a complex ring like this), but multiple consumers are likely.  Based on your other comment, do you intend to add support in this PR, or finish this PR based on this limited case?

lib/cuda/cudaCopyFromRingbuffer.cpp
```diff
@@ -0,0 +1,74 @@
...
+    cudaCommand(config, unique_name, host_buffers, device, "cudaCopyFromRingbuffer", "") {
+    _output_size = config.get<int>(unique_name, "output_size");
+    _ring_buffer_size = config.get<int>(unique_name, "ring_buffer_size");
+    _gpu_mem_input = config.get<std::string>(unique_name, "gpu_mem_input");
+    _gpu_mem_output = config.get<std::string>(unique_name, "gpu_mem_output");
+    signal_buffer = dynamic_cast<RingBuffer*>(
+        host_buffers.get_generic_buffer(config.get<std::string>(unique_name, "host_signal")));
+    signal_buffer->register_consumer(unique_name);
+
+    input_cursor = 0;
```
andrerenard: I think this might be a problem when GPU buffer depth != 1, since there will be a few copies of this object.   Would it be possible to simply register the consumer/producer once in `inst==0`, and use the return values from the ring buffer for the cursor?

andrerenard (dismissed):

lib/core/buffer.hpp
dstndstn: Done!

lib/core/buffer.cpp
dstndstn: Done (just broke out ringbuffer.cpp/hpp)

lib/core/buffer.cpp
dstndstn: Oopsies, fixed!

lib/core/buffer.cpp
dstndstn: The log level comes from the config, which currently isn't passed to the Buffer constructor (maybe it should be...), so I kept the set_log_level() in the bufferFactory, but moved the set_log_prefix to the GenericBuffer constructor.

lib/core/buffer.cpp
dstndstn: Done

lib/core/buffer.hpp
dstndstn: Good idea, done!

lib/core/buffer.hpp
dstndstn: I was thinking it could be convenient to handle elements != bytes, because the RingBuffer could log the logical units rather than the pointers and bytes.  But maybe that's too much extra complexity for not enough payoff!

lib/core/buffer.hpp
dstndstn: I ended up removing `is_basic()` as a member variable, instead adding a free function `is_frame_buffer`, and also pulled out some of the Dot and JSON formatting into member functions so that RingBuffer can provide its own values (in the future; I didn't implement anything yet).

lib/core/buffer.hpp
dstndstn: Good point -- I think I will hold off on renaming until a later PR, after I've added the multi-consumer, since other things may shake out in that process.

lib/core/buffer.hpp
dstndstn: I think I would like to get this PR in and then do multi-consumer in another (hopefully much smaller!) PR.

lib/cuda/cudaCopyFromRingbuffer.cpp
dstndstn: Yes, good catch

dstndstn: Fixed -- needs cudaCommandState to hold the cursor (since I'm not yet tracking the cursors in the ringbuffer, and adding multi-consumer is going to change everything).

lib/gpu/gpuProcess.cpp
dstndstn: Fixed

config/tests/multi-pipe-ringbuf.yaml
dstndstn: Yeah, we're going to have to be careful about that.  Kendrick wrote a stand-alone kernel that duplicates the functionality of cudaMemcpyAsync but operates without the copy engine - we may need to use that in some places.  I'm not going to worry too much about that here, since cudaCopy{To,From}Ringbuffer are really just demos of how to do the copies, and usually they would be integrated into a kernel wrapper directly, and then not require a(n async) copy.

lib/cuda/cudaCopyFromRingbuffer.cpp
dstndstn: yeah, this will require some thought!

dstndstn:
Hi @andrerenard , I pushed the multi-producer, multi-consumer ring buffer commits here, so I think this is ready for your re-review.  There is a small unit-test code for it, but it's not super extensive so far (doesn't test multi-threading, for example!), so that could be an area for more work if you feel it's required.  I have a demo pipeline for multi-consumer, will add one that does multi-producer, multi-consumer!

config/tests/multi-pipe-rechunk.yaml
```diff
@@ -0,0 +1,118 @@
...
+
+gen_data:
+    kotekan_stage: testDataGen
+    type: const8
+    values: [0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07]
+    #    array_shape: [64, 16]
+    #dim_name: ["X", "Y"]
+    out_buf: host_in_buffer
+
+gpuA:
```
andrerenard, line 57: I'm trying to decide if it would make sense to organize the config blocks something like this:
```
gpu_0:
    profiling: true
    log_profiling: true
    gpu_id: 0
    kernel_path: "../../lib/cuda/kernels/"
    
    main_pipeline: &main_pipeline
    # commands for the main pipeline
    
    mid_pipeline: &mid_pipeline
    # commands for mid pipeline

    gpu_main:
          kotekan_stage: cudaProcess
          commands: *main_pipeline
          # buffers

    gpu_mid:
          kotekan_stage: cudaProcess
          commands: *mid_pipeline
          # buffers
```

I think a lot of config like the `profiling` options, kernel path, etc.  Don't need to be repeated each time, for the different pipelines.    These blocks will get fairly long, but it might help to show this is tightly interconnected.

julia/Project.toml
```diff
@@ -14,6 +13,7 @@ JuliaFormatter = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
 Mustache = "ffc61752-8dc7-55ee-8c37-f3e9cdd09e70"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SixelTerm = "65edfddc-f399-4499-8369-a1bd38eee2ea"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
```
andrerenard, line 16: Did you intend to pull in these Julia/kernel changes?

lib/core/ringbuffer.hpp
```diff
@@ -0,0 +1,145 @@
...
+ * @conf metadata_pool The name of the metadata pool to associate with the buffer
+ */
+class RingBuffer : public GenericBuffer {
+public:
+    /**
+     * @brief Build a new RingBuffer for connecting kotekan stages
+     * that want to produce & consume data with different sizes or at
+     * different cadences.
+     *
+     * Currently, only a single producer and single consumer are supported!
```
andrerenard: No longer true. 
```suggestion
```

lib/cuda/cudaCommand.hpp
```diff
@@ -97,6 +97,9 @@ class cudaCommand : public gpuCommand {
     virtual cudaEvent_t execute_base(cudaPipelineState& pipestate,
                                      const std::vector<cudaEvent_t>& pre_events);
 
+    virtual bool should_execute(cudaPipelineState& pipestate,
```
andrerenard, line 100: Are we still going to need this with the ring buffer?

tools/docker/22.04/Dockerfile
```diff
@@ -199,7 +199,7 @@ ENV HDF5_PLUGIN_PATH /usr/local/lib/python3.10/dist-packages/hdf5plugin/plugins
 ENV KOTEKAN_BUILD_DIRNAME build-2204
 
 # Update Julia...
-RUN juliaup update
+RUN juliaup update && echo 2
```
andrerenard, line 202: Why `echo 2`?   Was this just to force a change?

tests/boost/test_ringbuffer.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+#include <boost/test/included/unit_test.hpp>
+#include <iostream>
+
+using kotekan::Config;
+using json = nlohmann::json;
+using namespace kotekan;
+using namespace std;
+
+BOOST_AUTO_TEST_CASE(test1) {
```
andrerenard, line 21: Thanks for adding unit tests!

andrerenard (changes requested):
Looking really good!   

Is it expected that all the kernels got regenerated (and look different now)?   I understand that the cxx templates needed to change and be regenerated, but I'm not sure why that caused so many other kernel level changes?

dstndstn:
> Looking really good!
> 
> Is it expected that all the kernels got regenerated (and look different now)? I understand that the cxx templates needed to change and be regenerated, but I'm not sure why that caused so many other kernel level changes?

I merged in Erik's #1154 changes on the 'chord' branch, which updated most of the auto-generated files.

config/tests/multi-pipe-rechunk.yaml
dstndstn: Oh, that's a good idea!  I think we'll get a good feel for this once we start layout out the real pipelines - eg the upchannelization & FRB chain is going to have like 6-7 pipelines, one per upchan factor, because they're going to operate at somewhat different cadences.

julia/Project.toml
dstndstn: This is from merging in #1154.  I'm not exactly sure why this change appears here rather than in #1154, but it resulted from auto-rebuilding the Julia kernels.

lib/core/ringbuffer.hpp
dstndstn: Fixed!

lib/cuda/cudaCommand.hpp
dstndstn: Good point, we should probably remove this whole mechanism after the dust settles!  *Possibly* including removing the `cudaPipelineState` class; the buffer metadata objects will ideally hold any information that has to flow down the pipeline.

I don't really want to scrap this just yet, because so far this change hasn't broken the old MVP setup!

tools/docker/22.04/Dockerfile
dstndstn: Yep!  The CI pipelines need a bit of work to handle the 'chord' branch analogously to the way they handle the 'develop' branch.

tests/boost/test_ringbuffer.cpp
dstndstn: :)

dstndstn:
I think I have addressed everything in the review but github is still showing it as "1 change requested".  I don't know how to get rid of that and get it merged!

andrerenard (approved):
Thanks for all your work on this huge change!  It's fantastic to finally have everything be C++ now, and I think we'll make very good use of the new functionality.   (I'm also going to write a new buffer for improved baseband dumps, and that's going to make good use of the buffer class structure here.)

PR #1152, opened 2023-11-30
CHORD #1146 -- add Julia code-generation (kernels & C++ templates) and F-engine simulator

dstndstn:
This changes the Cuda kernels and Kotekan CudaCommands for the Baseband beamformer, FRB beamformer, and Upchannelizer to be generated by the Julia code generator.

We still have both the Julia sources and the generated code outputs checked in -- re-running Julia to generate these codes is assumed to be done offline, not as part of the build process, using scripts in `julia/bin/{bb,frb,upchan}.sh`.

Also adds an FEngine stage that simulates the data arrays produced by the F-engine.


julia/src/Kotekan.jl
```diff
@@ -0,0 +1,5 @@
+module Kotekan
+
+greet() = print("Hello World!")
```
andrerenard: Do we need this?

julia/cxx/bb.cxx
```diff
@@ -0,0 +1,173 @@
+// CHORD 8-bit baseband beamformer
+// See <https://www.overleaf.com/project/6228adae742a3a2da1afe437>
```
andrerenard: Maybe reference the internally linked PDF, since most people won't have access to the overleaf?

julia/bin/plot_bb_beams.jl
```diff
@@ -0,0 +1,138 @@
+using CUDASIMDTypes
+using CairoMakie
+using HDF5
+using SixelTerm
+
+dir = "/tmp"
+prefix = "blue_"
+iter = "0000000"
```
andrerenard, line 12: It's fairly clear to me what this is doing and why these variables are set as they are.   But I could see this being confusing to someone who isn't also looking at the config files generating the input to this script.   Maybe just reference the config that's generating this data?

julia/cxx/README.md
```diff
@@ -0,0 +1,9 @@
+# Simple C++ kernel implementations
+
+This directory contains simple (straightforward) C++ implementations
```
andrerenard, line 3: This code might be more valuable as part of the CI process (e.g. as a pipeline the computes the CPU version of the GPU logic, and compares them.   I realize not all the tooling is in place yet, but we should keep this in mind for when it is.  Otherwise I worry we'll forget about keeping this code upto date with any kernel changes we might make later.

julia/bin/upchan.sh
```diff
@@ -0,0 +1,37 @@
...
+for U in $Ufactors; do
+    clang-format -i output-A40/upchan-U$U.cxx
+done
+
+# Format generated Julia code
+julia --project=@. --eval 'using JuliaFormatter; JuliaFormatter.format_file("output-A40")'
+
+# Copy kernel into Kotekan
+for U in $Ufactors; do
+    cp output-A40/upchan-U$U.cxx ../lib/cuda/cudaUpchannelizer_U$U.cpp
```
andrerenard, line 33: There are a fair number of assumptions about this using an A40.  Is the logic specific enough to the A40 that basically that's the only card this will run on, or would it be possible to be more general for say an A10/A30/L40/L40S, etc. if those endup getting used for some instrument?

lib/stages/FEngine.cpp
```diff
@@ -0,0 +1,531 @@
...
+#include <juliaManager.hpp>
+#include <string>
+#include <vector>
+#include <visUtil.hpp>
+
+#if !KOTEKAN_FLOAT16
+#warning "The F-Engine simulator requires float16 support"
+#else
+
+class FEngine : public kotekan::Stage {
```
andrerenard: We should have some documentation on what this does and generically how it works (including what config parameters it takes).   Also generally we've used separate header files for declarations, could we do that for consistency?

andrerenard (changes requested):
This looks great!  A few inline comments, some of them might not make sense to address at this point.

julia/src/Kotekan.jl
dstndstn: I added a comment to the effect that we kind-of need this file as a placeholder (some Julia tools want such a file and module to exist), and we may want to build it out in the future.

julia/cxx/bb.cxx
dstndstn: Good idea, done

julia/bin/plot_bb_beams.jl
dstndstn: Done!

julia/cxx/README.md
dstndstn: Actually, at some point in the distant past I copied versions of these codes into testing stages!  Eg, `lib/testing/gpuSimulateCudaBasebandBeamformer.cpp` is a stage that includes code from `bb.cxx`.  We could potentially even auto-generate these testing stages, and then run them all in CI!  I added a TODO comment to this effect.

julia/bin/upchan.sh
dstndstn: I don't know, but these kernels rely heavily on the exact set (and performance levels) of tensor core instructions, so I'm guessing that if a different card got used, one would have to carefully check the performance anyway, so that there's not a ton of value in theoretically targeting a broader set of cards.

lib/stages/FEngine.cpp
dstndstn: Done!

dstndstn:
I think I have addressed everything raised in your review, @andrerenard , but github is still saying "1 change requested"... I'm not sure how to proceed.

andrerenard (approved):
Sorry missed your comment from last week.  This looks great, thanks for the changes!

PR #1154, opened 2023-12-08
Implement upchannelizer kernels

eschnett:


dstndstn (approved):

dstndstn:
The failing CI builds are due to an issue where it's not using the updated Docker container where we have the gcc-12 compiler installed.

This PR looks good to me.  Overall, it would be good to have some remarks at the top of each auto-generated file to say which file it is generated from!

PR #1155, opened 2023-12-20
Add support for building kotekan on arm64

andrerenard:
This change allows for building kotekan on arm64 platorms with the CUDA and DPDK cmake options.  

This change has _not_ attempted to find all the subtle behavioural differences between x86_64 and arm64.   It simply allows for the code to be compiled, although simple pipelines do appear to work.   It is primarly intended to allow easier developement on Apple Silicon computers, and lay the foundation for future arm64 compatibility work. 

* Adds a Ubuntu 22.04 arm64 docker file with arm64 versions of the depdences
* Adds a number of missing preprocessor directives to exclude AVX(2) code.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,7 +40,7 @@ jobs:
       uses: actions/checkout@v3
 
     - name: Check for file changes
-      uses: tj-actions/changed-files@v37
+      uses: tj-actions/changed-files@v41
       id: changed-files
 
     - name: Login to GitHub Container Registry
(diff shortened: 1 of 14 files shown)


andrerenard:
@jbmertens  I looked into adding an ARM build to the CI, however it looks like ARM64 isn't yet supported by github actions runners.  I could cross-compile, but given this is low priority, I will hold off until github adds native support for ARM64 runners.

jbmertens (approved):
LGTM

PR #1156, opened 2023-12-21
Refactor metadata, and connect upchannelizer to ringbuffers

dstndstn:
This work started by trying to connect the upchannelizer kernels to the ringbuffers, but ended up refactoring the metadata along the way!

(Ridiculous commit history because of squash-merged commits to chord.)

- changed the `metadata` from a C struct to a C++ class with inheritance.  However, this ended up being messy because there were many places where the in-memory representation of a metadata object was used directly as a serialization format.  The new C++ classes add JSON serializers, but we will have to decide to use these rather than binary serializers.  As a result, this implementation has a nasty hack where we define C structs to define the binary serialization format, and pack data into those structs when serializing.
- `copyToRingbuffer` and `copyFromRingbuffer` got extended to copy directly between host memory and GPU ringbuffers.
- metadata interactions with ringbuffers not totally settled
- for "reasons", I updated the version of the "fmt" library that we include


eschnett:
The CI failures look weird. Pulling in the recent changes to `.github` from the `chord` branch might help.

dstndstn:
> The CI failures look weird. Pulling in the recent changes to `.github` from the `chord` branch might help.

Hmm, I just merged chord into this branch.

In this build
https://github.com/kotekan/kotekan/actions/runs/7852095589/job/21430037040?pr=1156
it looks like CMake thinks it has found ASDF, but the header file isn't found:
```
 -- ASDF found
...
[ 24%] Building CXX object lib/stages/CMakeFiles/kotekan_stages.dir/asdfFileWrite.cpp.o
/code/kotekan/lib/stages/asdfFileWrite.cpp:3:10: fatal error: asdf/asdf.hxx: No such file or directory
    3 | #include <asdf/asdf.hxx>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [lib/stages/CMakeFiles/kotekan_stages.dir/build.make:286: lib/stages/CMakeFiles/kotekan_stages.dir/asdfFileWrite.cpp.o] Error 1
 ```

**Comment by dstndstn** [comment]
In the current chord docker image,
```
> docker run -it ghcr.io/kotekan/kotekan/kotekan-2204:chord bash
```
I see that asdf is installed in `/usr/local`, and `pkg-config` says to include `/usr/local/include`, and the header file is directly in that directory.  But `asdfFileWrite.cpp` has an extra `asdf/` on the `#include`...  that's my guess of what is going wrong.  Oh, and `hpp` vs `hxx` suffix.

```
# pkg-config --cflags asdf-cxx
-I/usr/local/include
# ls /usr/local/include/asdf/asdf.hxx
ls: cannot access '/usr/local/include/asdf/asdf.hxx': No such file or directory
# ls /usr/local/include/asdf.hpp
/usr/local/include/asdf.hpp
```

But why am I hitting this?  If I merged in changes from chord, shouldn't it build?

On another note, do we need to update this other github target to push a new Docker container for the `chord` branch when a PR is merged?
https://github.com/kotekan/kotekan/blob/chord/.github/workflows/publish_docker.yml

**Comment by eschnett** [comment]
You are seeing an older version of ASDF (v7), not the new version (v8) that we need. I think this is because you are using the old Docker image. Since my changes to the `chord` branch worked, I think you may have changes to either the `Dockerfile` or the `CI.yaml` script that cause this.

If I understand correctly, our current mechanism to allow local Docker changes to the `chord` branch doesn't work. @jbmertens helped me fix things by pulling things directly into the `develop` branch. I didn't know there is also a `chord` tag for the Docker images, and I don't think this tag is currently working – we would need to pull Docker changes from `develop` back into `chord`, and I don't think we did this (unless @jbmertens did).

I solved this by removing all my local changes that relate to Docker, merging them directly into `develop`, and then running the `chord` CI checks with the (new) default image.

File: .github/workflows/main.yml [hunk 1]
```diff
@@ -67,8 +67,8 @@ jobs:
           fi
           echo "docker_build=1" >> $GITHUB_OUTPUT;
         else
-          echo "Dockerfile was not modified. Using develop branch image."
```
eschnett, line 70: @dstndstn This change might be holding you up.

dstndstn: thanks

julia/kernels/bb_template.cxx
```diff
@@ -240,7 +237,7 @@ cudaEvent_t cuda{{{kernel_name}}}::execute(cudaPipelineState& /*pipestate*/, con
         {{^hasbuffer}}
             {{^isoutput}}
                 CHECK_CUDA_ERROR(cudaMemcpyAsync({{{name}}}_memory,
-                                                 {{{name}}}_host.at(gpu_frame_index).data(),
+                                                 {{{name}}}_host.data(),
```
eschnett, line 240: Does this copy the complete ringbuffer to the GPU?

This code is for debugging only: It initializes output buffers with a poison value to detect cases where the kernel doesn't properly fill in all its outputs. A simple solution would be to just not copy anything to the GPU any more. Otherwise we should only copy those parts of the buffer that the kernel is allowed to write.

eschnett: Sorry, I was confused. This code is for input buffers, not output buffers, and it is for input buffers that are hidden from Kotekan. I don't think the baseband kernel has any such buffers. The FRB kernel does have such a buffer, and it is only used for the dish layout which is constant in time and which should be handled better anyway – it should be copied to the GPU only once, not for every call. I think your code here is inconsequential in the sense that it is never executed.

julia/kernels/bb_template.cxx
```diff
@@ -275,15 +272,15 @@ cudaEvent_t cuda{{{kernel_name}}}::execute(cudaPipelineState& /*pipestate*/, con
...
-        ERROR("cuLaunchKernel: Error number: {}: {}", err, errStr);
+        ERROR("cuLaunchKernel: Error number: {}: {}", (int)err, errStr);
     }
 
     // Copy results back to host memory
     // TODO: Skip this for performance
     {{#kernel_arguments}}
         {{^hasbuffer}}
             {{#isoutput}}
-                CHECK_CUDA_ERROR(cudaMemcpyAsync({{{name}}}_host.at(gpu_frame_index).data(),
```
eschnett, line 286: This is similar to my comment above, but this code handles output buffers that are hidden from Kotekan. These are the `info` output arguments for the kernel. I don't know if kernels can now be launched in an overlapping manner, and if so, we would need to make sure that these `info` output arguments are separate for each call. If the kernels remain sufficiently serialized then your code is correct.

julia/kernels/upchan_template.cxx
```diff
@@ -484,25 +484,29 @@ cudaEvent_t cuda{{{kernel_name}}}::execute(cudaPipelineState& /*pipestate*/, con
...
+    for (std::size_t i = 0; i < info_host.size(); ++i)
+        if (info_host[i] != 0)
             ERROR("cuda{{{kernel_name}}} returned 'info' value {:d} at index {:d} (zero indicates no error)",
-                info_host.at(gpu_frame_index)[i], i);
+                info_host[i], i);
 
     return record_end_event();
 }
 
 void cuda{{{kernel_name}}}::finalize_frame() {
```
eschnett, line 500: All other templates remove the `finalize_frame` function. Only the upchannelizer keeps them, and does something that looks important. Is something different happening here, or is the code just structured differently?

lib/core/buffer.cpp
```diff
@@ -52,14 +52,11 @@ GenericBuffer::GenericBuffer(const std::string& _buffer_name, const std::string&
 
 GenericBuffer::~GenericBuffer() {}
 
-bool GenericBuffer::set_metadata(int ID, metadataContainer* meta) {
+bool GenericBuffer::set_metadata(int ID, std::shared_ptr<metadataObject> meta) {
     assert(ID >= 0);
     assert(ID < num_frames);
     buffer_lock lock(mutex);
```
eschnett, line 58: What is this mutex used for? I thought it was only for reference counting. If so, then we wouldn't need it any more.

lib/core/metadata.cpp
```diff
@@ -0,0 +1,59 @@
+#include "metadata.hpp"
+
+#include "errors.h" // for CHECK_ERROR_F, CHECK_MEM_F
+
+#include <assert.h> // for assert
```
eschnett: Maybe use `cassert`, `cstdlib`, and `cstring` here instead?

Is `cstring` even necessary?

lib/core/metadata.hpp
```diff
@@ -0,0 +1,147 @@
...
+#define KOTEKAN_METADATA_HPP
+
+#include "factory.hpp"
+#include "kotekanLogging.hpp"
+
+#include "json.hpp" // for json
+
+#include <memory>
+#include <mutex>
+#include <stdint.h> // for uint32_t
```
eschnett: Maybe `cstdint` and `cstdio`? Or `iostream` instead?

eschnett: Actually, `size_t` should also be in `cstdint` or `cstddef`.

lib/core/metadata.hpp
```diff
@@ -0,0 +1,147 @@
...
+
+protected:
+    /// The underlying block af data that we allocate objects out of
+    void* data_block;
+
+    /**
+     * @brief An array to indicate the use state of each pointer in the @c metadata_objects array
+     * A value of 1 indicates the pointer is in use and should have a reference count > 0
+     */
+    std::vector<bool> in_use;
```
eschnett: I don't know how this vector is accessed, but C++ has a quirk: All containers are thread-safe in the sense that different threads can access different elements simultaneously, as long as the container itself is not modified (grown or shrunk etc.). This guarantee does not hold for `std::vector<bool>`. If different threads access it simultaneously the result is undefined. One has to use `std::vector<char>` instead. I don't know whether this is the case here, but I'm bringing it up anyway just to be safe.

lib/core/ringbuffer.cpp
```diff
@@ -2,8 +2,33 @@
 
 typedef std::lock_guard<std::recursive_mutex> buffer_lock;
 
-RingBuffer::RingBuffer(size_t sz, metadataPool* pool, const std::string& _buffer_name,
-                       const std::string& _buffer_type) :
+static void print_py_status(RingBuffer* rb) {
```
eschnett, line 9: I would use `std::ostringstream` here instead of string concatenation.

lib/metadata/chordMetadata.cpp
```diff
@@ -41,3 +43,123 @@ chordMetadata::chordMetadata() :
...
+    // by which the time samples have been downsampled relative to
+    // FPGA samples.
+    int32_t time_downsampling_fpga[CHORD_META_MAX_FREQ];
+};
+
+size_t chordMetadata::get_serialized_size() {
+    return chimeMetadata::get_serialized_size() + sizeof(chordMetadataFormat);
+}
+
+size_t chordMetadata::set_from_bytes(const char* bytes, size_t length) {
```
eschnett, line 94: Are these functions tested?

I don't think these functions work well. They implicitly depend on global `#define` e.g. for the number of frequencies, and there are not checks for these, and the case where they change isn't handled. I would simply rip out this kind of binary serialization, it's worse than common blocks in Fortran. There are nice libraries (e.g. https://uscilab.github.io/cereal/) that we could use instead if we needed. Otherwise I strongly prefer serializing to YAML, e.g. via ASDF.

lib/metadata/chordMetadata.hpp
```diff
@@ -164,50 +176,43 @@ struct chordMetadata {
     }
 };
 
-inline void chord_metadata_init(chordMetadata* c) {
-    bzero(c, sizeof(chordMetadata));
+inline void chord_metadata_init(std::shared_ptr<chordMetadata> meta) {
+    // ??
+    *meta = chordMetadata();
```
eschnett: Ditto on the `??`. This code looks weird. We have a nice constructor that initializes the metadata, I don't think we need an explicitly initialization function.

lib/metadata/chordMetadata.hpp
```diff
@@ -164,50 +176,43 @@ struct chordMetadata {
...
 }
 
-inline void chord_metadata_copy(chordMetadata* out, const chordMetadata* in) {
-    memcpy(out, in, sizeof(chordMetadata));
+inline void chord_metadata_copy(std::shared_ptr<chordMetadata> out,
```
eschnett: I would remove this function as well.

lib/metadata/chordMetadata.hpp
```diff
@@ -164,50 +176,43 @@ struct chordMetadata {
...
-    if (!mc)
-        return nullptr;
-    if (strcmp(mc->parent_pool->type_name, "chordMetadata")) {
-        WARN_NON_OO("Expected metadata to be type \"chordMetadata\", got \"{:s}\".",
-                    mc->parent_pool->type_name);
-        return nullptr;
-    }
-    return static_cast<const chordMetadata*>(mc->metadata);
+inline std::shared_ptr<chordMetadata> get_chord_metadata(Buffer* buf, int frame_id) {
+    std::shared_ptr<metadataObject> meta = buf->metadata[frame_id];
```
eschnett: Should there be an assert statement checking that the buffer has metadata and that the metadata is in chord format?

lib/stages/FEngine.cpp
```diff
@@ -491,9 +491,11 @@ void FEngine::main_thread() {
         } // if !skip_julia
 
         // Set metadata
-        chordMetadata* const E_metadata = get_chord_metadata(E_buffer, E_frame_id);
+        std::shared_ptr<chordMetadata> const E_metadata = get_chord_metadata(E_buffer, E_frame_id);
         chord_metadata_init(E_metadata);
-        E_metadata->chime.fpga_seq_num = frame_index;
+        // ???
+        E_metadata->frame_counter = frame_index;
```
eschnett, line 495: Do we still have a frame counter? That doesn't make sense in a ringbuffer?

tools/docker/22.04/Dockerfile
```diff
@@ -128,15 +128,34 @@ RUN apt-get update && \
...
+        -DBUILD_EXAMPLES=OFF \
+        -DBUILD_FUZZERS=OFF \
+        -DBUILD_TESTS=OFF \
+        && \
+    cmake --build build --parallel $(nproc) && \
+    cmake --install build && \
+    true) && \
+    rm -rf src
+
+# Install ASDF (depends on Blosc2)
```
eschnett, line 150: At some point we installed ASDF twice into the Docker image. That is corrected on `develop` (methinks). Did you check this Dockerfile whether the problem is removed here as well?

eschnett (commented):
220 kloc of changes! Thats 50% more than mine.

I left some comments, many of these are questions to double check.

julia/kernels/bb_template.cxx
dstndstn: Okay, thanks for checking that.  I have been meaning to figure out the updatable-config or REST calls, and kotekan protocols for handling constant or infrequently-changed input data.

This specific change was because of a change a while back (#1133/#1150) where we create a set of cudaCommand objects for each step in the GPU pipelines, and each cudaCommand is only responsible for handling one frame at a time - so there is rarely a reason for the cudaCommands to hold GPU-buffer-depth arrays.

julia/kernels/upchan_template.cxx
dstndstn: The upchannelizer is the only one that uses the ring-buffers so far, so it's different; the others will probably need finalize_frame() methods once they get ring-bufferized.  (Previously, they used to have finalize_frame() methods to free metadata, but since the metadata now uses shared_ptrs, that's not necessary any more.)

lib/core/buffer.cpp
dstndstn: It's used along with condition variables for signalling empty/full events.  (It's also used to protect a bunch of the methods, but I bet that some of those could now be dropped!)

lib/core/metadata.cpp
dstndstn: fixed!

lib/core/metadata.hpp
dstndstn: fixed!

lib/core/metadata.hpp
dstndstn: Good catch!  It turns out that vector wasn't needed at all any more! :)

lib/core/ringbuffer.cpp
dstndstn: Good idea, done!

lib/metadata/chordMetadata.cpp
dstndstn: I completely agree, we should serialize to a better format (this class has JSON serialization support).  But the CHIME code uses binary serialization extensively, so for backward compatibility I kept that.  We could just make the chordMetadata object serialize itself to/from JSON when requested to serialize itself to a binary array!

(I did at least assert that the #define'd sizes are the same)

lib/metadata/chordMetadata.hpp
dstndstn: Yes, good catch!  This was mindless conversion of old C code to C++.  I removed both these methods.

lib/metadata/chordMetadata.hpp
dstndstn: Good catch, done!

lib/metadata/chordMetadata.hpp
dstndstn: Good catch, done!

lib/stages/FEngine.cpp
dstndstn: This is an FPGA-level frame counter... this may not make sense any more when the data go through the ringbuffer.  We could copy the first frame counter, or zero it out...

tools/docker/22.04/Dockerfile
dstndstn: It's only installed once in the 22.04 container, and not at all in 18.04.

eschnett (approved):

PR #1028, opened 2022-05-29
New phase generation template

andrerenard:
Base code for the new beamforming phase generation. 

TODO:

- [X] Add new UTC to UT1 endpoint - Moved to #1031
- [X] Confirm element/frequency array indices
- [ ] Phase copy functions for HSA/OpenCL
- [X] Add array element positions file for CHIME - Move to #1031
- [X] Refactor `beam_coord` struct? -> Decided against this change

PR #1161, opened 2024-01-24
Beamforming phases update

andrerenard:
This change pulls together changes from #1028 and #1031 and bases them against the 2021.11 release instead of develop.

Key changes: 
* Compute tracking beam phases outside of the HSA commandObject logic. 
* Switch to @VishwangiShah312's phase generation functions: 
  * Moves the phase center to the middle of the array
  * Uses exact positions for feed
  * Adjusts for UT1
  * Updates phase solutions to every frame (~120ms) instead of every 1 second.
* Allow both the old and new beamformer phase algorithms to run on different beams.


andrerenard:
Closing this PR as we've merged the changes in https://github.com/kotekan/kotekan/pull/1261

PR #1167, opened 2024-02-01
Grab changes from the review of #1152 (chord #1146)

dstndstn:
Cherry-pick 99c2ee9 from dstn/develop/chord-1146: "address Andre's review: add various comments, and pull out FEngine.hpp header"


andrerenard:
@dstndstn do you also want me to review this, or was there something you wanted Erik to check?

dstndstn:
Looks like this has gone a bit stale, let me resolve the conflicts and then yes, if you could review that would be great!

dstndstn:
Hi @andrerenard , if you wouldn't mind reviewing this, that would be great!

tests/boost/test_ringbuffer.cpp
```diff
@@ -18,6 +19,23 @@ using json = nlohmann::json;
 using namespace kotekan;
 using namespace std;
 
+// Use a boost::test "Global Fixture" to set the locale for fmt...
+// https://www.boost.org/doc/libs/1_75_0/libs/test/doc/html/boost_test/tests_organization/fixtures/global.html
+struct GlobalFixture_Locale {
```
andrerenard: Should this be moved into `testUtils.hpp` since I assume more than one test might want to make use of it?

andrerenard (approved):
Approved, with one comment to consider.

tests/boost/test_ringbuffer.cpp
dstndstn: Good idea, done!

dstndstn:
The test failures look *really* unrelated...

andrerenard:
I'm going to try rerunning them.   We do have some tests that fail due to random number generator bad luck... But odd that all of them seemed to fail...

dstndstn:
The values seem to be very far off, so doesn't look like one of these random-sample things, maybe it's a real structural change we made by accident, but is it happening on the current head of the `chord` branch?

andrerenard:
Oh yeah, some of those values are way out...  Could the location settings somehow impact this?  Seems really unlikely.  I'm not seeing where this could have been added in the current head.  The last big PR to go in seemed to pass everything except the boost tests.

andrerenard:
Best guess would be something updated in the docker container, since we've been making a few changes to that.  @jbmertens Would you able to try and figure out what's happening here?   It's in the N2 pipeline, and I think something you've looked at recently?

jbmertens:
> Best guess would be something updated in the docker container, since we've been making a few changes to that. @jbmertens Would you able to try and figure out what's happening here? It's in the N2 pipeline, and I think something you've looked at recently?

Yeah, this does look like something that changed in Docker. The test passes on an older docker build from  # 1152. I'll try to see exactly what changed.

dstndstn:
Okay, I think this is _finally_ ready for review again!!

lib/stages/visTestPattern.cpp
```diff
@@ -238,6 +238,7 @@ void visTestPattern::main_thread() {
                     avg_err /= (float)num_bad;
 
                     // write frame report to outfile
+                    outfile.imbue(std::locale::classic());
```
andrerenard, line 241: So do we still need this?  I guess cannot hurt having it.

tests/boost/test_utils.hpp
```diff
@@ -4,8 +4,27 @@
...
 
 #include <boost/test/included/unit_test.hpp>
+#include <locale>
 #include <vector>
 
+// Use a boost::test "Global Fixture" to set the locale...
+// https://www.boost.org/doc/libs/1_75_0/libs/test/doc/html/boost_test/tests_organization/fixtures/global.html
+// Enable this in your test suite by adding:
+// BOOST_TEST_GLOBAL_FIXTURE(GlobalFixture_Locale);
+struct GlobalFixture_Locale {
```
andrerenard, line 14: Likewise, can we just remove this if we aren't going to use it anymore?

andrerenard (changes requested):

lib/stages/visTestPattern.cpp
dstndstn: Jim's opinion was to keep it, just to be explicit

tests/boost/test_utils.hpp
dstndstn: yes, definitely could.  I was setting the locale explicitly (to classic()) in kotekan.cpp, so I figured it would be good to at least have the mechanism to do the same in the tests, even if we're not using it.  What do you think?

andrerenard: That sounds reasonable.

andrerenard (approved):

PR #1169, opened 2024-02-06
CI: Switch to GCC-12 properly

eschnett:


eschnett:
We're now using `update-alternatives` instead of `rm` to make GCC 12 available. This should help https://github.com/kotekan/kotekan/pull/1159, where `cmake` currently says that we still have GCC 11 installed.

jbmertens (approved):
LGTM!

PR #1171, opened 2024-02-07
Make error macros safe to use as statements

eschnett:


jbmertens (approved):
LGTM! Good change to macros.

PR #1172, opened 2024-02-08
Reduce the E-field frame size from 32768 to 2048 time samples

eschnett:
This is a "temporary" change to speed up development and debugging.


dstndstn (approved):
Looks fine!

PR #1176, opened 2024-02-13
Baseband Receiver Updates

shinybrar:
The CHIME/FRB Receiver has been moved from the cfDn9.chime node on site to outrigger-buffer.chime.

### DIFF
--- a/config/chime_science_run_gpu.yaml
+++ b/config/chime_science_run_gpu.yaml
@@ -705,7 +705,7 @@ baseband_merge:
 
 # Transmit baseband data to the baseband receiver
 baseband_send:
-  server_ip: 10.6.213.19 # cfDn9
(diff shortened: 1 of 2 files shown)


shinybrar:
@andrerenard @jrs65 

- This PR is meant to against the [CHIME Site Release 2021.11b](https://github.com/kotekan/kotekan/releases/tag/2021.11b).
- Additionally, any downstream changes e.g. to Ansible Scripts to make sure on site release is current after this update also needs to be updated. If you can point me to the where that needs to be or documentation regarding it, I can go ahead and update them.

jrs65 (approved):

jrs65:
Although we probably should do, as far as I can tell we haven't actually specified the tag for the configs to install.

I think the only thing left to do is to actually create a tag for the version.

PR #1183, opened 2024-03-10
Update ASDF in Docker image (#1182)

eschnett:
Cherry-pick the change from `develop` to `chord`.


eschnett:
The Ubuntu 18.04 builds have been broken on the `chord` branch for some time. It would be tedious to correct this because it's mostly the lack of a newer GCC, and thus a lack for native `float16` support in C++, that causes this. The remedy would be to disable much of the new (actual CHORD) stuff in Kotekan. I suggest we instead disable the Ubuntu 18.04 tests.

dstndstn (approved):
Looks good!

PR #1187, opened 2024-03-19
Enable multiple GPU pipelines with Buffers for signalling (chord #1149)

dstndstn:
* make Buffer an actual C++ class, not a struct; insert a GenericBuffer base class
* add RingBuffer class
* enable multiple GPU-pipeline stages to be connected by signalling (possibly empty) Buffers
* add example multi-GPU-pipeline config files
* cudaProcesses share a cudaDeviceInterface
* make it so that gpuDeviceInterface does not know the gpu_buffer_depth


andrerenard:
This is now in develop via the chord->develop rename.

PR #1188, opened 2024-03-19
Use ringbuffer & metadata in cudaCorrelator kernel

dstndstn:
- assume the cudaCorrelator input is coming from a ring buffer
- propagate metadata from the input to the correlation matrix output
- fix testDataGen to produce output for more generator types (including random signed that we use in the test code)


lib/cuda/cudaCorrelator.hpp
```diff
@@ -41,13 +45,16 @@ class cudaCorrelator : public cudaCommand {
...
     ~cudaCorrelator();
+    int wait_on_precondition() override;
     cudaEvent_t execute(cudaPipelineState& pipestate,
                         const std::vector<cudaEvent_t>& pre_events) override;
+    void finalize_frame() override;
 
 protected:
 private:
     // Common configuration values (which do not change in a run)
     /// Number of elements on the telescope (e.g. 2048 - CHIME, 256 - Pathfinder).
```
eschnett: 128 for the pathfinder

lib/cuda/cudaCorrelator.hpp
```diff
@@ -56,15 +63,22 @@ class cudaCorrelator : public cudaCommand {
     // Number of time samples into each of the output correlation
     // triangles.  The number of output correlation triangles is the
     // length of the input frame divided by this value.
+    // must be a factor of 256.
```
eschnett: "factor" or "multiple" here?

lib/cuda/cudaCorrelator.cpp
```diff
@@ -33,25 +41,80 @@ cudaCorrelator::cudaCorrelator(Config& config, const std::string& unique_name,
...
+        //  num_freq x
+        //  num_elements x
+        //  num_elements x
+        //  2 complex components)
+        // In int32 format.
+        out_meta->type = chordDataType::int32;
+        out_meta->dims = 5;
+        out_meta->set_array_dimension(0, num_subintegrations, "Tc");
+        out_meta->set_array_dimension(1, _num_local_freq, "F");
+        out_meta->set_array_dimension(2, _num_elements, "E");
```
eschnett: Suggestion: change to "dish" and "polarization" instead of "element".

lib/cuda/cudaCorrelator.cpp
```diff
@@ -33,25 +41,80 @@ cudaCorrelator::cudaCorrelator(Config& config, const std::string& unique_name,
...
+        std::shared_ptr<chordMetadata> const out_meta = get_chord_metadata(out_mc);
+        *out_meta = *in_meta;
+        // Output shape is
+        // (nt_outer = samples_per_data_set / sub_integration_ntimes x
+        //  num_freq x
+        //  num_elements x
+        //  num_elements x
+        //  2 complex components)
+        // In int32 format.
+        out_meta->type = chordDataType::int32;
```
eschnett, line 112: There is a new metadata field `name` that describes the data itself, e.g. `E` for the voltage, `I` for the FRB intensity beams, etc. Maybe use `n²` here?

lib/stages/testDataGen.cpp
```diff
@@ -197,76 +197,54 @@ void testDataGen::main_thread() {
         gettimeofday(&now, nullptr);
         set_first_packet_recv_time(buf, frame_id, now);
 
+        std::shared_ptr<chordMetadata> chordmeta;
+        if (metadata_is_chord(buf, frame_id)) {
+            chordmeta = get_chord_metadata(buf, frame_id);
+            chordmeta->dims = (int)_array_shape.size();
+            for (int d = 0; d < chordmeta->dims; ++d)
+                chordmeta->dim[d] = _array_shape[d];
```
eschnett: Call `set_array_dimension` instead?

eschnett (changes requested):
I suggested changes in comments.

lib/cuda/cudaCorrelator.hpp
dstndstn: I think that was an old comment about the CHIME pathfinder!  Fixed!

lib/cuda/cudaCorrelator.hpp
dstndstn: Good call!, it should be multiple; fixed

lib/cuda/cudaCorrelator.cpp
dstndstn: Done!

lib/cuda/cudaCorrelator.cpp
dstndstn: Good idea, done!

lib/stages/testDataGen.cpp
dstndstn: good idea, done!

eschnett:
Mmmm, "fromage"...

dstndstn:
At first it was a typo but then I decided I liked it better!

eschnett (approved):

PR #1193, opened 2024-03-21
try to fix github action when merging back to chord

dstndstn:


jbmertens:
Per docs here,
https://docs.github.com/en/actions/learn-github-actions/variables
We might want the conditional to look something like
```
          if [ "${GITHUB_EVENT_NAME}" = "pull_request" || "${GITHUB_EVENT_NAME}" = "pull_request_target" ]; then
            echo "Dockerfile was not modified. Using image from branch ${GITHUB_BASE_REF}."
            IMAGE_NAME="${IMG_2204}:$(echo ${GITHUB_BASE_REF} | tr / .)"
          else
            echo "Dockerfile was not modified. Using image from branch ${GITHUB_REF_NAME}."
            IMAGE_NAME="${IMG_2204}:$(echo ${GITHUB_REF_NAME} | tr / .)"
          fi
```

jbmertens (approved):
Looks good (assuming it works)

dstndstn:
Let's see if it works!

PR #1198, opened 2024-04-02
cudaUtils: Correct C macros

eschnett:
- Evaluate the argument of `CHECK_CUDA_ERROR` and friends only once.
- Use parentheses.
- Wrap in `do ... while(0)`


jbmertens (approved):
Shades of 1171. LGTM!

PR #1202, opened 2024-05-08
testing for ubuntu 24.04

jbmertens:


.github/workflows/main.yml
```diff
@@ -26,10 +26,212 @@ env:
...
+    - name: Build kotekan
+      run: |
+        OPTS=(--rm --mount type=bind,src=$(pwd),target=/code/kotekan -w /code/kotekan/build-2404/ ${{ needs.build-docker-2404.outputs.docker_img_name }})
+        docker run "${OPTS[@]}" ccache -s
+        docker run "${OPTS[@]}" \
+          cmake -Wdev -Werror=dev -Wdeprecated -Werror=deprecated \
+            -DWERROR=ON \
+            -DCMAKE_LINK_WHAT_YOU_USE=ON \
+            -DCMAKE_BUILD_TYPE=Test \
+            -DUSE_ASDF=ON \
```
andrerenard, line 202: I think we should include `USE_CUDA` and `USE_OPENCL` here, since otherwise we could miss issues with the GPU related compile in 24.04.   Or we should have a separate build like in the 22.04 case for the GPU code.

jbmertens: Agreed - though, I was thinking to add these tests in a future PR?

andrerenard (approved):
If it makes more sense to break the PR, we can merge this now, and open another one for dealing with the CUDA related changes - as it appears those impact 22.04 as well.

.github/workflows/main.yml
jbmertens: Added!

andrerenard (approved):
Looks great!

PR #1204, opened 2024-06-10
Vendorize n2k and gputils

jbmertens:


lib/testing/testDataCheck.hpp
```diff
@@ -170,9 +173,11 @@ void testDataCheck<A_Type>::main_thread() {
         frames++;
 
         if (num_frames_to_test == frames) {
-            INFO("Test passed, exiting!");
-            TEST_PASSED();
-        }
+            if (!error) {
```
andrerenard: We should rethink this logic.   I like the idea of having it print the first `max_num_errors`, but we shouldn't treat `max_num_errors` as a "maximum acceptable number of errors", and exit for any number of errors in all cases (after printing the first max_num_errors).  Perhaps we should also rename `max_num_errors` to something like `num_errors_logged`?    On that note, the `error` boolean can likely be replaced with just `num_errors > 0`.

tools/docker/22.04/Dockerfile
```diff
@@ -8,8 +8,8 @@ ARG DEBIAN_FRONTEND=noninteractive
 # (see https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners).
 # This is to prevent the docker image getting build on a newer architecture and then failing when it
 # is loaded on an older one.
-ENV CFLAGS "-march=haswell"
-ENV CXXFLAGS "-march=haswell"
+ENV CFLAGS="-march=haswell"
+ENV CXXFLAGS="-march=haswell"
```
andrerenard, line 12: For the 22.04 image I'd be tempted to bump this up to a newer system.  But we'd have to lookup what the oldest system github CI is running these days.  Not to solve in this PR, just a note.

andrerenard (approved):
This all looks great, thanks!  I think we should just clean up the testDataCheck logic while we are in this space.

lib/testing/testDataCheck.hpp
```diff
@@ -126,30 +121,30 @@ void testDataCheck<A_Type>::main_thread() {
...
+                              std::abs((float)(first_value - second_value)),
+                              epsilon * std::abs((float)(first_value + second_value)));
                     }
                 }
-            } else { // N2 numbers are int
-                // INFO("Checking non float numbers-----------");
+            } else {
                 if (first_value != second_value) {
-                    if (num_errors++ < max_num_errors)
+                    error = true;
```
andrerenard: This `error` flag is kind of redundant, since `num_errors > 0` is equivalent, but I'll leave it to you if you want to remove before merging, or just leave it.

andrerenard (approved):

andrerenard (approved):
Ah, I see that extra was _already_ removed.

PR #1205, opened 2024-07-10
Changes to support the CHIME/Outrigger sites

andrerenard:
This change contains a number of updates to support the CHIME/Outrigger sites.  Some of this code was previously stuck in another branch due to issues with Ubuntu 18.04/22.04 compatibility.   This PR updates that branch to work with 22.04, as well as resolves a number of problems caused by updates to the ROCm/OpenCL API, and the DPDK version in 22.04. 

Key changes:

* Adds config files for the three Outrigger sites at KKO, GBO, and HCO
* Updates the OpenCL copy commands to work with the array of input buffers design used by the Outrigger sites
* Updates the OpenCL API to address some changes in the latest ROCm versions
* Fixes an issue with the OpenCL memory pointers being treated like normal pointers.   `cl_mem` pointers cannot be incremented directly on the host side and then passed to the OpenCL API (results in a seqfault in the ROCm runtime), instead they are now correctly generated with `clCreateSubBuffer`. 
* Fixes a race condition in the GPU process logic where the wrong `gpu_frame_id` could get set on the command objects, resulting in the wrong frames being released (and other possible issues).
* Disables profiling for OpenCL on the Outriggers due to what appears to be a performance regression in ROCm (see: https://github.com/ROCm/ROCm/issues/3357 ) 
* Correctly register host memory used in the zeroing command objects to prevent GPU pipeline stalls. 
* Fix an issue with the OpenCL process where profiling didn't get disabled when profiling was set to false in the config.
* Fix an potential issue with bufferMerge when used with buffers with more than one consumer (automatically force a deep copy).

Minor changes:
* Fix install scrips to include .j2 files.
* Fixes OpenCL find script for ROCm
* Updates the verify opencl test to use the old CHIME Pathfinder kernel, since the KVCorr kernel uses the now unsupported dot4 instructions. 
* Fixes and issue with the kotekan install RPATH
* Adds config string trimming to deal with `rocprof` extra output causing issues with the python yaml->json conversion.
* Adds new `get_array()` option to the config object, and fixes some warning with `bool` types
* Allow DPDK to have `null` packet handlers for ports we don't want to receive data on (used for some F-engine data dump configs). 
* Increase DPDK alignment window to deal with increased chance of alignment failure with increased links.
* Abstract OpenCL async copy operation (similar to the CUDA design)
* Fixes a few issues with the dot4 kernel (although note this might just get removed at some point due to recent hardware changes)
* Improves some baseband readout metrics
* Fixes a bug in the `gpu_port` in ICETelescope when using non-standard ports. 

TODO: Before merging:
- [x] Make the ordering in VisAccumulate a config option (Outrigger kernel uses a different ordering) 
- [ ] Make the buffer zero thread cpu affinity configurable
- [x] Decide on what to do with the site specific kotekan service files.


lib/core/buffer.cpp
```diff
@@ -328,7 +328,9 @@ int private_mark_frame_empty(Buffer* buf, const int id) {
         cpu_set_t cpuset;
         CPU_ZERO(&cpuset);
         // TODO: Move this to the config file (when buffers.c updated to C++11)
-        CPU_SET(5, &cpuset);
+        // Review note, this needs to be fixed in _this_ PR
```
jbmertens, line 331: Just flagging this... does something need to change here?

lib/dpdk/iceBoardHandler.hpp
```diff
@@ -93,7 +93,7 @@ class iceBoardHandler : public dpdkRXhandler {
         // We allow for the fact we might miss the first packet by upto 100 FPGA frames,
```
jbmertens, line 93: Update the comment too!

lib/stages/applyGains.cpp
```diff
@@ -545,7 +545,7 @@ std::optional<applyGains::GainData> applyGains::fetch_gains(std::string update_i
         std::tie(weights_1d, w_shape) = json_base64_to_array<uint8_t>(weight);
 
     } catch (std::exception& e) {
-        WARN("Failed to read gain and weight from cal broker response.: {}", e.what());
+        548("Failed to read gain and weight from cal broker response.: {}", e.what());
```
jbmertens, line 548: Is this change intentional?

jbmertens (approved):

PR #1209, opened 2024-08-26
Merge Erik's working branch

eschnett:


CMakeLists.txt
```diff
@@ -72,7 +73,8 @@ endif()
 # lots of warnings and all warnings as errors
 add_compile_options($<$<OR:$<COMPILE_LANGUAGE:C>,$<COMPILE_LANGUAGE:CXX>>:-Wall>)
 add_compile_options($<$<OR:$<COMPILE_LANGUAGE:C>,$<COMPILE_LANGUAGE:CXX>>:-Wextra>)
-add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wzero-as-null-pointer-constant>)
+# This warning produces too many false positives in the CUDA header files
```
jbmertens, line 76: Just flagging this - @andrerenard , @dstndstn , any objection to turning this off?

eschnett:
The parallel Python tests fail due to a 60-second timeout. Does anyone know what this means?

jbmertens:
Hi @eschnett - if I run these tests locally, they also hang (time out, zero cpu use).

> FAILED test_receiveflags.py::test_clear_flags - Failed: Timeout >60.0s
> ERROR test_applygains.py::test_eigen[file] - Failed: Timeout >60.0s
> ERROR test_vistestpattern.py::test_no_noise - Failed: Timeout >60.0s

Along with some other tests. Not sure exactly what is causing this.

I don't think these tests involve anything f-engine or even GPU related, so it must be something else? Most of the other changes seem pretty benign though 🤷

eschnett:
Can you tell me in which file `test_receiveflags` are defined?

eschnett:
How do I mark a test as "expected to fail"?

jbmertens (approved):
In the future, we'll want to make sure the file writer stage(s) are tested, and tests in this PR are run as part of CI. But for now, let's leave that for the future.

PR #1213, opened 2024-09-11
Add CHIME FRB beamformer

eschnett:
This excludes most of the auto-generated code to make the PR easy to read. This will build but not run.


dstndstn (approved):
This is much easier to see what is going on... but the code is still quite mystical to me!

So it... LGTM but I also don't believe I would see a glaring error!

PR #1214, opened 2024-09-11
Add libraries for GDAL file writing...

jbmertens:


andrerenard (approved):
LGTM!

Side note at some point I have it as a TODO to update the CMAKE to auto detect libraries so we don't need to always be remembering all the options we need.  But that's in the low priority task list.

PR #1215, opened 2024-09-13
Regenerate all kernels

eschnett:
Please approve without looking at the diff.


eschnett:
The test `test_no_noise_inputs` fails. Does this test use any of the Julia-generated kernels? If not there is no reason it should fail now. Maybe the failure is spurious?

eschnett:
It was spurious.

jbmertens (approved):

PR #1218, opened 2024-09-13
Switch to CHIME's offset encoding for 4+4-bit comples integers

eschnett:
This does not include any regenerated kernels.


dstndstn (approved):
Looks fine to me!

PR #1220, opened 2024-09-13
Regenerate kernels

eschnett:


dstndstn (approved):
In Erik We Trust

eschnett:
The CI is doing the heavy lifting. We should make the CI run some tests with the new kernels.

eschnett:
And now cmake complains about `lz4` missing??? There is nothing about cmake that changes in this PR.

PR #1219, opened 2024-09-13
Add lz4 naming in 22.04 docker image

jbmertens:

PR #1221, opened 2024-09-14
Eschnett/regenerate kernels

jbmertens:
Re-creating this pull request so it runs on an updated docker image...
(See PR #1220 and PR #1219 / lz4 library find failing)


jbmertens:
Do you not want to pull this in any more?
I'm looking into why the test was failing... guessing a false positive on a past test / something to do with log or output formatting rather than this PR itself.

jbmertens:
This is passing now. Would you like it merged @eschnett?
The error earlier seems spurious; it ran fine locally. Maybe some cache had to clear somewhere...

PR #1222, opened 2024-09-15
add some "direct" tests

jbmertens:


andrerenard (approved):

andrerenard (commented):
Actually I'm not sure if that works.  Because the last exit code will be the result of the `break` which will be a success?

andrerenard (approved):

PR #1224, opened 2024-09-16
Add F-Engine tests and benchmarks

eschnett:


config/fengine/include/fengine_pathfinder.j2
dstndstn, line 1: Are all these run_send_voltage stages the same?  Could they get factored out?

dstndstn (approved):
Looks fine to me!

jbmertens:
Do you want these tests run as part of the CI?
Things aren't running correctly on the GPU right now (#1222). Looking into it though.

config/fengine/include/fengine_pathfinder.j2
eschnett: They are quite similar. I tried factoring out but there are differences. It might work to do arithmetic as part of the Jinja2 stage. In particular I need `if` statements to disable some upchannelization stages for some instruments; this would combine all the input and output stages.

Maybe it is time for a Python wrapper that generates the YAML files.

eschnett:
The `*init*` stages take a long time to run. The `*test*` stages need the result from the `*init*` stages. And there is nothing that checks whether the output is correct yet. But yes, I want to run some tests automatically.

PR #1225, opened 2024-10-10
update n2k and gputils

jbmertens:


external/gputils/src_lib/test_utils.cu
```diff
@@ -372,6 +372,7 @@ INSTANTIATE_TEMPLATES(unsigned short);
 INSTANTIATE_TEMPLATES(unsigned char);
 INSTANTIATE_TEMPLATES(complex<float>);
 INSTANTIATE_TEMPLATES(complex<double>);
+INSTANTIATE_TEMPLATES(complex<int>);
```
eschnett, line 375: Technically speaking, the `INSTANTIATE_TEMPLATES` already provide the trailing semicolons, so the macro calls here should have not semicolon.

lib/testing/testDataCheck.hpp
```diff
@@ -141,7 +141,7 @@ void testDataCheck<A_Type>::main_thread() {
             }
         }
 
-        if (num_errors > 0) {
+        if (num_errors == 0) {
```
eschnett, line 144: That change looks like a bug.

jbmertens: I think it was a bug before. If there are no errors, print out some info. Otherwise, fail the test.

andrerenard:
Is this PR still relevant?

eschnett:
I think not – I think we have the current versions.

andrerenard:
Updated by other PRs.

eschnett:
Actually, I now see that these changes were excluded from other recent commits.

lib/testing/testDataCheck.hpp
geoffryan: Correct, this change fixes the bug.

jbmertens:
> Actually, I now see that these changes were excluded from other recent commits.

@eschnett - is this PR relevant though, or will those changes be pulled in by another PR?

eschnett:
I would need to check. This looks non-trivial.

jbmertens:
I think we can close this:
 - I'm pretty sure (after discussing with @geoffryan) that the current code in the chord branch includes the most recent changes form kendrick (i.e. this PR) *and* the most changes from Erik.
 - The diff github shows for this PR is not correct. (Known issue: "This is a result of the type of diff we use on GitHub. We use git’s three dot diff, which is the difference between the latest commit on the HEAD branch and the last common ancestor commit with the base branch.")
 - If I open a new PR, the diff is still misleading/"wrong".

So what is currently in the chord branch should be correct.

Except, I think we probably do want to remove the `argparse` file. (cc @eschnett @andrerenard  )

PR #1228, opened 2024-11-05
Create build_docker.yaml

jbmertens:


dstndstn:
taking a look...

dstndstn (approved):
It's a bit funny that the ".04" part of the Ubuntu version name is hard-coded ... looks like the immediate reason is because of a difference in naming scheme between the kotekan-2204 docker image names, and the tools/docker/22.04/Dockerfile path name?

Otherwise, looks fine, not that I have much experience in these files!

jbmertens:
Thanks! Yeah, just the 2204 vs 22.04. (And we only use LTS which is always 04).

PR #1233, opened 2025-01-07
CI updates

jbmertens:
- Update Blaze version
 - Keep docker images around on runner
 - Fix direct Kotekan test fail codes, add stub pipeline test


andrerenard (approved):
LGTM. Please add notes to description.

PR #1236, opened 2025-02-10
Implement generic metadata and nd-array types

eschnett:


eschnett:
The test failure for `build-full-tests-2404` seems unrelated.

lib/metadata/DataType.hpp
```diff
@@ -0,0 +1,198 @@
...
+        case uint16:
+            return 16;
+        case uint32:
+            return 32;
+        case uint64:
+            return 64;
+        case int4p4:
+            return 4;
+        case int4p4chime:
+            return 4;
```
andrerenard: Should these be 8-bits?   Same for `uint4p4`?   e.g. do you intend to index individual numbers or complex numbers?   It seems like based on the description each one of these is intended to represent 1 byte of data at a time.   Perhaps if we intend these to be 4-bits, they should just be generically `int4` and leave it to the user to infer complex or not?  otherwise it feels like we might want complex types for each size?

lib/metadata/Metadata.hpp
```diff
@@ -0,0 +1,186 @@
...
+
+// The Metadata class is roughly similar to a Python dictionary. It
+// holds typed key/value pairs. The keys are strings, and the values
+// are either boolen, integer, float, string, or a vector of these.
+//
+// Values cannot be overridden once set. This is a write-once data
+// structure.
+//
+// Internally, integers and floats are stored using 64 bits.
+class Metadata {
```
andrerenard, line 30: I like this as a possible replacement for the current metadata structures.   Although we will need some serialization and deserialization functions.  Doesn't necessarily need to be added now, but would be nice to have. since then we could start working on updating the network and basic file logic away form struct serialization.

lib/metadata/Metadata.hpp
```diff
@@ -0,0 +1,186 @@
...
+#include <vector>
+
+namespace chord {
+
+// The Metadata class is roughly similar to a Python dictionary. It
+// holds typed key/value pairs. The keys are strings, and the values
+// are either boolen, integer, float, string, or a vector of these.
+//
+// Values cannot be overridden once set. This is a write-once data
+// structure.
```
andrerenard, line 27: I'm not sure about making this write-once and not copyable.  There are places in the pipeline where it's nice to add or update a few parameters, but otherwise leave the rest of the metadata data intact while passing it on to the next stage.   There are also place the metadata gets copied between two branches of the pipeline and treated differently in each branch.

lib/metadata/DataType.hpp
```diff
@@ -0,0 +1,198 @@
...
+#if defined __FLT16_MAX__
+using float16_t = _Float16;
+#define KOTEKAN_FLOAT16 1
+#else
+// There is no float16 type
+#define KOTEKAN_FLOAT16 0
+#endif
+#endif
+
+namespace chord {
```
andrerenard: I'm not sure it makes sense to use namespace `chord` here, in theory these functions will get used by HIRAX and CHIME at some point.

lib/metadata/testSymbol.cpp
```diff
@@ -0,0 +1,37 @@
+#include <Symbol.hpp>
+#include <cassert>
+#include <iostream>
+#include <set>
+#include <unordered_set>
+
+using namespace chord;
+
+int main() {
```
andrerenard: Same here.

andrerenard (changes requested):
Looks great, but some comments inline.

lib/metadata/testSymbol.cpp
andrerenard: It seems one of my comments got lost.   This was to say these main() functions for tests should be converted to `boost::test`, see `tests/boost/*` as examples.   This way the tests will run as part of the CI process.

lib/metadata/DataType.hpp
eschnett: These are supposed to be SIMD types, combining two 4-bit values into one byte. What they represent (two real values or a single complex value) is not defined, that's up to the caller. I didn't want to call them `int4` because that doesn't convey the SIMD-ness of the values. `int4x2` might be another good name.

lib/metadata/Metadata.hpp
eschnett: Adding new attributes is possible, changing attribute values is not. We can add this, but we don't do that anywhere at the moment, and so it sounds like a nice design principle avoiding accidental errors.

Copying makes sense. Not sure whether I want a copy constructor for this or a more explicit operations since it's expensive. It's probably fine to stick with C++ semantics here and just allow a copy constructor.

lib/metadata/DataType.hpp
eschnett: True! I should have put this into `kotekan` instead.

However, we should use a namespace, just to avoid conflicts with other libraries.

lib/metadata/testSymbol.cpp
eschnett: Okay!

jbmertens (approved):

andrerenard (approved):
LGTM!  Thanks for the edits!

PR #1237, opened 2025-02-24
Fringestopping code

jbmertens:
- Fringestopping logic
 - Adds CHORD Telescope object
 - Metadata updates
 - Python tests


dstndstn (approved):

dstndstn:
It looks like we do have a couple of bugs exposed in the test cases.  I'll try to fix those tomorrow!

dstndstn:
I'm getting some test failures, including, eg, `test_time_downsample.py::test_structure`
where it's failing in the `timeDownsample` `main_thread` with a `Thread 21 "Downsample_test" received signal SIGFPE, Arithmetic exception.`

dstndstn:
Specifically:

```
0x0000558461133600 in timeDownsample::main_thread (this=0x558464926b30) at /code/kotekan/lib/stages/timeDownsample.cpp:76
76	            if (fpga_seq_start % (nsamp * frame.fpga_seq_length) != 0) {
```

It's getting `frame.fpgq_seq_length = 0`

But I guess this is new code, and we know it's not finished yet.  I'm investigating some possible regressions in old code.

dstndstn:
eg `test_valve.py::test_complete`:

`INFO: Fatal error message was: Failure copying frame: Metadata sizes must match for direct copy (src 200 != dest 160).`

(because of a `vis_pool` vs `N2_pool` difference)

geoffryan:
I don't believe the time_downsample tests reflect the last changes I made before leaving, can address tomorrow.

geoffryan:
After restoring some FakeVis and VisBuffer functionality (which may be removed in a later PR), all tests appear to pass.

The time <-> UT1 <-> ERA calculations are also now accurate to the nanosecond (our native time resolution, and far more than the accuracy of DUT1), invertible, and covered by unit tests.  Tests have been added for the ERA binning as well.   There are not yet automated tests for the fringestopped visibilities & coordinate transforms, but the visibilities have been checked offline.  Automated tests for these will be added in a future PR.

Note there is instability in the behaviour of test_vistestpattern.py:test_no_noise_inputs.  It seems to pass/fail in the Github workflow randomly.  It does not depend on anything in this PR, so I believe this behaviour is unrelated to the changes in this PR.

geoffryan (approved):

lib/stages/N2TimeDownsample.cpp
```diff
@@ -0,0 +1,340 @@
...
+                        size_t d_i = i % num_dishes;
+                        size_t d_j = j % num_dishes;
+                        output_frame.vis[idx] *= fringe_phase[d_i] * std::conj(fringe_phase[d_j]);
+
+                        idx++;
+                    }
+                }
+            }
+
+            for (size_t i = 0; i < nprod; i++) {
```
ssiegelx: Are there cases where the weight can be zero? Should we handle it explicitly, e.g.:
```
output_frame.weight[i] = (output_frame.weight[i] != 0.0) ? 1. / output_frame.weight[i]: 0.0;
```
I know that in the output data files, weight = 0.0 is used to indicate missing or fully masked sample.

lib/stages/N2TimeDownsample.cpp
```diff
@@ -0,0 +1,340 @@
...
+                output_frame.weight[i] = 1. / output_frame.weight[i];
+            }
+
+            for (uint32_t i = 0; i < num_eigenvectors; i++) {
+                for (uint32_t j = 0; j < num_elements; j++) {
+                    // Eigenvectors get phases too.
+                    int k = i * num_elements + j;
+                    size_t d_j = j % num_dishes;
+                    output_frame.evec[k] *= fringe_phase[d_j];
```
ssiegelx: We should check how the eigenvectors are phase referenced. Since the absolute phase of an eigenvector is arbitrary, I believe the kotekan eigendecomposition algorithm follows a convention where it references the phases with respect the first element. In that case, I think you’ll also want to match this convention during fringestopping, i.e.,
```
output_frame.evec[k] *= fringe_phase[d_i] * std::conj(fringe_phase[0]);
```

The same applies below when accumulating.

lib/stages/N2TimeDownsample.cpp
```diff
@@ -0,0 +1,340 @@
...
+                    // in case one of the dish phases is small.
+                    std::complex<double> w_doub = fringe_phase[d_i] * std::conj(fringe_phase[d_j]);
+
+                    // Now truncate the phase to a float to match vis[]
+                    // Have to be explicit about this, compiler complains
+                    // otherwise.
+                    N2::cfloat w_fs{(float)w_doub.real(), (float)w_doub.imag()};
+
+                    // Accumulate
+                    output_frame.vis[idx] += w_fs * frame.vis[idx];
```
ssiegelx:
```suggestion
                    output_frame.vis[idx] += w_fs * frame.n_valid_fpga_ticks_in_frame * frame.vis[idx];
```
Should we weight `frame.vis` by the number of sub-integrations that were averaged together to generate it? I believe this is the `frame.n_valid_fpga_ticks_in_frame` quantity?  Otherwise, a "complete" frame and a frame that is heavily masked due to RFI or packet loss would be given equal weight, even though the latter could be significantly noisier.

ssiegelx: This would change how the `vis` dataset is normalized below, and how the `weight`, `eval`, `evec`, and `erms` quantities are accumulated and normalized.

ssiegelx (commented):
The framework for specifying dish coordinates, performing the relevant coordinate transformations, and deriving the fringestop phase is very nicely done!  I just have a few comments/questions about the downsampling task.

andrerenard:
What's left (if anything) before we can merge this?  Just my final review?   I'm actually thinking we might be better to merge this before the metadata changes, as there is some unique metadata handling in here that may need to get reworked.  And the metadata PR (or something following it) might actually be a better place to do it, since this is already quite complex...

lib/stages/N2TimeDownsample.cpp
geoffryan: Correct!  This is addressed in commit b2a03ef .

lib/stages/N2TimeDownsample.cpp
geoffryan: Addressed in b1a2dd7.

lib/stages/N2TimeDownsample.cpp
geoffryan: Confirmed and fixed by e1d1f46.

geoffryan:
Once CI passes, this should be good to merge. Seth's suggestions have been incorporated, and some final refactoring on the N2 metadata.

PR #1240, opened 2025-05-09
Update CUDA linking/includes

jbmertens:


lib/metadata/CMakeLists.txt
```diff
@@ -14,3 +14,8 @@ add_library(
 
 target_link_libraries(kotekan_metadata PRIVATE kotekan_utils)
 target_include_directories(kotekan_metadata PUBLIC .)
+
+if(${USE_CUDA})
```
andrerenard, line 18: If this is needed by metadata, does the mean that kotekan won't build if CUDA isn't included?  or are those parts ifdef protected?

andrerenard (approved):
This all looks good.  But I would try a few configurations to make sure this is working in all cases.  Also see the inline comment, did we maybe create a cuda dependency for the base build?

lib/metadata/CMakeLists.txt
jbmertens: It is conditionally used here,
https://github.com/kotekan/kotekan/blob/chord/lib/metadata/DataType.hpp#L12
...so ifdef protected.

PR #1242, opened 2025-05-13
Update linking, includes, workflows, some HSA, some linting. NB: removes 1804

jbmertens:
This removes 18.04 support, and provides other changes for updated package/driver/toolkit/etc versions.


andrerenard (approved):
Yeah you can bypass the merge rules here to get around the docker rebuilds, let's just make sure we follow up with the fixes quickly.  

Can you just add to the title/description that this PR removes 18.04 support - since that's a fairly consequential change.

andrerenard:
And remove HSA test builds from the CI.

PR #1246, opened 2025-05-14
Update code from running new linter

jbmertens:


andrerenard (approved):
I agree with what the linter is doing here.  Seems like more of a bug fix than anything fundamental.

lib/testing/gpuBeamformSimulate.cpp
```diff
@@ -556,13 +556,13 @@ void gpuBeamformSimulate::main_thread() {
...
                             } // end for ff
-                        }     // end for tt
-                    }         // end for pol
+                        } // end for tt
+                    } // end for pol
                     cpu_final_output[out_id] = out_sq / 48. / BP[int((f + 8) % 16)];
 
                 } // end for freq
-            }     // end for time
-        }         // end for beam
```
andrerenard, line 565: I don't love this change to the linter, but we can live with it...  Although would be mildly tempted to remove the comment spacing rule, sometimes it's nice to line up comments like this.

jbmertens: Yeah, I don't like it either. This looks to be enforced by the `AlignTrailingComments: true` option, which I'm not sure we can turn off without seeing poor formatting elsewhere.

PR #1248, opened 2025-05-14
Rhaas/pulsar

rhaas80:
So mostly this is a "like so?" pull request, in particular the test code commit is just to ask what sorts of tests one would like to see.


rhaas80:
Hmm, the linter fails, though at least some (meaning "I have not found any that are not") of its warnings / errors are not in files that this pull request actually changes, eg.

https://github.com/kotekan/kotekan/actions/runs/15032589584/job/42248114071?pr=1248#step:6:537

jbmertens:
> Hmm, the linter fails, though at least some (meaning "I have not found any that are not") of its warnings / errors are not in files that this pull request actually changes, eg.
> 
> https://github.com/kotekan/kotekan/actions/runs/15032589584/job/42248114071?pr=1248#step:6:537

Don't worry about the iwyu errors, I'll address those in another PR soon.

Should the test_pulsar.py file be part of this PR? Maybe it could go in a subdirectory somewhere, although I'm not sure where is best. If you can make a pytest out of it, that would be neat.

rhaas80:
No `test_output.py` is not really part of what I'd propose to go in (obviously in the wrong location, completely undocumented, not called by anything). 

Everything in https://github.com/kotekan/kotekan/pull/1248/commits/f4fec36f8e8feb9dc07ea54f6e70f65b67fcd657 is really only there to request comments for what sort of unit tests one should add (ie anything that would test more than this)?

The actual code that would be inclusion ready is the C++ code in https://github.com/kotekan/kotekan/pull/1248/commits/4cfbb8b6986ca5a3cb8e2c25b57e0a7d515e54b9 though it is still a hack and it *does* disable some (never used) functionality. So again a "like so?" pull request.

lib/stages/pulsarNetworkProducer.hpp
```diff
@@ -0,0 +1,48 @@
...
+#define PULSAR_NETWORK_PRODUCER_H
+#include "Config.hpp" // for Config
+#include "Stage.hpp"
+#include "buffer.hpp"
+#include "bufferContainer.hpp" // for bufferContainer
+
+#include <string> // for string
+
+/**
+ * @class PulsarNetworkProducer
```
andrerenard, line 11: Likely worth including the word "test" or similar in the name of this stange, since it's only used for testing.   If we are keeping this stage for unit testing, it might be worth moving it to `lib/testing`, and update the comments in here.

lib/stages/pulsarNetworkProducer.cpp
```diff
@@ -0,0 +1,95 @@
...
+    struct timeval now;
+    int frame_id = 0;
+    uint64_t seq_num = 0;
+    stream_t stream_id;
+    stream_id.id = 0; // no idea what that does
+
+#define samples_per_dataset 16384
+#define num_pols 2
+#define num_freqs 16
+#define num_beams 16
```
andrerenard, line 51: Even if we just use this for testing, it's nice to have these as configurable options (with defaults)

lib/stages/pulsarPostProcess.hpp
```diff
@@ -104,7 +104,10 @@ class pulsarPostProcess : public kotekan::Stage {
     Buffer* pulsar_buf;
 
     /// Config variables
+    /// The number of GPUs to collect data from, each providing _num_freqs data
```
andrerenard: In the new CHIME design, I think we always expect the number of frequencies to be `>4*k` per GPU, so there isn't any need for logic to combine from multiple GPUs.

lib/stages/pulsarPostProcess.cpp
```diff
@@ -42,6 +43,13 @@ pulsarPostProcess::pulsarPostProcess(Config& config_, const std::string& unique_
 
     // Apply config.
     _num_gpus = config.get<uint32_t>(unique_name, "num_gpus");
+    // TODO: chord and the old chime Xengine code order input differently.
```
andrerenard: The new code _only_ needs to handle the new CHIME/CHROD ordering.  We can just get rid of the logic for the old system.   In fact we've just removed all the old CHIME GPU code.

lib/stages/pulsarPostProcess.cpp
```diff
@@ -52,10 +60,13 @@ pulsarPostProcess::pulsarPostProcess(Config& config_, const std::string& unique_
     _num_packet_per_stream = config.get<uint32_t>(unique_name, "num_packet_per_stream");
     _num_stream = config.get<uint32_t>(unique_name, "num_stream");
 
+    // RH: only 625 used in production CHIME, get rid of other format option
     assert(_timesamples_per_pulsar_packet == 625 || _timesamples_per_pulsar_packet == 3125);
```
andrerenard: The "modern kotekan" way to do this is to throw an `std::runtime_error` that will unwind the pipeline startup process and generate a nice error about it to the user (which can be an API).     Although actually given that we are only supporting the `625` case, we could just hard code it in.  There isn't a huge value in providing a configuration option that cannot be changed...

test_output.py
```diff
@@ -0,0 +1,99 @@
...
+#  frame[index + 2] = (i >> 8) & 0xff;
+#  frame[index + 3] = (i >> 0) & 0xff;
+# where
+
+samples_per_dataset = 16384
+num_pols = 2
+num_freqs = 16
+num_beams = 16
+
+for packet in range(0,5):
```
andrerenard, line 81: We can likely convert this into a pytest.   It should be possible to setup a mini-pipeline with the pytest logic, and then convert this from an output type script to a series of checks.

lib/stages/pulsarPostProcess.cpp
```diff
@@ -265,38 +303,79 @@ void pulsarPostProcess::main_thread() {
...
+                    assert(freq_block < num_out_frames);
+                    unsigned char* out_buf = (unsigned char*)out_frames[freq_block];
+                    const int gpu = thread_id / _num_freqs_per_gpu;
+                    uint8_t* in_buf_data = in_frame[gpu];
 
                     for (uint32_t psr = 0; psr < _num_pulsar_beams; ++psr) { // loop psr
                         for (uint32_t p = 0; p < _num_pol; ++p) {
                             uint32_t out_index = 0;
                             if (_timesamples_per_pulsar_packet == 3125) {
+                                assert(0 && "not supported anymore");
```
andrerenard: Just delete this entire block of code, and throw an error in the constructor (or remove this variable entirely).

lib/stages/pulsarPostProcess.cpp
```diff
@@ -265,38 +303,79 @@ void pulsarPostProcess::main_thread() {
...
+                               &psr_header, fpga_seq_num, &time_dummy,
+                               beam_coord,
+                               &thread_ids[freq_block * num_freqs_per_frame]);
+                }
+                time_now = time_dummy;
             }
 
             // now store samples into output buffer.
+            for (uint32_t i = 0; i < _num_gpus; ++i) {
+                assert(_samples_per_data_set * _num_pol * _num_beams * _num_freqs_per_gpu == in_buf[i]->frame_size);
```
andrerenard: This kind of check can actually go in the constructor, since the frame size is a property of the buffer, not the individual frames.   It's worth also checking the output buffer(s) frame size.

andrerenard (changes requested):
This all looks to be on the right track.   I think we can simplify the logic a little by removing compatibility with the existing CHIME kernels (remove the num_gpus, buffer sync, etc.).  

We should move the tests towards automated tests with the pytests framework.  Note that was designed around the `visFrameVIew` formatted buffers.  So it may require some modification. 

Once we are happy with this code, we should then update the network sending logic.  That one likely deserves a full rewrite, since the existing code will not work for the new instrument layout.  The layout there actually deserves some discussion, since the existing system uses a lot of VLANs to control packet flows, but we will be replacing some of the switches.  I will make a detailed ticket on how the new system should work.

rhaas80:
this is a somewhat ancient pull request. Will have to see if it is worthwhile keeping.

PR #1259, opened 2025-05-30
update gsl_lite

jbmertens:
GSL version update, since it's been throwing errors for a while. Nothing big here, just the namespace was renamed. The span accessor seems to have also changed.


lib/stages/RingMapMaker.cpp
```diff
@@ -154,7 +154,7 @@ void RingMapMaker::main_thread() {
                               input_frame.vis.begin() + (p + 1) * num_bl - 1,
                               special_vis.begin() + 1);
                     // Add missing cross-pol
-                    special_vis.at(0) = conj(input_frame.vis.at((p - 1) * num_bl));
+                    special_vis.at(0) = conj(input_frame.vis[(p - 1) * num_bl]);
```
jbmertens: TODO: look into

andrerenard (approved):
LGTM

PR #1260, opened 2025-06-03
get the clang build working

jbmertens:


.github/workflows/main.yml
```diff
@@ -461,23 +461,19 @@ jobs:
       env:
         CC: clang
         CXX: clang++
-      if: env.RUN_2404_CLANG_TESTS
       run: |
         OPTS=(--rm --mount type=bind,src=$(pwd),target=/code/kotekan -w /code/kotekan/build-2404/ ${{ needs.build-docker-2404.outputs.docker_img_name }})
         docker run "${OPTS[@]}" ccache -s
         docker run "${OPTS[@]}" cmake -Wdev -Werror=dev -Wdeprecated -Werror=deprecated \
-        -DWERROR=ON \
```
andrerenard, line 457: I only see one warning about unused dependencies.  Is it possible we can just exclude that one warning type and leave this on?   I think that's `-Wno-unused-direct-frontend-input-files`?

andrerenard (changes requested):

.github/workflows/main.yml
jbmertens: Ah, it wasn't actually using the clang compiler! There we go...

andrerenard:
Can I convert to draft until we fix the warnings?

lib/cuda/cudaShuffleAstron.cu
```diff
@@ -12,16 +12,17 @@ cudaShuffleAstron::cudaShuffleAstron(Config& config, const std::string& unique_n
...
-    _num_local_freq = config.get<int>(unique_name, "num_local_freq");
-    _samples_per_data_set = config.get<int>(unique_name, "samples_per_data_set");
-    _num_data_sets = config.get<int>(unique_name, "num_data_sets");
-    _block_size = config.get<int>(unique_name, "block_size");
-    _num_blocks = config.get<int>(unique_name, "num_blocks");
-    _buffer_depth = config.get<int>(unique_name, "buffer_depth");
-
-    _gpu_mem_voltage = config.get<std::string>(unique_name, "gpu_mem_voltage");
-    _gpu_mem_ordered_voltage = config.get<std::string>(unique_name, "gpu_mem_ordered_voltage");
+    _num_elements = config.get_value(unique_name, "num_elements").get<int>();
```
andrerenard, line 15: I'm curious why this needed to be rewritten this way in the `.cu` file.  Although since we don't compile `.cu` files directly much, it's likely not a big deal.

andrerenard (approved):
LGTM.

lib/cuda/cudaShuffleAstron.cu
jbmertens: Something to do with how clang handles templating... codex's explanation here:

> Config::get is implemented as two SFINAE’d overloads (one for arithmetic types, one for non‑arithmetic/bool). In CUDA translation units compiled with clang (or nvcc+clang host), those overloads can become brittle because CUDA compilers parse code twice (host/device), and clang’s CUDA front‑end is stricter about template instantiation and header usage inside device-compilation passes. Switching to get_value(...).get() avoids that overload set entirely and steers clear of those CUDA-specific pitfalls

Note that the get<int>() is a json call, so arithmetic operations won't be resolved. I didn't think that was significant since we aren't using Astron for CHORD anyways (and I considered deleting the file). But let me know if this is indeed a problem, since there's probably another better solution.

jbmertens:
I'm going to merge this in, but note the restriction on [cudaShuffleAstron.cu](https://github.com/kotekan/kotekan/pull/1260/files/3b438b2e1d5f76d841d3fb10e570aa74a6461e61#diff-5a3b348662292a6e52fcdc045eb0fd6fc1f342281e3aef9d84b1451dd0d2c235) above.

Also note that there are a number of additional compiler warnings that should be fixed or suppressed at some point.
https://github.com/kotekan/kotekan/actions/runs/17522952941/job/49770009424?pr=1260#logs

PR #1262, opened 2025-06-24
Update dataset broker tests

ljgray:
chime-experiment/comet#107 modifies the command line invocation to run comet. This PR updates the relevant tests to accommodate the updated syntax.

Also, the `comet_broker` fixture was being defined in multiple places, pretty much identically. I've moved the fixture to `conftest.py` and updated the relevant tests accordingly.

The dataset broker tests are being skipped here, but I think that's expected. They should get run in the dataset broker test suite (https://github.com/chime-experiment/comet/blob/master/.github/workflows/main.yaml#L94).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -14,6 +14,12 @@ import subprocess
 import os
 import sys
 import faulthandler
+import re
+import signal
+import shutil
+import tempfile
+
+from time import sleep
 
 faulthandler.enable()
 _fh_interval = os.environ.get("FAULTHANDLER_DUMP_INTERVAL")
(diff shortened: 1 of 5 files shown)


ljgray:
This is based against the `chord` branch at the moment, but I think that these tests should be included in ~`2021.11b`~ `2021.11_hotfix` as well, since that's currently running on CHIME.

ketiltrout (approved):
I think the changes here are fine, but we need to figure out what branch to use in the comet tests.

This shouldn't be added to `2021.11b` (nor `2021.11c`, which we currently have deployed to the telescope) because those are tags and not branches.

The best might be to append it to the `2021.11_hotfix` branch (where both of the above tags point), and switch the comet tests to use the head of that branch in its CI.

I don't know if it should be also applied to the chord branch or not.

ljgray:
> The best might be to append it to the `2021.11_hotfix` branch (where both of the above tags point), and switch the comet tests to use the head of that branch in its CI.

Yeah I think this is best for now.

> I don't know if it should be also applied to the chord branch or not.

Yeah, not sure about that. It should be easy to rebase this PR either way.

I've been asked to hold off on deploying this until the x-engine upgrade is done, so I think this is sort of on hold anyway.

ljgray:
@ketiltrout Given that we're making these changes as part of the upgrade, we should probably get this merged in. Any blockers?

PR #1264, opened 2025-07-07
configTracker

jbmertens:
- [x] Add logic to write configs
- [x] Add CI test(s) and existing fix broken ones
- [x] Add a compatibility flag for headers without the update bit
- ~~Add logic to handle configs from FPGAs?~~


lib/core/configTracker.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+        conn.send_json_reply(return_json);
+    }
+
+private:
+    /// Constructor, we don't want anyone to call this
+    ConfigTracker() = default;
+
+    /// List of hash/config pairs
+    std::map<std::string, nlohmann::json> _configs;
```
jbmertens: Change to something like: <host:port, <hash, config>>

lib/core/configTracker.cpp
```diff
@@ -0,0 +1,133 @@
...
+    }
+    std::string hash = canonicalHash(filtered_json);
+    
+    // Store the config in the map with its hash
+    // Note: This is a raw JSON object, not a Config object.
+    {
+        std::lock_guard<std::mutex> lock(_lock);
+
+        if (_configs.count(hash)) {
+            DEBUG_NON_OO("ConfigTracker: config already exists with hash: {}", hash);
```
jbmertens: Make this a fatal error.

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,150 @@
...
+     * @brief Registers this class with the REST server, creating the
+     *        /trackers end point
+     * @param rest_server The server to register with.
+     */
+    void register_with_server(restServer* rest_server) {
+        using namespace std::placeholders;
+        rest_server->register_get_callback("/config_tracker",
+                                           std::bind(&ConfigTracker::trackers_callback, this, _1));
+    }
+
```
jbmertens: Add file dump functionality

jbmertens:
Remaining issues [moved to description]

docs/sphinx/dev/dev_config_tracker.rst
```diff
@@ -0,0 +1,50 @@
+ConfigTracker
+=============
+
+Overview
+--------
+``ConfigTracker`` is a singleton service that records the *startup-time* configuration
```
andrerenard, line 6: Thanks for the excellent documentation!  

Do note, we will also want to use this to pull in the F-engine config as well, so it's not just limited to kotekan.   I'm not yet sure where it will be best to put the call get the F-engine config.  It could possibly be in the telescope object?   That could however be a detail added after merging this PR.

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,626 @@
...
+        check_num_configs_consistent();
+    }
+
+    /**
+     * @brief Write all configuration data to disk with detailed error reporting.
+     *
+     * @param directory The directory to write files to
+     * @return Number of configurations successfully written
+     */
+    size_t writeConfigsToDisk(const std::string& directory) const {
```
andrerenard: I'm tempted to make the errors in here fatal, or at least throw an exception on error so the caller can decide how critical is it.   Generally I would consider an inability to save out the state of the system to be a critical error, since otherwise we might keep recording data which doesn't have a reproducible state - and not notice it.

lib/core/kotekanLogging.hpp
```diff
@@ -48,127 +51,134 @@ enum class logLevel {
...
+#define DEBUG(m, ...)                                                                              \
     do {                                                                                           \
-        if (_member_log_level > 3) {                                                               \
-            internal_logging(LOG_DEBUG, __log_prefix, fmt(m), ##a);                                \
-        }                                                                                          \
+        if (_member_log_level > 3)                                                                 \
+            internal_logging(LOG_DEBUG, __log_prefix, fmt(m), ##__VA_ARGS__);                      \
     } while (0)
-#define DEBUG_NON_OO(m, a...)                                                                      \
+#define DEBUG2(m, ...)                                                                             \
```
andrerenard, line 70: Looks like we have some extra white space in this file.

lib/stages/bufferSend.cpp
```diff
@@ -162,6 +180,9 @@ void bufferSend::main_thread() {
             DEBUG("Sent frame: {:s}[{:d}] to {:s}:{:d}", buf->buffer_name, frame_id, server_ip,
                   server_port);
 
+            /// Set to true sine we've set a transmission.
```
andrerenard:
```suggestion
            /// Set to true since we've set a transmission.
```

andrerenard (changes requested):
This is looking very good.   Thank you for having very clear documentation. 

I think it just needs the changes to the bufferSend/Recv we talked about yesterday, and then it's ready to go.  We'll likely also want to add automatic calls to write out (or otherwise include this config tree into science data files), but that could be a separate PR.

lib/core/configTracker.hpp
jbmertens: Sure - I'll change this to be fatal.

lib/core/kotekanLogging.hpp
jbmertens: The line continuations are lined up ok I think. clang-format seems ok with this at least?

docs/sphinx/dev/dev_config_tracker.rst
```diff
@@ -0,0 +1,73 @@
...
+JSONs of all Kotekan instances participating in a pipeline. Each unique
+``(host, port)`` pair has exactly one configuration entry, keyed by a canonical
+hash of the JSON (with any ``updatable_config`` pruned).
+
+The tracker exposes two REST endpoints:
+
+- ``GET /config_tracker_configs`` — returns the stored configurations (optionally filtered by ``hash`` query arg).
+- ``GET /config_tracker_hashes`` — returns the map of config-hash → ``host:port``.
```
rhaas80, line 14: is the extra indirection `host:port` -> `hash` needed? Assuming there's not going to be duplicate `host::post` pairs then the hash does not really seem to be much shorter than `host:port`. Is this hostname or IP?

docs/sphinx/dev/dev_config_tracker.rst
```diff
@@ -0,0 +1,73 @@
...
+
+Why this exists
+^^^^^^^^^^^^^^^
+Downstream writers can persist the configuration that produced their data.
+Instead of shipping full configs with every frame, the sender only flags changes.
+The receiver then pulls missing configs via REST and caches them locally using the tracker.
+
+Tracker-combined-hash
+---------------------
+The tracker also maintains a **combined MD5** over all stored ``json_hash`` values.
```
rhaas80, line 24: is the algorithm used (MD5) something that is user-relevant (and fixed)?

docs/sphinx/dev/dev_config_tracker.rst
```diff
@@ -0,0 +1,73 @@
...
+This summarizes the *set* of known configurations on a node. It is inexpensive to compute
+and is used by the streaming stages to signal whether configuration state has changed
+since the last transmission.
+
+Operational Flow
+----------------
+
+1. **Startup registration**
+   The local node registers its full filtered config with the tracker and exposes the REST endpoints.
```
rhaas80: how does it do so? Not with the two endpoints document above, does it?

docs/sphinx/dev/dev_config_tracker.rst
```diff
@@ -0,0 +1,73 @@
...
+
+2. **Sending data**
+   The sender compares its current tracker-combined-hash to the one last sent. If changed,
+   it sets a flag in the frame header (``config_tracker_update = true``) to prompt receivers
+   to refresh configuration state.
```
rhaas80: the mechanism is opaque and not documented here it seems. This refers to Kotekan's buffer_send calls?

docs/sphinx/dev/dev_config_tracker.rst
```diff
@@ -0,0 +1,73 @@
...
+
+3. **Receiving data**
+   Upon seeing ``config_tracker_update = true``, the receiver calls
+   ``getUpstreamConfigs(client_ip, client_port)`` to retrieve any missing configs by hash.
```
rhaas80: where is `getUpstreamConfigs` defined?

docs/sphinx/dev/dev_config_tracker.rst
```diff
@@ -0,0 +1,73 @@
...
+   The receiver blocks further processing until required configs are present locally.
+
+Threading & Safety
+------------------
+- All public methods that touch shared state take internal locks.
+- The tracker is process-local; network exchange happens via the REST client/servers under the hood.
+- Hash collisions are unlikely in practice; if the same hash is mapped to different endpoints,
+  execution aborts to avoid misattribution.
```
rhaas80: why use the hash?

docs/sphinx/dev/dev_config_tracker.rst
```diff
@@ -0,0 +1,73 @@
...
+
+Per-Connection REST Ports
+-------------------------
+When receiving frames over the network, ``bufferRecv`` may need to pull upstream configurations
+from the sender’s REST server (only when the config tracker is enabled).
```
rhaas80: are we concerned with the extra (and unpredictable) network latency this can imply when doing a `bufferRecv`?

docs/sphinx/dev/dev_config_tracker.rst
```diff
@@ -0,0 +1,73 @@
...
+    type: bufferRecv
+    listen_port: 11024
+    use_config_tracker: true
+    upstream_rest_endpoints:
+      - "10.1.2.3:13000"
+      - "192.168.5.10:14080"
+
+Notes
+- This setting is only meaningful when ``use_config_tracker: true``.
+- If a client IP is not listed, the default port ``12048`` is used.
```
rhaas80: if the IP is not listed? What IP does it connect to then?

kotekan/kotekan.cpp
```diff
@@ -514,7 +514,6 @@ int main(int argc, char** argv) {
 
     restServer& rest_server = restServer::instance();
     std::vector<std::string> address_parts = regex_split(bind_address, ":");
-    // TODO validate IP and port
```
rhaas80, line 544: so it does validate, or we no longer validate?

lib/core/configTracker.cpp
```diff
@@ -0,0 +1,21 @@
+/**
+ * @file
+ * @brief Translation unit for ConfigTracker.
+ *
+ * The ConfigTracker implementation is header-only by design to keep related
+ * logic colocated and enable inlining where appropriate. This file provides
+ * a dedicated translation unit for the class and documents the intent; it also
+ * centralizes common includes used by the tracker.
```
rhaas80, line 8: hmm. The header file contains only the class definition as far as I can tell, does it? So this source file will always translate to an empty object file, will it?

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+#include <string>     // for string
+#include <sys/stat.h> // for stat
+
+namespace kotekan {
+
+/**
+ * @class ConfigTracker
+ * @brief Kotekan core component that tracks the (startup-time) configurations through a pipeline.
+ *
+ * This class must be registered with a kotekan REST server instance by
```
rhaas80: the class or its functions? Since the class does not inherit from anything it would seem the latter.

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+    ConfigTracker(const ConfigTracker&) = delete;
+    void operator=(const ConfigTracker&) = delete;
+
+    ~ConfigTracker() {}
+
+    /**
+     * @brief Struct to hold a (host, port) pair.
+     */
+    struct HostPort {
+        std::string host;
```
rhaas80, line 76: so the hostname. Since DNS is not case sensitive (but the hash functions is) this could lead to situations where differences in case lead to different hashes for the same system.

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+        uint16_t port;
+
+        bool operator==(const HostPort& other) const {
+            return host == other.host && port == other.port;
```
rhaas80, line 80: see above

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+        }
+        // < operator for std::map sorting
+        bool operator<(const HostPort& other) const {
+            return std::tie(host, port) < std::tie(other.host, other.port);
```
rhaas80, line 84: see above

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+
+        // Default constructor
+        ConfigInfo() = default;
+
+        // Constructor with all parameters
+        ConfigInfo(const nlohmann::json& config, const std::string& json_hash,
+                   const std::string& kotekan_version, const std::string& kotekan_build_branch,
+                   const std::string& kotekan_git_commit_hash,
+                   const std::string& kotekan_cmake_options) :
+            config(config), json_hash(json_hash), kotekan_version(kotekan_version),
```
rhaas80, line 116: so there is no check that the `json_hash` passed in actually is the one for the `config` data? Or is it separate from that anyway?

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+     * configuration JSON object.
+     */
+    struct ConfigInfo {
+        nlohmann::json config;
+        std::string json_hash;
+
+        std::string kotekan_version;
+        std::string kotekan_build_branch;
+        std::string kotekan_git_commit_hash;
+        std::string kotekan_cmake_options;
```
rhaas80, line 106: descriptions missing for all members.

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+                   const std::string& kotekan_version, const std::string& kotekan_build_branch,
+                   const std::string& kotekan_git_commit_hash,
+                   const std::string& kotekan_cmake_options) :
+            config(config), json_hash(json_hash), kotekan_version(kotekan_version),
+            kotekan_build_branch(kotekan_build_branch),
+            kotekan_git_commit_hash(kotekan_git_commit_hash),
+            kotekan_cmake_options(kotekan_cmake_options) {}
+
+        // Constructor from JSON
+        explicit ConfigInfo(const nlohmann::json& j) {
```
rhaas80, line 122: why no `from_json`  to match `nlohmann`'s convention (admittedly they use free-standing functions)?

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+     * @brief Check if the number of configs and hashes are consistent.
+     *
+     * This function checks if the number of configurations stored in _configs
+     * matches the number of hashes stored in _config_hashes.
+     *
+     * @returns True if the sizes are consistent, false otherwise.
+     */
+    bool check_num_configs_consistent() const {
+        std::lock_guard<std::mutex> lock(_lock);
+        // _configs and _config_hashes should always have the same size
```
rhaas80: how can they differ? If direct access is granted, replace with accessor function that ensures this property? If no direct access is possible, make this function private?

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+
+        if (_configs.size() != _config_hashes.size()) {
+            FATAL_ERROR_NON_OO(
+                "ConfigTracker: _configs and _config_hashes have different sizes: {} vs {}",
+                _configs.size(), _config_hashes.size());
```
rhaas80: :+1:  for actually showing the non-matching sizes.

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+            return false;
```
rhaas80: does it ever get here? the macro name `FATAL_ERROR_NON_OO` would indicate it does not (so one could add the correct attribute etc to it to tell the compiler do).

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+        }
+
+        return true;
```
rhaas80: a return values seems no needed assuming that `FATAL_ERROR_NON_OO` is indeed fatal.

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+     */
+    std::size_t n_configs() const {
+        check_num_configs_consistent();
+
+        std::lock_guard<std::mutex> lock(_lock);
+        return _configs.size();
+    }
+
+    /**
+     * @brief Get the canonical hash of a (jsonified) config.
```
rhaas80: is there non-canonical one we would provide? If not, I'd remove the word "canonical".

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+     *
+     * This function generates a consistent hash for a given configuration JSON object.
+     * The JSON object should have any "updatable_config" fields stripped before hashing.
```
rhaas80: these fields cannot be identified by this function and skipped while hashing I assume?

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+     * (This function only checks for that, it does not strip them itself.)
```
rhaas80: why? If it can check, it should be able to strip.

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+     *
+     * @param filtered_json The configuration JSON object to hash.
+     * @returns The canonical hash as a string.
+     */
+    std::string jsonHash(const nlohmann::json& filtered_json) const {
+        std::stringstream ss;
+
+        // nlohmann::json::dump() uses an alpha-ordered map for objects, so the
+        // config should be serialized in a consistent order.
```
rhaas80: should be? This is not guaranteed or has not been at least verified?

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+
+        // In order for this to hash configs correctly, this assumes the updatable_config
+        // field is removed, and versioning information has been added.
+        if (filtered_json.contains("updatable_config")) {
+            FATAL_ERROR_NON_OO(
+                "ConfigTracker: jsonHash called with updatable_config field present.");
```
rhaas80: so it *is* fatal

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+        }
+
+        // Stick to a string dump; less likely to run into floating point issues?
```
rhaas80: why? I'd have expected that a binary dump (storing the full double precision bit pattern) would be less sensitive.

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+        MD5(reinterpret_cast<const unsigned char*>(serialized.c_str()), serialized.size(),
+            md5_result);
+#pragma GCC diagnostic pop
+
+
+        std::stringstream md5_ss;
+        for (int i = 0; i < MD5_DIGEST_LENGTH; ++i)
+            md5_ss << std::hex << std::setw(2) << std::setfill('0')
```
rhaas80: isn't it sufficient to to `<< std::hex << std::setw(2) << std::setfill('0')` once? it does not that it gets reset after the statement.

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+        // maintain compatibility.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+        MD5(reinterpret_cast<const unsigned char*>(ss.str().c_str()), ss.str().size(), md5_result);
+#pragma GCC diagnostic pop
+
+        // Convert the MD5 result to a hex string
+        std::stringstream md5_ss;
+        for (int i = 0; i < MD5_DIGEST_LENGTH; ++i)
+            md5_ss << std::hex << std::setw(2) << std::setfill('0')
```
rhaas80: see above

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+                   << static_cast<int>(md5_result[i]);
+        {
+            std::lock_guard<std::mutex> lock(_lock);
+            // Store the combined hash
+            _tracker_hash = md5_ss.str();
+        }
+
+        DEBUG_NON_OO("ConfigTracker: Combined hash set to: {}", _tracker_hash);
```
rhaas80: does not protect access.

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+     * representing the current tracker state.
+     */
+    std::string getTrackerHash() const {
+        std::lock_guard<std::mutex> lock(_lock);
+        return _tracker_hash;
+    }
+
+    /**
+     * @brief Insert JSON configuration into the tracker.
+     * This function checks requires "updatable_config" fields have been removed.
```
rhaas80: does it check? Could it strip itself?

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+     * This is an overload that allows using a ConfigInfo object directly.
+     *
+     * @param host The host where the kotekan instance with the configuration is running.
+     * @param port The port where the kotekan instance with the configuration is running.
+     * @param config_info The configuration information to insert.
+     * @throws Error if a config with the same host and port already exists
```
rhaas80: of what type?

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+                    || _configs[host_port].kotekan_version != config_info.kotekan_version
+                    || _configs[host_port].kotekan_build_branch != config_info.kotekan_build_branch
+                    || _configs[host_port].kotekan_git_commit_hash
+                           != config_info.kotekan_git_commit_hash
+                    || _configs[host_port].kotekan_cmake_options
+                           != config_info.kotekan_cmake_options) {
+                    FATAL_ERROR_NON_OO("ConfigTracker: conflicting configuration data present for "
+                                       "host: {}, port: {}",
+                                       host, port);
+                    return;
```
rhaas80: never reached

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+            _config_hashes.emplace(config_info.json_hash, host_port);
+        }
+        // Sanity check to ensure that the number of configs is consistent
+        check_num_configs_consistent();
+
+        // Update the combined hash
+        setTrackerHash();
+        DEBUG_NON_OO("ConfigTracker: inserted config for host: {}, port: {}, hash: {}", host, port,
+                     config_info.json_hash);
+        DEBUG_NON_OO("ConfigTracker: _tracker_hash: {}", _tracker_hash);
```
rhaas80: does not protect access

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+    }
+
+    /**
+     * @brief Insert JSON configuration into the tracker.
+     * Creates a ConfigInfo object from the parameters.
+     * This function will strip "updatable_config" fields.
```
rhaas80: see above

rhaas80: actually no, not see above but: "Why is this different from above?"

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+     *
+     * @param host The host where the kotekan instance with the configuration is running.
+     * @param port The port where the kotekan instance with the configuration is running.
+     * @param config_json The JSON configuration to insert.
+     * @param kotekan_version The version of Kotekan.
+     * @param kotekan_build_branch The build branch of Kotekan.
+     * @param kotekan_git_commit_hash The git commit hash of Kotekan.
+     * @param kotekan_cmake_options The CMake options used to build Kotekan.
```
rhaas80, line 201: since it calls `insertConfig(std::string host, uint16_t port, ConfigInfo config_info)` which can `throw`, should that be documented here?

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+
+    /**
+     * @brief Determine if a config exists in the _configs map.
+     *
+     * @param hash The hash of the config.
+     * @returns True if the config exists, false otherwise.
+     */
+    bool hasConfig(std::string host, uint16_t port) const {
+        std::lock_guard<std::mutex> lock(_lock);
+        return _configs.count({host, port}) != 0;
```
rhaas80: does an explicit `!= 0` while line 283 above relies on the implicit conversion.

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+    }
+
+    /**
+     * @brief check hasConfig using a hash string instead of host and port.
+     */
+    bool hasConfig(std::string hash) const {
+        std::lock_guard<std::mutex> lock(_lock);
+        return _config_hashes.count(hash) != 0;
```
rhaas80: see above

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+            std::string host_port_str = host_port.host + ":" + std::to_string(host_port.port);
+            return_json[host_port_str] = nlohmann::json::object();
+
+            // Add the config JSON and metadata to the return JSON
+            return_json[host_port_str]["config"] = info.config;
+            return_json[host_port_str]["json_hash"] = info.json_hash;
+            return_json[host_port_str]["kotekan_version"] = info.kotekan_version;
+            return_json[host_port_str]["kotekan_build_branch"] = info.kotekan_build_branch;
+            return_json[host_port_str]["kotekan_git_commit_hash"] = info.kotekan_git_commit_hash;
+            return_json[host_port_str]["kotekan_cmake_options"] = info.kotekan_cmake_options;
```
rhaas80, line 280: can there really be more than one result for a single hash?

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+    void trackers_hashes_callback(connectionInstance& conn) {
+        check_num_configs_consistent();
+
+        nlohmann::json return_json = {};
+        std::lock_guard<std::mutex> lock(_lock);
+        for (const auto& hash : _config_hashes) {
+            // Add the hash and its corresponding host:port to the return JSON
+            return_json[hash.first] = nlohmann::json::object();
+            return_json[hash.first]["host"] = hash.second.host;
+            return_json[hash.first]["port"] = hash.second.port;
```
rhaas80, line 303: why not return the string `host_port.host + ":" + std::to_string(host_port.port)` that is used by `trackers_configs_callback`?

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+        }
+        conn.send_json_reply(return_json);
+    }
+
+    /**
+     * @brief Registers this class with the REST server, creating the
```
rhaas80: registers this object I'd say.

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+     *        /trackers end point
+     * If a hash is provided, it will return the config for that hash.
+     * @param rest_server The server to register with.
```
rhaas80, line 312: seems to be either incorrect or missing the description for the 2nd endpoint, namely the one that returns all known `host:port` pairs.

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+            ERROR_NON_OO(
+                "ConfigTracker: Failed to get config hashes from upstream host: {}, port: {}", host,
+                port);
+            return;
+        }
+
+        // Check if the response contains any hashes
+        if (reply.second.empty()) {
+            ERROR_NON_OO("ConfigTracker: No configs found at upstream host: {}, port: {}", host,
+                         port);
```
rhaas80, line 349: is that always an error? Even very early on (ie are we guaranteed that that the upstream tracker has some configs before we query it first)?

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+
+        // Convert the reply string to a JSON object
+        nlohmann::json response_json;
+        try {
+            response_json = nlohmann::json::parse(reply.second);
+        } catch (const nlohmann::json::parse_error& e) {
+            ERROR_NON_OO("ConfigTracker: Failed to parse JSON response from upstream host: {}, "
+                         "port: {}. Error: {}",
+                         host, port, e.what());
+            return;
```
rhaas80, line 363: include the faulty string (at least as a debug option)?

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+                continue;
+            }
+            // Parse the response JSON
+            nlohmann::json config_response_json;
+            try {
+                config_response_json = nlohmann::json::parse(reply.second);
+            } catch (const nlohmann::json::parse_error& e) {
+                ERROR_NON_OO("ConfigTracker: Failed to parse JSON response for hash: {} from "
+                             "upstream host: {}, port: {}. Error: {}",
+                             hash, upstream_host, upstream_port, e.what());
```
rhaas80, line 410: see above

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+
+            // Check if the response contains the config under the owner
+            // (upstream_host:upstream_port)
+            std::string host_port_str = upstream_host + ":" + std::to_string(upstream_port);
+            if (config_response_json.contains(host_port_str)) {
+                ConfigInfo info = ConfigInfo(config_response_json[host_port_str]);
+                insertConfig(upstream_host, upstream_port, info);
+            } else {
+                // If the config was not found, log an error or take appropriate action
+                ERROR_NON_OO("ConfigTracker: Config not found for hash: {}", hash);
```
rhaas80, line 427: it never takes any action

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+     * @brief Write all configuration data to disk with detailed error reporting.
+     *
+     * @param directory The directory to write files to
+     * @return Number of configurations successfully written
+     */
+    size_t writeConfigsToDisk(const std::string& directory) const {
+        // Check if directory exists and is writable
+        struct stat info;
+        if (stat(directory.c_str(), &info) != 0) {
+            std::string err = "Directory does not exist: " + directory;
```
rhaas80: not necessarily, could be any other error. `man 2 stat` says

```
 On error, -1 is returned, and errno is set to indicate the error.
 ```
 
 so `strerror` is our friend here.

File: lib/core/configTracker.hpp [hunk 55]
```diff
@@ -0,0 +1,650 @@
...
+            FATAL_ERROR_NON_OO("ConfigTracker: {}", err);
+        }
+        if (!(info.st_mode & S_IFDIR)) {
```
rhaas80: maybe better use ` S_ISDIR(m)`? The correct test would be:

```
if ((info.st_mode & S_IFMT) == S_IFDIR)
```

if I understand `man inode`'s example correctly:

```
       The following mask values are defined for the file type:
           S_IFMT     0170000   bit mask for the file type bit field

           S_IFSOCK   0140000   socket
           S_IFLNK    0120000   symbolic link
           S_IFREG    0100000   regular file
           S_IFBLK    0060000   block device
           S_IFDIR    0040000   directory
           S_IFCHR    0020000   character device
           S_IFIFO    0010000   FIFO

       Thus, to test for a regular file (for example), one could write:

           stat(pathname, &sb);
           if ((sb.st_mode & S_IFMT) == S_IFREG) {
               /* Handle regular file */
           }
```

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+            std::string err = "Path is not a directory: " + directory;
+            FATAL_ERROR_NON_OO("ConfigTracker: {}", err);
+        }
+
+        std::lock_guard<std::mutex> lock(_lock);
+        size_t written = 0;
+
+        for (const auto& [host_port, config_info] : _configs) {
+            // Create filename - use underscore instead of colon for portability
```
rhaas80, line 458: In which case one also should not rely on `/` being the path separator. At least if we are concerned about he win32 API https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file

Are we really expecting to run any of this on macOS or Windows?

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,650 @@
...
+                    file << config_info.to_json().dump(4, ' ', false,
+                                                       nlohmann::json::error_handler_t::strict);
+
+                    if (!file.good()) {
+                        FATAL_ERROR_NON_OO("ConfigTracker: Write failed");
+                    }
+                } // file closed
+
+                // Atomically rename temp file to final name
+                if (std::rename(temp_filename.c_str(), filename.c_str()) != 0) {
```
rhaas80, line 484: will fail on windows or be non-atomic (`MoveFile` does not allow existing targets, https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-movefile?redirectedfrom=MSDN ).

The C++ standard does not guarantee overwriting either (https://cppreference.net/cpp/io/c/rename.html):

> If new_filename exists, the behavior is implementation-defined.

Are we expecting to run this on Windows?

lib/core/kotekanLogging.hpp
rhaas80: tab vs space? (no tabs, pleeeaaasseee).

lib/core/kotekanLogging.hpp
```diff
@@ -48,127 +51,134 @@ enum class logLevel {
...
+
+/**
+ * \enum logLevel
+ * \brief Log level
+ * \note Both DEBUG and DEBUG2 are removed entirely when building in release mode.
+ * \note The macros support fmt's python style string formatting only.
+ * \note The deprecated macros with a `_F` suffix are to be used in C code only and only offer
+ *       printf-style string formatting. They can be found in errors.h.
+ */
+enum class logLevel {
```
rhaas80, line 173: use these `enum` values in the various `ERROR` etc macros in their `if(log_level > foo` condition?

lib/core/kotekanMode.cpp
```diff
@@ -14,6 +15,7 @@
 #include "metadataFactory.hpp"   // for metadataFactory
 #include "prometheusMetrics.hpp" // for Metrics
 #include "restServer.hpp"        // for restServer, connectionInstance
+#include "version.h" // for get_kotekan_version, get_cmake_build_options, get_git_commit_hash
```
rhaas80, line 17: `.h` and not `.hpp`?

lib/core/restServer.cpp
```diff
@@ -56,9 +60,105 @@ restServer::~restServer() {
     }
 }
 
+// Validation function that accepts both IPs and hostnames
+bool restServer::isValidAddress(const std::string& address) {
```
rhaas80: document the valid characters. DNS allows any octet in the name.

lib/core/restServer.hpp
```diff
@@ -139,6 +139,24 @@ class restServer {
      */
     static restServer& instance();
 
+    /**
+     * @brief Checks if the given string is a valid address (IPv4, IPv6, or hostname)
+     */
+    bool isValidAddress(const std::string& address);
```
rhaas80: add a `const` to this function (and others that don't change their parent object)?

lib/stages/bufferRecv.cpp
```diff
@@ -61,6 +64,37 @@ bufferRecv::bufferRecv(Config& config, const std::string& unique_name,
     num_threads = config.get_default<uint32_t>(unique_name, "num_threads", 1);
     connection_timeout = config.get_default<int>(unique_name, "connection_timeout", 60);
     drop_frames = config.get_default<bool>(unique_name, "drop_frames", true);
+    use_config_tracker = config.get_default<bool>(unique_name, "use_config_tracker", true);
+
+    // Optional per-connection upstream REST port overrides (list of "host:port" strings)
```
rhaas80: if `parts.size() != 2` aren't the ports not required then?

rhaas80: or should I read "port" is 'host:port' pair?

lib/stages/bufferRecv.cpp
```diff
@@ -61,6 +64,37 @@ bufferRecv::bufferRecv(Config& config, const std::string& unique_name,
...
+    if (config.exists(unique_name, "upstream_rest_endpoints")) {
+        try {
+            auto entries =
+                config.get<std::vector<std::string>>(unique_name, "upstream_rest_endpoints");
+            for (const auto& ep : entries) {
+                auto parts = regex_split(ep, ":");
+                if (parts.size() != 2) {
+                    WARN("Invalid upstream_rest_endpoints entry: {:s} (expected host:port)", ep);
+                    continue;
```
rhaas80: should this not abort instead of using an expected host:port value?

lib/stages/bufferRecv.cpp
```diff
@@ -61,6 +64,37 @@ bufferRecv::bufferRecv(Config& config, const std::string& unique_name,
...
+                uint16_t port = 0;
+                try {
+                    int p = std::stoi(parts[1]);
+                    if (p < 0 || p > 65535) {
+                        throw std::out_of_range("port out of range");
+                    }
+                    port = static_cast<uint16_t>(p);
+                } catch (const std::exception& e) {
+                    WARN("Invalid port in upstream_rest_endpoints entry: {:s} ({:s})", ep,
+                         e.what());
```
rhaas80: this just catches the throw from three lines above, yes? Why use a throw/catch pair then?

lib/stages/bufferRecv.cpp
```diff
@@ -391,71 +437,106 @@ void connInstance::internal_read_callback() {
...
                         return;
                     }
-                    if (metadata_size != buf_frame_header.metadata_size) {
-                        ERROR("Metadata size does not match between server and client!");
-                        decrement_ref_count();
-                        close_instance();
+                    bytes_read += n;
+                    DEBUG2("Header read bytes: {:d}, bytes_read {:d}, expected: {:d}", n,
+                           bytes_read, sizeof(bufferFrameHeader));
+                    assert(bytes_read == sizeof(bufferFrameHeader));
```
rhaas80, line 446: hmm, so this does not actually allow multiple reads and always assumes `bytes_read` starts from 0?

lib/stages/bufferRecv.cpp
```diff
@@ -391,71 +437,106 @@ void connInstance::internal_read_callback() {
...
+                    n = read(fd, (void*)(((int8_t*)&buf_frame_header) + bytes_read),
+                             sizeof(bufferFrameHeaderNoConfigTracker) - bytes_read);
+                    if (n <= 0) {
+                        handle_error("reading header", errno, n);
                         return;
                     }
+                    bytes_read += n;
+                    DEBUG2("Header read bytes: {:d}, bytes_read {:d}, expected: {:d}", n,
+                           bytes_read, sizeof(bufferFrameHeaderNoConfigTracker));
+                    assert(bytes_read == sizeof(bufferFrameHeaderNoConfigTracker));
```
rhaas80, line 465: same as above

lib/stages/bufferRecv.cpp
```diff
@@ -391,71 +437,106 @@ void connInstance::internal_read_callback() {
...
                 if (n <= 0) {
                     handle_error("reading header", errno, n);
                     return;
                 }
                 DEBUG2("Metadata read bytes: {:d}", n);
                 bytes_read += n;
-                if (bytes_read >= buf_frame_header.metadata_size) {
-                    assert(bytes_read == buf_frame_header.metadata_size);
+                if (bytes_read >= metadata_size) {
+                    assert(bytes_read == metadata_size);
```
rhaas80, line 508: somewhat strange construction:

```
if(a >= b)
  assert(a==b)
  a == 0;
```

this could miss the beginning of the next package. Shouldn't this be `bytes_read -= metadata_size` if one wants to reset the stream pointer?

lib/stages/bufferRecv.cpp
```diff
@@ -391,71 +437,106 @@ void connInstance::internal_read_callback() {
...
                     return;
                 }
                 bytes_read += n;
                 DEBUG2("Frame read bytes: {:d}, total read: {:d}", n, bytes_read);
-                if (bytes_read >= buf_frame_header.frame_size) {
-                    assert(bytes_read == buf_frame_header.frame_size);
+                if (bytes_read >= frame_size) {
+                    assert(bytes_read == frame_size);
                     state = connState::finished;
                     bytes_read = 0;
```
rhaas80, line 524: same as above

lib/stages/bufferRecv.hpp
```diff
@@ -56,6 +57,10 @@ class connInstance;
  * @conf num_threads         Int, default 1.  The number of worker threads to use
  * @conf connection_timeout  Int, default 60.  Number of seconds before timeout on transfer
  * @conf drop_frames         Bool, default true.  Whether to drop frames when buffer fills.
+ * @conf upstream_rest_endpoints  List[str], default empty. Optional list of
+ *        "host:port" entries specifying non-standard upstream REST ports to use for
+ *        particular senders. If the client IP (as seen by bufferRecv) matches a host in this
```
rhaas80, line 65: it does not match IP (addresses) does it? It matches host names.

rhaas80: since this is a replacement, `host:port` seems a bit confusing to me (since it is not the full pair but just a replacement, note: why do we need the replacement and cannot provide the port as part of the list of hostnames that are being used?). Something like `host -> port` instead or so?

lib/stages/bufferRecv.hpp
```diff
@@ -125,9 +130,15 @@ class bufferRecv : public kotekan::Stage {
...
     bool drop_frames;
 
+    /// Optional per-host overrides for upstream REST ports
+    std::map<std::string, uint16_t> upstream_rest_port_overrides;
+
     /// A lock on the current frame, since many systems may ask for the next frame
     std::mutex next_frame_lock;
 
+    /// Whether to use the config tracker
+    bool use_config_tracker;
```
rhaas80: make this `const` if possible (since it cannot ever be changed)?

lib/stages/bufferRecv.hpp
```diff
@@ -298,15 +310,20 @@ class connInstance : public kotekan::kotekanLogging {
     /// The start time of a new frame read
     double start_time;
 
-    /// The buffer transfer header
-    struct bufferFrameHeader buf_frame_header;
+    /// Whether to use the config tracker
+    bool use_config_tracker;
```
rhaas80: I am confused. Another one of these? What about the one in lin 140?

lib/stages/bufferSend.cpp
```diff
@@ -93,45 +99,84 @@ void bufferSend::main_thread() {
...
 
             auto meta = buf->get_metadata(frame_id);
+            auto metadata_size = meta->get_serialized_size();
+            auto frame_size = buf->frame_size;
+            size_t header_len;
 
-            header.frame_size = buf->frame_size;
-            header.metadata_size = meta->get_serialized_size();
+            // Send header
+            // Use NoTrack header version if there is a config setting to enable it.
```
rhaas80: NoTrack?

lib/stages/bufferSend.cpp
```diff
@@ -93,45 +99,84 @@ void bufferSend::main_thread() {
...
+                DEBUG2("frame_size: {:d}, metadata_size: {:d}", header.frame_size,
+                       header.metadata_size);
+
+                // Recover from partial sends
+                DEBUG2("Sending header");
+                while ((n = send(socket_fd, &((uint8_t*)&header)[n_sent], header_len - n_sent,
+                                 MSG_NOSIGNAL))
+                       > 0) {
+                    n_sent += n;
+                }
```
rhaas80, line 151: I think writing 0 bytes is technically allowed and not an error.

lib/stages/bufferSend.hpp
```diff
@@ -20,26 +21,41 @@
 
 /**
  * @struct bufferFrameHeader
- * @brief Internal struct for sending the transfer details.
+ * @brief Stable wire header for transfer details.
+ *
+ * Uses fixed-width types and explicit padding to make the layout
+ * consistent across compilers/ABIs without relying on packing pragmas.
```
rhaas80, line 28: not actually guaranteed. A compiler is still allowed to align, nothing in the standard actually guarantees that eg:

```
struct s {
  int a, b;
};
```

is the same size as a `int[2]`.

lib/stages/bufferSend.hpp
```diff
@@ -20,26 +21,41 @@
...
  */
-#pragma pack()
 struct bufferFrameHeader {
     uint32_t metadata_size;
     uint32_t frame_size;
+    /// 0 = no update, 1 = update
+    uint32_t config_tracker_update;
 };
+static_assert(sizeof(bufferFrameHeader) == 12, "bufferFrameHeader must be 12 bytes");
```
rhaas80: :+1: 

Well it does not have to be 12 bytes, it is just unexpected if it is not.

lib/stages/bufferSend.hpp
```diff
@@ -20,26 +21,41 @@
...
- * @brief Sends a buffer and metadata over TCP.
+ * @struct bufferFrameHeaderNoConfigTracker
+ * @brief Wire header without the config tracker flag.
+ */
+struct bufferFrameHeaderNoConfigTracker {
+    uint32_t metadata_size;
+    uint32_t frame_size;
+};
+static_assert(sizeof(bufferFrameHeaderNoConfigTracker) == 8,
+              "bufferFrameHeaderNoConfigTracker must be 8 bytes");
```
rhaas80: see above

lib/stages/bufferSend.hpp
```diff
@@ -114,6 +136,9 @@ class bufferSend : public kotekan::Stage {
     /// Set to true if there is an active connection
     std::atomic<bool> connected;
 
+    /// Set to true if this is the first transmission
+    std::atomic<bool> first_transmission_sent;
```
rhaas80, line 142: an alternative to this would be for the client to request a new config blob on startup. This would avoid issues of the server itself restarting and sending the blob to the client unexpectedly (can this happen?)

tests/boost/run_boost_tests.sh
```diff
@@ -0,0 +1,95 @@
...
+
+for exe in "${EXECUTABLES[@]}"; do
+    echo "Running $exe..."
+    # Use temporary files to capture output
+    tmp_output=$(mktemp)
+    tmp_stderr=$(mktemp)
+
+    if [ $VERBOSE -eq 1 ]; then
+        # Verbose mode
+        output=$(timeout "$TIMEOUT" "$exe" 2>&1)
```
rhaas80, line 63: merges stdout and stderr, does this confuse the testing framework? (Complicated) workarounds exist in that case, that do show ouptut "life" without first spooling to a file.

tests/boost/run_boost_tests.sh
```diff
@@ -0,0 +1,95 @@
...
+fi
+
+# Validate timeout is a positive integer
+if ! [[ "$TIMEOUT" =~ ^[0-9]+$ ]] || [ "$TIMEOUT" -le 0 ]; then
+    echo "Error: Timeout must be a positive integer" >&2
+    exit 1
+fi
+
+# Find all executable files in the directory that start with "test_"
+mapfile -t EXECUTABLES < <(find "$TEST_DIR" -maxdepth 1 -type f -executable -name "test_*")
```
rhaas80, line 42: no symbolic links are expected for executables (those would not match `f`)?

tests/boost/run_boost_tests.sh
```diff
@@ -0,0 +1,95 @@
...
+    fi
+    echo "----------------"
+done
+
+# Summary
+echo "Summary: $passed passed, $failed failed"
+
+# List failed tests
+if [ ${#failed_tests[@]} -gt 0 ]; then
+    echo "Failed tests:"
```
rhaas80: this goes to stdout but the names go to stderr in line 91. Probably not what we want.

rhaas80 (commented):
mostly docs that I comment on. I don't think I found an actual bug.

kotekan/kotekan.cpp
jbmertens: It does now perform some basic validation (restServer::isValidAddress).

lib/core/kotekanMode.cpp
jbmertens: Yes, it is c-compatible code I think.
We could re-name it everywhere, but 🤷

lib/core/kotekanLogging.hpp
jbmertens: No tabs, I think the diff is just because the line continuations were re-aligned.

lib/stages/bufferSend.hpp
jbmertens: I believe this happens, yes. But, re-sending is a good check that the server has started up with the correct configuration data. Asking when making a connection is perhaps better than using this flag, anyways?

tests/boost/run_boost_tests.sh
jbmertens: They could exist in principle, but the boost tests expected in the boost test directory should all be executables. We can change this, if that changes.

docs/sphinx/dev/dev_config_tracker.rst
jbmertens: It should happen just at startup, which is slow anyways, so I don't believe so.
The additional header information... @andrerenard , is there a case this matters?

tests/boost/run_boost_tests.sh
jbmertens: The testing framework looks ok with this, since the return code should propagate through,
https://github.com/kotekan/kotekan/actions/runs/17621839847/job/50069157736?pr=1264#step:8:896
But it would still be a good idea to watch out for erroneously missing or failing tests, as happened with pytest.

jbmertens: The script exit code should propagate through in the test framework, so this should only matter if trying to process the output locally. But we should keep an eye out for erroneously missing/failing tests just in case.
https://github.com/kotekan/kotekan/actions/runs/17621839847/job/50069157736#step:8:900

lib/stages/bufferRecv.cpp
jbmertens: Yes, the idea was to override using a host:port pair, to override the port at the given host.

lib/stages/bufferRecv.cpp
jbmertens: Yes, good point.

lib/core/configTracker.hpp
jbmertens: Never on windows!

lib/core/configTracker.hpp
jbmertens: No windows. Should be ok on macos?

docs/sphinx/dev/dev_config_tracker.rst
andrerenard: Generally I expect this will be ok.  The send/recv logic tends to happen on large data frames with low cadence, where it's easy for the sender system to buffer frames on the time scale of making an extra transfer of this size. 

Having said that, there are hypothetical situations where this does become an issue (small and high frequency frames), but this design is no worse than the dataset manager approach which pulled this data off a central server - it just created an external latency down stream of the recv stage.  If it actually becomes a problem, we can possibly move this to be fully side channel, but that creates another set of synchronization issues with the data files.

lib/core/restServer.cpp
andrerenard: Since this always binding to a host IP, e.g. 127.0.0.1, one of the host's IPs, or 0.0.0.0.  I'm not sure we actually need to make it work with hostnames.   Unlike the restClient or possibly bufferSend.

lib/stages/bufferSend.hpp
andrerenard: The server (bufferRecv) shouldn't be able to restart without breaking the connection, so bufferSend will have to open a new connection on server restart.  As long as we send any first time information after the formation of a connection, we should be ok I think.

docs/sphinx/dev/dev_config_tracker.rst
jbmertens: Avoiding network sends / comparison of potentially large config jsons. Happy to consider alternatives, though.

lib/core/configTracker.hpp
jbmertens: Mainly for debugging/internal verification, I want to call it from boost tests.

lib/core/configTracker.hpp
jbmertens: Downstream shouldn't ask for an update unless flagged to by upstream.

lib/core/kotekanLogging.hpp
jbmertens: Ideally yes, should be another PR/fix though.

lib/core/restServer.cpp
jbmertens: Disallowed non-ip, except localhost, which tests use.

lib/stages/bufferRecv.cpp
jbmertens: punting (need to look into, but this change should leave the current behavior unchanged at least)

lib/stages/bufferRecv.cpp
jbmertens: punting

lib/stages/bufferRecv.cpp
jbmertens: Agreed. But also punting.

docs/sphinx/dev/dev_config_tracker.rst
jbmertens: Just there to make sure host/port/hash are all consistent. The same node could reboot and launch a kotekan instance with a bad config.

docs/sphinx/dev/dev_config_tracker.rst
jbmertens: Open to other options, really it's just for a checksum.

lib/stages/bufferRecv.hpp
jbmertens: One for the stage, one for the connInstance.

lib/stages/bufferSend.hpp
jbmertens: Yeah. Adjusted the comment, but I don't have a better solution 🤷

kotekan/kotekan.cpp
```diff
@@ -539,7 +539,11 @@ int main(int argc, char** argv) {
...
+    if (address_parts.at(0) == "" || address_parts.size() != 2) {
```
geoffryan, line 544: Are IPv6 addresses allowed here?  The split on ":" will work differently there, as there might be more than 2 ":" delimited sections.

kotekan/kotekan.cpp
```diff
@@ -539,7 +539,11 @@ int main(int argc, char** argv) {
...
+        ERROR_NON_OO("The bind address {:s} is not valid, it should be in the form ip:port",
+                     bind_address.c_str());
+        exit(-1);
+    }
     rest_server.start(address_parts.at(0), std::stoi(address_parts.at(1)));
```
geoffryan, line 549: If the address is IPv6, there might be multiple ":" delimited substrings, right?   In that case the port should be the last section, not necessarily the second.

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,696 @@
...
+     */
+    void insertRawConfig(std::string host, uint16_t port, const nlohmann::json& config_json,
+                         const std::string& kotekan_version,
+                         const std::string& kotekan_build_branch,
+                         const std::string& kotekan_git_commit_hash,
+                         const std::string& kotekan_cmake_options) {
+        // Strip updatable config fields before hashing
+        nlohmann::json filtered_json;
+        for (auto& [key, value] : config_json.items()) {
+            if (key != "updatable_config") {
```
geoffryan: I might be thinking too much in the yaml, there might be some config parsing I'm unaware of, but I think this only strips out fields called "updatable_config" at the root of the config tree.  Stages/objects can contain "updatable_config" fields internally, which will point to a root-level object, but those objects are not called "updatable_config". They contain a field called "kotekan_update_endpoint" with the value "json".  See config/tests/fringestop.yaml for an example.

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,696 @@
...
+
+    /**
+     * @brief Check if a config exists in the _configs map.
+     *
+     * @param hash The hash of the config.
+     * @returns True if the config exists, false otherwise.
+     */
+    bool hasConfig(std::string host, uint16_t port) const {
+        std::lock_guard<std::mutex> lock(_lock);
+        return _configs.count({host, port});
```
geoffryan: should this be `return _configs.count({host, port}) > 0;` just to be explicit?

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,696 @@
...
+    }
+
+    /**
+     * @brief Check if a config exists using a hash string instead of host and port.
+     */
+    bool hasConfig(std::string hash) const {
+        std::lock_guard<std::mutex> lock(_lock);
+        return _config_hashes.count(hash);
```
geoffryan: Same as above.

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,696 @@
...
+                             hash, upstream_host, upstream_port, e.what());
+                continue;
+            }
+
+            // Check if the response contains the config under the owner
+            // (upstream_host:upstream_port)
+            std::string host_port_str = upstream_host + ":" + std::to_string(upstream_port);
+            if (config_response_json.contains(host_port_str)) {
+                ConfigInfo info = ConfigInfo(config_response_json[host_port_str]);
+                _insertConfig(upstream_host, upstream_port, info);
```
geoffryan, line 424: Should we re-compute the hash for the received config and make sure it matches what was requested?  Not sure how intensive that calculation is.

lib/core/configTracker.hpp
```diff
@@ -0,0 +1,696 @@
...
+     * This function generates a hash for a given configuration JSON object.
+     * In the context of the configTracker, the JSON object must have any
+     * "updatable_config" fields stripped before hashing. (This function only
+     * checks for that, and errors, expecting a caller to supply
+     * consistent information.)
+     *
+     * @param filtered_json The configuration JSON object to hash.
+     * @returns The canonical hash as a string.
+     */
+    std::string _jsonHash(const nlohmann::json& filtered_json) const {
```
geoffryan, line 542: The hash here only hashes the config, not the build parameters, git commit, etc.  Should they all be included?  That is, should the hash be of (most of a) ConfigInfo, instead of just the config json?

lib/core/restServer.cpp
```diff
@@ -56,9 +59,91 @@ restServer::~restServer() {
...
+    return false;
+}
+
+bool restServer::canBindToAddress(const std::string& address, const u_short port) const {
+    DEBUG_NON_OO("restServer: Checking if we can bind to address {:s}:{:d}", address, port);
+
+    // Use getaddrinfo to resolve the address (works for both IPs and hostnames)
+    struct addrinfo hints, *result = nullptr;
+    std::memset(&hints, 0, sizeof(hints));
+    hints.ai_family = AF_UNSPEC;     // Allow both IPv4 and IPv6
```
geoffryan: isValidAddress() only allows IPv4 though?

lib/core/restServer.cpp
```diff
@@ -56,9 +59,91 @@ restServer::~restServer() {
     }
 }
 
+// Validation function that accepts both IPs and 'localhost'
+bool restServer::isValidAddress(const std::string& address) const {
+    // First check if it's a valid IPv4 address
+    struct sockaddr_in sa4;
+    if (inet_pton(AF_INET, address.c_str(), &(sa4.sin_addr)) == 1) {
```
geoffryan, line 66: canBindToAddress() and the rest server docs allow IPv6 though?

lib/stages/bufferRecv.cpp
```diff
@@ -57,6 +60,25 @@ bufferRecv::bufferRecv(Config& config, const std::string& unique_name,
     connection_timeout = config.get_default<int>(unique_name, "connection_timeout", 60);
     drop_frames = config.get_default<bool>(unique_name, "drop_frames", true);
 
+    // Optional per-connection upstream REST port overrides for a given host (list of "host:port"
+    // strings)
+    if (config.exists(unique_name, "upstream_rest_endpoints")) {
+        auto entries = config.get<std::vector<std::string>>(unique_name, "upstream_rest_endpoints");
+        for (const auto& ep : entries) {
+            auto parts = regex_split(ep, ":");
+            if (parts.size() != 2) {
```
geoffryan, line 69: Assumes IPv4, is that in the docs?

tests/boost/test_configTracker.cpp
```diff
@@ -0,0 +1,159 @@
...
+    // Some json
+    json j1 = {{"key4", "value4"},
+               {"key5", {{"subkey3", "subvalue3"}, {"subkey4", "subvalue4"}}},
+               {"key6", "value6"}};
+
+    // Another identical json with updatable config
+    json j2 = {{"key4", "value4"},
+               {"key5", {{"subkey3", "subvalue3"}, {"subkey4", "subvalue4"}}},
+               {"key6", "value6"},
+               {"updatable_config", {{"key7", "value7"}, {"key8", "value8"}}}};
```
geoffryan: This is not always how `updatable_config` appears. It can appear inside other fields, and the actual time dependent data may not be called `updatable_config` at all.  Needs more test cases.

tests/ci-scripts/nightly/test_hello_world.sh
geoffryan, line 1: Once this works, should this be replaced with a test that performs a useful task?

tools/lint.sh
```diff
@@ -10,6 +10,17 @@ fi
 # kotekan root directory (infer from this script location)
 KOTEKAN_DIR="$(dirname "$(dirname "$(readlink -f "$0")")")"
 
+# Prefer system-wide kotekan Python env for Python tooling (e.g., black)
+# Detect a usable black command, prioritizing /opt/kotekan_env when present.
+BLACK_CMD=""
+if [ -x /opt/kotekan_env/bin/black ]; then
```
geoffryan, line 16: Would it be easier to just activate the virtual environment (and deactivate afterwards) if it hasn't been activated?

geoffryan (changes requested):
Overall looks good!  Comments above, mainly worried about `updatable_config` fields, I don't think they'll all get stripped out here.  Could you add a test with a realistic config that uses updatable config (I have an example in `config/tests/fringestop.yaml') and confirm the correct fields get stripped?

I'm a little worried about in a real pipeline the startup causing congestion while everyone is waiting on these blocking config tracker calls.  It feels like maybe buffer_recv should ping its senders for config once or twice during startup? 

 Or in the pipeline config we could add a send/recv stage that just sends a tiny amount of data to the receivers, something that executes fast and would get the config up-to-date while waiting for the bigger buffers to fill up and do their sends.

lib/core/configTracker.hpp
andrerenard: I think it's ok to not strip those values out.  The values that change and therefore break the config tracking model here, are all in the root of the config tree at `updatable_config`.   The references are static, so shouldn't be a concern.

kotekan/kotekan.cpp
andrerenard: FWIW I don't ever see us using IPv6.   We are basically always using private ranges, and there are more than enough IPs there for any system we are going to build.

lib/core/configTracker.hpp
geoffryan, line 1: The REST endpoint to request a config uses the hash as a key, which requires the configTracker to include a map<hash, host-port> to do the lookup quickly.  It seems like it would be simpler to request a config via the upstream host-port directly, and then to check that the received hash is as expected.

This is not a deal breaker, just a potential simplification.

geoffryan (changes requested):

lib/core/configTracker.hpp
geoffryan: Ah, maybe I wasn't aware that was a constraint.  I have been setting up configs like:

```yaml
my_telescope:
    updatable_config: /earth_rotation_data
    ...

earth_rotation_data:
    kotekan_update_endpoint: json
    my_data: [...]
```

But should that `earth_rotation_data` field be under an `updatable_config` field?  That does not seem to be enforced.

kotekan/kotekan.cpp
geoffryan: That's fine! Then [`docs/sphinx/dev/dev_rest_endpoints.rst`](https://github.com/kotekan/kotekan/blob/jbm/cm/docs/sphinx/dev/dev_rest_endpoints.rst) should be updated, it says IPv6 is an acceptable address form.

lib/core/restServer.cpp
jbmertens: Yes, thanks docs need to be updated. (Originally was going to allow ipv6, but that changed after discussion.)

kotekan/kotekan.cpp
jbmertens: ipv6 should not be allowed per discussions with @andrerenard

lib/core/configTracker.hpp
jbmertens: Agreed!

tests/boost/test_configTracker.cpp
jbmertens: Good point, need to discuss in-person!

tests/ci-scripts/nightly/test_hello_world.sh
jbmertens: Yes. I had in mind Erik's tests that take a long time, and/or iwyu, but scope creep...

lib/core/configTracker.hpp
jbmertens: Intentional -- consistency of other fields is also checked in _insertConfig. They are left separate for explicit comparison.

lib/core/configTracker.hpp
jbmertens: Good check.

tools/lint.sh
jbmertens: I think so - I didn't want to tailor the script to the ctc too much, but I'm not sure there's a clean way to avoid that.

geoffryan (approved):
Nice job with the updatable config check, lgtm!

andrerenard (approved):

PR #1266, opened 2025-07-25
iwyu nightly test workflow updates

jbmertens:
Updated workflow for iwyu:
 - misc version updates
 - Updated (and removed) dockerfiles
 - Updated actions scripts


andrerenard (approved):
LGTM!   One note though, we don't seem to have included a note in the `LICENSE` file about the origin/license attached to the external `.imp` files.   If you can track down the original source, it would be good to put a note about their license in the license file for completeness.

jbmertens:
Test fails are now due to unrelated cuda api updates. So, merging this ...

Issue #327, opened 2018-12-04
Port HDF5 components to new version of HighFive


jrs65:
~~There's been some substantial improvements to HighFive which we can't use because we are stuck on my old fork. Updating my fork to the current HighFive master is easy, the problem is we then have to port over all the chunk and filter support.~~

Now I think all of the functionality we need has been upstreamed into HighFive. We should try and move over to upstream the next time we touch the kotekan HDF5 code.

PR #1268, opened 2025-07-30
Changes to base Kotekan

eschnett:


eschnett:
The current build failures are there because this PR updates Kotekan's internals but does not update the science code I wrote. How do you want to handle this?

copy-kotekan-to-cx66.sh
jbmertens, line 1: Might've included this on accident.

CMakeLists.txt
```diff
@@ -269,31 +269,8 @@ if(${USE_GDAL})
...
-            INSTALL_COMMAND ""
-            LOG_DOWNLOAD ON)
-        ExternalProject_Get_Property(highfive source_dir)
-        set(HIGHFIVE_PATH ${source_dir})
-    else()
-        # just to satisfy dependencies (dependencies on highfive are needed to make sure it is
-        # cloned before attempting to build kotekan)
-        add_custom_target(highfive COMMENT "Highfive path supplied manually.")
-    endif()
-    message("HDF5 enabled; using HighFive from: " ${HIGHFIVE_PATH})
```
jbmertens, line 295: Thanks for changing this!

tools/docker/24.04/Dockerfile
```diff
@@ -1,196 +1,219 @@
...
+        pytest-xdist \
+        requests \
+        setuptools \
+        sphinx==6.2.* \
+        sphinx_rtd_theme==2.0.* \
+        sphinxcontrib-plantuml==0.25 \
+        tabulate \
+        wheel
+
+#        pyyaml
```
jbmertens: Are you sure this is unused?

tools/docker/24.04/Dockerfile
```diff
@@ -1,196 +1,219 @@
...
-
+# TODO:
+# - start from 
+# - don't `apt upgrade`
+# - install all apt packages at the same time
+# - install all pip packages at the same time
+# - why gcc 12?
+# - install a specific julia version
+# - use UNIX line endings
+FROM nvidia/cuda:12.9.1-devel-ubuntu24.04
```
jbmertens: I'm OK with this, as long as everything is working.
Can we update to cuda 13?

tools/docker/24.04/Dockerfile
```diff
@@ -1,196 +1,219 @@
...
-
-# Do nothing when the container launches
-CMD ["/bin/bash"]
-
+# TODO:
+# - start from 
+# - don't `apt upgrade`
+# - install all apt packages at the same time
+# - install all pip packages at the same time
+# - why gcc 12?
```
jbmertens: Previously, issues with dependencies, which might no longer be relevant.

.github/workflows/main.yml
```diff
@@ -191,7 +191,7 @@ jobs:
             -DCMAKE_BUILD_TYPE=Test \
             -DUSE_ASDF=ON \
             -DUSE_GDAL=ON \
-            -DUSE_HDF5=ON -DHIGHFIVE_PATH=/code/build/HighFive \
+            -DUSE_HDF5=ON \
```
jbmertens, line 194: Changes related to removing highfive or compile commands should ideally be reflected in any related docs, too.

tools/docker/24.04/Dockerfile
jbmertens: (This seems to be already installed by something else.)

tools/docker/24.04/Dockerfile
jbmertens: Just looking at this more -- a few notes,
 - The docker image doesn't seem to contain nvidia drivers, so I can't actually run code using just this image.
 - There isn't a version 13.0 image yet.
 - I do have some concern that nvidia images will be removed while CHORD is still running. So I might prefer to build our own images to maintain longer-term control over versions. (But since CHORD isn't deployed yet, right now I'd like to keep everything up to date.)
(@andrerenard @eschnett)

tools/docker/24.04/Dockerfile
```diff
@@ -1,196 +1,219 @@
...
+        libfftw3-dev \
+        libgdal-dev \
+        libhdf5-dev \
+        libhdf5-serial-dev \
+        liblapack-dev \
+        liblapacke-dev \
+        liblz4-dev \
+        libnuma-dev \
+        libopenblas-dev \
+        libopencl-clang-17-dev \
```
jbmertens: Is there a benefit to grouping these all up?
I believe they were grouped up before to make it logically clear which packages were relevant for which libraries/functionality.

tools/docker/24.04/Dockerfile
```diff
@@ -1,196 +1,219 @@
...
+        -DCMAKE_INSTALL_PREFIX=/usr/local \
+        -DBUILD_BENCHMARKS=OFF \
+        -DBUILD_EXAMPLES=OFF \
+        -DBUILD_FUZZERS=OFF \
+        -DBUILD_TESTS=OFF \
+        && \
+    cmake --build cmake-build --parallel $(nproc) && \
+    cmake --install cmake-build && \
+    true) && \
+    rm -rf src
```
jbmertens, line 160: blosc is no longer found by ldconf with these changes.

jbmertens (changes requested):

tools/docker/24.04/Dockerfile
jbmertens: Marking this as resolved since it is working.

tools/docker/24.04/Dockerfile
jbmertens: PR 1272 addresses this.

jbmertens:
Just a note -- most tests are passing with #1272 , but not with -DUSE_CUDA=ON. 
There are a couple still not passing though.

CMakeLists.txt
jbmertens: But, unfortunately some of the writer tests are failing now with this change. Flagging this for discussion on Monday @eschnett @andrerenard  --

andrerenard: I've looked over the CHIME config files, and as far as I can tell we don't actually use HDF5/HighFive in any of the real-time production configs.  We are mostly following the pattern of "raw" files followed by python conversion scripts.  This applies to baseband, visibilities, and HFB (absorber) data. 

However, we _do_ use the `*Transpose` stages as a part of gossec, which basically generates config files and gets kotekan to run them.   So the actual transpose and HDF5 file write actually happens in kotekan there.   

I believe most of Richard's changes did eventually get merged: https://github.com/BlueBrain/HighFive/pull/78

However it doesn't appear that the `extensible-datasets` was ever pulled into HighFive.   I'm not entirely sure why we didn't try to get that one merged, it's not a very big change

I'm checking with the CHIME team to see if anyone remembers the exact details on that. 

Currently we haven't decided if long term CHIME will use the CHORD file writers and transport logic (or lack there of), or if we'll keep support for the `*Transpose` classes.   

For now I'd be tempted to just "vendor" Richard's branch into kotekan if we don't like having to pull it.  It's header only, so would be fairly easy to merge in, and the license allows that.   Unless we need to have support for the latest HighFive for something CHORD wants to do? 

Long term the solution is either to move entirely to the CHORD way of doing things (not yet finished), or to update the HighFive version and get those changes merged in.

lib/core/buffer.cpp
```diff
@@ -425,45 +420,47 @@ void Buffer::print_full_status() {
             status_string[i] = '_';
     }
 
-    INFO_F("Full Frames (X)                : %s", status_string);
+    INFO_F("%-40s : %s", "Full Frames (X)", status_string);
```
andrerenard, line 423: Not really in the scope of this PR, but we should likely replace these `*_F` logging macros with the newer ones in kotekanLogging.

copy-kotekan-to-cx66.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
andrerenard: I don't think we should include this in the PR.

lib/core/kotekanLogging.cpp
```diff
@@ -70,9 +70,17 @@ void kotekanLogging::set_log_level(const std::string& s_log_level) {
...
 }
 
+logLevel kotekanLogging::get_log_level() const {
+    return logLevel(_member_log_level);
+}
+
 void kotekanLogging::vset_error_message(const fmt::basic_string_view<char> format,
                                         fmt::format_args args) {
-    fmt::format_to_n(__err_msg, __max_log_msg_len, fmt::vformat(format, args));
+    // Note: We should protect `__err_msg` with a lock
```
andrerenard, line 79: Do we want to do that as part of this PR, or just change that to a `TODO:`?

lib/core/ringbuffer.cpp
```diff
@@ -9,8 +9,8 @@ using fmt::group_digits;
 
 // This prints out a Python literal used for making plots in post-processing, for debugging
 // and illustration purposes.
+__attribute__((unused))
```
andrerenard: Should we have a unit test or similar if we want to keep this, but not use it in any code for now?

lib/core/ringbuffer.cpp
```diff
@@ -42,54 +42,52 @@ RingBuffer::RingBuffer(std::ptrdiff_t sz, std::shared_ptr<metadataPool> pool,
...
-    print_py_status(this);
+    DEBUG("Ringbuffer {:s}: wait_without_claiming({:s}[{:d}]): "
+          "requested bytes: {}, initial bytes available: {}",
+          buffer_name, name, inst, group_digits(sz), group_digits(first_write_head - read_head));
+    DEBUG("Ringbuffer {:s}: wait_without_claiming({:s}[{:d}]): waiting...", buffer_name, name,
+          inst);
     print_full_status();
-    return std::optional<std::ptrdiff_t>(sz);
+    full_cond.wait(lock, [&]() { return shutdown_signal || first_write_head - read_head >= sz; });
+    DEBUG("Ringbuffer {:s}: wait_without_claiming({:s}[{:d}]): waiting done.", buffer_name, name,
```
andrerenard: I wonder if we should make these `DEBUG2`.   Setting `DEBUG` globally in the config (which we sometimes do) will result in a lot of new messages with these outputs, unless we also suppress it by overriding at the buffer level.

lib/core/ringbuffer.cpp
```diff
@@ -183,93 +176,100 @@ void RingBuffer::finish_read(const std::string& name, const int inst, const std:
...
 void RingBuffer::print_full_status() {
-    buffer_lock lock(mutex);
-    INFO("  status: size {}, last_read_tail {}, first_write_head {}, "
-         "available to read: {}",
-         group_digits(size), group_digits(last_read_tail), group_digits(first_write_head),
-         group_digits(first_write_head - last_read_tail));
+    std::unique_lock<std::recursive_mutex> lock(mutex);
+    static std::recursive_mutex print_mutex;
+    std::lock_guard<std::recursive_mutex> print_lock(print_mutex);
+    INFO_F("--------------------- %s ---------------------", buffer_name.c_str());
```
andrerenard, line 255: Why use `INFO_F` here?

lib/cuda/NDArrayBuffer.hpp
```diff
@@ -0,0 +1,236 @@
...
+
+private:
+    std::ptrdiff_t length_in_bytes() const {
+        return ndarray.get_size() * ndarray.value_type_size;
+    }
+
+public:
+    // Metadata:
+
+    std::shared_ptr<const chordMetadata> get_metadata() const {
```
andrerenard, line 172: This might be out of scope of this PR, but we'd discussed moving some of the these properties into the buffer itself, and out of the frame-by-frame metadata.   Although I haven't yet carefully worked out if there are some frame-by-frame dependent components here?

lib/cuda/NDArrayBuffer.hpp
```diff
@@ -0,0 +1,236 @@
...
+#include <ostream>
+#include <sstream>
+#include <string>
+#include <utility>
+#include <vector>
+
+enum buffer_type_t { standard = 0, do_once = 1 << 0 };
+
+template<typename T, std::size_t D>
+class NDArrayBuffer : public kotekan::kotekanLogging {
```
andrerenard, line 60: It would be nice to have some class level documentation on how this works, and what it's intended to be used for.  And how it interacts with the metadata object.

lib/cuda/NDArrayBuffer.hpp
```diff
@@ -0,0 +1,236 @@
...
+        void* const ptr =
+            is_do_once ? cuda_command.get_device().get_gpu_memory(buffer_name_device, size_in_bytes)
+                       : cuda_command.get_device().get_gpu_memory_array(
+                             buffer_name_device, get_instance_num(),
+                             cuda_command.get_gpu_buffer_depth(), size_in_bytes);
+        return static_cast<T*>(ptr);
+    }
+
+public:
+    NDArrayBuffer(const std::string& buffer_name, const std::string& quantity,
```
andrerenard, line 91: This seems like it could also use some documentation to explain now to use it.

lib/cuda/NDArrayRingBuffer.hpp
```diff
@@ -0,0 +1,559 @@
...
+
+public:
+    NDArrayRingBuffer(const std::string& buffer_name, const std::string& quantity,
+                      const std::array<std::ptrdiff_t, D>& extents,
+                      const std::array<kotekan::Symbol, D>& dimnames, cudaCommand& cuda_command) :
+        // metadata
+        buffer_name(buffer_name),                            // e.g. "bb_beams"
+        buffer_name_host("host_" + buffer_name + "_buffer"), // e.g. "host_bb_beams_buffer"
+        buffer_name_device(buffer_name + "_buffer"),         // e.g. "bb_beams_buffer"
+        signal_buffer_name("host_" + buffer_name + "_ringbuffer"),
```
andrerenard, line 167: There seem to be some hard-coded conventions here, that I assume would need to be also followed in the config?  Should we be deriving these from the config?   Alternatively we should add something to the documentation to explain the naming requirements.

lib/cuda/NDArrayRingBuffer.hpp
```diff
@@ -0,0 +1,559 @@
...
+    std::ptrdiff_t end() const noexcept {
+        return m_end;
+    }
+    std::ptrdiff_t size() const noexcept {
+        return end() - begin();
+    }
+};
+
+template<typename T, std::size_t D>
+class NDArrayRingBuffer : public kotekan::kotekanLogging {
```
andrerenard, line 120: Comments on how this should be used.

lib/metadata/DataType.hpp
```diff
@@ -28,6 +32,143 @@ using float16_t = _Float16;
...
+        return (std::int8_t)(bits << 4) >> 4;
+    }
+
+    friend std::ostream& operator<<(std::ostream& os, const int4x2_t x) {
+        return os << x[0] << "," << x[1];
+    }
+};
+
+// offset-encoded (stored is value + 8), low and high values swapped
+struct int4x2chime_t {
```
andrerenard: I'm not sure `chime` is the right term here, since at least for now the CHORD F-engine is still using this model, and we'd decided a few months back to keep it for all projects (e.g. HIRAX) for consistency).

lib/stages/hdf5FileRead.cpp
```diff
@@ -0,0 +1,220 @@
+#include "hdf5Files.hpp"
+
+#include <Stage.hpp>
+#include <StageFactory.hpp>
+#include <algorithm>
+#include <cassert>
+#include <chordMetadata.hpp>
+#include <cstdint>
+#include <highfive/highfive.hpp>
```
andrerenard, line 9: Ah I see you are planning to use highfive as well.   That may complicate the CHIME dependence on Richard's version.

lib/stages/hdf5FileRead.cpp
```diff
@@ -0,0 +1,220 @@
...
+                const std::shared_ptr<metadataObject> metadata = buffer->get_metadata(frame_id);
+                if (!metadata)
+                    FATAL_ERROR("Buffer \"{:s}\" frame {:d} does not have metadata",
+                                buffer->buffer_name, frame_id);
+                assert(metadata);
+                if (!metadata_is_chord(metadata))
+                    FATAL_ERROR("Metadata of buffer \"{:s}\" frame {:d} is not of type CHORD",
+                                buffer->buffer_name, frame_id);
+                assert(metadata_is_chord(metadata));
+                const std::shared_ptr<chordMetadata> meta = get_chord_metadata(metadata);
```
andrerenard, line 96: This will likely all break with @rhaas80 's changes, but I suppose getting this in first still makes sense.

andrerenard (changes requested):

copy-kotekan-to-cx66.sh
eschnett: Thanks.

tools/docker/24.04/Dockerfile
eschnett: I kept things alphabetical to make it easier to see whether a certain package is already installed.

copy-kotekan-to-cx66.sh
eschnett: Removed.

lib/core/kotekanLogging.cpp
eschnett: Let's leave it as a todo.

lib/cuda/NDArrayRingBuffer.hpp
eschnett: Correct, this sets conventions. What do you mean by "deriving these from the config"? Do you want to list all four different buffer names in each kernel? I think our configs are too long and repetitive already, I am trying to take a step to simplify them and to help people avoid errors by typos that often lead to very weird error messages downstream.

lib/metadata/DataType.hpp
eschnett: So e.g. `int4x2_swapped_withoffset_t`? (That's the name in `CUDASIMDTypes.jl`.)

lib/cuda/NDArrayRingBuffer.hpp
andrerenard: I wasn't sure if you are listing all 4 in the config and needed this to match, or if you are hiding these extra buffers from the config, and just listing the one "buffer_name" in the config.  If it's the latter, then I think this works well.

.github/workflows/main.yml
eschnett: Done.

lib/core/buffer.cpp
eschnett: Ok.

lib/metadata/DataType.hpp
andrerenard: It's kind of a long name, but I think it's at least a lot more descriptive than `chime`.  Maybe just shorten to `int4x2_swapped_offset_t`, since it's generally understood what `offset` means in the project.

lib/core/ringbuffer.cpp
eschnett: Done.

lib/core/ringbuffer.cpp
eschnett: What else do you suggest?

lib/cuda/NDArrayBuffer.hpp
eschnett: Yes, we plan to move this out of the metadata. However, we are currently allocating and deallocating the `NDArrayBuffer` etc. within each kernel so that the respective metadata doesn't survive yet. These helper functions here transfer the metadata between the Kotekan metadata and the new buffer wrapper classes.

lib/core/ringbuffer.cpp
andrerenard: `INFO_NON_OO()` if you want this code to rely on the default log_level, that uses the `fmt` style format strings we use everywhere else.    `INFO_F()` was mostly intended to support the few bits of C code we still had floating around, and likely isn't actually needed anymore.

CMakeLists.txt
andrerenard: I missed this ticket: https://github.com/kotekan/kotekan/issues/327 It looks like Richard didn't merge in his changes because HighFive upstream had everything we needed.  @eschnett if you happened to make changes to `*Transpose` and the related file writers, perhaps you could add them here and we can see if that fixes the tests?

tools/docker/24.04/Dockerfile
jbmertens: Resolving this per #1272 .

CMakeLists.txt
tristpinsm: My impression is that the `extensible-datasets` fork of `HighFive` was needed in order to implement the `extend_time` method on `visFileH5`. I don't believe we actually use that anywhere since only `gossec` writes out HDF5 and those files have their time axis well-defined in the first place. So if you can re-implement or deprecate that method using the upstream `HighFive` I'm guessing you could drop Richard's fork.

jbmertens: Hopefully the changes in #1274 address this. Please have a double check though @andrerenard @tristpinsm

jbmertens (approved):
Changes look good to me with #1274 pulled in.

andrerenard (approved):
Thanks for all the changes.  This is looking good to merge.

andrerenard:
When you squash merge this, could you just include a summary of what's changed here.  And maybe a note that this commit will not build, as it will require the next commit with the autogen changes.

jbmertens:
Note that this PR doesn't build with -DUSE_CUDA=ON, since it is missing autogenerated code (future PR).
CI tests still pass locally with this flag off.

PR #1273, opened 2025-08-15
remove chimeMetadata

rhaas80:
this pull request adds json backends for some metada items and generically removes use of chimeMetadata from all code.

It is not a full solution since it does not move all items into json.


lib/metadata/jsonMetadata.hpp
```diff
@@ -0,0 +1,151 @@
...
+
+#pragma pack()
+
+// json based metadata structures to hold "physics" metadata, ie. data that
+// various at most once per frame
+
+namespace jsonMetadata {
+// these must match chimeMetadata and choordMetadata for now
+// in namespace for C++ const
+#define MAX_NUM_BEAMS 20
```
andrerenard, line 22: This max beams logic was largely in place to deal with the fact we used structs which had compile time defined size.   We can remove that now that we have a dynamic structure.

lib/metadata/jsonMetadata.hpp
```diff
@@ -0,0 +1,151 @@
...
+const int CHORD_META_MAX_FREQ = 1024;
```
andrerenard, line 23: Same with this parameter.

lib/metadata/chordMetadata.hpp
```diff
@@ -392,6 +399,128 @@ class chordMetadata : public metadataObject {
...
+    int32_t get_lost_timesamples() const {
+        return metadata[jsonMetadata::LOST_TIMESAMPLES].template get<int32_t>();
+    }
+
+    void set_lost_timesamples(int32_t lost_timesamples) {
+        // very much non-atomic, due to json dict entry creation
+        metadata[jsonMetadata::LOST_TIMESAMPLES] = lost_timesamples;
+    }
+
+    void atomic_add_lost_timesamples(const int32_t lost_samples) {
```
andrerenard, line 266: We need to rethink metadata changes where more than one producer could be setting values.   This is very much an edge case, but still needs to be accounted for.   My vote would be to add a mutex to the metadata object and then we can do this safely.

lib/metadata/chordMetadata.hpp
```diff
@@ -392,6 +399,128 @@ class chordMetadata : public metadataObject {
...
+    dset_id_t get_dataset_id() const {
+        return metadata[jsonMetadata::DATASET_ID].template get<dset_id_t>();
+    }
+
+    void set_dataset_id(const dset_id_t dset_id) {
+        metadata[jsonMetadata::DATASET_ID] = dset_id;
+    }
+
+private:
+    jsonMetadata::metadata metadata;
```
andrerenard, line 384: We will need serialization/deserialization functions for this class.   Specifically for stages like `bufferSend`/`bufferRecv` we currently only perform "shallow copies" which worked for the flat struct model, but don't work with this dynamic structure.   Of course for CHIME compatibility we'll also want to have a "legacy" version, which will likely have some conflicts with what Jim is working on now...

lib/metadata/chordMetadata.hpp
```diff
@@ -392,6 +399,128 @@ class chordMetadata : public metadataObject {
         this->onehot_index[dim] = i;
         strncpy(this->onehot_name[dim], name.c_str(), CHORD_META_MAX_DIMNAME);
     }
+
```
andrerenard: General comment on this class.   We now seem to have a slightly odd mix of old struct based logic for some items, and new json backed methods.  This is going to have some strange effects with serialization/deserialization.   As much as it's a huge pain, my feeling is we need to either commit entirely to changing this all over, or at the very least create serialization/deserialization which addresses _both_ the old components and the new components. 

Also in terms of naming.  I think we should rename this `frameMetadata`, since it's basically just taking on all possible types of frame metadata in the system.   We might someday want to breakup the classes just for clarity, but that's not obviously an improvement.

lib/metadata/chordMetadata.hpp
```diff
@@ -392,6 +399,128 @@ class chordMetadata : public metadataObject {
...
+        metadata[jsonMetadata::RFI_NUM_BAD_INPUTS] = rfi_num_bad_inputs;
+    }
+
+    /// The number of FPGA frames flagged as containing RFI.
+    /// NOTE: This value might contain overlap with lost samples, so it can count
+    /// missing samples as samples with RFI.  For renormalization this value
+    /// should NOT be used, use @c lost_timesamples instead.
+    /// This value will be filled even if RFI zeroing is disabled.
+    int32_t get_rfi_flagged_samples() const {
+        return metadata[jsonMetadata::RFI_FLAGGED_SAMPLES].template get<int32_t>();
```
andrerenard, line 248: What does the runtime error message look like when a get value doesn't exist?  Should we create a generic `get()` function that creates nice error messages about where things failed, and then throws the exception in case the caller wants to handle the case of no value set.

lib/metadata/jsonMetadata.hpp
```diff
@@ -0,0 +1,151 @@
...
+
+struct coarseFreq {
+        int nfreq;
+        int coarse_freq[CHORD_META_MAX_FREQ];
+
+        // TODO: turn into a from_json function?
+        coarseFreq(const metadata& md) {
+        md[NFREQ].get_to(nfreq);
+        const auto& freq(md[COARSE_FREQ]);
+        if(freq.size() > MAX_NUM_BEAMS)
```
andrerenard: Shouldn't this be max frequencies not beams.   But also I think we should be able to get rid of hard coded max sizes since we are backing the metadata with a dynamic structure.

andrerenard (commented):
Not finished the review, but leaving some early comments.  

I think we are going to have to take a careful look at places where metadata is generically handled.   Specifically `bufferSend/bufferRecv`, some of the file writing logic, and the buffer functions `copy_metadata` and `move_metadata` which used to rely on being able to shallow copy the structure.   The current test environment is mostly geared to testing individual stages, which I think is why we aren't seeing these changes throw errors.  That should change with some of the work @jbmertens is doing with the CI self-hosted runners, but I don't think has very good coverage of these generic parts of the framework at the moment.    It might be worth having some simple metadata passing pipelines get added to the `config/ci-tests` if that's now working.  

We also need to consider how to address all the metadata types used by "CHORD" (and thus the kernels we need for CHIME).  To merge this in, we are going to need to either create some hybrid serialization logic, or commit to fully converting all the "CHROD" metadata types to use json under the hood, and then serialize that.

lib/metadata/chordMetadata.hpp
rhaas80: lost samples is currently one such where there are multiple threads accumulating onto the same metadata :-(

lib/metadata/chordMetadata.hpp
rhaas80: yup, for anything to go forward this needs to all move into json. This pull request (and the work for it) was mostly to see if there are already objections in how the getters / setters work. I will move the leftover bits to json.

lib/metadata/chordMetadata.hpp
rhaas80: I'd expect it would be a bus-error (if eg the atomic `int` needs alignment that a regular `int` violates). Really what we want is a mutex, such as the one removed in cd1d4ebf .  :-)

rhaas80: I think it's a C++ exception, out of range is what the docs say: https://json.nlohmann.me/home/exceptions/#jsonexceptionout_of_range403

rhaas80:
@andrerenard so tests are failing on the CI system due to CI system issues it seems.

Since I have re-formatted and auto-generated code in here, should I create two pull requests:

1. to actually be reviewed without auto-gen and reformat, this will always fail the tests
2. to actually commit, incl. the auto-gen and reformatting, this is about 400 changed files and lots of noise

lib/metadata/chordMetadata.hpp
rhaas80: this one is the one that actually causes a SEGFAULT (due the cast to `atomic_int*` I assume).

rhaas80: added a `mutex` but this is still not thread safe unless `nlohmann_json` itself is thread safe.

lib/metadata/chordMetadata.hpp
rhaas80: true, well it now inherits from `MetadataObject` so there's a serialize function to overload

rhaas80: I have kept the serialize to bytes functionality intact (and working) though, so if the buffers actually used that (they should...) then they'd still work.

lib/metadata/chordMetadata.hpp
rhaas80: the onehot code was removed, with approval by @dstndstn

lib/metadata/jsonMetadata.hpp
rhaas80: the `max_num_beam` define is still used in the to-bytes serialized structure (`chordMetadataFmt` or so I think)

lib/metadata/jsonMetadata.hpp
rhaas80: yup

lib/metadata/BeamMetadata.hpp
```diff
@@ -29,8 +30,8 @@ class BeamMetadata : public metadataObject {
     int64_t fpga_seq_start;
     /// The GPS time of @c fpga_seq_start.
     timespec ctime;
-    /// Stream identifier
-    stream_t stream_id;
+    /// Frequency indices in frame
+    std::vector<int> coarse_freq;
```
andrerenard, line 34: We might want to leave this as a `stream_t` actually.  We only really use this metadata in the `beamExtract` stage, which will get updated to make it's output identical to what it outputs in CHIME now.   As a result we may need to reverse lookup the appropriate streamIDs.   Since we have not yet made our way to this part of the project, we could just leave this as `stream_t` and return to generating a valid stream_id in BeamExtract later.  With a big TODO note in that code.

lib/core/metadataFactory.cpp
```diff
@@ -84,11 +83,6 @@ std::shared_ptr<metadataPool> metadataFactory::new_pool(const std::string& pool_
                                     pool_type);
     }
 
-    if (pool_type == "chimeMetadata") {
```
andrerenard, line 87: Other than to make sure things compiled without CHIME metadata, I'm not sure there is value to removing it entirely.  We are going to use this structure eventually in the interface layers.   Perhaps rename to `legacyChimeMetadata`?

lib/metadata/chordMetadata.hpp
andrerenard: Hmm, that's a good question about `nlohmann_json`.  I don't think it's modification safe because in theory one of the underlying structures could change shape in the middle of a store.   Which actually creates a larger problem for all possible co-current update/insertions.  This means we need to be very strict about doing deep copies anytime we have pipeline divergence (although that was always the case, but for different reasons).  And multiple producer systems need to lock against all possible concurrent updates (or only have one thread set the metadata).      I think the updated the packet capture code to only set metadata from one thread, but I need to double check that...  

Having said all that, I think updating a value like this is safe with the lock, as long as the underlying structure doesn't change shape mid-store.   I'll need a bit more time to audit to code to see if this problem will actually show up anywhere...

The one issue I can see actually is that normally the value would be zeroed in metadata structure, so setting an initial 0 value wasn't needed.   However here the first call could throw an exception for a missing value unless it's initialized to zero in the code that calls it.   We could fix that either by updating the calling code, or checking for the value's existence before setting, and initializing to zero if needed.

lib/metadata/chordMetadata.hpp
```diff
@@ -80,59 +89,11 @@ class chordMetadata : public metadataObject {
...
-    int64_t half_fpga_sample0[CHORD_META_MAX_FREQ];
-
-    // Time sampling -- for each coarse frequency channel, the factor
-    // by which the time samples have been downsampled relative to
-    // FPGA samples.
-    int time_downsampling_fpga[CHORD_META_MAX_FREQ];
-
     // Dish layout
     int ndishes;                                  // number of dishes
     int n_dish_locations_ew, n_dish_locations_ns; // number of possible dish locations
```
andrerenard, line 96: In the interest of this change not being too crazy large, I'm ok with leaving these kind of variables for now, but we should really factor this out to the telescope object.

lib/metadata/chordMetadata.hpp
```diff
@@ -50,28 +56,31 @@ class chordMetadata : public metadataObject {
...
     kotekan::DataType type;
 
-    /// Track the number of lost fpga samples in each gpu sub-integration
-    int lost_fpga_samples[CHORD_META_MAX_FREQ][CHORD_META_MAX_VIS_SAMPLES];
-    /// Track the number of rfi-flagged samples in each gpu sub-integration
-    int rfi_flagged_samples[CHORD_META_MAX_FREQ][CHORD_META_MAX_VIS_SAMPLES];
-
     int dims;
     int dim[CHORD_META_MAX_DIM];
     char dim_name[CHORD_META_MAX_DIM][CHORD_META_MAX_DIMNAME]; // "F", "T", "D", etc
```
andrerenard, line 83: TODO note, I think this should move into the buffer class in some capacity so we don't need to carry it around like this, but that is another project...

andrerenard (changes requested):
I'm a bit hung-up on the atomic issue + wider possible JSON thread safety issues.  My current read is that it's actually ok given the way things are currently used.  However I need to spend a few more hours double checking the code.  It's likely the case that we need to add some software protection against metadata misuse to catch ourselves if we do something unsafe...  Possibly by limiting how stages are allowed to mutate metadata, ideally without just a big mutex on everything, but I need to think about this a bit more.

rhaas80:
hmm, a copy `metadata_a = metadata_b` would work, but a `memcpy` which is likely what it does will not. Hmm it copies using `metadataObject` shared pointers. I *am* suspicious:

```
void GenericBuffer::private_copy_metadata(int dest_frame_id, GenericBuffer* src, int src_frame_id) {
    std::shared_ptr<metadataObject> from_metadata_container = src->metadata[src_frame_id];
    std::shared_ptr<metadataObject> to_metadata_container = metadata[dest_frame_id];
    if (from_metadata_container->get_object_size() != to_metadata_container->get_object_size()) {
        WARN("Metadata sizes don't match, cannot copy metadata!!");
        return;
    }
    *to_metadata_container = *from_metadata_container;
}
```

lib/metadata/chordMetadata.hpp
rhaas80: using the non-data stuff of NDArray preferably I'd say.

lib/core/metadataFactory.cpp
rhaas80: likely, yes. There will not be any factory produced chimeMetadata though. Only the struct that is produced for legacy buffer writers (if there's any that need to write chimeMetadata). In which case I will revert the commit that deletes the files I guess.

lib/metadata/BeamMetadata.hpp
rhaas80: ok. I'll give it a try once the review is done. Not sure if I actually have a `stream_id` to copy in given that data that is available to `BeamMetadata`.

lib/metadata/chordMetadata.hpp
rhaas80: I added protections using a `mutex` (standard one so far, so not recursive) to all getters and setters. Waiting for the CI tests to finish now.

rhaas80:
superseded by #1308

PR #1274, opened 2025-08-21
fix for boost data and h5 transpose tests

jbmertens:


tests/boost/test_DataType.cpp
```diff
@@ -16,7 +16,6 @@ BOOST_AUTO_TEST_CASE(test1) {
     std::cout << "    uint32:      " << uint32 << "\n";
     std::cout << "    uint64:      " << uint64 << "\n";
     std::cout << "    int4x2:      " << int4x2 << "\n";
-    std::cout << "    int4x2chime: " << int4x2chime << "\n";
```
eschnett, line 19: I do not understand this line? It seems unrelated and probably unintended.

lib/utils/visFileArchive.hpp
```diff
@@ -8,12 +8,27 @@
...
 #include <highfive/H5File.hpp>     // for File
+#include <hdf5.h>                  // H5Dwrite, etc.
 #include <map>                     // for map
 #include <memory>                  // for allocator, unique_ptr
 #include <stddef.h>                // for size_t
 #include <string>                  // for string
 #include <vector>                  // for vector
 
+template<typename T>
+static inline void write_hyperslab_raw(HighFive::DataSet ds,
```
eschnett: I think this can be modelled in HighFive by creating a `Selection` and then writing it. Selections also have a `write_raw` function.

I don't think it's important to change the code – I assume the current code works – but it might be worthwhile to add a respective comment for future reference.

tests/boost/test_DataType.cpp
jbmertens: This was causing the test to fail (unable to compile with the line). I think there might be some uncommitted code that made it work?

lib/utils/visFileArchive.hpp
jbmertens: I played with this for a bit yesterday and again just now -- HighFive doesn't seem to have a `write` or `write_raw` that works for all of the types/ here,
https://github.com/kotekan/kotekan/blob/83e205688fe1a079d1a6c0feb01e8bb9b5bbdfbc/lib/utils/visFileArchive.cpp#L155
There might be a way around this though.

eschnett: HighFive has
```C++
    template <typename T>
    void write_raw(const T* buffer,
                   const DataType& mem_datatype,
                   const DataTransferProps& xfer_props = DataTransferProps());
```

If I see correctly, this works with a dataset, or a hyperslab selection of a dataset, and allow specifying an arbitrary memory datatype, arbitrary buffer pointer, and arbitrary transfer properties.

tests/boost/test_DataType.cpp
eschnett: This enum constant is now called `int4x2_swapped_withoffset` instead. Can you use this name instead of removing the line?

lib/utils/visFileArchive.hpp
jbmertens: Ok thanks, there we go ...

eschnett (approved):

andrerenard (approved):
LGTM

PR #1276, opened 2025-09-03
Regenerate kernels

eschnett:


external/gputils/README.md
jbmertens, line 1: For gputils and n2k, the current chord branch contains only copies of the files that are relevant for an installation of gputils/n2k. (For details see the pull-n2k script and/or the gputils/n2k makefiles). If we aren't using the additional files, I'd prefer to keep just the files needed for kotekan. If we do indeed want to include everything, let's leave a note somewhere about what kotekan needs/uses.

Either way, it would be helpful to keep a readme and/or script with notes/commands about a preferred procedure for updating gputils and n2k, and what differs here from those repos.

eschnett:
There is an error calling `memset` on `float16`. I understand where the C++ compiler is coming from, truly I do, and I also understand why CUDA would make the bits of a `float16` value protected instead of public, it all makes kind of sense, but in the end this is just creating a huge headache for no actual benefit.

What should we do here? Write an external function in C where memsetting is allowed? Casting the `float16*` to a `void*` and interspersing a `volatile` to confuse the C++ compiler enough to allow it? Downgrade GCC?

Maybe adding the option `-Wno-error=class-memaccess` will do the trick...

OLD/report.ncu-rep
jbmertens, line 1: Guessing this was added by accident?

lib/gpu/gpuCommand.hpp
```diff
@@ -187,6 +188,10 @@ class gpuCommand : public kotekan::kotekanLogging {
     /// State that is shared by instances of this command (at one point in the pipeline)
     std::shared_ptr<gpuCommandState> command_state;
 
+    /// Ensure serial execution of calls to the same instance
+    std::mutex execute_mutex;
+    std::unique_lock<std::mutex> execute_lock;
+
```
andrerenard: Do we actually want to keep this after having fixed the race condition?

lib/stages/hdf5FileWrite.hpp
```diff
@@ -1,55 +0,0 @@
...
-
-#include "Config.hpp"
-#include "Stage.hpp" // for Stage
-#include "bufferContainer.hpp"
-
-#include <stdint.h> // for uint32_t
-#include <string>   // for string
-
-/**
- * @class hdf5FileWrite
```
andrerenard, line 12: Is removing this intentional?

OLD/report.ncu-rep
andrerenard, line 1: Could you maybe add a note in this folder to say what this is for?   I can see getting confused by it in the future..

jbmertens (changes requested):
As of writing this there are a few compiler warnings to clean up, plus the comments above, but otherwise LGTM!

andrerenard (commented):
This is a truly impressive amount of auto generated code.

OLD/report.ncu-rep
eschnett: Yes. Removed.

lib/gpu/gpuCommand.hpp
eschnett: No, I thought I had removed these.

lib/stages/hdf5FileWrite.hpp
eschnett: Yes, this is intentional.

OLD/report.ncu-rep
eschnett: This is not good for anything. It's gone now.

eschnett:
> This is a truly impressive amount of auto generated code.

It would be much more impressive if someone had hand-written this code.

external/gputils/README.md
eschnett: I removed the files which we don't use in Kotekan (tests, makefiles, Python bindings) and added a README describing how to update these directories.

andrerenard (approved):
I'm happy to merge this if it's ready @eschnett ?

jbmertens (approved):
If the `waiting_for_max_frames` variable change (slack discussion) can be included that would be good -- but otherwise let's prioritize getting the PR merged in asap.

PR #1277, opened 2025-09-05
Unregister functionality

jbmertens:


andrerenard (changes requested):
On second reading, I'm afraid there are a lot of problems/missing features with this code as codex produces. 

The idea I had was more along the lines of the following behaviour:

When the producer stages reach a set number of frames, then exist their `main_thread()`.  This triggers an unregistration of the buffers they are producers off. 
The consumers of those buffers continue to receive frames as usual, however, once there are no "full" frames in the buffer _and_ there are no producers, then the `wait_for_full()` calls return the shutdown condition (-1).   This should cause the consumers to also exit their `main_thread()` and unregister any buffers they produce into. 
Only `kotekanMode` needs to then watch for when all buffers lose their producer/consumer status.   Which should be equivalent to all `main_threads()` becoming joinable.    There should be no need to "bound" the writers, as the initial producers should be the ones to determine the number of frames to get processed.  Then those frames get "flushed" through the pipeline.  

This requires some modification at the stage and buffer level.    Ideally the current pattern of getting a buffer from the container, registering on it, and now adding it to a list for unregistering, could be consolidated into one base class function.

There is an additional wrinkle in this, which is the GPU command objects and the GPU control "ringBuffers".   Those need to behave in a similar way, but with the added complexity that those consumers read a variable amount of data.  So the exit condition is when they ask for more data than ever _will_ be produced into the ring.   That might actually be more of the complex parts to get right.

jbmertens:
Poking at this a bit more - another complication occurs when buffer send/recv is in the pipeline.
In #1264, some header information is added for the config tracker. But it's a full uint32, for padding reasons, so most of the bits aren't actually used... we could use additional bits in there to signal an upstream node is disconnecting?

PR #1279, opened 2025-09-12
Updating GPU Simulator & testing for N2K

geoffryan:
Updates the `testDataGen` and `gpuSimulateN2k` stages to work with updated `chordMetadata`. Updates `verify_cuda_n2k` config setup to work with updated n2k framework.  This test now passes and is moved to `ci-tests`

### DIFF
--- a/config/ci-tests/verify_cuda_n2k.yaml
+++ b/config/ci-tests/verify_cuda_n2k.yaml
@@ -75,7 +75,7 @@ gen_data:
     out_buf: host_voltage_buffer
     array_shape: [ 2048, 16, 2, 64 ]
     dim_name: ["T", "F", "P", "D" ]
-    num_frames: -1
+    num_frames: buffer_depth  # Must be at least what check_data expects
 
 run_send_voltage:
     gpu_0:
--- a/config/ci-tests/verify_cuda_n2k.yaml
+++ b/config/ci-tests/verify_cuda_n2k.yaml
@@ -124,21 +124,3 @@ check_data:
     first_buf: host_correlation_buffer
     second_buf: simulated_correlation_buffer
     num_frames_to_test: buffer_depth
-
-write_n2k_correlation:
-   kotekan_stage: hdf5FileWrite
-   base_dir: data
-   file_name: n2k_corr
-   prefix_hostname: false
-   max_frames: -1
-   skip_writing: false
-   in_buf: host_correlation_buffer
-
-write_simulated_correlation:
-    kotekan_stage: hdf5FileWrite
-    base_dir: data
-    file_name: sim_corr
-    prefix_hostname: false
-    max_frames: -1
-    skip_writing: false
-    in_buf: simulated_correlation_buffer
--- a/config/ci-tests/verify_cuda_n2k.yaml
+++ b/config/ci-tests/verify_cuda_n2k.yaml
@@ -69,7 +69,7 @@ host_voltage_ringbuffer:
 gen_data:
     # type: const
     # value: 34 # 1i # 34 # = 2 + 2i
-    type: random_signed
+    type: random_signed_offset  # const_offset
     value: 120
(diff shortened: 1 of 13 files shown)


config/ci-tests/verify_cuda_n2k.yaml
```diff
@@ -120,3 +124,21 @@ check_data:
     first_buf: host_correlation_buffer
     second_buf: simulated_correlation_buffer
     num_frames_to_test: buffer_depth
+
+write_n2k_correlation:
+   kotekan_stage: hdf5FileWrite
+   base_dir: data
+   file_name: n2k_corr
+   prefix_hostname: false
+   max_frames: -1
```
jbmertens: Let's make sure this has an end condition (just need to write a few frames).

lib/testing/testDataCheck.hpp
```diff
@@ -141,7 +141,8 @@ void testDataCheck<A_Type>::main_thread() {
             }
         }
 
-        if (num_errors > 0) {
+        //if (num_errors > 0) {
```
jbmertens: This is a (correct) bugfix, this line can be deleted.

tests/boost/CMakeLists.txt
```diff
@@ -131,3 +131,6 @@ add_custom_target(
     test_timeUtil_script ALL
     DEPENDS ${PROJECT_BINARY_DIR}/timeUtil.py
     COMMENT "Generating timeUtil.py script")
+
+# add_executable(test_CHORDTelescope test_CHORDTelescope.cpp)
```
jbmertens: Looks like a few extra lines here.

lib/testing/gpuSimulateN2k.cpp
```diff
@@ -112,7 +118,38 @@ void gpuSimulateN2k::main_thread() {
...
+        meta_out->set_array_dimension(1, _num_local_freq, "F");
+        meta_out->set_array_dimension(2, (_num_elements/16) * (_num_elements/16 + 1)/2, "DPhi");
+        meta_out->set_array_dimension(3, 16, "DPlo1");
+        meta_out->set_array_dimension(4, 16, "DPlo2");
+        meta_out->set_array_dimension(5, 2, "C");
+        meta_out->set_strides_simple();
+        meta_out->nfreq = _num_local_freq;
+        assert(meta_out->nfreq <= CHORD_META_MAX_FREQ);
+        meta_out->sample0_offset = 0;
+        meta_out->offset_downsampling = 1;
```
jbmertens: Will this (and does it need to) work for both CHIME and CHORD conventions?
I'm happy to have the test even just for CHORD.

jbmertens (changes requested):

config/ci-tests/verify_cuda_n2k.yaml
geoffryan: Done.  Although should note the setup will hang if fewer frames are generated than the checkData expects.  checkData signals a shutdown once its seen enough frames.  Added a note about this to the config.

lib/testing/testDataCheck.hpp
geoffryan: Done.

tests/boost/CMakeLists.txt
geoffryan: Yep. Removed.

config/ci-tests/verify_cuda_n2k.yaml
jbmertens: Ah yeah, I see the end condition there now. In that case, do we need to write any files?

lib/testing/gpuSimulateN2k.cpp
geoffryan: This uses the CHIME convention for voltage encoding everywhere, so should be fine on this front.

jbmertens (approved):

PR #1283, opened 2025-09-16
Support long FRBs in F-engine simulator

eschnett:


jbmertens (approved):
I can't really comment much on julia code, but LGTM!

PR #1288, opened 2025-09-17
Correct FRB data generation

eschnett:


eschnett:
I don't understand the build failure. I do not even see an error message.

jbmertens (approved):

jbmertens:
It looks like the github worker hung, I had no problem locally. The build also passes after re-running.
LGTM!

PR #1278, opened 2025-09-05
cmake refactor

jbmertens:
is cmake fun now with codex??

  - CMake modularization: Introduced cmake/Options.cmake, Toolchain.cmake, Summary.cmake, and Features/* modules (GPU, IO, Math, Julia) to refactor and centralize configuration.
  - Find modules reorganized: Moved Find*.cmake files into cmake/Modules/; added new modules (e.g., FindGDAL.cmake, FindHighFive.cmake).
  - Top-level updates: Simplified CMakeLists.txt to include the new modules; adjusted lib/*/CMakeLists.txt to align with the refactor.
  - Feature detection UX: Added colorized feature summary and clearer -DUSE_<FEATURE>=ON|OFF controls; improved toolchain handling and flags.
  - Docs and README: Updated build instructions and compiling docs to reflect new CMake flow and feature summary behavior.
  
  
  TODO: 
 - [x] Do a quick pass and make sure we're listing all (kotekan-specific, non-cmake) options 
 - [x] TODO: fix OpenCL vs CUDA conflict, remove HIP.
 - [x] Fix BOOST vs DPDK conflict if it's easy?
 - [x] Then the default for most options will be auto-detect.
 - [x] Add a developer note to add new options both to the cmake summary, and docs, and version info.

PR #1289, opened 2025-09-18
Fix cmake regressions

jbmertens:
Fix regressions introduced in #1278


andrerenard (approved):
Well that got more complicated than expected with the `ktk_tristate_option()` now being needed....  But overall looks good to me.

PR #1290, opened 2025-09-18
get act and dpdk working

jbmertens:
- Add some logic to help act run locally
 - Fix cmake logic for dpdk

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,8 +22,8 @@ on:
 
 env:
   REGISTRY: ghcr.io
-  IMG_IWYU: "ghcr.io/kotekan/kotekan/kotekan-iwyu"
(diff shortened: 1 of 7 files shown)


.github/workflows/main.yml
```diff
@@ -231,6 +235,7 @@ jobs:
     - uses: actions/checkout@v4
 
     - name: Login to GitHub Container Registry
+      if: ${{ !env.ACT }} # Don't log in if running locally
```
andrerenard, line 238: Do we still output the logs to screen?

andrerenard (approved):
LGTM

.github/workflows/main.yml
jbmertens: Should just be flushed to the terminal when running act (up to options, maybe)?, so yes.

PR #1292, opened 2025-09-19
HDF5 plugins set programatically

geoffryan:
This fixes bugs in the HDF5 cmake logic and sets the directory where plugins are searched programmatically instead of by environment variables.

On configure, the directory is set by the following in order of precedence:

1. the `HDF5_PLUGIN_DIR` cmake variable
2. the `HDF5_PLUGIN_PATH` environment variable (at time of configure)
3. the`hdf5plugin` plugin package in the active python environment
4. common system locations

This path is passed to`kotekan` on compilation, and then is passed to hdf5 at runtime via `H5PLprepend()`.  This allows kotekan to run without the `HDF5_PLUGIN_PATH` environment variable set.

### DIFF
--- a/cmake/Features/FeatureIO.cmake
+++ b/cmake/Features/FeatureIO.cmake
@@ -60,7 +60,7 @@ endif()
 
 
 # HDF5 + HighFive + HDF5Plugin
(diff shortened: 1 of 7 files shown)


jbmertens (approved):
After linting, LGTM!

andrerenard (approved):
LGTM

PR #1295, opened 2025-09-24
Add CI tests for the pathfinder pipeline

eschnett:


jbmertens:
Please see also #1307 which makes significant CI changes.
In particular, a /data directory is mounted for storing data.

jbmertens (commented):
I do see some errors in the actual run,
```
HDF5-DIAG: Error detected in HDF5 (1.10.10) thread 1:
  #000: ../../../src/H5F.c line 412 in H5Fopen(): unable to open file
```
Is this acceptable? I wouldn't expect a clean exit from kotekan here.

eschnett:
Hmm. This can be fine. The reader stages read files until there are no more files, then stop producing data. The HDF5 error message shouldn't be there, instead there should be an informative message stating that the reader has run out of files.

The possible error is that a path name is wrong and that the reader doesn't read any files at all. I will have to check for and catch this.

jbmertens (approved):

PR #1298, opened 2025-09-24
test_baseband_readout: skip in podman

rhaas80:
hangs, most likely for the same reason as in Docker

### DIFF
--- a/tests/test_baseband_readout.py
+++ b/tests/test_baseband_readout.py
@@ -27,9 +27,16 @@ def in_container() -> bool:
     except OSError:
         return False
 
+def is_podman():
+    try:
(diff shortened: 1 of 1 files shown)


jbmertens:
This should be fixed per #1290
https://github.com/kotekan/kotekan/pull/1290/files
Is that fix ok, or would you prefer to merge this?

rhaas80:
> This should be fixed per #1290 https://github.com/kotekan/kotekan/pull/1290/files Is that fix ok, or would you prefer to merge this?

nope, if that one handles podman then fine by me. I had not noticed the pull request so my branch is still starting from an older version of `chord`.

PR #1301, opened 2025-09-25
Work around missing 1-bit correlator for CHIME and HIRAX

eschnett:
Just assume there is no packet loss at the moment. Which is true. At the moment.

### DIFF
--- a/lib/cuda/cudaPL1bitCorrelator.cpp
+++ b/lib/cuda/cudaPL1bitCorrelator.cpp
@@ -6,6 +6,7 @@
 #include <cstddef>
 #include <cstring>
 #include <cudaCommand.hpp>
+#include <cudaMemsetInt.hpp>
 #include <div.hpp>
 #include <memory>
 #include <metadata.hpp>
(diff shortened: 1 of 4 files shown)


lib/cuda/cudaPL1bitCorrelator.cpp
```diff
@@ -247,16 +249,22 @@ cudaEvent_t cudaPL1bitCorrelator::execute(cudaPipelineState& /*pipestate*/,
...
+            n2k_counts_memory, (const ulong*)pl_expanded_mask_memory,
+            (const uint*)rfi_RFImask_memory, rfimask_fstride,
+            T,   // number of time samples before correlation
+            F,   // number of frequency channels
+            Sds, // number of stations (after downsampling by 8)
+            Nds, // downsampling factor of counts array, relative to baseband
+            device.getStream(cuda_stream_id));
+    } else {
+        // These cases are not yet implemented in n2k. Pretend that there is no packet loss.
+        cudaMemsetInt(n2k_counts_memory, Nds, n2k_counts.get_ndarray().size());
```
jbmertens, line 268: Worth adding a user-facing warning here?

jbmertens (approved):
LGTM

lib/cuda/cudaPL1bitCorrelator.cpp
eschnett: Done.

PR #1302, opened 2025-09-25
Initial support for CHIME F-Engine packet capture with the upgraded hardware/software. 

andrerenard:
- Updates the packet capture code for CHIME to support more than 4 ports. 
- Allows for the use of `chordMetadata`
- Adds a new GPU command, `cudaCopyNtoRingbuffer` to merge a list of buffers with one frequency each into a single ring with 16 frequencies (will require a transpose and reorder kernel to use with the CHIME kernels).
- Cleans up the buffer status logging to allow for per-buffer log levels
- Updates the CPU affinity for the frame zeroing threads to use the config file
- Removes support for versions of DPDK older than 19.11 
- Adds PCIe block-list support to get around issues with Nvidia/Mellanox cards getting automatically included in the list of DPDK enabled cards. 

Known issues/todo:
- When using chordMetadata the zeroing/reset logic happens in the critical packet loop, which is resulting in considerable packet loss.   Hopefully the refactored metadata will improve this.  Although likely this will require reworking the critical packet loop to remove buffer/metadata handling entirely. 
- The packet loss array uses the old CHIME format, and has not been updated to the new PL CHORD+CHIME array format.
- This needs to be integrated with the transpose and reorder kernel before we can start using the new CHIME kernels.


config/chime_upgrade_network.j2
```diff
@@ -0,0 +1,208 @@
...
+  # Format is index = lcore, value = cpu core
+  lcore_cpu_map: [4,5,6,7,36,37,38,39]
+  main_lcore_cpu: 3
+  status_cadence: 60
+  # Set to true when using a switch to replicate packets
+  using_switch: true
+  alignment: samples_per_frame * 4
+  # List of PCIe devices we don't want to attach too
+  # Required to exclude the ConnectX 6 Lx output NICs. 
+  pcie_block_list:
```
rhaas80, line 111: this list the device by its PICe location, does it? not by some sort of identifier of the device itself (vendor id, serial number)?

config/chime_upgrade_network.j2
```diff
@@ -0,0 +1,208 @@
...
+    out_buf: host_voltage_buffer_{{ id }}
+{%- endfor %}
+
+
+run_send_voltage:
+  gpu_0:
+    kotekan_stage: cudaProcess
+    gpu_id: 0
+    buffer_depth: gpu_buffer_depth
+    frame_arrival_period: 390625/samples_per_frame
```
rhaas80, line 190: comment on magic number "390625"?

config/chime_upgrade_network.j2
```diff
@@ -0,0 +1,208 @@
...
+      ring_buffer_size: gpu_buffer_depth * samples_per_frame * num_elements * num_freq_gpu_buf
+      in_bufs: 
+{%- for id in range(16) %}
+        - host_voltage_buffer_{{ id }}
+{%- endfor %}
+      signal_buf: gpu_voltage_ringbuffer
+      gpu_mem_output: blocked_voltage_buffer
+
+
+# Hex dump the host_voltage_buffer_12 (for debugging)
```
rhaas80: 12 is just any random buffer or special?

lib/core/buffer.cpp
```diff
@@ -595,7 +601,7 @@ uint8_t* Buffer::wait_for_empty_frame(const std::string& producer_name, const in
     // The second condition stops us from using a buffer we've already filled,
     // and forces a wait until that buffer has been marked as empty.
     DEBUG2("wait_for_empty_frame({:s}[{:d}]): waiting...", producer_name, ID);
-    print_full_status();
+    // print_full_status();
```
rhaas80: remove

lib/core/ringbuffer.cpp
```diff
@@ -233,7 +233,7 @@ void RingBuffer::finish_write(const std::string& name, const int inst, const std
     DEBUG2("finish_write({:s}[{:d}]): "
            "produced bytes: {}",
            name, inst, group_digits(sz));
-    // print_full_status();
+    print_full_status();
```
rhaas80, line 236: is this `print_full_status()` correct here? produces output to screen one assumes.

lib/cuda/cudaCopyNtoRingbuffer.cpp
```diff
@@ -0,0 +1,196 @@
...
+                }
+            }
+        }
+    }
+
+    _gpu_mem_output = config.get<std::string>(unique_name, "gpu_mem_output");
+
+    signal_buffer = dynamic_cast<RingBuffer*>(
+        host_buffers.get_generic_buffer(config.get<std::string>(unique_name, "signal_buf")));
+    assert(signal_buffer);
```
rhaas80: add an error message, at least something like:

```
assert(signal_buffer && "Buffer given by signal_buf is not a RingBuffer");
```

lib/cuda/cudaCopyNtoRingbuffer.cpp
```diff
@@ -0,0 +1,196 @@
...
+    if (instance_num == 0)
+        signal_buffer->register_producer(unique_name);
+
+    set_command_type(gpuCommandType::COPY_IN);
+    gpu_buffers_used.push_back(std::make_tuple(_gpu_mem_output, false, false, true));
+}
+
+cudaCopyNtoRingbuffer::~cudaCopyNtoRingbuffer() {
+    for (size_t i = 0; i < in_buffers.size(); ++i) {
```
rhaas80: why use `for(int i` type looping here by range-based loops before? The index `i` is not used for anything.

lib/cuda/cudaCopyNtoRingbuffer.cpp
```diff
@@ -0,0 +1,196 @@
...
+        Buffer* buf = in_buffers[i];
+        if (buf && buf->frame_size) {
```
rhaas80: for `buf` to be `NULL` line 34 above must not have triggered, so this would be a bug, yes? so instead use an `assert(buf)` maybe?

lib/cuda/cudaCopyNtoRingbuffer.cpp
```diff
@@ -0,0 +1,196 @@
...
+            if (cudaSuccess == cudaHostGetFlags(&flags, buf->frames[instance_num])) {
+                CHECK_CUDA_ERROR(cudaHostUnregister(buf->frames[instance_num]));
+            }
+        }
+    }
+}
+
+int cudaCopyNtoRingbuffer::wait_on_precondition() {
+    // Wait for all input buffers to have data
+    for (size_t i = 0; i < in_buffers.size(); ++i) {
```
rhaas80: same comment about `int i` vs range based loop as above

lib/cuda/cudaCopyNtoRingbuffer.cpp
```diff
@@ -0,0 +1,196 @@
...
+        Buffer* buf = in_buffers[i];
```
rhaas80: I note no check of `buff == NULL` here.

lib/cuda/cudaCopyNtoRingbuffer.cpp
```diff
@@ -0,0 +1,196 @@
...
+        DEBUG("Waiting for input data frame {} from buffer {}", gpu_frame_id, buf->buffer_name);
+        uint8_t* frame = buf->wait_for_full_frame(unique_name, gpu_frame_id % buf->num_frames);
+        if (!frame)
+            return -1;
```
rhaas80, line 72: really a `return` or a `continue` instead (there could be valid frames after this one, could there?)

lib/cuda/cudaCopyNtoRingbuffer.cpp
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    // Wait for ringbuffer space
+    DEBUG("Waiting for ringbuffer space for writing to become available ({} bytes)",
+          total_input_size);
+    std::optional<size_t> val =
+        signal_buffer->wait_for_writable(unique_name, instance_num, total_input_size);
+    DEBUG("Ringbuffer space for writing is now available");
+    if (!val.has_value())
+        return -1;
```
rhaas80, line 84: explicitly invalidate `output_cursor`?

lib/cuda/cudaCopyNtoRingbuffer.cpp
```diff
@@ -0,0 +1,196 @@
...
+    output_cursor = val.value();
+    return 0;
+}
+
+cudaEvent_t cudaCopyNtoRingbuffer::execute(cudaPipelineState& pipestate,
+                                           const std::vector<cudaEvent_t>& pre_events) {
+    (void)pipestate;
```
rhaas80: instead of C style cast to void, use

```
udaEvent_t cudaCopyNtoRingbuffer::execute(cudaPipelineState& /*pipestate*/,
                                           const std::vector<cudaEvent_t>& pre_events) {
```

instead?

lib/cuda/cudaCopyNtoRingbuffer.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+    // Compute total size and wrapping
+    size_t start = output_cursor % signal_buffer->size;
+
+    record_start_event();
+
+    // Merge all input buffers into the ringbuffer sequentially
+    size_t offset = 0;
+    std::vector<std::shared_ptr<chordMetadata>> metas;
+    for (size_t i = 0; i < in_buffers.size(); ++i) {
```
rhaas80: comment about looping

lib/cuda/cudaCopyNtoRingbuffer.cpp
```diff
@@ -0,0 +1,196 @@
...
+        Buffer* buf = in_buffers[i];
```
rhaas80: no check for `NULL`

lib/cuda/cudaCopyNtoRingbuffer.cpp
```diff
@@ -0,0 +1,196 @@
...
+        int buf_index = gpu_frame_id % buf->num_frames;
+        void* host_memory_frame = (void*)buf->frames[buf_index];
```
rhaas80: is this cast required? Doesn't even C++ let you cast any pointer to `void*` automatically?

lib/cuda/cudaCopyNtoRingbuffer.cpp
```diff
@@ -0,0 +1,196 @@
...
+        // Allocate a new metadata object on the signal buffer
+        signal_buffer->allocate_new_metadata_object(0);
+        auto meta_ring = std::dynamic_pointer_cast<chordMetadata>(signal_buffer->metadata[0]);
+        if (!meta_ring)
+            throw std::runtime_error(
+                "cudaCopyNtoRingbuffer: could not get chordMetadata from signal buffer");
+
+        // Merge metadata from all input buffers
+        // NB This is highly specific to CHIME.
+        for (uint i = 0; i < in_buffers.size(); ++i) {
```
rhaas80: comment about looping

rhaas80: `uint`?

lib/cuda/cudaCopyNtoRingbuffer.cpp
```diff
@@ -0,0 +1,196 @@
...
+            auto meta_in = std::dynamic_pointer_cast<chordMetadata>(
+                in_buffers[i]->get_metadata(gpu_frame_id % in_buffers[i]->num_frames));
+            if (!meta_in)
+                throw std::runtime_error(
+                    "cudaCopyNtoRingbuffer: input buffer has no chordMetadata");
+            if (i == 0) {
```
rhaas80, line 147: add comment about this being the first loop iteration?

lib/cuda/cudaCopyNtoRingbuffer.cpp
```diff
@@ -0,0 +1,196 @@
...
+            meta_ring->coarse_freq[i] = meta_in->coarse_freq[0];
+            assert(meta_ring->sample0_offset == meta_in->sample0_offset);
+        }
+        // Debug log the merged metadata with the data set above
+        INFO("cudaCopyNtoRingbuffer: Merged metadata frequency list: {}",
+             fmt::join(std::vector<int>(meta_ring->coarse_freq,
+                                        meta_ring->coarse_freq + in_buffers.size()),
+                       ", "));
+        first_time = false;
+    }
```
rhaas80, line 175: `metas` is not used.

lib/cuda/cudaCopyNtoRingbuffer.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+    return record_end_event();
+}
+
+void cudaCopyNtoRingbuffer::finalize_frame() {
+    DEBUG("finalize_frame() for frame {}, releasing metadata on GPU output buffer {}", gpu_frame_id,
+          _gpu_mem_output);
+    for (size_t i = 0; i < in_buffers.size(); ++i) {
```
rhaas80: looping

lib/cuda/cudaCopyToRingbuffer.cpp
```diff
@@ -49,6 +49,8 @@ cudaCopyToRingbuffer::cudaCopyToRingbuffer(Config& config, const std::string& un
         if (in_buffer->frame_size) {
             uint flags;
             // only register the memory if it isn't already...
+            // TODO This logic doesn't work for host buffers with more then num_gpu_frames frames
+            // (which is likely)
```
rhaas80: should I be worried?

lib/dpdk/dpdkCore.cpp
```diff
@@ -73,29 +70,17 @@ dpdkCore::dpdkCore(Config& config, const string& unique_name, bufferContainer& b
...
-#else
-    num_system_ports = rte_eth_dev_count();
-#endif
 
     // This default works well for ICE boards,
     // but we might change this to something more genertic
     memset((void*)&port_conf, 0, sizeof(struct rte_eth_conf));
-    uint32_t max_rx_pkt_len = config.get_default<uint32_t>(unique_name, "max_rx_pkt_len", 5000);
-#ifndef OLD_DPDK
+    uint32_t max_rx_pkt_len = config.get_default<uint32_t>(unique_name, "max_rx_pkt_len", 9000);
```
rhaas80, line 78: has this change in default been reflected in the doxygen docs?

lib/dpdk/dpdkCore.hpp
```diff
@@ -140,6 +140,7 @@ class dpdkRXhandler : public kotekan::kotekanLogging {
  * @conf   jumbo_frame      Bool. Default true Enable support for Jumbo frames
```
rhaas80, line 140: incorrect default for `max_rx_pkt_len` documented (see comment above)

lib/dpdk/iceBoardShuffle.hpp
```diff
@@ -43,6 +44,8 @@
  *
  * @conf  fpga_dataset          String. The dataset ID for the data being received from
  *                              the F-engine.
+ * @conf  link_group            Int.  The link group this handler is a member of.
```
rhaas80: unsigned int I assume actually?

lib/dpdk/iceBoardShuffle.hpp
```diff
@@ -143,10 +146,19 @@ class iceBoardShuffle : public iceBoardHandler {
...
 
+    /// The maximum number of link groups.
+    static const uint32_t link_group_max = 8;
+
     /// The stream_ids for all iceBoardShuffle objects.
-    /// This might be an issue in the case of multiple indepdent
-    /// shuffle operations. In that case this will need to be factored out.
-    static ice_stream_id_t all_stream_ids[shuffle_size];
+    /// There are 4 stream_ids per link group.
+    inline static ice_stream_id_t all_stream_ids[link_group_max][shuffle_size];
```
rhaas80, line 154: I am confused. What does `inline` on a data member do?

lib/dpdk/iceBoardShuffle.hpp
```diff
@@ -354,24 +379,33 @@ inline bool iceBoardShuffle::check_stream_id() {
...
+            || all_stream_ids[link_group][i].link_id != link_id) {
+            /// Print out all the stream IDs for this group.
+            for (uint32_t j = 0; j < shuffle_size; ++j) {
+                ERROR("link_group {:d} subid {:d} stream_id: crate {:d} slot {:d} link {:d} "
+                      "dpdk_port {:d}",
+                      link_group, j, all_stream_ids[link_group][j].crate_id,
+                      all_stream_ids[link_group][j].slot_id, all_stream_ids[link_group][j].link_id,
+                      port);
+            }
             FATAL_ERROR("One of the link_ids or slot_ids don't match! There is a cabling problem.");
```
rhaas80: At least one of...

lib/dpdk/iceBoardShuffle.hpp
```diff
@@ -407,22 +441,58 @@ inline bool iceBoardShuffle::advance_frames(uint64_t new_seq, bool first_time) {
...
-        // We take the stream ID only from the first pair of crates,
-        // to avoid overwriting it on different ports.
-        // This makes the stream ID unique for down stream stages.
-        if (port_stream_id.crate_id / 2 == 0) {
             ice_stream_id_t tmp_stream_id = port_stream_id;
             // Set the unused flag to store the post shuffle freq bin number.
             tmp_stream_id.unused = i;
-            ice_set_stream_id_t(out_bufs[i], out_buf_frame_ids[i], tmp_stream_id);
+            tmp_stream_id.crate_id = tmp_stream_id.crate_id % 2;
+            meta->coarse_freq[0] = tel.to_freq_id(ice_encode_stream_id(tmp_stream_id));
```
rhaas80, line 479: single frequency only?

lib/dpdk/iceBoardShuffle.hpp
```diff
@@ -433,18 +503,43 @@ inline bool iceBoardShuffle::advance_frames(uint64_t new_seq, bool first_time) {
...
+        for (int i = 0; i < 4; i++) {
+            tmp_stream_id.unused = i;
+            tmp_stream_id.crate_id = tmp_stream_id.crate_id % 2;
+            meta->coarse_freq[i] = tel.to_freq_id(ice_encode_stream_id(tmp_stream_id));
+        }
+
+        meta->type = kotekan::uint8;
+
+        std::strncpy(meta->dim_name[0], "T", sizeof meta->dim_name[0]);
+        meta->dim[0] = lost_samples_buf->frame_size; // One byte per time sample
```
rhaas80, line 543: no name for the data?

rhaas80, line 543: not setting `dims`, maybe other fields missing.

rhaas80 (changes requested):

config/chime_upgrade_network.j2
andrerenard: Yep it uses PCIe addresses.  Perhaps not ideal, but given our largely uniform hardware, it didn't seem worth doing address lookups on some other property.

config/chime_upgrade_network.j2
andrerenard: Added a comment, this could maybe be dervied from the Telescope object in the code instead of the config, but there are cases it's nice to be able to set this independent of the F-engine parameters (e.g. in testing).

config/chime_upgrade_network.j2
andrerenard: Just a random buffer so I could peek at the data.  I'll remove this entirely, since it's not likely to be used by anyone else.

lib/core/buffer.cpp
andrerenard: Actually I ment to leave that in.   Erik had added a bunch of status messages, but they overwhelm the output for these large pipelines.   Hence my other changes to allow one to selectively tune the logging on a per-buffer basis.

lib/core/ringbuffer.cpp
andrerenard: Yeah it produces data to screen.  Was commented out I assume because it was producing too much output, should be ok to have it back now with the logging changes.

lib/cuda/cudaCopyNtoRingbuffer.cpp
andrerenard: Done.

lib/cuda/cudaCopyNtoRingbuffer.cpp
andrerenard: No harm in that check.

lib/cuda/cudaCopyNtoRingbuffer.cpp
andrerenard: Likely I was using i at some point, and never updated it.

lib/cuda/cudaCopyNtoRingbuffer.cpp
andrerenard: Also yeah only need to check frame_size here + an assert.

lib/cuda/cudaCopyNtoRingbuffer.cpp
andrerenard: I have not coded any logic to track frames in the sequency which are missing.  So yes, I want to return here and not attempt to transfer a partially filled frame to the GPU.

lib/cuda/cudaCopyNtoRingbuffer.cpp
andrerenard: Done.

lib/cuda/cudaCopyNtoRingbuffer.cpp
andrerenard: Done.

lib/cuda/cudaCopyNtoRingbuffer.cpp
andrerenard: Done.

lib/cuda/cudaCopyNtoRingbuffer.cpp
andrerenard: Done.

lib/cuda/cudaCopyNtoRingbuffer.cpp
andrerenard: Done.

lib/cuda/cudaCopyNtoRingbuffer.cpp
andrerenard: I think the "modern/correct way" is to use `static_cast<void*>`, so switched to that. Although yes, the cast isn't strictly required.

lib/cuda/cudaCopyNtoRingbuffer.cpp
andrerenard: Yeah should be `size_t`, `i` is actually useful here.

lib/cuda/cudaCopyNtoRingbuffer.cpp
andrerenard: Ah yes, removed.

lib/cuda/cudaCopyNtoRingbuffer.cpp
```diff
@@ -0,0 +1,196 @@
...
+                                          nullptr, nullptr);
+            device.async_copy_host_to_gpu(rb_memory, (char*)host_memory_frame + first, second,
+                                          cuda_stream_id, nullptr, nullptr, nullptr);
+        }
+
+        offset += sz;
+    }
+
+    // We only need to set the metadata on the ring buffer once.
+    if (first_time && instance_num == 0) {
```
andrerenard, line 129: Thinking about this more we should implement a cross-check of the metadata on every call to this function.

lib/cuda/cudaCopyNtoRingbuffer.cpp
andrerenard: Done.

lib/cuda/cudaCopyToRingbuffer.cpp
andrerenard: Actually I think all this register stuff can be removed.  The framework level calls to `cudaProcess::register_host_memory()` in the buffer mapping should take care of it...  If they aren't then as currently written this would cause a performance issue for the frames that aren't captured by this logic when the CPU and GPU frames have different lenghts.   I'm just going to remove it entirely, if it causes performance issues some place, then that's a bug and we should address it at the framework level, and not make the copy process deal with it.

lib/core/buffer.cpp
rhaas80: it's commented out though, you are saying to leave in the commented out code? Sometimes done as a hint for testers I guess.

lib/dpdk/iceBoardShuffle.hpp
andrerenard: That's done to avoid needing an out of class definition (which was removed in this PR).   There are likely more elegant ways to do this, but since the DPDK code needs to be extremely efficient, I wanted to avoid requiring dynamic structures...

lib/dpdk/iceBoardShuffle.hpp
andrerenard: Done.

lib/dpdk/iceBoardShuffle.hpp
andrerenard: Yep, each output frame is only one frequency.  This isn't something that's likely to change,    `iceBoardShuffle` is very specific to the CHIME F-engine design.

lib/dpdk/iceBoardShuffle.hpp
andrerenard: Yes.

lib/dpdk/iceBoardShuffle.hpp
andrerenard: Thanks for catching that.

andrerenard: Done.

andrerenard:
Comments mostly addressed, need to run some tests to make sure I didn't break the system performance by removing the register logic.

lib/core/buffer.cpp
andrerenard: It's been changed back to uncommented, but I think because there is now no longer a change here, github is displaying in strangely.

andrerenard:
I think I've addressed all the comments.  I also added some metadata consistency checking logic, and confirmed that my removal of the cuda memory registration logic was safely handled already in the framework buffer mapping logic.  Ready for re-review.

lib/dpdk/dpdkCore.hpp
rhaas80: `max_rx_pkt_len Int. Default 5000` still incorrect.

rhaas80:
https://github.com/kotekan/kotekan/pull/1302/files#r2403361840 remains unresolved. Fine otherwise.

rhaas80 (approved):

PR #1303, opened 2025-09-30
1 bit correlator quickfix

geoffryan:
Updates the `cudaPL1bitCorrelator` to set correct `time_downsampling_fpga` values.


geoffryan:
Oops, wrong branch.

PR #1305, opened 2025-10-01
Update Julia plotting scripts

eschnett:


jbmertens (approved):
In Erik's julia code we trust

PR #1306, opened 2025-10-07
New telescope configuration "smallfinder"

eschnett:
The "smallfinder" is a smaller version of the "pathfinder" for CHORD,
using fewer frequencies so that it can run on out CI infrastructure.
Probably.


lib/cuda/cudaBasebandBeamformer_chime.cpp
```diff
@@ -6,35 +6,23 @@
...
-#include <cudaDeviceInterface.hpp> // for cudaDeviceInterface
-#include <div.hpp>                 // for mod, round_down, div, div_noremainder
-#include <fmt.hpp>                 // for compile_string_to_view
-#include <functional>              // for function
-#include <limits>                  // for numeric_limits
-#include <map>                     // for map
-#include <memory>                  // for shared_ptr, __shared_ptr_access
-#include <string>                  // for basic_string, allocator, string, operator+, operator<
-#include <vector>                  // for vector
+#include <DataType.hpp>
```
jbmertens, line 9: This might be a regression of recent IWYU changes.
Can we keep the headers included, unless they actually aren't used? Or, re-run the lint tool with iwyu, `./tools/lint.sh -i ON` .

lib/cuda/cudaBasebandBeamformer_chord.cpp
```diff
@@ -6,35 +6,23 @@
...
-#include <cudaDeviceInterface.hpp> // for cudaDeviceInterface
-#include <div.hpp>                 // for mod, round_down, div, div_noremainder
-#include <fmt.hpp>                 // for compile_string_to_view
-#include <functional>              // for function
-#include <limits>                  // for numeric_limits
-#include <map>                     // for map
-#include <memory>                  // for shared_ptr, __shared_ptr_access
-#include <string>                  // for basic_string, allocator, string, operator+, operator<
-#include <vector>                  // for vector
+#include <DataType.hpp>
```
jbmertens, line 9: Similar here, let's make sure the header changes are correct.

lib/cuda/cudaBasebandBeamformer_hirax.cpp
```diff
@@ -6,35 +6,23 @@
  * Do not modify this C++ file, your changes will be lost.
  */
```
jbmertens, line 8: And check headers here, and down below too

jbmertens (changes requested):

lib/cuda/cudaBasebandBeamformer_chime.cpp
eschnett: These changes actually look correct. They correspond to the C++ templates.

I do not know why these changes are not already there on the `chord` branch. Possibly the regenerated kernels were not pushed when their source code was modified?

lib/cuda/cudaBasebandBeamformer_chord.cpp
eschnett: Looks correct

lib/cuda/cudaBasebandBeamformer_hirax.cpp
eschnett: Yes, looks correct

jbmertens (approved):

PR #1307, opened 2025-10-08
CI updates

jbmertens:
- [x] Pin julia version (hopefully fix build issues)
- [x] Use matrix-style jobs
- [x] Use own container for runners
- [x] Fix documentation build errors
- [x] Use bash script instead of pytest for running boost tests
- [x] Make sure /data is mounted
- [x] .gitattributes!
- [x] IWYU for autogen code?
- [x] Fix n2k warnings
- [x] linter fixes


tools/docker/24.04/Dockerfile
```diff
@@ -195,10 +195,11 @@ RUN mkdir src && \
     true) && \
     rm -rf src
 
-# Install Julia
-RUN curl -fsSL https://install.julialang.org | sh -s -- -y
-ENV PATH="$PATH:/root/.juliaup/bin"
-RUN juliaup add 1.11 && juliaup default 1.11
+# Install Julia 1.11.2 directly (no juliaup for no automatic update checks)
+RUN wget https://julialang-s3.julialang.org/bin/linux/x64/1.11/julia-1.11.2-linux-x86_64.tar.gz && \
```
andrerenard: I think this question was raised on the call, but do we need to add this to the PATH, or does the fact we've symlinked it into `/usr/local/bin` sufficiently take care of that?   I guess just check that we are auto detecting in this location?

tests/boost/run_boost_tests.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
```
andrerenard, line 1: Might be nice to have a short description of what this does, and how it gets used.

external/n2k/src_lib/s0_kernel.cu
```diff
@@ -85,7 +85,7 @@ __device__ uint _cmask(int b)
 //   long F;                                  // number of freq channels
 //   long S;                                  // number of stations (= dish+pol pairs)
 //   long Nds;                                // time downsampling factor
-//   long out_fstride4;                       // freq stride of s0 array (ulong4 stride, not ulong stride)
+//   long out_fstride4;                       // freq stride of s0 array (ulong4_16a stride, not ulong stride)
```
andrerenard, line 88: Are changes to this file supposed to be in this PR?

andrerenard (approved):
LGTM.

external/n2k/src_lib/s0_kernel.cu
jbmertens: I could issue another PR, but figured I'd fix the deprecation warnings here.

tools/docker/24.04/Dockerfile
jbmertens: The `ln -s /opt/julia-1.11.2/bin/julia /usr/local/bin/julia` bit seems sufficient for Kotekan to compile, at least.

PR #1308, opened 2025-10-10
update all metadata in chordmetadata to use json base storage

rhaas80:


andrerenard (approved):
Let's see what breaks :)

PR #1311, opened 2025-10-21
Vis transpose writer

jbmertens:
TODO:

 - [x] Ensure frequency ids map to an index in array output
        (Will expect everything between freq_id start and stop)
 - [x] Use HDF5 instead of zarr?
 - [x] Write config data
 -  ~Write gains (or reference to)~
 - [x] Write/note element indexes (maybe in config?)
        (Write a full map of indexes)
 - [x] Check consistency with CHIME format
 - [x] add CI test(s) (ongoing)
 
Tangential changes:
 - N2 pipeline plumbing: N2Accum should output an N2 type buffer
 - FATAL_ERROR should fail with an error


tests/boost/test_timeUtil.cpp
geoffryan, line 1: I don't think this test should rely on a specific python interpreter being present on a specific path. When the test is run, the user should just have the relevant venv activated.

geoffryan: A better solution in the future would be to move to something like `uv` and having a kotekan venv defined within kotekan itself.

jbmertens: Agreed; I could revert this change here, this was mainly for convenience running tests locally via script on the ctc. (It doesn't _rely_ on the specific path, just a preference.) Short of `uv` I'm not sure what a better solution would be.

geoffryan:
- For the frequencies: the incoming N2 frames are each tagged with their own `freq_id` (which is linearly related to the physical frequency). They won't come in order and may not be contiguous, the total number of frequencies to expect should probably be a config parameter to the N2-writer. The map of array index to `freq_id` in the output file can probably follow the CHIME convention (Sec 4.2.2), placed in `~index_map/freq` although we could make two or three arrays (one for `freq_id`, one for central freq (in MHz), and one for channel width (in MHz)), instead of packaging all together into one? These data can be extracted from the Telescope once you have the `freq_id`.
 - The file format should be HDF5, it's what we're using everywhere else.
 - element indexes should largely follow the CHIME convention (Sec 4.2.1). The data for both of these fields can/will be supplied by the Telescope. In `index_map`:
     - a field `input` of length `num_elements` (or `num_dishes`) that ids (and possibly contains info like position) all of the inputs, mostly the main array dishes + polarizations, but could include a small number of other sources (the atomic clock, the 26m, an RFI or noise channel, something from CHIME, etc). This could also be several arrays `input_id`, `input_grid_position`, `input_pol`, etc. This will largely (entirely) be present in the config, as well as a label/id for lookup into the layout DB.
      - a single `prod` field of type `<int64_t>[num_prod, 2]` or two fields `prod_i`, `prod_j` (or 1/2, a/b) of type `<int64_t>[num_prod,]`, which for every element in the vis matrix give the index into the `input` array(s) corresponding to one end of the baseline in this correlation. We should be very clear about whether the first or second element is the one that has been conjugated.
 - The digital gains (applied by the FPGA/ADCs) should be included following CHIME (sec 4.1.1.5 and 4.1.1.6).
 - The calibration gains... should perhaps be stored by reference? Something like the CHIME `update_id` field? They may change during the acquisition a file corresponds to, and the boundaries of those changes won't line up with the file IO. On the other hand... a single self-describing file is very nice!

tests/boost/test_timeUtil.cpp
geoffryan: For now, the script running tests should invoke the `venv`.  If we wanted a work-around here, i'd go more for checking an environment variable like `KOTEKAN_PY_ENV` through which we could supply the path to a virtual environment, that the cpp code could invoke before calling python.

jbmertens: Unfortunately the venv only exists on the cluster, not the docker image nor the runner, so the venv would similarly need to be invoked as a preference In general, the environment both kotekan and the test is run in are expected to have relevant libraries, etc, installed. The main kotekan.cpp file also has this issue, since python is called for yaml->json parsing.

Punting this to #1328

jbmertens:
> - For the frequencies: the incoming N2 frames are each tagged with their own `freq_id` (which is linearly related to the physical frequency).  They won't come in order and may not be contiguous, the total number of frequencies to expect should probably be a config parameter to the N2-writer.  The map of array index to `freq_id` in the output file can probably follow the CHIME convention (Sec 4.2.2), placed in ~`index_map/freq` although we could make two or three arrays (one for `freq_id`, one for central freq (in MHz), and one for channel width (in MHz)), instead of packaging all together into one?  These data can be extracted from the Telescope once you have the `freq_id`.

Yeah, though we still need a way to get array indices in the first place, which the current telescope object doesn't give. I'd prefer to have the telescope object infer it from the config rather than set it by hand.

> - The file format should be HDF5, it's what we're using everywhere else.

Well, there's been lots of churn on this. I think it's straightforward to make it a config option 🤷 

> - element indexes should largely follow the CHIME convention (Sec 4.2.1). The data for both of these fields can/will be supplied by the Telescope.  [...]
> - The digital gains (applied by the FPGA/ADCs) should be included following CHIME (sec 4.1.1.5 and 4.1.1.6).

Agreed we should follow CHIME conventions as appropriate per the CHIME doc. This will need to be added to the telescope object as well.

> - The calibration gains... should perhaps be stored by reference?  Something like the CHIME `update_id` field?  They may change during the acquisition a file corresponds to, and the boundaries of those changes won't line up with the file IO.  On the other hand... a single self-describing file is very nice!  

I don't think we want to pass the calibration gains themselves downstream, so a reference is ok with me. But self-contained (not just self-describing) would be nice... but unfortunately we can't until we have something like a gains broker.

geoffryan:
I have been getting a similar error as 48ba988 and fd3ae95 on #1330, and with a different test (`test_eigenvis.py` instead of `test_n2_accumulate.py`), but also on the `gcc_release_pytest` run.  The test does not seem to be able to get past starting kotekan (or at least sends no output after that point).  I have not been able to ID a specific cause in the config, test setup, etc that triggers this.  A second incidence of this bug is interesting... the only real similarity in the tests is that they both use N2 buffers and make a `CHORDTelescope`.

jbmertens:
Splitting into multiple PRs.

PR #1315, opened 2025-10-21
Fix Release build and add to CI

geoffryan:
Fixes the Release kotekan build (largely by suppressing errors from unused variable warnings from asserts) and adds a Release build to CI testing.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -217,33 +217,6 @@ jobs:
               -DUSE_DPDK=OFF
               -DWITH_TESTS=ON -DWITH_BOOST_TESTS=ON
             post_build: ""   # Any clang-specific tests to run?
-          
-          - id: release_clang
-            name: "Release build with clang"
-            ccache_key: kotekan-release-clang
-            cmake_defs: &build_release_clang_opts >-
-              -DCMAKE_BUILD_TYPE=Release
-              -DCMAKE_C_COMPILER=clang
-              -DCMAKE_CXX_COMPILER=clang++
-              -DWERROR=ON
-              -DCCACHE=ON
-              -DUSE_OMP=ON
-              -DUSE_NUMA=ON
-              -DNO_MEMLOCK=ON
-              -DUSE_CUDA=ON
-              -DUSE_HIP=OFF
-              -DUSE_OPENCL=ON
-              -DUSE_ASDF=ON
-              -DUSE_HDF5=ON
-              -DUSE_GDAL=ON
-              -DUSE_AIRSPY=ON
-              -DUSE_LAPACK_BLAZE=ON
-              -DUSE_FFTW=ON
-              -DUSE_JULIA=ON
-              -DUSE_DPDK=ON
-              -DWITH_TESTS=ON
-              -DWITH_BOOST_TESTS=OFF
-            post_build: ""  # clang tests?
 
           - id: docs
             name: "Build Documentation"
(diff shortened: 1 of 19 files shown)


lib/metadata/chordMetadata.hpp
```diff
@@ -152,9 +152,14 @@ class chordMetadata : public metadataObject {
...
 #pragma GCC diagnostic push
 #if GCC_VERSION > 80000
+#pragma GCC diagnostic ignored "-Warray-bounds"
 #pragma GCC diagnostic ignored "-Wstringop-truncation"
 #endif
-        strncpy(this->name, name.c_str(), CHORD_META_MAX_DIMNAME);
+        // strncpy(&(this->name[0]), name.c_str(), CHORD_META_MAX_DIMNAME);
+        int len = name.length() < CHORD_META_MAX_DIMNAME ? name.length() : CHORD_META_MAX_DIMNAME;
+
+        for (int i = 0; i < len; i++)
```
eschnett, line 156: This should fill the unused space (if any) with `NUL` since there is otherwise no way to know how long the actual string is.

.github/workflows/main.yml
```diff
@@ -173,6 +173,37 @@ jobs:
               -DUSE_AIRSPY=ON -DUSE_LAPACK_BLAZE=ON -DUSE_FFTW=ON -DUSE_JULIA=ON -DUSE_DPDK=ON
               -DWITH_TESTS=ON -DWITH_BOOST_TESTS=OFF
             post_build: ""   # Any OpenCL-specific commands to test?
+
```
eschnett, line 176: You are adding multiple builds and tests. I think we do want to build with gcc. Not sure about clang, but why not. I'm not sure we need to run tests for all builds because this will lead to longer CI times. Did you check how long it takes to run CI now, before and afterwards?

lib/metadata/DataType.hpp
```diff
@@ -31,11 +31,11 @@ using float16_t = _Float16;
 
 namespace kotekan {
 
-#ifdef DEBUGGING
```
eschnett, line 34: Did you check the CI run times with this change?

lib/stages/asdfFiles.hpp
```diff
@@ -47,7 +47,7 @@ inline ASDF::scalar_type_id_t chord2asdf(const kotekan::DataType type) {
         case kotekan::float64:
             return ASDF::id_float64;
         default:
-            assert(0);
+            throw std::runtime_error("chord2asdf given bad DataType: " + type_to_string(type));
```
eschnett: If the type is "bad", will `type_to_string` work, or will it throw as well? Maybe better to return the integer enum value instead.

lib/stages/hdf5Files.cpp
```diff
@@ -44,7 +44,7 @@ HighFive::DataType chord2hdf5(const kotekan::DataType type) {
         case kotekan::float64:
             return HighFive::AtomicType<double>();
         default:
-            assert(0);
```
eschnett, line 47: Ditto.

lib/testing/gpuSimulateCudaUpchannelize.cpp
```diff
@@ -159,7 +159,7 @@ constexpr T sinc(const T x) {
 
 // array indexing
 
-constexpr int Eidx(int c, int d, int f, int p, int t) {
```
eschnett, line 162: It's weird that the function takes `c (specifying which component, `real` or `imag`, to return) and then ignores this argument. That argument probably shouldn't be there... But that's not for this PR. Maybe add a respective comment to the code instead?

lib/testing/gpuSimulateCudaUpchannelize.cpp
```diff
@@ -168,7 +168,7 @@ constexpr int Eidx(int c, int d, int f, int p, int t) {
     return d + D * f + D * F * p + D * F * P * t;
 }
 
-constexpr int Ebaridx(int c, int d, int fbar, int p, int tbar) {
```
eschnett, line 171: Ditto.

eschnett (changes requested):

.github/workflows/main.yml
geoffryan: The current chord branch takes ~18min to complete CI, the longest time is taken by the full build with gcc and pytest testing, which takes ~17m: ~2m to initialize containers, ~10m to build kotekan, ~5m to run the tests.  

With the added tests here, the full CI time goes to ~20m. The full build with gcc & pytest takes about the same time. The new longest CI test is the release clang build, which takes a full 20m just to compile (it does no other testing).  This doesn't seem like too bad of a regression to me, seems worth it to maintain compatibility with a large  compiler family, although compile times have gotten long recently and that should be addressed, but probably as a different PR ;)

lib/metadata/DataType.hpp
geoffryan: No appreciable change.

lib/metadata/chordMetadata.hpp
geoffryan: Yes.  Fixed in 61c1862

lib/stages/asdfFiles.hpp
geoffryan: Good catch, fixed in 61c1862

lib/stages/hdf5Files.cpp
geoffryan: As above, fixed in 61c1862

lib/testing/gpuSimulateCudaUpchannelize.cpp
geoffryan: Comment added in 61c1862

lib/testing/gpuSimulateCudaUpchannelize.cpp
geoffryan: ditto 61c1862

geoffryan:
Noting here for posterity: this globally turns off the `unused-variables` warning, because there are a few of these in auto-generated code that only get used in asserts.    In the future, we should update the generators, and allow the unused-variables warning again.

eschnett (approved):

PR #1316, opened 2025-10-21
use ndarray to record frame shape and axis labels

rhaas80:


lib/metadata/NDArray.cpp
```diff
@@ -38,4 +39,42 @@ void GenericNDArray::output_metadata(std::ostream& os) const {
...
 }
 
+// this templated function implemets a reurives 2D loop calling itself until
+// my_rank == rank and my_datatype == datatype, then it creates a NDArray of that
+// rank and type. This resolves the runtime values into compile time constants.
+template<DataType my_datatype, size_t my_rank>
+static inline std::shared_ptr<GenericNDArray>
+make_NDArray(const DataType datatype, const size_t rank, const std::vector<std::ptrdiff_t>& extents,
+             const std::vector<Symbol>& dimnames, void* data) {
+    // recursse until datatype matches
```
eschnett: `recursse` -> `recurse`

lib/stages/hdf5FileWrite.cpp
```diff
@@ -104,33 +104,32 @@ class hdf5FileWrite : public kotekan::Stage {
      * @param full_path The full path, including file name, for the file.
      * @param frame uint8_t pointer to the frame data
      * @param meta shared pointer to the chordMetadata object for this frame.
+     * @param frame_desc shared pointer to the NDarray description for this frame.
      */
     void write_chord(const std::string& full_path, const std::uint8_t* const frame,
-                     const std::shared_ptr<const chordMetadata> meta) {
+                     const std::shared_ptr<const chordMetadata> meta,
```
eschnett, line 110: It's customary to pass `std::shared_ptr` (like `std::string`) as `const &` if possible, unlike plain C pointers. I think this would work here. This saves a few atomic operations (not really important here) and meets the expectations of C++ programmers.

eschnett (approved):
I left some minor comments.

I wonder how large the code is that dispatches onto the array type and rank.

I assume the Julia kernel templates also need updating?

lib/stages/hdf5FileWrite.cpp
rhaas80: I'll fix `const std::shared_ptr<const GenericNDArray>` here and `const std::shared_ptr<const chordMetadata>` in the commit that deal with `chordMetadata` and thread safety (since I believe I am touching these files anyway there).

rhaas80:
> I left some minor comments.

I addressed the typo and the `const shared_ptr` for `GenericNDArray` (but not `chordMetadata` will do that one in the pull request on `chordMetadata`). 

> I wonder how large the code is that dispatches onto the array type and rank.

`NDArray.o` on cx67 (which is mostly just that templated dispatcher) is 1,883,808 bytes without debug symbols (running `strip` over the object file). A C++ function produced by this:

```perl
print "switch (datatype) {\n";
foreach my $type ( "uint1x8", "uint4x2", "uint8", "uint16", "uint32", "uint64", "int4x2", "int4x2_swapped_withoffset", "int8", "int16", "int32", "int64", "float16", "float32", "float64") {
print "case kotekan::DataType::$type:\n";
print "switch (rank) {\n";
foreach my $rank (1,2,3,4,5,6,7,8,9,10) {
print "case $rank:\n";
print "return std::shared_ptr<GenericNDArray>(new NDArray<GetType_t<kotekan::DataType::$type>, $rank>(
                extents, dimnames, static_cast<GetType_t<kotekan::DataType::$type>*>(data)));\n"
}
print "default: assert(0); return nullptr;\n";
print "}\n";
print "break;\n";
}
print "default: assert(0);return nullptr;\n";
print "}\n";
```

is 1,838,480 bytes.So both are about the same size.  Does `NDArray` actually need to be templated? One would assume that code that uses it on the GPU copies the respective array dimensions into a (compile time known) set of local variables anyway (and similar on the CPU side). 
 
> I assume the Julia kernel templates also need updating?

Yes,  I was not really able to see how / where they get the information from `chordMetadata`. `cudoCopyFromRingbuffer` has been updated, but it relies on the GPU having a `chordMetadata` with the array information, which is obviously not helpful.

PR #1318, opened 2025-10-22
Add `has_` functions for metadata

eschnett:
This allows checking whether a metadata key exists.

### DIFF
--- a/lib/metadata/chordMetadata.hpp
+++ b/lib/metadata/chordMetadata.hpp
@@ -246,6 +246,11 @@ public:
     }
 
     // TODO: remove this, it's redundant
+    bool has_gps_time() const {
+        return has_fpga_seq_num();
+    }
+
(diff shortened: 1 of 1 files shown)


geoffryan (changes requested):
There are optional (json-based) fields that do not yet have `has` methods:

- `beam_coord`
- `gps_time` (based on presence of `fpga_seq_num`)
- `rfi_num_bad_inputs`
- `rfi_flagged_samples`
- `lost_timesamples`
- `first_packet_recv_time`
- `stream_id`
- `dataset_id`

I think several (most) of these will be removed or altered in the future, but for now it would be useful to have full coverage.

eschnett:
Implemented these, except `has_gps_time` because it always exists.

geoffryan (approved):
lgtm!

PR #1321, opened 2025-10-23
checked access in chordMetadata and N2 hdf5 writing

geoffryan:
Replaces the `chordMetadata` and `jsonMetadata` getters with `at` accesses instead of `[]`.  This will throw exceptions on invalid accesses instead of an assertion error.

Also adds capability to `hdf5FileWrite` to write buffers with `N2Metadata`.

### DIFF
--- a/lib/stages/hdf5FileWrite.cpp
+++ b/lib/stages/hdf5FileWrite.cpp
@@ -1,5 +1,7 @@
-#include "Config.hpp"          // for Config
-#include "DataType.hpp"        // for type_to_string
+#include "Config.hpp"   // for Config
+#include "DataType.hpp" // for type_to_string
+#include "N2FrameView.hpp"
+#include "N2Metadata.hpp"
 #include "buffer.hpp"          // for Buffer
 #include "bufferContainer.hpp" // for bufferContainer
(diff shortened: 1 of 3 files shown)


lib/stages/hdf5FileWrite.cpp
```diff
@@ -160,112 +401,26 @@ class hdf5FileWrite : public kotekan::Stage {
...
-                        std::vector<int>(meta->dish_index, meta->dish_index
-                                                               + meta->n_dish_locations_ns
-                                                                     * meta->n_dish_locations_ew));
+                // Fetch metadata
+                const std::shared_ptr<const metadataObject> mc = buffer->get_metadata(frame_id);
+                if (!mc)
+                    FATAL_ERROR("Buffer \"{:s}\" frame {:d} does not have metadata",
+                                buffer->buffer_name, frame_id);
+                assert(mc);
+                if (!(metadata_is_chord(mc) || metadata_is_N2(mc)))
```
eschnett: Can you move this error message into a new `else` branch below?

eschnett (changes requested):

lib/stages/hdf5FileWrite.cpp
geoffryan: Fixed in c3edcba

eschnett (approved):

PR #1322, opened 2025-10-24
improve chordMetadata thread safety

rhaas80:


lib/metadata/BasebandMetadata.cpp
geoffryan, line 1: The `from_json` should use the checked access `.at()` instead of unchecked `[]`.

lib/metadata/BeamMetadata.cpp
geoffryan, line 1: As above, the `from_json` should use the checked access `.at()` instead of unchecked `[]`.

lib/metadata/HFBMetadata.cpp
geoffryan, line 1: As above, the `from_json` should use the checked access `.at()` instead of unchecked `[]`.

lib/metadata/N2Metadata.cpp
geoffryan, line 1: As above, the `from_json` should use the checked access `.at()` instead of unchecked `[]`.

lib/metadata/chordMetadata.hpp
geoffryan, line 1: Does this need a `void to_json(nlohmann::json& j, const chordMetadata& m)` and `void from_json(const nlohmann::json& j, BasebandMetadata& m)` ?

Should the `to_json` include (perhaps temporarily) the non-json'd members?

lib/utils/CHORDTelescope.cpp
geoffryan, line 1: As above, the from_json should use the checked access .at() instead of unchecked [].

geoffryan (changes requested):
Looks good! Can't spot any problems with `deepCopy`.  A few comments on the `json`:

1. I think the `from_json` methods should use checked accesses `.at()`, unless there's a pressing reason not to.
2. The `to_json` (I believe) always constructs a fresh `json` (with no existing members), so the checked `emplace()` can be used instead of `[]`.
3. `chordMetadata` is missing to/from json functions.
4. While existing (and future) CI tests will cover the `deepCopy` functionality, I don't think the to/from json is covered by any testing.  I think the right place for this would be in `tests/boost`.  Could you add some quick tests there that ensure the metadata  types can survive the round trip through json?

lib/metadata/chordMetadata.hpp
rhaas80: it should, it must.

rhaas80:
> 1. I think the `from_json` methods should use checked accesses `.at()`, unless there's a pressing reason not to.

No reason why I should not use `.at()`. Fixed.

> 2. The `to_json` (I believe) always constructs a fresh `json` (with no existing members), so the checked `emplace()` can be used instead of `[]`.

Sounds good. Fixed.

> 3. `chordMetadata` is missing to/from json functions.

hmm, let's see. Oh right, it has the `to_json` member function (must have, it's pure-virtual in `metadataObject`) but no free-standing to/from json code. That will actually be some minutes of work since I also need to serialize / de-serialize that leftover structure members (eg the array description)....

> 4. While existing (and future) CI tests will cover the `deepCopy` functionality, I don't think the to/from json is covered by any testing.  I think the right place for this would be in `tests/boost`.  Could you add some quick tests there that ensure the metadata  types can survive the round trip through json?

Sure. Tests provided in commit 31176d7d1

That triggered a bunch of other changes since eg the to / from bytes serializations were also not quite complete yet. The unit test above checks both serialization to / from json and to / from bytes.

geoffryan (approved):
Fantastic!  LGTM

rhaas80:
Very good. I'll merge just after #1327 (and once I've removed access to `name`).

lib/metadata/chordMetadata.hpp
```diff
@@ -46,6 +46,7 @@ const int CHORD_META_MAX_VIS_SAMPLES = 64;
 class chordMetadata : public metadataObject {
 public:
     chordMetadata();
+    bool operator==(const chordMetadata& other);
```
rhaas80: actually should this be `operator==() const`?

PR #1323, opened 2025-10-27
N2Accumulate Version 0

geoffryan:
This updates the `N2Accumulate` stage to take `n2k`-formatted input, as well as work with updated `chordMetadata`. It also contains miscellaneous bug fixes in the `n2k` correlator wrapper classes, tests for N2Accumulate as well as the `n2k` correlator kernels, and some general updated test stages.


config/ci-tests/verify_cuda_n2k_pl1bitcorr.yaml
```diff
@@ -162,3 +162,4 @@ check_data:
     first_buf: host_counts_buffer
     second_buf: simulated_counts_buffer
     num_frames_to_test: buffer_depth
+    check_metadata: True
```
jbmertens: Out of scope for this PR, but we should have an example config file that includes comments/docs for available stages, options, etc.

lib/cuda/cudaCorrelator.cpp
```diff
@@ -127,17 +127,25 @@ cudaEvent_t cudaCorrelator::execute(cudaPipelineState&, const std::vector<cudaEv
...
 
-    // Since we do not use a ring buffer we need to set `meta->sample0_offset`
+    // The input ringbuffers do not contain time-dependent data,
+    // so we must reconstruct it here. (fpga_seq_num and sample0_offset)
     // TODO: do this automatically in `NDArrayRingBuffer`
-    out_meta->set_sample0_offset(voltage.get_read_valid().begin());
+
+    // Assuming the voltage is coming in at the raw sample rate.
+    out_meta->set_fpga_seq_num(voltage.get_read_valid().begin());
+    out_meta->set_sample0_offset(out_meta->get_fpga_seq_num() / _sub_integration_ntime);
```
eschnett, line 133: This division by `_sub_integration_ntime` is wrong. Below you increase `out_time_downsampling_fpga` by `_sub_integration_ntime`, and this will also be applied to `sample0_offset`, so you would downsample twice:

```
    // The actual (possibly fractional) time sample index is calculated as follows:
    //     T_actual = (sample0_offset + T / offset_downsampling + half_fpga_sample0[F] / 2) /
    //                time_downsampling_fpga[F]
    // where `T` is the time sample index (the slowest varying index)
    // and `F` is the coarse frequency index.
```

lib/cuda/cudaPL1bitCorrelator.cpp
```diff
@@ -207,21 +207,37 @@ cudaEvent_t cudaPL1bitCorrelator::execute(cudaPipelineState& /*pipestate*/,
...
+    // The input ringbuffers do not contain time-dependent data,
+    // so we must reconstruct it here. (fpga_seq_num and sample0_offset)
     // TODO: do this automatically in `NDArrayRingBuffer`
-    out_meta->set_sample0_offset(rfi_RFImask.get_read_valid().begin());
+
+    // The RFI mask has a fast time index of size 1024, so has an apparent
+    // time downsampling of 1024. To get the needed outgoing fpga_seq_num,
+    // we need to undo that.
+    out_meta->set_fpga_seq_num(1024 * rfi_RFImask.get_read_valid().begin());
+    out_meta->set_sample0_offset(
```
eschnett, line 211: Ditto, the division by `n2k_sub_integration_ntime` is wrong.

lib/cuda/cudaPL1bitCorrelator.cpp
```diff
@@ -207,21 +207,37 @@ cudaEvent_t cudaPL1bitCorrelator::execute(cudaPipelineState& /*pipestate*/,
...
+        out_time_downsampling_fpga[f] =
+            n2k_sub_integration_ntime * div_noremainder(in_time_downsampling_fpga[f], 64);
+        out_half_fpga_sample0[f] =
+            in_half_fpga_sample0[f] + out_time_downsampling_fpga[f] - in_time_downsampling_fpga[f];
     }
-    out_meta->set_time_downsampling_fpga(time_downsampling_fpga);
+    out_meta->set_time_downsampling_fpga(out_time_downsampling_fpga);
+    out_meta->set_half_fpga_sample0(out_half_fpga_sample0);
 
-    n2k_counts.set_to_poison(0xff);
```
eschnett, line 224: Why are you disabling poisoning?

lib/metadata/N2Metadata.hpp
```diff
@@ -107,41 +107,24 @@ inline std::shared_ptr<N2Metadata> get_N2_metadata(const std::shared_ptr<metadat
     return std::static_pointer_cast<N2Metadata>(mc);
 }
 
+inline std::shared_ptr<const N2Metadata>
+get_N2_metadata(const std::shared_ptr<const metadataObject> mc) {
```
eschnett: This should probably take a `const std::shared_ptr &` instead.

lib/stages/N2Accumulate.cpp
```diff
@@ -37,54 +38,114 @@ REGISTER_KOTEKAN_STAGE(N2Accumulate);
 N2Accumulate::N2Accumulate(Config& config, const std::string& unique_name,
                            bufferContainer& buffer_container) :
     Stage(config, unique_name, buffer_container, std::bind(&N2Accumulate::main_thread, this)),
+    _tel(Telescope::instance().cast<CHORDTelescope>()),
     skipped_frame_counter(Metrics::instance().add_counter(
         "kotekan_N2accumulate_skipped_frame_total", unique_name, {"freq_id", "reason"})) {
 
-    auto& tel = Telescope::instance();
+    // auto& tel = Telescope::instance();
```
eschnett: Remove this line

lib/stages/N2Accumulate.cpp
```diff
@@ -37,54 +38,114 @@ REGISTER_KOTEKAN_STAGE(N2Accumulate);
...
     // number of frequencies in frame
-    _num_freq_in_frame = config.get<int32_t>(unique_name, "num_freq_in_frame");
+    _num_freq_per_n2k_frame = config.get<int64_t>(unique_name, "num_freq_per_n2k_frame");
+    assert(_num_freq_per_n2k_frame > 0);
+
+    // accumulation setup
+    _num_n2k_samples_to_accumulate =
+        config.get<int64_t>(unique_name, "num_n2k_samples_to_accumulate");
+
+    assert(_num_n2k_samples_to_accumulate > 0);
```
eschnett: It's probably better to `std::abort` here instead of using `assert` (which is disabled in production) or calling `FATAL_ERROR` (which will return so the code below still executes).

lib/stages/N2Accumulate.cpp
```diff
@@ -37,54 +38,114 @@ REGISTER_KOTEKAN_STAGE(N2Accumulate);
...
+    _n_fpga_samples_per_n2k_frame = config.get<int64_t>(unique_name, "samples_per_data_set");
+    _n_fpga_samples_per_n2k_correlation = config.get<int64_t>(unique_name, "sub_integration_ntime");
+    assert(_n_fpga_samples_per_n2k_frame > 0);
+    assert(_n_fpga_samples_per_n2k_correlation > 0);
+    assert(_n_fpga_samples_per_n2k_frame % _n_fpga_samples_per_n2k_correlation == 0);
+
+    _n_integrations_per_n2k_frame =
+        _n_fpga_samples_per_n2k_frame / _n_fpga_samples_per_n2k_correlation;
+
+    //_n_vis_samples_per_N2_output_frame =
```
eschnett: No need to keep old code.

lib/stages/N2Accumulate.cpp
```diff
@@ -37,54 +38,114 @@ REGISTER_KOTEKAN_STAGE(N2Accumulate);
...
-    // blocked matrix x frequencies from the GPU)
-    _vis = std::vector<int32_t>(2 * _num_N2_products_freqs, 0); // vis with complex as 2 ints
-    _vis_even = std::vector<int32_t>(2 * _num_N2_products_freqs,
-                                     0); // store even vis matrix for differencing
-    _weights = std::vector<int32_t>(_num_N2_products_freqs, 0); // real-valued weights
+
+    // Number of elements (polarization x dish) in the array.
+    _num_elements = config.get<int64_t>(unique_name, "num_elements");
+    assert(_num_elements > 0);
+    // Check its consistent with the count & correlation blocksizes.
```
eschnett: "its" -> "it's"

lib/stages/N2Accumulate.cpp
```diff
@@ -37,54 +38,114 @@ REGISTER_KOTEKAN_STAGE(N2Accumulate);
...
+    _rfi_downsampling_factor = config.get<int64_t>(unique_name, "rfi_downsampling_factor");
+
+    // _n2k_correlation_num_products * _num_freq_per_n2k_frame = _num_N2_products *
+    // _num_freq_per_n2k_frame; Number of products to accumulate _num_accum_products =
+    // N2::get_num_prod(_num_elements);
+
+    // Initializing these here using the computed _n2k_correlation_num_products *
+    // _num_freq_per_n2k_frame (accumulate the full, blocked matrix x frequencies from the GPU)
+    _vis = std::vector<int32_t>(2 * _num_freq_per_n2k_frame * _n2k_correlation_num_products,
+                                0); // vis with complex as 2 ints
```
eschnett, line 148: Use `std::complex<std::int32_t>` instead?

lib/stages/N2Accumulate.cpp
```diff
@@ -37,54 +38,114 @@ REGISTER_KOTEKAN_STAGE(N2Accumulate);
...
+    _vis_even = std::vector<int32_t>(2 * _num_freq_per_n2k_frame * _n2k_correlation_num_products,
+                                     0); // store even vis matrix for
```
eschnett: Incomplete comment?

lib/stages/N2Accumulate.cpp
```diff
@@ -130,71 +215,85 @@ void N2Accumulate::main_thread() {
...
             }
 
+            int64_t corr_offset = vis_samp_n * corr_stride_t;
+
             // Actual accumulation over
-            for (size_t d = 0; d < 2 * _num_N2_products_freqs; ++d) {
-                _vis[d] += input[d];
+            for (int64_t d = 0; d < 2 * _n2k_correlation_num_products * _num_freq_per_n2k_frame;
+                 ++d) {
+                _vis[d] += corr[d + corr_offset];
```
eschnett: Add some index checks? I recommend checking the actual buffer size (where buffer pointers are obtained) and ensuring here that the buffer size, converted to number of elements, is at least / exactly equal to the loop bounds here.

lib/stages/N2Accumulate.cpp
```diff
@@ -130,71 +215,85 @@ void N2Accumulate::main_thread() {
...
             } // d
 
             // If we're working on an even sample, store it for differencing
             // with an odd sample. If odd, add to the _weights matrix.
             // Potential optimization: copying vis_even is only really
             // necessary if we've started accumulating a new frame
             if (vis_sample_num_abs % 2 == 0) {
-                std::copy(input, input + 2 * _num_N2_products_freqs, _vis_even.begin());
+                std::copy(corr, corr + 2 * _n2k_correlation_num_products * _num_freq_per_n2k_frame,
```
eschnett: Same here, check that all buffers have the right size.

lib/stages/N2Accumulate.cpp
```diff
@@ -130,71 +215,85 @@ void N2Accumulate::main_thread() {
...
             } else {
-                for (size_t d = 0; d < _num_N2_products_freqs; ++d) {
-                    int32_t dr = _vis[2 * d + 0] - _vis_even[2 * d + 0];
-                    int32_t di = _vis[2 * d + 1] - _vis_even[2 * d + 1];
-                    _weights[d] += (dr * dr + di * di);
+                for (int64_t d = 0; d < _n2k_correlation_num_products * _num_freq_per_n2k_frame;
+                     ++d) {
+                    float dr = corr[corr_offset + 2 * d + 0] - _vis_even[2 * d + 0];
+                    float di = corr[corr_offset + 2 * d + 1] - _vis_even[2 * d + 1];
+                    _weights[d] += dr * dr + di * di;
```
eschnett, line 341: This could be a one-liner: https://en.cppreference.com/w/cpp/numeric/complex/norm.html

lib/stages/N2Accumulate.cpp
```diff
@@ -205,12 +304,58 @@ void N2Accumulate::main_thread() {
...
+            // Furthermore, the RFI mask is only computed every rfi_downsampling_factor
+            // samples, so t can be incremented by rfi_downsampling_factor.
+            // rfi_downsampling factor itself must be divisible by 32, so we know
+            // tbit will always be 0.
+            for (int64_t f = 0; f < _num_freq_per_n2k_frame; ++f) {
+                for (int64_t t = vis_samp_n * _n_fpga_samples_per_n2k_correlation;
+                     t < (vis_samp_n + 1) * _n_fpga_samples_per_n2k_correlation;
+                     t += _rfi_downsampling_factor) {
+
+                    int64_t thi = ((uint64_t)t) / 1024;
```
eschnett, line 401: The casts to `uint64_t` are superfluous.

lib/stages/N2Accumulate.cpp
```diff
@@ -219,57 +364,120 @@ bool N2Accumulate::output_and_reset(N2::frameID& in_frame_id, N2::frameID& out_f
...
+        for (int64_t ihi = 0; ihi < _n2k_correlation_lin_blocks; ihi++) {
+            // Lower triangular blocks
+            for (int64_t jhi = 0; jhi <= ihi; jhi++) {
+                for (int64_t ilo = 0; ilo < _n2k_correlation_blocksize; ilo++) {
+                    for (int64_t jlo = 0; jlo < _n2k_correlation_blocksize; jlo++) {
+                        // 2D indices into the N2K matrix.
+                        int64_t i = ilo + _n2k_correlation_blocksize * ihi;
+                        int64_t j = jlo + _n2k_correlation_blocksize * jhi;
+
+                        // Only procede if we're in the *true* lower-triangular section of the
```
eschnett: "procede" -> "proceed"

lib/stages/N2Accumulate.cpp
```diff
@@ -219,57 +364,120 @@ bool N2Accumulate::output_and_reset(N2::frameID& in_frame_id, N2::frameID& out_f
...
+                        //    3| 6  7  8  9            3|          9
+                        //
+                        // vis_N2(i, j) = vis_n2k(j, i)*
+
+                        int64_t n2_idx = N2::cmap(j, i, _num_elements);
+
+                        // Populate the visibility matrix
+                        N2::cfloat v = {(float)_vis[2 * idx],
+                                        -(float)_vis[2 * idx + 1]}; // conjugate
+                        out_vis.vis[n2_idx] = ins * v;
```
eschnett: Another one-liner: `out_vis[vis[n2_idx] = ins * _vis[idx].conj()`.

lib/stages/rawFileWrite.cpp
```diff
@@ -144,6 +144,7 @@ void rawFileWrite::main_thread() {
         // which can contain a "final" signal.
         if (_exit_after_n_files != 0 && file_num >= _exit_after_n_files) {
```
eschnett: Does this use the "official" shutdown mechanism or is this duplicating it? The official mechanism ensures that all writers write their requested number of files, not just one of the writers.

lib/stages/testDataGen.cpp
```diff
@@ -217,11 +222,15 @@ void testDataGen::main_thread() {
         std::vector<int64_t> half_fpga_sample0(coarse_freq.size());
         std::vector<int> time_downsampling_fpga(coarse_freq.size());
         for (int f = 0; f < static_cast<int>(coarse_freq.size()); f++) {
-            coarse_freq[f] = f;
+            if (_manual_freq_ids.size() > 0)
+                coarse_freq[f] = _manual_freq_ids[f % _manual_freq_ids.size()];
```
eschnett, line 227: Prefer `.at()` over `[]`. For reasons.

lib/stages/testDataGen.cpp
```diff
@@ -217,11 +222,15 @@ void testDataGen::main_thread() {
...
+            else
+                coarse_freq[f] = f;
             freq_upchan_factor[f] = 1;
-            half_fpga_sample0[f] = 0;
+            half_fpga_sample0[f] = _meta_time_downsample_factor - 1;
```
eschnett, line 231: That looks weird – shouldn't it be `_meta_time_downsample_factor / 2` or somesuch?

lib/testing/applyGenPL.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+        int fstride = _num_elements;
+        int tstride = _num_local_freq * fstride;
+        int fstride_pl = _num_elements / 8;
+        int tstride_pl = (_num_local_freq / 4) * fstride_pl;
+
+        for (int t = 0; t < _samples_per_data_set; t++) {
+            for (int f = 0; f < _num_local_freq; f++) {
+                for (int e = 0; e < _num_elements; e++) {
+                    int pl_t = t >> 1;
```
eschnett: `%` and `/` have the same speed as `&` and `>>`.

lib/testing/applyGenPL.cpp
```diff
@@ -0,0 +1,104 @@
...
+                    int pl_t_bit = pl_t & 0x3F; // last 6 bits to index into a
+                                                // uint64_t
+                    int pl_t_slow = pl_t >> 6;  // the rest are for the array
+                    int pl_f = f >> 2;          // downsample by 4
+                    int pl_e = e >> 3;          // downsample by 8
+
+                    int pl_idx = tstride_pl * pl_t_slow + fstride_pl * pl_f + pl_e;
+                    int pl = (plmask[pl_idx] >> pl_t_bit) & 0x1;
```
eschnett, line 124: I strongly recommend checking the buffer sizes before the loop. Otherwise the segfaults depend on typos in the yaml files.

lib/testing/gpuSimulateN2kCorr.cpp
```diff
@@ -188,21 +192,30 @@ void gpuSimulateN2kCorr::main_thread() {
         meta_out->set_array_dimension(4, 16, "DPlo2");
         meta_out->set_array_dimension(5, 2, "C");
         meta_out->set_strides_simple();
+
+        meta_out->set_fpga_seq_num(meta_in->get_fpga_seq_num());
+        meta_out->set_sample0_offset(meta_in->get_sample0_offset() / _sub_integration_ntime);
```
eschnett, line 278: No division

lib/testing/gpuSimulateN2kCorr.cpp
```diff
@@ -188,21 +192,30 @@ void gpuSimulateN2kCorr::main_thread() {
...
         std::vector<int> coarse_freq(_num_local_freq);
         std::vector<int> time_downsampling_fpga(_num_local_freq);
+        std::vector<int64_t> half_fpga_sample0(_num_local_freq);
+        const std::vector<int> coarse_freq_in = meta_in->get_coarse_freq();
         const std::vector<int> time_downsampling_fpga_in = meta_in->get_time_downsampling_fpga();
+        const std::vector<int64_t> half_fpga_sample0_in = meta_in->get_half_fpga_sample0();
+
         for (int f = 0; f < _num_local_freq; f++) {
-            coarse_freq[f] = f; // TODO: set some actual frequency indices and a stream_id
+            coarse_freq[f] = coarse_freq_in[f];
```
eschnett, line 289: Recommend `.at()` instead of `[]` by default, unless in a loop with explicit checking before the loop.

lib/testing/gpuSimulateN2kPL1bitCorr.cpp
```diff
@@ -194,9 +198,29 @@ void gpuSimulateN2kPL1bitCorr::main_thread() {
...
+        //  sample0_offset(out) = sample0_offset(in) * dim[0](out) / dim[0](in)
+        //
+        //     dim[0](out) = num_integrations = samples_per_data_set / sub_integration_ntime
+        //     dim[0](in) = samples_per_data_set / 64
+        //
+        //  sample0_offset(out) = sample0_offset(in) * 64 / sub_integration_ntime
+
         meta_out->set_fpga_seq_num(meta_in->get_fpga_seq_num());
-        meta_out->set_sample0_offset(meta_in->get_sample0_offset());
+        meta_out->set_sample0_offset((64 * meta_in->get_sample0_offset()) / _sub_integration_ntime);
```
eschnett: No rescaling

lib/testing/gpuSimulateN2kPL1bitCorr.cpp
```diff
@@ -205,12 +229,16 @@ void gpuSimulateN2kPL1bitCorr::main_thread() {
...
         std::vector<int64_t> half_fpga_sample0(coarse_freq.size());
         std::vector<int> time_downsampling_fpga(coarse_freq.size());
+
         for (int f = 0; f < static_cast<int>(coarse_freq.size()); f++) {
             coarse_freq[f] = coarse_freq_in[f];
             freq_upchan_factor[f] = freq_upchan_factor_in[f];
-            half_fpga_sample0[f] = half_fpga_sample0_in[f];
-            time_downsampling_fpga[f] = time_downsampling_fpga_in[f] * _sub_integration_ntime;
+            time_downsampling_fpga[f] =
+                (time_downsampling_fpga_in[f] / 64) * _sub_integration_ntime;
```
eschnett: Check divisibility?

lib/testing/gpuSimulateN2kPLExpand.cpp
```diff
@@ -152,19 +152,25 @@ void gpuSimulateN2kPLExpand::main_thread() {
...
-        meta_out->set_array_dimension(2, ne, "P");
-        meta_out->set_array_dimension(3, ne, "D8");
+        meta_out->set_array_dimension(2, 2, "P");
+        meta_out->set_array_dimension(3, ne / 2, "D8");
+        meta_out->set_array_dimension(4, 8, "Tlo64");
         meta_out->set_strides_simple();
 
         meta_out->set_fpga_seq_num(meta_in->get_fpga_seq_num());
-        meta_out->set_sample0_offset(meta_in->get_sample0_offset());
+        meta_out->set_sample0_offset(2 * meta_in->get_sample0_offset());
```
eschnett, line 176: No rescalling

lib/testing/testN2kGen.cpp
```diff
@@ -0,0 +1,423 @@
...
+    num_integrations = samples_per_data_set / sub_integration_ntime;
+
+    corr_lin_blocks = num_elements / corr_blocksize;
+    count_lin_blocks = (num_elements / 8) / count_blocksize;
+
+    corr_num_blocks = (corr_lin_blocks * (corr_lin_blocks + 1)) / 2;
+    count_num_blocks = (count_lin_blocks * (count_lin_blocks + 1)) / 2;
+}
+
+const std::shared_ptr<chordMetadata> testN2kGen::get_new_metadata(Buffer* buf, frameID frame_id) {
```
eschnett: It does not make sense to return a `const` value. const-ness is a property of the object holding the value, not of the type of the value.

lib/testing/testN2kGen.cpp
```diff
@@ -0,0 +1,423 @@
...
+    meta->set_sample0_offset(seq_num / sub_integration_ntime);
+    meta->set_offset_downsampling(1);
+
+    std::vector<int> coarse_freq(num_local_freq);
+    std::vector<int> freq_upchan_factor(num_local_freq);
+    std::vector<int> time_downsampling_fpga(num_local_freq);
+    std::vector<int64_t> half_fpga_sample0(num_local_freq);
+
+    for (int f = 0; f < num_local_freq; f++) {
+        coarse_freq[f] = freq_ids[f % freq_ids.size()];
```
eschnett, line 221: `.at()`

lib/testing/testN2kGen.cpp
```diff
@@ -0,0 +1,423 @@
...
+    }
+
+    meta->set_coarse_freq(coarse_freq);
+    meta->set_freq_upchan_factor(freq_upchan_factor);
+    meta->set_time_downsampling_fpga(time_downsampling_fpga);
+    meta->set_half_fpga_sample0(half_fpga_sample0);
+    assert(meta->get_nfreq() <= CHORD_META_MAX_FREQ);
+}
+
+void testN2kGen::set_counts_metadata(std::shared_ptr<chordMetadata> meta, uint64_t seq_num) {
```
eschnett: `const std::shared_ptr<chordMetadata>& meta`?

lib/testing/testN2kGen.cpp
```diff
@@ -0,0 +1,423 @@
...
+                                     int& jlo, int& block_idx) {
+
+    ihi = i / blocksize;
+    jhi = j / blocksize;
+    ilo = i % blocksize;
+    jlo = j % blocksize;
+
+    // Only reference lower triangular blocks (jhi <= ihi), so swap if jhi > ihi
+    if (jhi > ihi) {
+        int temp = ihi;
```
eschnett: `std::swap`

lib/testing/testN2kGen.cpp
```diff
@@ -0,0 +1,423 @@
...
+                        assert(count_block_idx == (ihi * (ihi + 1)) / 2 + jhi);
+
+                        for (int ilo = 0; ilo < count_blocksize; ilo++) {
+                            for (int jlo = 0; jlo < count_blocksize; jlo++) {
+                                int idx = jlo + ilo * count_blocksize + count_block_idx * db_count
+                                          + f * df_count + t * dt_count;
+
+                                if (count_type == "const") {
+                                    if (count_value_array.size() > 0) {
+                                        count[idx] = count_value_array[count_val_idx
```
eschnett, line 386: Please check the buffer sizes. They are taken from yaml code.

lib/testing/testN2kGen.cpp
```diff
@@ -0,0 +1,423 @@
...
+                    for (int jhi = 0; jhi <= ihi; jhi++) {
+
+                        assert(count_block_idx == (ihi * (ihi + 1)) / 2 + jhi);
+
+                        for (int ilo = 0; ilo < count_blocksize; ilo++) {
+                            for (int jlo = 0; jlo < count_blocksize; jlo++) {
+                                int idx = jlo + ilo * count_blocksize + count_block_idx * db_count
+                                          + f * df_count + t * dt_count;
+
+                                if (count_type == "const") {
```
eschnett, line 384: String comparisons in an inner loop?

lib/testing/testN2kGen.cpp
```diff
@@ -0,0 +1,423 @@
...
+                        assert(corr_block_idx == (ihi * (ihi + 1)) / 2 + jhi);
+
+                        for (int ilo = 0; ilo < corr_blocksize; ilo++) {
+                            for (int jlo = 0; jlo < corr_blocksize; jlo++) {
+                                int idx = 2 * (jlo + ilo * corr_blocksize)
+                                          + corr_block_idx * db_corr + f * df_corr + t * dt_corr;
+
+                                if (corr_type == "const") {
+                                    if (corr_value_array.size() > 0) {
+                                        corr[idx + 0] =
```
eschnett, line 425: `std::complex`

lib/testing/testN2kGen.hpp
```diff
@@ -0,0 +1,102 @@
...
+    std::array<int32_t, 2> corr_max;
+    int32_t count_value;
+    std::vector<int32_t> count_value_array;
+    int32_t count_min;
+    int32_t count_max;
+    int32_t rfi_value;
+    std::vector<int32_t> rfi_value_array;
+    int32_t rfi_min;
+    int32_t rfi_max;
+    bool mul_correlation_by_counts;
```
eschnett, line 72: What does this do?

lib/testing/testN2kGen.cpp
```diff
@@ -0,0 +1,423 @@
...
+
+                        assert(corr_block_idx == (ihi * (ihi + 1)) / 2 + jhi);
+
+                        for (int ilo = 0; ilo < corr_blocksize; ilo++) {
+                            for (int jlo = 0; jlo < corr_blocksize; jlo++) {
+                                int idx = 2 * (jlo + ilo * corr_blocksize)
+                                          + corr_block_idx * db_corr + f * df_corr + t * dt_corr;
+
+                                if (corr_type == "const") {
+                                    if (corr_value_array.size() > 0) {
```
eschnett, line 424: `!corr_value_array.empty()`

lib/testing/testN2kGen.hpp
```diff
@@ -0,0 +1,102 @@
...
+    Buffer* corr_buf;
+    Buffer* count_buf;
+    Buffer* rfi_buf;
+    std::string corr_name;
+    std::string corr_type;
+    std::string count_name;
+    std::string count_type;
+    std::string rfi_name;
+    std::string rfi_type;
+    std::array<int32_t, 2> corr_value;
```
eschnett, line 64: Are these complex numbers?

lib/testing/testN2kGen.hpp
```diff
@@ -0,0 +1,102 @@
...
+    int corr_blocksize;  // ALWAYS 16
+    int count_blocksize; // ALWAYS 8
+    int corr_lin_blocks;
+    int count_lin_blocks;
+    int corr_num_blocks;
+    int count_num_blocks;
+    int num_integrations;
+
+    const std::shared_ptr<chordMetadata> get_new_metadata(Buffer* buf, frameID frame_id);
+    void set_correlation_metadata(std::shared_ptr<chordMetadata> meta, uint64_t seq_num);
```
eschnett: `shared_ptr` -> `const shared_ptr&`

lib/testing/testN2kGen.hpp
```diff
@@ -0,0 +1,102 @@
...
+    bool mul_correlation_by_counts;
+    int32_t samples_per_data_set;
+    int32_t sub_integration_ntime;
+    int32_t num_frames;
+    int32_t num_local_freq;
+    int32_t num_elements;
+    std::vector<uint32_t> freq_ids;
+    uint32_t seed;
+
+    int corr_blocksize;  // ALWAYS 16
```
eschnett: If always 16, then:
`constexpr int corr_blocksize = 16`

lib/utils/CHORDTelescope.cpp
```diff
@@ -667,7 +670,7 @@ freq_id_t CHORDTelescope::to_freq_id(stream_t, uint32_t) const {
 
 // TODO: This is a stub to satisfy inheritance and must be updated.
```
eschnett, line 671: Is this really a stub? If so, why not abort instead? Or is this real code with a wrong comment?

lib/utils/CHORDTelescope.cpp
```diff
@@ -667,7 +670,7 @@ freq_id_t CHORDTelescope::to_freq_id(stream_t, uint32_t) const {
...
 double CHORDTelescope::to_freq(freq_id_t freq_id) const {
-    return freq_id * (GIGA / dt_ns);
+    return freq_id * (GIGA / (double)dt_ns);
```
eschnett, line 673: In which unit is the frequency measured? This calculation looks wrong. Exahertz?

tests/boost/test_CHORDTelescope.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include <sys/wait.h>
+#include <time.h>
+#include <vector>
+
+using kotekan::Config;
+using json = nlohmann::json;
+using kotekan::configUpdater;
+using kotekan::restServer;
+
+const std::string default_config_str = "{\
```
eschnett: https://en.cppreference.com/w/cpp/language/string_literal.html, in particular `R""`.

eschnett (changes requested):

lib/cuda/cudaCorrelator.cpp
geoffryan: We decided in-person that the division is the correct thing to do as implemented, but the proper solution is to fix this indexing in a future PR.

lib/cuda/cudaPL1bitCorrelator.cpp
geoffryan: That was a mistake, adding back in.

lib/stages/N2Accumulate.cpp
geoffryan: done.

lib/metadata/N2Metadata.hpp
geoffryan: Done.

lib/stages/N2Accumulate.cpp
geoffryan: Replaced throughout the constructor with FATAL_ERROR and std::abort().

lib/stages/N2Accumulate.cpp
geoffryan: Correct, removed!

lib/stages/N2Accumulate.cpp
geoffryan: It's done.

lib/stages/N2Accumulate.cpp
geoffryan: completed.

lib/stages/N2Accumulate.cpp
geoffryan: Added size checks in the constructor, and a few (possibly redundant) checks at the beginning of this section.  Found a bug in the process!

lib/stages/N2Accumulate.cpp
geoffryan: Buffers are checked in the constructor, and extra checks on the accumulation vectors have been added at the beginning of this section.

lib/stages/N2Accumulate.cpp
geoffryan: This is a (very slight) optimization that simplifies the assembly output.  Added an explanatory comment.

lib/stages/N2Accumulate.cpp
geoffryan: compleetd

lib/stages/rawFileWrite.cpp
geoffryan: This was an exiting mechanism in this stage already.  Updated now to use the `waiting_for_max_frames` mechanism.

lib/stages/testDataGen.cpp
geoffryan: Nope, the "2" is absorbed by the "half".  "half_sample0" should always be "new_downsample - old_downsample" and here "old_downsample" is 1.

Of course, this will be simplified in the upcoming refactor of this mechanism ;)

lib/testing/gpuSimulateN2kCorr.cpp
geoffryan: See above.

lib/testing/gpuSimulateN2kPL1bitCorr.cpp
geoffryan: See above.

lib/testing/gpuSimulateN2kPLExpand.cpp
geoffryan: It should be correct as implemented, see above discussion.

lib/testing/testN2kGen.hpp
geoffryan: Vestigial variables, removed.

lib/testing/testN2kGen.hpp
geoffryan: Fixed.

lib/stages/testDataGen.cpp
geoffryan: These vectors are all constructed immediately before this block with explicitly compatible sizes, and the `_manual_freq_ids` is size-checked in its access, so I think here the `[]` is ok?

lib/testing/gpuSimulateN2kCorr.cpp
geoffryan: Added an explicit check before the loop.

lib/testing/testN2kGen.cpp
geoffryan: Done.

lib/testing/testN2kGen.cpp
geoffryan: yep.

lib/testing/testN2kGen.cpp
geoffryan: These are constructed in this function to have the right size, and the `%` checks the access into `freq_ids`, which must have at least 1 element in the constructor.

lib/testing/testN2kGen.cpp
geoffryan: This stage does not have to be fast, it just has to be clear and correct, which I think this accomplishes.

lib/testing/testN2kGen.hpp
geoffryan: done

lib/testing/testN2kGen.cpp
geoffryan: C++ is amazing.

lib/cuda/cudaCorrelator.cpp
```diff
@@ -69,7 +67,7 @@ cudaCorrelator::cudaCorrelator(Config& config, const std::string& unique_name,
                                               n2k_dimnames, *this);
     }()),
     n2correlator(_num_elements, _num_local_freq) {
-    if (_samples_per_data_set % _sub_integration_ntime)
+    if (_num_times % _sub_integration_ntime)
```
jbmertens: _num_times could probably also use a better name, or docstring (number of times of what, and does this coincide with other num_times elsewhere)?

lib/testing/gpuSimulateN2kPL1bitCorr.cpp
```diff
@@ -200,9 +204,29 @@ void gpuSimulateN2kPL1bitCorr::main_thread() {
         /* test that things are consistent */
         meta_out->check_frame_desc(output_buf->get_frame_desc(output_frame_id));
 
+        //
```
jbmertens: Just checking this comment was left here intentionally.

lib/testing/testDataCheck.hpp
```diff
@@ -24,6 +25,98 @@
 #include <type_traits> // for is_same, enable_if
 #include <vector>      // for vector
 
+#define CHECK_META_SCALAR_INT_DIRECT(FIELD, META1, META2, ERR_COUNT, BUF_NAME1, FRAME_ID1,         \
+                                     BUF_NAME2, FRAME_ID2)                                         \
+    do {                                                                                           \
+        INFO("Checking meta field {:s}", #FIELD);                                                  \
+        if ((META1)->FIELD != (META2)->FIELD) {                                                    \
+            ERROR("metadata {:s}[{:d}] {:s} != {:s}[{:d}] {:s}; values: {:d} {:d}", (BUF_NAME1),   \
```
jbmertens: Will we ever want _NON_OO versions of these?
Maybe we could let the user specify if they want this called at the level of ERROR, FATAL_ERROR, WARN.

tests/boost/test_CHORDTelescope.cpp
geoffryan: Again, amazing C++.

lib/testing/applyGenPL.cpp
geoffryan: And its certainly clearer. Done.

lib/testing/applyGenPL.cpp
geoffryan: Added checks in the constructor.

lib/testing/gpuSimulateN2kPL1bitCorr.cpp
geoffryan: `div_noremainder` to the rescue.

lib/testing/testN2kGen.cpp
geoffryan: done.

lib/testing/testN2kGen.hpp
geoffryan: Yes.  This stage is supposed to emulate the `cudaCorrelator`, which encodes complexes as 2 ints, so I'm doing the same here.

(also this is read from the json, and we need write a tiny bit of boilerplate to read complex values from json, where they will be represented by two Numbers)

lib/utils/CHORDTelescope.cpp
geoffryan: I'm going to punt on this for my next PR, which will be revising the Telescope and getting it ready for production.

lib/utils/CHORDTelescope.cpp
geoffryan: As implemented this produces the frequency in Hz, I realize now the other telescopes produce a number in MHz.  That should be in the name of the function, but regardless, this is currently a bug.

I'm going to punt on this for my next PR, which will be revising the Telescope and getting it ready for production.

lib/testing/testN2kGen.cpp
geoffryan: Here we'll stick with the two ints like in `cudaCorrelator`

lib/stages/N2Accumulate.cpp
geoffryan: `std::complex<>` is implementation-defined for integer types.  Some things work (subtraction), some things don't (assigning from literals), and some things implicitly cast to a double (norm).  I'm going to say we should stick with the explicit integers for now.

lib/stages/N2Accumulate.cpp
geoffryan: Not quite, see above.

lib/stages/N2Accumulate.cpp
geoffryan: Still a 2-liner, but moved to conj() to be more explicit.

geoffryan:
Most suggestions implemented, much more bounds-checking throughout.  

Will not be doing the complex-type changes on integer values (see comments above), and punting the CHORDTelescope updates to the next PR.

eschnett (approved):

PR #1324, opened 2025-10-28
Use new CHIME packet capture buffer layout

eschnett:


config/fengine/include/fengine_chime.j2
```diff
@@ -228,13 +242,24 @@ run_send_voltage:
...
             host_voltage_ringbuffer: host_voltage_ringbuffer
         commands:
-        - name: cudaCopyToRingbuffer
-          input_size: num_dishes * num_polarizations * num_frequencies * num_times
+        #TODO - name: cudaCopyToRingbuffer
+        #TODO   input_size: num_dishes * num_polarizations * num_frequencies * num_times
+        #TODO   ring_buffer_size: buffer_depth * input_size
+        #TODO   in_buf: host_voltage_buffer
+        #TODO   signal_buf: host_voltage_ringbuffer
+        #TODO   gpu_mem_output: voltage_buffer
```
rhaas80: are these TODOs still relevant?

lib/cuda/cudaCopyNToRingbuffer.cpp
```diff
@@ -139,51 +139,56 @@ cudaEvent_t cudaCopyNToRingbuffer::execute(cudaPipelineState& /*pipestate*/,
...
-            throw std::runtime_error(
-                "cudaCopyNToRingbuffer: could not get chordMetadata from signal buffer");
+        // Get the first input metadata
+        const auto meta_in0 = std::dynamic_pointer_cast<chordMetadata>(
+            in_buffers.at(0)->get_metadata(gpu_frame_id % in_buffers.at(0)->num_frames));
+        if (!meta_in0)
+            throw std::runtime_error("cudaCopyNToRingbuffer: input buffer has no chordMetadata");
+
+        // Copy metadata (because we modify it)
+        auto meta_ring = std::make_shared<chordMetadata>(*meta_in0);
```
rhaas80, line 149: that will fail with https://github.com/kotekan/kotekan/pull/1322 which makes the copy (and assignment) operators private and only leaves `deepCopy` exposed. That then requires some dancing around things, eg:

```
auto meta_ring = std::make_shared<chordMetadata>();
meta_ring->deepCopy(meta_in0);
```

Fine right now though (since that pull request is not yet in).

lib/stages/FEngine.cpp
```diff
@@ -104,6 +104,9 @@ FEngine::FEngine(kotekan::Config& config, const std::string& unique_name,
     // Dish reordering
     scatter_indices(config.get_default<std::vector<int>>(unique_name, "scatter_indices", {})),
 
+    // Input buffers
+    receive_chime(config.get_default<bool>(unique_name, "use_chime_input_buffers", false)),
```
rhaas80, line 108: missing from doxygen.

lib/stages/FEngine.hpp
```diff
@@ -104,6 +104,9 @@ class FEngine : public kotekan::Stage {
     // Dish reordering
     const std::vector<int> scatter_indices;
 
+    // Input buffers
+    const bool receive_chime;
```
rhaas80, line 109: comment could be better. This flag indicates whether the input buffers use CHIME layout of CHORD layout, does it?

rhaas80 (approved):
nothing actually wrong. One option not described in doxygen. One code block will fail when next chordMetadata pull request is merged in due to chordMetadata copy / assignment becoming private.

config/fengine/include/fengine_chime.j2
eschnett: Removed.

lib/stages/FEngine.cpp
eschnett: Added.

lib/cuda/cudaCopyNToRingbuffer.cpp
eschnett: I'll try to be speedy.

lib/stages/FEngine.hpp
eschnett: Yes. Comment improved.

PR #1326, opened 2025-10-29
[VLBI Real-time Tracking Beamformer] Update kotekan config to use additional input (chan_id=32)

aaronpearlman:
Summary of Changes:

- On Oct. 27, LayoutDB was updated so that CHIME input (chan_id=32) is no longer a non-sky channel (i.e., NoiseSource object) and now accurately reflects that it is a usable sky channel (i.e., CHIMEAntenna object).
- To enable this input to be used in the beamformer, the corresponding feed position in the config has been changed to the value from LayoutDB, as queried through `ch_util` (i.e., it is no longer [0.0, 0.0], which would indicate an unused feed in the beamformer).
- Comments have been added in the config files to make it more clear which input each of the feed positions corresponds to.


config/chime_science_run_gpu.yaml
```diff
@@ -4057,1035 +4057,1035 @@ input_reorder:
...
-   [33.04511018536640420962, 36.38267898048457027471],
-   [33.04548788848250495676, 36.68747874646275164423],
-   [33.04586559159860570389, 36.99227851244094011918],
-   [33.04624329471470645103, 37.29707827841912148870],
-   [33.04662099783080719817, 37.60187804439730996364],
-   [33.04699870094690794531, 37.90667781037549133316],
-   [33.04737640406300869245, 38.21147757635367980811],
-   [33.04775410717910943958, 38.51627734233186117763],
-   [33.04813181029521018672, 38.82107710831004254715]]
+feed positions:
```
rikvl:
```suggestion
feed_positions:
```

config/chime_science_run_gpu_no_output.yaml
```diff
@@ -4057,1035 +4057,1035 @@ input_reorder:
...
-   [33.04511018536640420962, 36.38267898048457027471],
-   [33.04548788848250495676, 36.68747874646275164423],
-   [33.04586559159860570389, 36.99227851244094011918],
-   [33.04624329471470645103, 37.29707827841912148870],
-   [33.04662099783080719817, 37.60187804439730996364],
-   [33.04699870094690794531, 37.90667781037549133316],
-   [33.04737640406300869245, 38.21147757635367980811],
-   [33.04775410717910943958, 38.51627734233186117763],
-   [33.04813181029521018672, 38.82107710831004254715]]
+feed positions:
```
rikvl:
```suggestion
feed_positions:
```

rikvl (commented):
I think I found a typo: missing underscore in `feed_positions`

aaronpearlman:
You're right - My bad! Thank you very much for catching this, @rikvl!

config/chime_science_run_gpu_no_output.yaml
aaronpearlman: Fixed

config/chime_science_run_gpu.yaml
aaronpearlman: Fixed

ketiltrout:
Before we deploy this, can we figure out what to do about #1329 so we can potentially fix that, too at the same time?

ketiltrout:
I've merged the keep-warm changes (resolving #1329).  We should remember to tag this once this is merged, too.

python/scripts/generate_chime_kotekan_config_feed_positions.py
```diff
@@ -0,0 +1,69 @@
+import numpy as np
```
andrerenard, line 18: We should have a little documentation here about how this should be used.  But otherwise this looks good to me.

andrerenard (approved):

python/scripts/generate_chime_kotekan_config_feed_positions.py
ketiltrout: This has not been resolved.  Are there plans to do anything about this?  We want to deploy this on Thursday.

aaronpearlman: I added the documentation to this script in aff3ca9cb4e4b2c99588d0ccd1f02b75f2e902e6.

PR #1327, opened 2025-10-29
New field `quantity_name` in `NDArray`

eschnett:


lib/core/buffer.hpp
```diff
@@ -543,11 +544,12 @@ class Buffer : public GenericBuffer {
      * @param[in] extents Array extentds in the D dimensions
      * @param[in] dimnames Array axis labels in the D dimensions
      */
-    void allocate_new_frame_desc(int frame_id, kotekan::DataType value_type, size_t rank,
```
rhaas80, line 546: this removes the `rank` argument of `create` and breaks the symmetry between the "runtime" `create` and the templated `create`. 

While it is true that one can get the rank from the `size()` of `extents` the same is true for the templated version isn't it? So I would either keep it in both or remove it in both.

rhaas80 (changes requested):
Not really "must" addressed in the sense of "fixed" but at least "commented on".

lib/core/buffer.hpp
eschnett: In the templated version, the type and the rank are the template parameters, so they have to be specified at compile time. They cannot be inferred from the arguments to the constructor which are only known at run time.

In the non-templated version, the rank can be inferred from the constructor arguments. All calls to the constructor read something like `NDArray(..., dims.size(), dims, ...)`, which looked off.

eschnett:
Given how we are using NDArray in the buffers it seems to me that the templated version is overkill. A non-templated version which stores the array size in a `std::vector` and with a templated access that returns a `T*` where the caller specifies the type `T` (and which is checked at run time) would make more sense.

lib/core/buffer.hpp
rhaas80: yes, the calls are a bit redundant. I think even for the templated version inference is possible (for the `create` function at least, probably not for the `NDArray` type), ie this compiles (and runs):

```c++
#if 0
g++ -std=c++17 infer.cc
exit 0
#endif

#include <array>
#include <iostream>

template <size_t D> void fun(const std::array<int, D>& input) {
  std::cout << input.size() << "\n";
}

int main() {
  std::array<int, 2> foo = {1,2};
  fun(foo);
  return 0;
}
```

So I could have gotten rid of it in both instances actually (and gotten rid of the `T` template as well since the `T*` pointer already determines it).

However it seems that the templating may go away anyway, rendering the symmetry argument moot (https://github.com/kotekan/kotekan/pull/1327#issuecomment-3463201226) anyway :-)

rhaas80 (approved):

rhaas80:
@eschnett last call: merge now or have conflicts with #1322

geoffryan (commented):
I'm a little concerned/curious about creating `Symbol`s for the name, which is a bit of an expensive operation with the locks and whatnot.  Will these be created every time a stage creates a buffer (ie. every time a cuda kernel executes), or only once when the buffer is getting set up?  I worry about the short-cadence kernels spamming the Symbol table.

eschnett:
The idea of a symbol is that you create it once during the run-time and then remember it. If that doesn't work for us then we should get rid of it and use strings instead. Strings need atomic operations (also not cheap, but cheaper than a lock) every time a string is copied.

geoffryan:
It seems like we're mostly constructing Symbols from strings though so the lock gets engaged every time.  Should `chordMetadata::get_name()` return a Symbol instead of a string?

rhaas80:
> It seems like we're mostly constructing Symbols from strings though so the lock gets engaged every time. Should `chordMetadata::get_name()` return a Symbol instead of a string?

well do we need the optimization that the Symbol provides? They are like X11's Atoms in that they make comparison fast at the cost of creation (and reduce storage space, making each symbol use a constant amount of storage when referred to). 

Right now though: they are implemented and unless we find they are slowing things down unacceptably for some reasons: why change a mostly working setup?

PR #1330, opened 2025-11-03
Manage Python with uv

geoffryan:
Adds `uv` (https://docs.astral.sh/uv/) to manage the kotekan Python environment.  This makes the Python environment needed to build, run, and test kotekan completely defined within kotekan itself.  `uv` creates a local virtual environment and installs all Python dependencies itself.

- Adds `pyproject.toml` defining Python environment, `uv.lock` (managed by `uv` to specify exact environment), and `.python-version` to pin a particular required Python range (for now, >= 3.12 and < 3.13)
- C++ calls (in `kotekan/kotekan.cpp` and `tests/boost/test_timeUtil.cpp`) to Python use `uv run`
- pytest suite is run with ~~`uv run --env-file env_test pytest ...`.  The `env_test` file sets the `PYTHONPATH` to point to the `python/kotekan` modules.~~ `uv run pytest ...`.  The `python/` directory is added to pytest's path in `pyproject.toml`
- Github actions install and setup `uv` before compiling `kotekan`.  

Sharp edges:
- Makes `uv` a necessary dependency to run `kotekan`.  (Could be avoided by having `kotekan.cpp` check if `uv` exists, or make it a compile-time option)
- Users should run tests, etc, with `uv run ...` not `python ...`
- The `kotekan` package in `python/` should be updated to build with `uv`, or split into the installed version and the script version (which is used for the pytest suite).  This will allow it to be declared as a regular dependency in `uv` ~~and avoid the need for the awkward `tests/env_test` file manually setting `PYTHONPATH`.~~


.github/workflows/main.yml
```diff
@@ -146,11 +146,11 @@ jobs:
               -DWITH_TESTS=ON -DWITH_BOOST_TESTS=ON
             post_build: |
               # Run pytest tests
-              PYTHONPATH=$GITHUB_WORKSPACE/python
+              # PYTHONPATH=$GITHUB_WORKSPACE/python
```
jbmertens: Feel free to delete if this is no longer needed!

.github/workflows/main.yml
```diff
@@ -260,6 +291,17 @@ jobs:
           echo "==========="
           lscpu
 
+      - name: Install uv
```
jbmertens: I think we'd want to move this to the dockerfile, but good for proof of concept

pyproject.toml
```diff
@@ -0,0 +1,44 @@
+[project]
+name = "kotekan-pyenv"
+version = "0.1.0"
+requires-python = ">=3.12,<3.13"
+dependencies = [
```
jbmertens, line 5: Will just want to double-check this captures everything in both the docker image and sysops repos.

jbmertens (commented):
A few comments, but I like the idea!

pyproject.toml
jbmertens: (And follow up by having the sysops repo refer to this...)

PR #1332, opened 2025-11-03
Fix ccache and readthedocs builds

jbmertens:


geoffryan (approved):
CI times more than halved: LGTM!

It is hard to tell on this end whether the RTD build is successful, but this certainly doesn't make that worse and the ccache fix is much needed.

PR #1336, opened 2025-11-04
Telescope and N2 updates

geoffryan:
- Updates CHORDTelescope fields for more consistent/informative naming.
- Defines a format for dish information, including indexing, to be passed through the config and stored in `CHORDTelescope`.  
- Adds functionality to `N2Buffer` to report the order of visibility matrix elements in the buffer.
- Provides functionality for `CHORDTelescope` and `N2Buffer` to report their information to a visibility file writer.
- Improves testing of `CHORDTelescope`


config/tests/fringestop.yaml
```diff
@@ -275,22 +276,26 @@ samples_per_dataset: 32768
...
-    [151.0, 0.0, 0.0],
-    [0.0, 204.0, 0.0],
-    [151.0, 204.0, 0.0]]
   inst_lat_deg: 49.32075144444
   inst_long_deg: -119.62081125
   inst_grid_x_axis: [1.0, 0.0, 0.0]
   inst_grid_y_axis: [0.0, 1.0, 0.0]
-  inst_alt_deg: 90.0
-  inst_dish_alt_axis: [1.0, 0.0, 0.0]
+  inst_coelev_deg: 0.0
```
jbmertens: Very pedantic, but related to naming feedback... inst_* is short for instrument_*? Is that redundant with this being a sub-item of telescope?

geoffryan: Exactly the type of pedantry I was looking for.  This has been bothering me too.

- `inst_long_deg` --> `origin_lon_deg`  (`origin_longitude_deg`?)
- `inst_lat_deg` --> `origin_lat_deg` (`origin_latitude_deg`?)
- `inst_grid_x_axis` --> `grid_x_axis`  (~~`grid_ew_axis`?~~ the "x" axis is not exactly E/W, x is more correct)
- `inst_grid_y_axis` --> `grid_y_axis`  (~~`grid_ns_axis`?~~ same as above)
- `inst_dish_elev_axis` --> `dish_elev_axis`
- `inst_dish_vert_axis` --> `dish_vert_axis`
- `inst_coelev_deg` --> `dish_coelev_deg`

How about this?

jbmertens: These namings sound even better. I can also imagine nesting details?
```
telescope:
    name: CHORDTelescope
    location_grid:
        origin_long_deg:
        origin_lat_deg:
        x_axis:
        ....
    dishes:
        elev_axis:
        vert_axis:
        separation_ew_m: 6.3 # only good for grid telescopes
        separation_ns_m: 8.5
        details: [
            {idx: 0, ew_idx: 0, ns_idx: 0, pos_disp_m: [0.0, 0.0, 0.0], coelev_disp_deg: 0.0, type: 0, label: "CHORD-A01"},
            .....
```
A drawing for understanding what all of these mean would be helpful, too!

geoffryan: There aren't so many of these that I think the nesting is necessary, for now I prefer the flat namespace and self-describing names.  Certainly need more documentation in kotekan on what this all means (including a diagram or 3!)

jbmertens:
Do we also want to move the relevant chordMetadata fields to the telescope in this PR?

geoffryan:
Re: removing `chordMetadata` fields, that would be a good idea!  Might need some input from @eschnett on getting the fengine simulator running with the telescope fields.

config/tests/fringestop.yaml
jbmertens: How about at least making the dish information a subset of the telescope object, but otherwise flat?
(I think it is functionally equivalent, but at least makes it clear that the telescope object "owns" this information, and other pieces of the pipeline should probably access it through the telescope object.)

jbmertens (approved):

config/tests/fringestop.yaml
geoffryan: Sure!

geoffryan: Done.

geoffryan:
Currently the redundant `chordMetadata` fields: `ndishes`, `n_dish_locations_ew`, `n_dish_locations_ns`, `dish_index`, and `get_dish_index()` are used by some of the auto-generated cuda kernels.  They should be removed, and possibly some `CHORDTelescope` functionality added to replace them in the cuda kernels, but I believe that should be a separate PR.

config/ci-tests/run_n2accum.yaml
```diff
@@ -122,9 +103,9 @@ fake_rfimask_buffer:
     metadata_pool: main_pool
 
 n2_buffer:
-    kotekan_buffer: standard
+    kotekan_buffer: N2
+    layout: 0
```
jbmertens: Could we pass this in as a string in the config?

config/tests/fringestop.yaml
```diff
@@ -267,6 +267,7 @@ log_level: debug
 
 num_data_sets: 1
 num_elements: 14
+num_dishes: 7
```
jbmertens, line 181: Can we infer this from other dish data in the telescope object?

config/ci-tests/run_n2accum.yaml
jbmertens: And maybe add a comment with the options, #FullUpperTri or RedundantBaselineAvg

lib/stages/N2Accumulate.cpp
```diff
@@ -259,36 +255,38 @@ void N2Accumulate::main_thread() {
...
-            */
-            t_frame_s = 0; // TODO: move this logic to telescope
+        std::shared_ptr<chordMetadata> counts_metadata =
+            get_chord_metadata(in_counts_buf, in_counts_frame_id);
+        std::shared_ptr<chordMetadata> rfimask_metadata =
+            get_chord_metadata(in_rfimask_buf, in_rfimask_frame_id);
+
+        // Check synchronization
+        if (frame_metadata->get_fpga_seq_num() != counts_metadata->get_fpga_seq_num()) {
+            FATAL_ERROR("Correlation buffer {:s}[{:d}] seq={:d} has lost synchronization with "
```
jbmertens, line 269: @andrerenard - any idea if this is likely to occur? I'm thinking some edge case of a node restarting, N2 frames were sent but counts weren't... we might want to try to handle this gracefully, try to sync things back up somehow. Keep incrementing through frames in whichever buffer has the "oldest" data, until things are synced back up?
(But then, this'll be running on the GPU node anyways?)

lib/stages/N2TimeDownsample.cpp
```diff
@@ -190,6 +198,12 @@ void N2TimeDownsample::main_thread() {
             // Set the output to target EOP.
             output_frame.eop = eop_target;
 
+            // Set the output absolute time index, from the difference between the current ERA bin
+            // and the ERA bin at startup
+            output_frame.abs_time_idx =
```
jbmertens, line 205: So these won't necessarily be sequential?

lib/stages/N2Accumulate.cpp
```diff
@@ -448,12 +446,15 @@ bool N2Accumulate::output_and_reset(N2::frameID& in_frame_id, N2::frameID& out_f
         meta->num_elements = _num_elements;
         meta->num_prod = _N2_num_products;
         meta->num_ev = 0;
+        meta->layout = N2Layout::FullUpperTri;
+        meta->abs_time_idx = _accum_fpga_start_tick
+                             / (_vis_samples_in_out_frame * _n_fpga_samples_per_n2k_correlation);
```
jbmertens, line 451: This is... probably in practice... always right even when _n_fpga_samples_per_n2k_correlation varies a bit (thinking to ERA-aligned bins). But, in principle it could be wrong:
Can we add documentation noting the thought process behind this, and maybe even enforce a limit on the _n_fpga_samples_per_n2k_correlation so this remains valid? Or issue an error otherwise. (bin size >~ 4*frame size or something?)

lib/utils/CHORDTelescope.cpp
```diff
@@ -501,6 +545,38 @@ void CHORDTelescope::fringestop_phases_1d(double freq_Hz, const struct EOP& eop,
     }
 }
 
+void CHORDTelescope::get_input_maps(dishInputFields& input) const {
```
jbmertens, line 554: get, or set? (or build, or something)

lib/utils/CHORDTelescope.cpp
```diff
@@ -501,6 +545,38 @@ void CHORDTelescope::fringestop_phases_1d(double freq_Hz, const struct EOP& eop,
...
+        input.ew_idx.push_back(_dish_info_table[i].ew_idx);
+        input.ns_idx.push_back(_dish_info_table[i].ns_idx);
+        input.feed_pos_disp_m.push_back(_dish_info_table[i].feed_pos_disp_m);
+        input.coelev_disp_deg.push_back(_dish_info_table[i].coelev_disp_deg);
+        input.type.push_back(static_cast<int64_t>(_dish_info_table[i].type));
+        input.label.push_back(_dish_info_table[i].label);
+    }
+}
+
+uint64_t CHORDTelescope::get_num_stacks() const {
```
jbmertens, line 582: Maybe issue a throw here for now, unless this is actually called.

lib/utils/CHORDTelescope.cpp
```diff
@@ -536,55 +612,59 @@ struct EOP CHORDTelescope::get_EOP_at_idx(uint64_t i) const {
 struct EOP CHORDTelescope::get_EOP_at_time(const timespec& ts_target) const {
```
jbmertens: More idiomatic to drop the `struct` specifier for these declarations/definitions.

lib/utils/CHORDTelescope.hpp
```diff
@@ -18,23 +18,99 @@
...
+ * @brief Enum for denoting the type of dish input into Kotekan
  */
-struct EOP {
-    int64_t t_inst;        // Instrument time, nanoseconds, UNIX epoch.
-    int64_t t_ut1;         // UT1 time, nanoseconds, J2000(UT1) epoch.
-    double delta_UT1_inst; // Diff between UT1 and Instrument time, seconds
-    double ERA_deg;        // Earth Rotation Angle, degrees
-    double xp_as;          // Polar Motion x', in arcseconds.
-    double yp_as;          // Polar Motion y', in arcseconds.
+enum class InputType : int64_t {
```
jbmertens: DishInputType? (or dishInputType?)

lib/utils/N2FrameView.cpp
```diff
@@ -92,3 +93,17 @@ void N2FrameView::copy_data(N2FrameView frame_to_copy_from, const std::set<N2Fie
...
 }
+
+void N2FrameView::get_prod_maps(std::vector<N2::prod_ctype>& prods) {
+
+    switch (layout) {
+        case N2Layout::FullUpperTri:
+            get_prod_maps_FullUpperTri(prods, num_elements, num_prod);
+            break;
+        default:
+            throw std::runtime_error(
```
jbmertens: Maybe issue a kotekan error instead/also.

lib/utils/N2FrameView.hpp
```diff
@@ -103,14 +111,12 @@ class N2FrameView : public FrameView {
     /**
      * @brief The sizes of the fields in the N2FrameView.
      */
-    static std::vector<std::pair<N2Field, size_t>> get_field_sizes(uint32_t num_elements_in,
-                                                                   uint32_t num_ev_in) {
-
-        size_t num_prod = N2::get_num_prod(num_elements_in);
+    static std::vector<std::pair<N2Field, size_t>>
+    get_field_sizes(uint32_t num_elements_in, uint32_t num_ev_in, size_t num_prod_in) {
```
jbmertens, line 115: Adding `*prod*` back in here is looking forward to stacking, or is there another reason?

lib/utils/timeUtil.hpp
```diff
@@ -1,53 +1,114 @@
...
+a time-like variable which exactly tracks the rotation of the Earth.  To convert to/from UT1, the
+IERS distributes a value UT1-UTC which can be added to UTC to get the UT1 time. Because Kotekan's
+instrument time is not exactly UTC, we rely on a value `delta_UT1_inst` to give us the conversion
+from instrument time to UT1 directly. This value should be computed taking into account leap seconds
+and the time at which the instrument started.  Internally UT1 time is represented as nanoseconds
+since 2451545.0 JD(UT1), which we call J2000(UT1).
+- *Earth Rotation Angle* (denoted `era`, `ERA`, or ERA). The rotational phase of the Earth,
+internally stored as a double in degrees in the interval [0.0, 360).  Computed from UT1 via IERS
+Covenventions (2010, 2012 update), Chapter 5, Eq. 5.14).
+- *Earth Rotation number* (denoted `nrot`, or `num_rot`). An integer, the number of full revolutions
```
jbmertens: Sidereal revolutions or rotations, right?

jbmertens (changes requested):

config/tests/fringestop.yaml
geoffryan: tl;dr Maybe, but for now we shouldn't.

We could in principle (the length of `dish_inputs`) but I think it's better to not.  The GPU kernels require the number of dishes in the pipeline to be a multiple of 32, but especially in the run-up to Pathfinder (and for testing) we won't always have that number of real inputs.  `num_dishes` is also read by cuda kernels, and has the multiple-of-32 constraint enforced. It also must be equal to `num_elements / 2`, which is read by many kernels and stages (and must be a multiple of 64 in the GPU). For now, both `num_dishes` and `num_elements` should stay in the config.

The telescope uses `num_dishes` to know how many dishes the pipeline is configured with.  It creates a table for all the dishes, and fills the entries present in the config `dish_input` with their values, the rest are explictly marked `Fake` and given default values.

lib/stages/N2Accumulate.cpp
geoffryan: `N2Accumulate` will, I think *always* be running on a GPU node, it's the GPU -> CPU bridge in the N^2 pipeline.  So, this check should only fire if something catastrophic happens, or if the config is very badly mis-specified.  In either case I don't think we should try to recover (for instance, by churning through old frames until everything is up to date).

lib/stages/N2TimeDownsample.cpp
geoffryan: They will be sequential so long as the ERA bins are reached sequentially.  If for some reason `N2TimeDownsample`'s input buffer is already downsampled coarser than its configured `num_bins_per_rotation`, then it will skip over bins.  The indexing will still be monotonic, unless ERA starts decreasing instead of increasing.  But if that's the case, no one will care what Kotekan is doing (because of the earthquakes, tidal waves, and shattering of the Earth's crust).

lib/stages/N2Accumulate.cpp
geoffryan: ERA-aligned bins (which are not implemented in `N2Accumulate` yet) will use a different mechanism to set `abs_time_idx`, a mirror of the one in `N2TimeDownsample`. Some refactoring of this code will happen when we implement that, of course.

I think, for accumulating by frames, this is always correct. This is really just setting `abs_time_idx = `fpga_start_tick / frame_length_fpga_ticks`.  We're guaranteed from the GPU kernels to always be ingesting frames containing the same number of fpga ticks, ie. `frame_length_fpga_ticks` will be the same for all frames leaving `N2Accumulate` (in this accumulating-by-frame mode).

lib/utils/CHORDTelescope.cpp
geoffryan: I think `get` is correct here, the caller is retrieving information from the Telescope.  They're just doing it via a reference instead of a return value.

lib/utils/CHORDTelescope.cpp
geoffryan: done.

lib/utils/CHORDTelescope.cpp
geoffryan: My C is showing :p

geoffryan: (done)

lib/utils/N2FrameView.hpp
geoffryan: No other reason, just plumbing in some code paths that stacking can use.  Without the `datasetManager`, the N2Frames themselves need to know what layout they're in (so they know how to set `num_prod` correctly).

lib/utils/timeUtil.hpp
geoffryan: Yes, clarified a bit.  The second sentence is really the technical definition, this keeps track of how may times ERA has rolled over from 360 -> 0.  It's the integer part of the UT1 -> ERA calculation.

lib/utils/N2FrameView.cpp
geoffryan: FrameViews don't inherit from `kotekanLogging`, and other code paths here throw exceptions, so I think keeping the throw is ok.  Adding a kotekan error with the same message to this (and the other throw in the header).

config/ci-tests/run_n2accum.yaml
geoffryan: Yep. Will work on this.

lib/stages/N2Accumulate.cpp
jbmertens: Is this true for HIRAX, outriggers, everything else that might urn kotekan? It would make sense, but I'm not sure it's guaranteed.
So this is probably "good enough", but want confirmation from @andrerenard .

lib/stages/N2TimeDownsample.cpp
jbmertens: How about adding a check to make sure downsampling is actually... downsampling then.
But I'm thinking more along the lines of when the frame and bin sizes are comparable (see the comment below).

lib/stages/N2Accumulate.cpp
jbmertens: Agreed, for accumulating a fixed number of frames, this is good.
When not, this might run into issues if the bin and frame sizes are comparable.

jbmertens (approved):
Looks good - lets merge this in, just a few very minor to-do items for the future.

lib/stages/N2Accumulate.cpp
geoffryan: `N2Accumulate` currently only accumulates via a fixed number of frames.  When we upgrade it for ERA bins (soon!) we will also compute an associated `abs_time_index` which is appropriate for that binning.  This specific code will only be used in fixed-number-of-frames accumulation.

lib/stages/N2TimeDownsample.cpp
geoffryan: Check added!

lib/utils/CHORDTelescope.hpp
geoffryan: Refactored to DishType

lib/stages/N2Accumulate.cpp
jbmertens: From Andre, to keep in mind for future changes--
"CHIME did have a concept of dropping vis data before visAccumulate for dropping entire RFI contaminated frames. We could choose to do that, in which case we have to be careful about the problem you mention, or we could have the tresholding done in N2Accum. In general it's not safe to assume buffers from different sources (particularly network ones) will be in sync. However in this instance, they should be in sync."

PR #1337, opened 2025-11-04
Add chime_only_baseband config

ketiltrout:
I found this on csBfs.  Last edited 17 March 2021.

Do we want to keep it?


andrerenard (approved):
I don't remember why we used this, we could keep it for posterity, but I don't see us using it again before we upgrade, at which point it would likely need to be updated.

ketiltrout:
Maybe it's best to just leave it off, then?

PR #1338, opened 2025-11-05
kotekan: add code to dump a backtrace on error

rhaas80:


rhaas80:
@eschnett says (https://chime-experiment.slack.com/archives/CBXE5SMNK/p1762356408204959) ok to merge.

lib/core/backtrace.cpp
```diff
@@ -0,0 +1,114 @@
...
+    kill(pid, signum);
+}
+
+void request_backtraces() {
+    signal(SIGQUIT, signal_handler);
+    signal(SIGILL, signal_handler);
+    signal(SIGABRT, signal_handler);
+    signal(SIGFPE, signal_handler);
+    signal(SIGBUS, signal_handler);
+    signal(SIGSEGV, signal_handler);
```
eschnett: Julia uses signals for its parallel garbage collector. I know about `SIGSEGV` – there may be others. There needs to be a way to disable setting up the signal handlers, or at least this signal handler.

eschnett (dismissed):

lib/core/backtrace.cpp
rhaas80: SEGV for garbage collection? Hmm, that will be tough. Trapping SEGV is one of the more useful things to do to see exactly where one has the bad memory access. 

https://cnuernber.github.io/libjulia-clj/signals.html would seem to indicate that SIGINT is used. Note that kotekan's main traps that one:

https://github.com/kotekan/kotekan/blob/95c67743b83317b75252c0768b571638057cbae1/kotekan/kotekan.cpp#L429

So there's code that calls out to Julia in kotekan (the F-Engine simulator I guess but are there others)? 

I can certainly try and add a (global) config option to select which signals to trap.

rhaas80: functionality to select which signals to trap has been added in b1dd80184bd159fcbd9d1a799ce0009bd8dfba27 , it default to no traps

jbmertens: This seems like a good solution to me, maybe we can discuss briefly on today's PI call, but I'm approving to speed things up.

jbmertens (approved):

PR #1339, opened 2025-11-06
CI workflow updates

jbmertens:
- Updated OpenCL build name
 - modified release build configurations for gcc and clang
 - Removed some redundant build steps, moved more to github workers rather than self-hosted
 - added architecture verification / compile flags (hopefully helps ccache)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -171,7 +171,7 @@ jobs:
               bash $GITHUB_WORKSPACE/tests/boost/run_boost_tests.sh -v -t "${TEST_TIMEOUT}" $GITHUB_WORKSPACE/build-2404/tests/boost
 
           - id: opencl
-            name: "OpenCL build"
+            name: "OpenCL-enabled build"
             ccache_key: kotekan-opencl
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -326,79 +338,28 @@ jobs:
...
-              -DUSE_JULIA=ON
-              -DUSE_DPDK=ON
-              -DWITH_TESTS=ON
-              -DWITH_BOOST_TESTS=OFF
-            post_build: ""
           
-          - id: release_gcc_yaml_tests
+          - id: release_build_yaml_tests
             name: "Build release kotekan"
             ccache_key: selfhosted-kotekan-release-gcc  # re-using key from build
```
geoffryan, line 346: This comment is old and doesn't apply any more (and was likely wrong when I wrote it)

geoffryan (approved):
LGTM once checks pass.

.github/workflows/main.yml
jbmertens: Will update in a future PR!

PR #1341, opened 2025-11-08
This commit appears large, but the bulk of it is automated whitespace linting with no functional changes. The actual modifications are small and focused:

djgormley:
- Improved CMake output formatting for better readability
- Added Python3 find_package helper to resolve build configuration issues
- Fixed Julia package lookup conflict
- Added directory support for Ubuntu 24.04 compatibility

No functionality was removed or altered—these changes only improve build system robustness and code formatting consistency.


jbmertens:
Thanks for the PR! I'm looking here to ignore most whitespace changes:
https://github.com/kotekan/kotekan/pull/1341/files?w=1

The functional changes (julia, find packages, etc) look mostly good. Could you submit a separate PR with those so it's easier to review?

For the formatting changes, the codebase is set up to use clang-format-18, which might revert some of this. For other formatting, if there's something we want to enforce, let's consider changes to linter settings (for clang-format, black, cmake-lint). Yamls could also use a linter? But I think we'll generally want to allow anything the linters allow so the formatting doesn't change regularly.

andrerenard (commented):
I think your linter might be setup incorrectly.   The CI should be tracking code formatting, so you shouldn't have needed to make this many changes.  Ideally just remove those changes so it's easier for us to see your changes.  Your code will pass CI as long as your new code has been linted.

djgormley:
Sounds good I'll submit separately.

PR #1342, opened 2025-11-08
build/docs: synchronize Sphinx and Doxygen, fix Julia and CMake warnings

djgormley:
### Summary
- **Docs:** Add `add_dependencies(sphinx doxygen)` to ensure Sphinx runs after Doxygen, fixing intermittent “Cannot find file: index.xml” build failures.
- **Build:** Remove premature `find_package(Julia)` call that caused false "Julia not found" warnings.
- **CMake:** Set policy `CMP0074` to `NEW` to silence `find_package(CUDAToolkit)` warnings.
- **Plugins:** Update HDF5 plugin search paths for Ubuntu 24.04 (`libH5Zblosc.so`).
- **Docs (Doxygen):** Exclude build and template directories from parsing.
- **Misc:** Formatting fixes in LAPACKE and Summary modules, disable placeholder test.

_No functional runtime changes; affects documentation and configuration only._


jbmertens:
Thanks, this mostly looks good!

For julia, we do want to make sure we're finding julia if it's available/requested as on. Were you getting build errors?

djgormley:
> Thanks, this mostly looks good!
> 
> For julia, we do want to make sure we're finding julia if it's available/requested as on. Were you getting build errors?

Yes.

This block in lib/CMakeLists.txt was running `find_package(Julia)` prematurely, before the main `FeatureJulia.cmake` script had a chance to run.

This caused a race condition where the `lib` check would fail, printing a "Julia not found" warning, even when the main script would find it successfully moments later.

Removing this redundant check makes `lib` respect the top-level detection logic and cleans up the contradictory build log.

jbmertens:
For julia - thanks, let's remove the comment as well.

It looks like a .venv is in this PR, plus an `apply_iwyu_fixes.py` binary that may have been committed by accident. Let's add .venv to .gitignore.

For formatting/whitespace, we'll generally want to stick to the existing linters. The formatting they prefer isn't always the prettiest, but at least keep things consistent, and is usually "good enough". Major changes will make it hard to see code history.

djgormley:
> For julia - thanks, let's remove the comment as well.
> 
> It looks like a .venv is in this PR, plus an `apply_iwyu_fixes.py` binary that may have been committed by accident. Let's add .venv to .gitignore.
> 
> For formatting/whitespace, we'll generally want to stick to the existing linters. The formatting they prefer isn't always the prettiest, but at least keep things consistent, and is usually "good enough". Major changes will make it hard to see code history.

Thanks.  Bare with me for a minute while I figure out how to clean this up.

djgormley:
Alight.  I think it's all tidied up.

lib/CMakeLists.txt
```diff
@@ -4,16 +4,10 @@ project(kotekan_libs)
...
-if("${USE_JULIA}" STREQUAL "AUTO" OR "${USE_JULIA}" STREQUAL "ON")
-    find_package(Julia QUIET)
-    if(Julia_FOUND)
-        message(STATUS "Julia support enabled")
-    else()
-        message(WARNING "${ColorYellow}Julia not found; Julia features disabled in"
-                        " kotekan_lib${ColorReset}")
-    endif()
-endif()
+# Julia is a convenient language to implement stages for testing and development if("${USE_JULIA}"
```
jbmertens: Let's remove this comment.

jbmertens (changes requested):
LGTM, let's just remove the commented code.

lib/CMakeLists.txt
djgormley: Done.  I ran linter.sh and it fixed up the whitespacing in a few files as well.  Thanks.

PR #1343, opened 2025-11-09
chore: apply IWYU include cleanup and formatting updates

djgormley:
Ran include-what-you-use across core, CUDA, metadata, and stage modules.
Removed unused headers, added missing includes, and fixed forward declarations.
No functional changes.


djgormley:
Bare with me this is a WIP.

PR #1346, opened 2025-11-10
chore(iwyu): Apply header reordering and comment updates

djgormley:
Enforced IWYU policies by running the linter with reordering and comment update flags enabled.

### DIFF
--- a/lib/core/buffer.cpp
+++ b/lib/core/buffer.cpp
@@ -1,32 +1,33 @@
 #include "buffer.hpp"
 
-#include <assert.h>            // for assert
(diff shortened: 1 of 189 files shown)


djgormley:
This is now only IWYU header formatting changes.

jbmertens:
Thanks for these changes -- we had decided (informally) to run the iwyu linter only at release time, since it takes a while to run so is excessive for every PR.
We can merge this in now, but should probably add documentation to that effect, perhaps in the lint script.

jbmertens (approved):

PR #1348, opened 2025-11-11
fixes here and there

rhaas80:
Mostly harmless, except the change to `testDataGen` that will make up an `array_shape` given the frame size and type size if no `array_shape` is given. This removes one code path in the `onehot` test data. 

If that needs to be retained then one needs extra options.

### DIFF
--- a/lib/utils/Telescope.cpp
+++ b/lib/utils/Telescope.cpp
@@ -11,11 +11,11 @@ Telescope::Telescope(const std::string& log_level) {
 }
 
 const Telescope& Telescope::instance() {
-    if (tel_instance == nullptr) {
+    if (!tel_instance()) {
         FATAL_ERROR_NON_OO("Telescope singleton must be configured before use.");
     }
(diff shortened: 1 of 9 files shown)


lib/utils/Telescope.cpp
```diff
@@ -25,11 +25,18 @@ const Telescope& Telescope::instance(const kotekan::Config& config) {
...
     }
 
-    tel_instance = FACTORY(Telescope)::create_unique(telescope_name, config, "/telescope");
+    tel_instance() = FACTORY(Telescope)::create_unique(telescope_name, config, "/telescope");
 
-    return *tel_instance;
+    return *tel_instance();
 }
 
+std::unique_ptr<Telescope>& Telescope::tel_instance() {
```
jbmertens, line 33: @geoffryan noticed this yesterday as well!
This complements #1349.

jbmertens (approved):

PR #1349, opened 2025-11-11
Add reset() to Telescope to clean up REST internals

geoffryan:
Fixes a race condition on `kotekan` exit that can cause the program to hang.  The Telescope (CHORDTelescope in particular) was calling `restServer::remove_json_callback()` in its destructor, potentially after the `restServer` itself had been destructed.  This PR adds a `reset()` function to the Telescope analogous to `configUpdater::reset()` which handles the REST cleanup before the destructor is called.  `Telescope::reset()` is called in the `kotekanMode` destructor which is guaranteed to be called before the `restServer` destructor.

### DIFF
--- a/lib/core/kotekanMode.cpp
+++ b/lib/core/kotekanMode.cpp
@@ -47,6 +47,7 @@ kotekanMode::kotekanMode(Config& config_) : config(config_) {
 kotekanMode::~kotekanMode() {
 
     configUpdater::instance().reset();
+    Telescope::instance().reset();
     restServer::instance().remove_get_callback("/config");
     restServer::instance().remove_get_callback("/buffers");
     restServer::instance().remove_get_callback("/pipeline_dot");
--- a/lib/testing/fakeGpu.hpp
+++ b/lib/testing/fakeGpu.hpp
@@ -101,6 +101,8 @@ class FakeTelescope : public Telescope {
 public:
     FakeTelescope(const kotekan::Config& config, const std::string& path);
 
+    void reset() const override {}
+
     // Dummy freq map implementations
     freq_id_t to_freq_id(stream_t stream_id, uint32_t ind) const override;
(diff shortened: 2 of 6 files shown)


jbmertens (approved):

lib/core/kotekanMode.cpp
```diff
@@ -47,6 +47,7 @@ kotekanMode::kotekanMode(Config& config_) : config(config_) {
 kotekanMode::~kotekanMode() {
 
     configUpdater::instance().reset();
+    Telescope::instance().reset();
```
rhaas80, line 50: since it was not kotekanMode that caused Telescope to register itself with resetServer, it should also not cause deregistration I'd say.

rhaas80:
This seems to be no longer required with 2237334 I'd probably close the pull request so as to not add code that is not used anymore.

lib/core/kotekanMode.cpp
geoffryan: `Telescope` is constructed via the `Telescope::instance(config)` call in `kotekanMode::initialize_stages()`, which does cause `Telescope` to register with the restServer.  I think this makes `kotekanMode` responsible for Telescope and a reasonable place to call its clean up.  This mirrors the behaviour of `configUpdater`, which is also managed by `kotekanMode`.

geoffryan:
Would still be used (called), but I agree this seems no longer necessary. Closing!

jbmertens:
~Calling~
~`    restServer::instance().remove_json_callback(_unique_name + "/time0_ns");`~
~May still be a good idea?~
Never mind, I see it is still there.

jbmertens:
Sorry, re-opening again! There is still an issue with the remove callback,
`remove_json_callback` should be `remove_get_callback`.

geoffryan:
Closing this PR. Will address the `remove_json` -> `remove_get` in a different quick PR.

PR #1351, opened 2025-11-13
allocate framedesc race

rhaas80:
This fixes a race condition when allocating a frame description. 

Other commits are mostly cleanup.

### DIFF
--- a/lib/core/buffer.cpp
+++ b/lib/core/buffer.cpp
@@ -484,6 +484,7 @@ void Buffer::mark_frame_full(const std::string& producer_name, const int ID) {
                       buffer_name, ID);
                 is_full[ID] = false;
                 metadata[ID].reset();
+                frames_desc.at(ID).reset();
                 set_empty = true;
                 private_reset_consumers(ID);
             }
@@ -581,8 +582,10 @@ bool Buffer::private_mark_frame_empty(const int ID) {
         private_reset_consumers(ID);
         broadcast = true;
     }
-    if (metadata[ID])
+    if (metadata[ID]) {
         metadata[ID].reset();
+        frames_desc.at(ID).reset();
+    }
     return broadcast;
 }
 
--- a/lib/core/buffer.cpp
+++ b/lib/core/buffer.cpp
@@ -208,7 +208,7 @@ Buffer::Buffer(int num_frames, size_t len, std::shared_ptr<metadataPool> pool,
                bool zero_new_frames) :
     GenericBuffer(_buffer_name, _buffer_type, pool, num_frames), frame_size(len),
     // By default don't zero buffers at the end of their use.
-    _zero_frames(false), frames(num_frames, nullptr), frames_desc(num_frames, nullptr),
+    _zero_frames(false), frames(num_frames, nullptr), frames_desc(nullptr),
     is_full(num_frames, false), last_arrival_time(0), use_hugepages(_use_hugepages),
     mlock_frames(_mlock_frames), numa_node(_numa_node) {
     assert(num_frames > 0);
@@ -482,7 +482,6 @@ void Buffer::mark_frame_full(const std::string& producer_name, const int ID) {
                       buffer_name, ID);
                 is_full[ID] = false;
                 metadata[ID].reset();
-                frames_desc.at(ID).reset();
                 set_empty = true;
                 private_reset_consumers(ID);
             }
@@ -582,7 +581,6 @@ bool Buffer::private_mark_frame_empty(const int ID) {
     }
     if (metadata[ID]) {
         metadata[ID].reset();
-        frames_desc.at(ID).reset();
     }
     return broadcast;
 }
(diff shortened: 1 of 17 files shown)


jbmertens (approved):
I can't comment on [iceBoardShuffle.hpp](https://github.com/kotekan/kotekan/pull/1351/files#diff-ee4cc2e38ce1de927b6a3debfff9827582433605819f0d24c8c732e962ca3c9f), but changes look good to me otherwise.

andrerenard (commented):
I should have asked on the original PR, but do we actually need to be setting the frame description (NDArray layout) with every frame?   I'd imagined this as more of a static property of the buffer, since I don't think we ever dynamically change what's stored by a buffer?   I think setting it at runtime has some good logic too it (although I think it could also be set in the config), but we likely only need to do it once.  Unless you have a counter example?

rhaas80:
@jbmertens asked the same question. As far as I recall I was told (by @andrerenard was my recollection, but he seems to disagree :-) ) that indeed per frame data was requested. 

Per buffer would be much easier to implement. At least as long as this is not pushed all the way to the yaml config file

andrerenard:
> @jbmertens asked the same question. As far as I recall I was told (by @andrerenard was my recollection, but he seems to disagree :-) ) that indeed per frame data was requested.
> 
> Per buffer would be much easier to implement. At least as long as this is not pushed all the way to the yaml config file

Yeah I don't think we've introduce any code where the dimensions change.  Although I think we may have some code where the dimensions only get finalized at runtime (with some of the upchannalization code).  So leaving it as a runtime set parameter I think makes sense.  Let's bring it up at the call today, just in case someone else is hoping to use that functionality.   Otherwise, I think it's a lot of overhead to set this every time for something that should be static after startup.

geoffryan:
Just adding a vote here for these being once-per-buffer objects!  It would simplify a lot of code to be able to just look this up from the buffer, instead of building it independently from the config and then checking that the buffer matches.

lib/core/buffer.hpp
```diff
@@ -526,43 +528,58 @@ class Buffer : public GenericBuffer {
...
                                  const std::array<std::ptrdiff_t, D>& extents,
                                  const std::array<kotekan::Symbol, D>& dimnames) {
-        frames_desc.at(frame_id) =
-            std::make_shared<kotekan::NDArray<T, D>>(quantity_name, extents, dimnames, nullptr);
+        buffer_lock lock(mutex);
+        if (!frames_desc)
+            frames_desc =
+                std::make_shared<kotekan::NDArray<T, D>>(quantity_name, extents, dimnames, nullptr);
+        else {
+            assert(D == frames_desc->get_rank());
```
jbmertens: Consider issuing kotekan errors; what behavior do we want in production here?

lib/core/buffer.cpp
```diff
@@ -210,7 +208,7 @@ Buffer::Buffer(int num_frames, size_t len, std::shared_ptr<metadataPool> pool,
                bool zero_new_frames) :
     GenericBuffer(_buffer_name, _buffer_type, pool, num_frames), frame_size(len),
     // By default don't zero buffers at the end of their use.
-    _zero_frames(false), frames(num_frames, nullptr), frames_desc(num_frames, nullptr),
+    _zero_frames(false), frames(num_frames, nullptr), frames_desc(nullptr),
```
jbmertens, line 211: If frames_desc is tailored to frames containing NDArrays, can we ~make this a part of an NDArray buffer type, rather than Buffer~? Integrate this initialization into the buffer factory? Template the GenericNDArray type in the `frames_desc` 😬? Trying to think what is best here. I'm wondering if we can use this to hold structural information for other buffer types, since not everything is an NDArray.

rhaas80: buffers get created before the stages that use them, so the information would then have to be present in the yaml files for kotekan to use it. This was discussed and rejected during the last software call, so this should address the question here. 

Everything in buffers can be an NDArray. Might a one-dim one with type `uint8`, but that is still an NDArray. :-)

Right now the frames are:

```
    /// The array of frames (the actual data we are carrying)
    std::vector<uint8_t*> frames;
```

pointing to the pool of frames. In an actual `NDArray` (with data, which is not what we want, since it would be one `NDArray` per frame again) this would be the 

```
    // Pointer to the array data
    T* m_data;
```

member for `T = uint8_t`.

lib/core/buffer.hpp
rhaas80: I could output an `ERROR()` message sure. I guess with generic producers like `testDataGen` there may also be generic consumers that would both register themselves.

lib/core/buffer.cpp
jbmertens: > Everything in buffers can be an NDArray. Might a one-dim one with type uint8, but that is still an NDArray. :-)

Well, sure, but then the GenericNDArray fields are of limited use, i.e. further structural data will remain in metadata for vis/N2/hfb frame(view)s. Future task for handling that.

lib/core/buffer.hpp
rhaas80: done in 14298940a

PR #1352, opened 2025-11-13
style(yaml): whitespace & duplicate cleanup

djgormley:


jbmertens:
.yaml formatting improvements are great!
But if there is a standard we want to enforce, we should add a linter, rather than periodic corrections by hand. Maybe we can add this, both as a dependency, and as part of the lint.sh script?
https://yamllint.readthedocs.io/en/stable/

djgormley:
> .yaml formatting improvements are great! But if there is a standard we want to enforce, we should add a linter, rather than periodic corrections by hand. Maybe we can add this, both as a dependency, and as part of the lint.sh script? https://yamllint.readthedocs.io/en/stable/

To be honest I was just doing some of this janitor work to get used to the system while I build my RFI kernel.  I agree that we should have a linter for all of the other file types, but I have my own fish to fry.

PR #1353, opened 2025-11-13
Style: Python formatting

djgormley:
Formatting that was originally included in a now closed PR.
Performed via `lint.sh`.


jbmertens:
The lint.sh currently marks this as incorrect. If a version (i.e. of `black`) has changed, that change should be reflected in the dockerfile.

djgormley:
> The lint.sh currently marks this as incorrect. If a version (i.e. of `black`) has changed, that change should be reflected in the dockerfile.

Ugh, I was trying to break the review up into organized segments.  This all works really well.

jbmertens:
> > The lint.sh currently marks this as incorrect. If a version (i.e. of `black`) has changed, that change should be reflected in the dockerfile.
> 
> Ugh, I was trying to break the review up into organized segments. This all works really well.

There's probably a good way to do this, but it'll need to be sequential, since the changes between PRs are interdependent. Something like follows?
1) Initial PR with updated black in docker, plus any python changes that result.
2) After (1) is merged, update documentation settings and style, in a PR that builds.
3) After (2) is merged, a new PR with the updated clang-format / lint script, plus resulting changes.
4) After (3) is merged, any final changes by-hand (yaml?)
For each step, I'd suggest making sure the lint script and builds run correctly on the code locally, before issuing a PR.

PR #1354, opened 2025-11-13
Style: Cpp formatting

djgormley:
Additional formatting courtesy of `lint.sh` + `iwyu'.

### DIFF
--- a/kotekan/kotekan.cpp
+++ b/kotekan/kotekan.cpp
@@ -2,7 +2,7 @@
 #include "StageFactory.hpp"       // for StageFactoryRegistry, StageMaker
 #include "basebandApiManager.hpp" // for basebandApiManager
 #include "errors.h"               // for ReturnCode, get_error_message, __enable_syslog, exit_k...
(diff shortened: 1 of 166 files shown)


jbmertens:
As mentioned in the other PR, let's try to avoid linting with IWYU except at release time.

jbmertens:
The current c++ style linter appears to disagree with some of this, and it seems to cause build errors.
If there is a version (or other) change of something, that should be included in this PR, and we should make sure builds don't have errors in the new version.

djgormley:
> The current c++ style linter appears to disagree with some of this, and it seems to cause build errors. If there is a version (or other) change of something, that should be included in this PR, and we should make sure builds don't have errors in the new version.

> As mentioned in the other PR, let's try to avoid linting with IWYU except at release time.

It works with the new linter which uses more strict compiler settings.

jbmertens:
Some of these errors look like genuine syntax errors, so the new linter may be doing something incorrectly.
We'll also want to avoid touching autogenerated code.

PR #1345, opened 2025-11-09
fix(cmake): Improve build dependencies and package discovery logic

djgormley:
* **Documentation Build:**
    * Added `add_dependencies(sphinx doxygen)` to ensure Doxygen completes before Sphinx starts. This fixes intermittent build failures related to "Cannot find file: index.xml".

* **Package Discovery:**
    * Set CMake policy `CMP0074` to `NEW` to silence warnings from `find_package(CUDAToolkit)`.
    * Updated HDF5 plugin search paths to correctly locate `libH5Zblosc.so` on Ubuntu 24.04.
    * Removed a premature `find_package(Julia)` call that was causing false "Julia not found" warnings.

* **Console Output:**
    * Corrected message formatting for the LAPACKE discovery and the final build configuration summary.

PR #1357, opened 2025-11-14
CI: cpu/gpu docker targets; reusable workflows.

jbmertens:
Github seems to have modified their runners so they now use a few more GB, leaving less free for us. This now causes our jobs to run out of memory, mainly due to the size of gpu tools.
To fix this, the dockerfile is split into cpu-only and incl. gpu targets. We can run CPU-only jobs with much smaller images for cpu (pytest, boost test) tests on github.

Changes include:
 - Add multiple targets for the dockerfile: cpu and gpu (gpu=last=default)
 - Re-factor CI builds into reusable workflows
 - Use ubuntu packages for blosc rather than src
 - Modify some jobs and code to allow cpu-only compiles/runs
 - Clean up space on github Ubuntu runners before building docker images
 - Corrections in cmake summary, airspy is auto by default
 - cmake extra julia check cleanup (#1345)
 - Docker should now build on diff with latest commits (or base-head diff), not always with base-head diff
 - Split up some jobs for parallel (faster) runs
 - monitor buffer and send/recv yaml test update
 - Only test a single frame in some gpu tests (faster local runs), less verbose output


.github/workflows/build_push_docker.yml
geoffryan, line 1: Could this file get a plain text comment at the top documenting its purpose, who calls it, major necessary (or optional) parameters to set when doing so?

.github/workflows/iwyu.yaml
geoffryan, line 1: Could use a comment on the top explaining when its run, what it does, possibly if its currently part of the automatic workflow, etc.

.github/workflows/manual_docker.yaml
geoffryan, line 1: This file could use a top level comment with when/how/if this action is triggered, by who, what parameters it may need, etc.

.github/workflows/scheduled_tasks.yml
geoffryan, line 1: I am not sure what.. this does exactly? Other than say that it ran.  As with the others, a doc comment at the top saying its purpose, if its currently running or a placeholder for future actions, etc.

.github/workflows/test_kotekan_build.yml
```diff
@@ -0,0 +1,101 @@
...
+        type: string
+      cmake_defs:
+        required: true
+        type: string
+      post_build:
+        required: true
+        type: string
+
+jobs:
+  build:
```
geoffryan: I think this could use a name other than "build", since it also runs the tests, etc. Maybe "run" or "test" or "ci-test"?

.github/workflows/test_kotekan_build.yml
```diff
@@ -0,0 +1,101 @@
...
+        if: inputs.cmake_defs
+        run: |
+          set -euo pipefail
+          mkdir $GITHUB_WORKSPACE/build-2404 && cd $GITHUB_WORKSPACE/build-2404
+          ccache -s
+          cmake -DARCH=haswell -Wdev -Werror=dev -Wdeprecated -Werror=deprecated ${{ inputs.cmake_defs }} ..
+          make -j $(nproc)
+          ccache -s
+
+      - name: Post-build commands
```
geoffryan: Just a quibble.  Could we call this something like "Run Commands" instead of "Post-build Commands"?

.github/workflows/test_kotekan_build.yml
geoffryan, line 1: As with the others, an explanatory comment documenting how other actions interact with this would be valuable.

.github/workflows/main.yml
geoffryan, line 1: Could use a top-level documentation comment.

.github/workflows/main.yml
```diff
@@ -11,326 +11,221 @@ on:
...
+            cmake_defs: >-
               -DCMAKE_BUILD_TYPE=Test -DWERROR=ON
               -DCMAKE_LINK_WHAT_YOU_USE=ON -DCCACHE=ON
               -DUSE_OMP=ON -DUSE_NUMA=ON -DNO_MEMLOCK=ON
-              -DUSE_CUDA=ON -DUSE_HIP=OFF -DUSE_OPENCL=OFF
+              -DUSE_CUDA=OFF -DUSE_HIP=OFF -DUSE_OPENCL=OFF
               -DUSE_ASDF=ON -DUSE_HDF5=ON -DUSE_GDAL=ON
               -DUSE_AIRSPY=ON -DUSE_LAPACK_BLAZE=ON -DUSE_FFTW=ON -DUSE_JULIA=ON -DUSE_DPDK=OFF
-              -DWITH_TESTS=ON -DWITH_BOOST_TESTS=ON
+              -DWITH_TESTS=ON -DWITH_BOOST_TESTS=OFF
```
geoffryan, line 64: This uses the same `ccache_key` as `test_gcc_basic` but uses different `cmake_defs`, is that ok?

.github/workflows/main.yml
```diff
@@ -11,326 +11,221 @@ on:
...
             post_build: |
-              # Run pytest tests
               PYTHONPATH=$GITHUB_WORKSPACE/python
```
geoffryan, line 66: I'm not sure this line does anything? To keep the change with later lines it needs an `export`. Regardless, it should be unnecessary, the `Dockerfile` already sets `PYTHONPATH` to the correct directory.

.github/workflows/main.yml
```diff
@@ -11,326 +11,221 @@ on:
...
+            cmake_defs: >-
+              -DCMAKE_BUILD_TYPE=Test -DWERROR=ON
+              -DCMAKE_LINK_WHAT_YOU_USE=ON -DCCACHE=ON
+              -DUSE_OMP=ON -DUSE_NUMA=ON -DNO_MEMLOCK=ON
+              -DUSE_CUDA=OFF -DUSE_HIP=OFF -DUSE_OPENCL=OFF
+              -DUSE_ASDF=ON -DUSE_HDF5=ON -DUSE_GDAL=ON
+              -DUSE_AIRSPY=ON -DUSE_LAPACK_BLAZE=ON -DUSE_FFTW=ON -DUSE_JULIA=ON -DUSE_DPDK=OFF
+              -DWITH_TESTS=ON -DWITH_BOOST_TESTS=OFF
+            post_build: |
+              PYTHONPATH=$GITHUB_WORKSPACE/python
```
geoffryan, line 83: Again, pythonpath set doesn't stick here.

.github/workflows/main.yml
```diff
@@ -11,326 +11,221 @@ on:
...
-              -DWITH_TESTS=ON -DWITH_BOOST_TESTS=OFF
+              -DWITH_TESTS=ON -DWITH_BOOST_TESTS=ON
             post_build: ""   # nothing to run
 
-          - id: full
-            name: "Full build + pytests tests"
-            ccache_key: kotekan-full-pytest
-            cmake_defs: &full_build_cmake_opts >-
+          # Test gcc build + serial pytests
+          - id: test_gcc_pytests
```
geoffryan: This `id` should include "serial"

.github/workflows/main.yml
```diff
@@ -11,326 +11,221 @@ on:
...
 
-          - id: boost
-            name: "Full build + boost tests"
-            ccache_key: kotekan-full-boost
-            runner: ubuntu-latest
-            container_image: ""
-            container_options: ""
-            cmake_defs: *full_build_cmake_opts
+          # Test gcc build + parallel pytests
+          - id: test_gcc_pytests
```
geoffryan: This `id` should include "parallel"

.github/workflows/main.yml
```diff
@@ -11,326 +11,221 @@ on:
...
+            ccache_key: kotekan-test-gcc-cpu
```
geoffryan, line 73: again, re-used ccache_key, but possibly different cmake defs?

.github/workflows/main.yml
```diff
@@ -11,326 +11,221 @@ on:
...
+              -DWITH_TESTS=ON -DWITH_BOOST_TESTS=ON
             post_build: |
               TEST_TIMEOUT=30
               bash $GITHUB_WORKSPACE/tests/boost/run_boost_tests.sh -v -t "${TEST_TIMEOUT}" $GITHUB_WORKSPACE/build-2404/tests/boost
 
-          - id: opencl
-            name: "OpenCL-enabled build"
-            ccache_key: kotekan-opencl
+          # Release gcc build + serial pytests
+          - id: release_gcc_pytests
```
geoffryan: `id` includes "serial"?

.github/workflows/main.yml
```diff
@@ -11,326 +11,221 @@ on:
...
-            ccache_key: kotekan-release-gcc
-            cmake_defs: &build_release_gcc_opts >-
+            post_build: |
+              PYTHONPATH=$GITHUB_WORKSPACE/python
+              PYTEST_TIMEOUT=120
+              cd $GITHUB_WORKSPACE/tests
+              pytest -v -x --timeout="${PYTEST_TIMEOUT}" -m serial
+
+          # Release gcc build + parallel pytests
+          - id: release_gcc_pytests
```
geoffryan: `id` should include parallel

.github/workflows/main.yml
```diff
@@ -11,326 +11,221 @@ on:
...
               -DUSE_OMP=ON -DUSE_NUMA=ON -DNO_MEMLOCK=ON
-              -DUSE_CUDA=ON -DUSE_HIP=OFF -DUSE_OPENCL=OFF
+              -DUSE_CUDA=OFF -DUSE_HIP=OFF -DUSE_OPENCL=OFF
               -DUSE_ASDF=ON -DUSE_HDF5=ON -DUSE_GDAL=ON
-              -DUSE_AIRSPY=ON -DUSE_LAPACK_BLAZE=ON -DUSE_FFTW=ON -DUSE_JULIA=ON -DUSE_DPDK=ON
+              -DUSE_AIRSPY=ON -DUSE_LAPACK_BLAZE=ON -DUSE_FFTW=ON -DUSE_JULIA=ON -DUSE_DPDK=OFF
               -DWITH_TESTS=ON -DWITH_BOOST_TESTS=OFF
             post_build: |
-              # Run pytest tests
               PYTHONPATH=$GITHUB_WORKSPACE/python
```
geoffryan, line 132: needs an `export`, or trust the Dockerfile.

.github/workflows/main.yml
```diff
@@ -11,326 +11,221 @@ on:
...
+              -DUSE_AIRSPY=ON -DUSE_LAPACK_BLAZE=ON -DUSE_FFTW=ON -DUSE_JULIA=ON -DUSE_DPDK=OFF
               -DWITH_TESTS=ON -DWITH_BOOST_TESTS=OFF
-            post_build: ""   # Any OpenCL-specific commands to test?
-          
-          - id: release_pytest_gcc
-            name: "Release build + pytest"
-            ccache_key: kotekan-release-gcc
-            cmake_defs: &build_release_gcc_opts >-
+            post_build: |
+              PYTHONPATH=$GITHUB_WORKSPACE/python
```
geoffryan, line 115: needs `export`, or just trust the dockerfile.

.github/workflows/main.yml
```diff
@@ -11,326 +11,221 @@ on:
...
+              PYTEST_TIMEOUT=120
+              cd $GITHUB_WORKSPACE/tests
+              pytest -v -x --timeout="${PYTEST_TIMEOUT}" -m serial
+
+          # Release gcc build + parallel pytests
+          - id: release_gcc_pytests
+            ccache_key: kotekan-release-gcc-cpu
```
geoffryan, line 122: Re-using ccache_key, possibly with different cmake defs?

config/ci-tests/README.md
```diff
@@ -25,5 +25,5 @@ Failed: 1
 Timed out: 1
 
 Failed tests:
-  - config/ci-tests/test_broken.yaml (exit code: 1)
+  - config/ci-tests/batch/test_broken.yaml (exit code: 1)
```
geoffryan: "batch" -> "cpu_batch"?

config/ci-tests/gpu_batch/verify_cuda_n2k.yaml
```diff
@@ -272,13 +269,13 @@ check_correlation_data:
     kotekan_stage: testDataCheckInt
     first_buf: host_correlation_buffer
     second_buf: simulated_correlation_buffer
-    num_frames_to_test: 2*buffer_depth
+    num_frames_to_test: 1
```
geoffryan: Running this for more than a buffer depth is important, have found bugs that way. Please revert to `2 * buffer_depth`.

config/ci-tests/gpu_batch/verify_cuda_n2k.yaml
```diff
@@ -272,13 +269,13 @@ check_correlation_data:
...
     check_metadata: True
 
 check_counts_data:
     kotekan_stage: testDataCheckInt
     first_buf: host_counts_buffer
     second_buf: simulated_counts_buffer
-    num_frames_to_test: 2*buffer_depth
+    num_frames_to_test: 2
```
geoffryan: See above, should remain `2 * buffer_depth`

config/ci-tests/gpu_batch/verify_cuda_n2k_corr.yaml
```diff
@@ -164,5 +163,5 @@ check_data:
     kotekan_stage: testDataCheckInt
     first_buf: host_correlation_buffer
     second_buf: simulated_correlation_buffer
-    num_frames_to_test: buffer_depth
+    num_frames_to_test: 1
```
geoffryan: Should remain `buffer_depth`

lib/stages/monitorBuffer.cpp
```diff
@@ -26,26 +27,107 @@ STAGE_CONSTRUCTOR(monitorBuffer) {
...
+
+        if (consume_frames) {
+            for (auto& state : buffer_states) {
+                while (!stop_thread) {
+                    struct timespec deadline;
+                    clock_gettime(CLOCK_REALTIME, &deadline);
+                    int status = state.buffer->wait_for_full_frame_timeout(
+                        consumer_name, state.next_frame_id, deadline);
+                    if (status == -1) {
+                        goto end_loop;
```
geoffryan, line 75: whoa! I thought this was C++, not K&R C :p

lib/stages/monitorBuffer.cpp
```diff
@@ -26,26 +27,107 @@ STAGE_CONSTRUCTOR(monitorBuffer) {
...
+    double monitor_start_time = e_time();
+    const std::string& consumer_name = get_unique_name();
+
     while (!stop_thread) {
+
+        if (consume_frames) {
+            for (auto& state : buffer_states) {
+                while (!stop_thread) {
+                    struct timespec deadline;
+                    clock_gettime(CLOCK_REALTIME, &deadline);
```
geoffryan: CLOCK_REALTIME is not reliable for purposes like this (it can return negative time differences in the future, etc), CLOCK_MONOTONIC_RAW would be a better choice here.

geoffryan: Except that `last_arrival_time` in the Buffer is set using `e_time()` from `lib/util/util.hpp` which calls `gettimeofday` from `<sys/time.h>` which also shouldn't be used for serious things...  For now, this should probably just also use `e_time()` to be consistent.

geoffryan (changes requested):
Looks great!  Mostly requesting some documentation for the CI system, and a few quibbles about some individual tests and stages.

tests/notest_baseband_readout.py
geoffryan, line 1: Instead, do you want to just move this to `notest_baseband_readout.py` as we did for a couple other tests? This feels easy to miss later, it should be obvious we're not running this test.

.github/workflows/scheduled_tasks.yml
jbmertens: On this branch, nothing, but hopefully it'll make it into the main branch eventually, then can be used to trigger docker builds manually through github ui. (comment added)

.github/workflows/test_kotekan_build.yml
jbmertens: Wanted to keep it short, since the job names are getting a bit out of hand. "test" seems good

config/ci-tests/gpu_batch/verify_cuda_n2k.yaml
jbmertens: Any objection moving this to a scheduled overnight test then?

lib/stages/monitorBuffer.cpp
jbmertens: Haha, just following code patterns already in here... I think @andrerenard  had a reason for this?

.github/workflows/main.yml
jbmertens: ~When running jobs locally with act, and not in a container, it will be needed. (Not that I've tested that for this PR...)~
Right, these do run in containers.

.github/workflows/main.yml
jbmertens: I thought about changing it, but this groups them up in the actions sidebar 🤷

config/ci-tests/gpu_batch/verify_cuda_n2k.yaml
jbmertens: Reverting it for now.

.github/workflows/main.yml
geoffryan: It merges them in the sidebar, but since they all share the "test" name you can't tell which is which without actually clicking through and seeing the text output, even their long names are identical.  If they're separate runs they should have separate names.

.github/workflows/main.yml
geoffryan: Same as above.

tests/test_compression.py
geoffryan, line 1: This is a substantial rewrite of this test with code that you state is not convincingly correct.  For now just revert this change.  If running this test becomes a bottle neck, we can come back to it and speed it up in a way we're confident is correct.

tests/test_receiveflags.py
```diff
@@ -339,7 +339,7 @@ def test_start_time(tmpdir_factory):
         [3],
     ]
     frame_flags = [[1, 0, 1, 1, 0], [1, 0, 1, 1, 1], [1, 1, 0, 1, 1], [1, 1, 1, 0, 1]]
-    ts = [time.time() + 15, time.time() + 17, time.time() + 18.5]
+    ts = [time.time() + 6, time.time() + 8, time.time() + 10]
```
geoffryan: Is it important that one of the original times was not an integer?  Maybe change the 10 to a 9.5?

tests/test_receiveflags.py
```diff
@@ -401,7 +401,7 @@ def test_start_time_out_of_order(tmpdir_factory):
...
```
geoffryan: 10 -> 9.5 ?

tests/test_receiveflags.py
```diff
@@ -575,7 +575,7 @@ def test_dset_id_change(tmpdir_factory):
...
```
geoffryan: 10 -> 9.5?

geoffryan (changes requested):
Just a couple requests, not comfortable with substantially changing tests unless we're sure we're not losing coverage.

geoffryan (approved):

PR #1360, opened 2025-11-18
fix reading of header data in bufferRecv

rhaas80:


lib/stages/monitorBuffer.cpp
```diff
@@ -93,7 +93,14 @@ void monitorBuffer::main_thread() {
 
         sleep(1);
         double cur_time = e_time();
-        for (Buffer* buf : buffers) {
+        for (size_t i = 0; i < buffers.size(); ++i) {
+            if (consume_frames && clean_exit_after_frames > 0) {
```
jbmertens, line 98: Helpful to add a comment here on why this was added (commit message is good)

jbmertens (commented):
LGTM.
Can we move the config/tests/test_send_recv.yaml to config/ci-tests/cpu_batch as part of this PR?

lib/stages/monitorBuffer.cpp
rhaas80: ok

rhaas80:
> LGTM. Can we move the config/tests/test_send_recv.yaml to config/ci-tests/cpu_batch as part of this PR?

sure. will have to run clang firmat on ctc once more. apparently debian's clang-format-18 is different from Ubuntu's :-(

jbmertens:
> > LGTM. Can we move the config/tests/test_send_recv.yaml to config/ci-tests/cpu_batch as part of this PR?
> 
> sure. will have to run clang firmat on ctc once more. apparently debian's clang-format-18 is different from Ubuntu's :-(

Oh, that's... annoying.
Maybe that was causing some of the issues in #1354 .

lib/stages/bufferSend.hpp
```diff
@@ -19,33 +19,31 @@
 #include <stdint.h>           // for uint32_t
 #include <string>             // for string, basic_string
 
+/**
+ * @struct bufferFrameHeaderNoConfigTracker
+ * @brief Wire header without the config tracker flag.
+ */
+struct bufferFrameHeaderNoConfigTracker {
```
andrerenard, line 26: `#pragma pack` worth doing there?

lib/stages/bufferSend.hpp
```diff
@@ -19,33 +19,31 @@
...
  * @struct bufferFrameHeader
  * @brief Stable wire header for transfer details.
  *
  * Uses fixed-width types and explicit padding to (hopefully) make the layout
  * consistent across compilers/ABIs without relying on packing pragmas.
  */
-struct bufferFrameHeader {
-    uint32_t metadata_size;
-    uint32_t frame_size;
+struct bufferFrameHeader : public bufferFrameHeaderNoConfigTracker {
```
andrerenard, line 40: And here?

andrerenard (approved):
This specific change looks good.  

However, given that we now have different protocols, we should really have some kind of version/protocol checking.   Ideally we'd expand the header to include some kind of cookie + protocol version.   Then at least we could generate errors if we have a mismatch between system.   That doesn't have to be part of this PR, but something we should add.   We'll have to think about how much to maintain old CHIME recv compatibility, or if we just have a legacy bufferSend for that.

jbmertens (approved):

jbmertens:
Approved, though would still like the send_recv yaml test moved. Can be a future change.

rhaas80:
> LGTM. Can we move the config/tests/test_send_recv.yaml to config/ci-tests/cpu_batch as part of this PR?

Done in 965028b15

lib/stages/bufferSend.hpp
rhaas80: maybe but it won't do anything for correctness. The standard does guarantee that a pointer to a derived class can be used as a pointer to its base class (in C this would be that a pointer to a struct can be used as a pointer to the struct's first member).

So it may possibly save some bytes (doesn't right now, there's static asserts on the expected sizes), but it won't help anything else (and it would be on the order of a couple of bytes saved).

lib/stages/bufferSend.hpp
rhaas80: same as above.

rhaas80:
> This specific change looks good.
> 
> However, given that we now have different protocols, we should really have some kind of version/protocol checking. Ideally we'd expand the header to include some kind of cookie + protocol version. Then at least we could generate errors if we have a mismatch between system. That doesn't have to be part of this PR, but something we should add. We'll have to think about how much to maintain old CHIME recv compatibility, or if we just have a legacy bufferSend for that.

As far as CHIME's receiver nodes are concerned, since they won't update their kotekan instance, there must be a way to produce the exact same network format as before. That is the two size fields and then the metadata and payload. No version information possible for this legacy format.

There's little consistency checks between independent kotekan instances. Eg it is also up to the user that the yaml config files describe compatible settings anyway.

PR #1361, opened 2025-11-18
Symbol: Add missing const in formatter

eschnett:


rhaas80 (commented):
looks ok to me

rhaas80 (approved):
looks good to me. Apparently I did not check "Approve".

PR #1362, opened 2025-11-18
Symbol: use c_str instead of data when passing along a C-like string

rhaas80:


rhaas80:
basically, with `std::string tst = "123";` one has that `tst.data()` is `{'1','2','3'}` and no terminating `'\0'`.

lib/metadata/Symbol.cpp
```diff
@@ -29,7 +29,7 @@ const char* Symbol::lookup_or_insert(const char* str) {
 }
 
 Symbol::Symbol(const std::string_view& str) : value(lookup_or_insert(str.data())) {}
```
eschnett: What about this line? If `string_view` has no `c_str` member then we need to convert to `string` or remove this constructor.

rhaas80: I think those are fine.

As far as I understand the docs for `std::string_view`' at https://en.cppreference.com/w/cpp/string/basic_string_view.html, its `data()` function returns the pointer that one stored inside it before, so as long as that pointer was a `NUL` terminated C string then this is ok. 

On the other hand `std::string` makes copy of the C string's content, but not of the terminating `NUL` (constructor 7 in https://en.cppreference.com/w/cpp/string/basic_string/basic_string.html).

So `data()` is different for:

```
std::string s("foo");
std::string_view sv("bar");
```

since the later is just a pointer to the 4 bytes already allocated for `"bar\0"` but the former is 3 bytes `f`, `o`, `o`.

Very confusing.

eschnett: This is a general constructor for a `string_view`. We don't know how the string view was created. Maybe it points just into the middle of a `std::string`, and in this case it will not be NUL terminated.

rhaas80: While we do indeed not know how the `string_view` has been constructed, I would say that this would be a failure of whichever code created the string view. Note that there is no constructor of `string_view` that takes a `std::string` as an argument (https://en.cppreference.com/w/cpp/string/basic_string_view/basic_string_view.html), they all take either `char*` or explicit pointer and length (there is a conversion from `std::string` to `std::string_view` but that should not be used in our case since we have an explicit constructor for `Symbol` taking `std::string` as an argument). So in this sense a `string_view` seems to be not so much a view (substring) into a `std::string` but a C++ view of a C string.

Just to be explicit, `string_view` does not have a `c_str()` member function, the `data()` function is the only one that can give access to the underlying array (other than say taking the address of the return of `front()` or `[0]` or so).

If we do want to support use cases where a subset of a given `char` array (but no terminating `NUL`) is supported the one has to change `lookup_or_insert` to not take a `char *` as an argument but instead a `string_view` so that pointer and count can be provided. `std::string` contains a conversion operator to `std::string_view` so that single `lookup_or_insert` would work for all cases (`std::string_view`, `char *` via `std::string_view`'s constructor from `char*`, and `std::string` via its conversion to `std::string_view`).

Is that what you would have in mind?

rhaas80: Something like 798df949b ?

eschnett: That look good!

rhaas80: > That look good!

Very good. GitHub requires approval by at least one reviewer, if you could do that then I will squash and merge.

eschnett (approved):

PR #1363, opened 2025-11-19
Use telescope object

eschnett:


lib/utils/CHORDTelescope.cpp
```diff
@@ -875,12 +914,27 @@ void CHORDTelescope::set_dish_info(const kotekan::Config& config, const std::str
...
-}
 
+    using std::max;
+
+    // Create and fill dish grid
+    _dish_grid = dishGrid(_num_dishes_x, _num_dishes_y);
+    for (int dish = 0; dish < _num_dishes; ++dish) {
+        const dishInfo& dish_info = get_dish_at_idx(dish);
+        if (dish_info.type == DishType::Fake)
+            continue;
```
geoffryan, line 931: You might want to include Fake dishes here, if, for instance we're running pathfinder but only have 27 dishes plugged in?

Alternately we could add a DishType for "Unplugged" or something?

lib/utils/CHORDTelescope.hpp
```diff
@@ -291,6 +346,12 @@ class CHORDTelescope : public Telescope {
      **/
     int get_num_dishes() const;
 
+    int get_num_dishes_x() const;
+    int get_num_dishes_y() const;
+
+    double get_dish_separation_x_m_DONOTUSETHIS() const;
+    double get_dish_separation_y_m_DONOTUSETHIS() const;
```
geoffryan: I'm fine keeping these, don't really need the DONOTUSETHIS tag?

lib/utils/CHORDTelescope.cpp
eschnett: You allow duplicate dish indices for fake dishes. I don't know what that means. For our GPU kernels we need exactly 64 thingies, and each thingie needs to have an index between 0 and 63 without duplication. If we use fake dishes as fill-ins then fake dishes need to have much of the same properties as real dishes.

eschnett:
@geoffryan Can you give this another look? I think this is ready to go, I only need to update the other test pipelines, but I want to make sure we agree on the `CHORDTelescope` changes.

geoffryan (commented):
Looks good to me!

config/fengine/include/fengine_smallfinder.j2
geoffryan, line 1: Bunch of TODOs here, but I'm assuming that's for the next PR?

geoffryan (approved):
Looks good to merge!

config/fengine/include/fengine_smallfinder.j2
eschnett: Yes, these require updates to the FRB kernels.

PR #1365, opened 2025-11-21
Misc changes to support N2 writer

jbmertens:
- update eop -> bin_eop and time_center_eop
 - "layout" -> "vis_layout"
 - Kotekan should exit gracefully on FATAL_ERROR, hard exit on EXIT_ERROR
 - Other misc. cleaning


lib/metadata/N2Metadata.hpp
```diff
@@ -57,8 +57,15 @@ struct N2MetadataFormat {
...
-    struct EOP eop;
+    /// Earth Orientation Parameters at the center of the integration/accumulation time
+    struct EOP time_center_eop;
+    /// Earth Orientation Parameters within an integration/accumulation bin
+    struct EOP bin_eop;
+    // bin start/end info for convenience
+    double bin_start_ERA_deg; /// Earth Rotation Angle at start of bin
+    double bin_end_ERA_deg;   /// Earth Rotation Angle at end of bin
+    int64_t bin_start_LAST;   /// local apparent sidereal time (nanoseconds) at start of bin
+    int64_t bin_end_LAST;     /// local apparent sidereal time (nanoseconds) at end of bin
```
geoffryan, line 68: are nanoseconds (and an integer) the appropriate unit (and type) for LST?  I thought i t was more of an hour-angle thing, in a double, or possibly, like, nanoarcseconds if in an int.

lib/stages/N2Accumulate.cpp
```diff
@@ -162,47 +167,22 @@ N2Accumulate::N2Accumulate(Config& config, const std::string& unique_name,
...
-
-    // Check buffer frame sizes
+    // Ensure incoming buffer frame sizes are correct
     size_t in_corr_frame_size = 2 * sizeof(int32_t) * _n2k_correlation_num_products
                                 * _num_freq_per_n2k_frame * _n_integrations_per_n2k_frame;
     size_t in_counts_frame_size = sizeof(int32_t) * _n2k_counts_num_products
                                   * _num_freq_per_n2k_frame * _n_integrations_per_n2k_frame;
     size_t in_rfimask_frame_size = _num_freq_per_n2k_frame * _n_fpga_samples_per_n2k_frame / 8;
-    size_t out_n2_frame_size =
-        N2FrameView::calculate_frame_size(_num_elements, 0, _N2_num_products);
```
geoffryan, line 184: This check is important and should remain.  N2Accumulate has not idea if the num_elements it sees is the same one the N2Buffer saw.

lib/stages/N2Accumulate.cpp
```diff
@@ -40,108 +40,113 @@ REGISTER_KOTEKAN_STAGE(N2Accumulate);
...
-    if (!_packet_loss_is_scalar)
-        FATAL_ERROR("N2Accumulate configured to use packet loss matrix, which is not implemented.");
-    assert(_packet_loss_is_scalar);
+    out_buf = get_buffer("out_buf");
+    out_buf->register_producer(unique_name);
 
-    // sampling information
-    _n_fpga_samples_per_n2k_frame = config.get<int64_t>(unique_name, "samples_per_data_set");
-    _n_fpga_samples_per_n2k_correlation = config.get<int64_t>(unique_name, "sub_integration_ntime");
+    // Ensure outgoing buffer is of type N2 (correct size implicit)
```
geoffryan: Correct size is not implicit.  The N2Buffer and this stage may see different num_elements, etc.  That would likely mean the config is bad, but there's nothing stopping it, unless we check here.

lib/stages/N2Accumulate.hpp
```diff
@@ -76,29 +76,32 @@ class N2Accumulate : public kotekan::Stage {
...
 
-    // Frame and vis sample durations in nanoseconds
-    // uint64_t _in_frame_duration_nsec;
-    // uint64_t _in_frame_vis_duration_nsec;
+    int64_t _rfi_downsampling_factor; ///< Downsampling factor for RFI mask
 
-    static constexpr int64_t _n2k_correlation_blocksize = 16; // THIS IS ALWAYS 16
-    static constexpr int64_t _n2k_counts_blocksize = 8;       // THIS IS ALWAYS 8
+    int64_t _num_elements; ///< Total number of telescope elements (~2 * num dishes)
+    int64_t _num_ev;       ///< Number of eigenvalues/vectors
```
geoffryan: Do we need this? The stage produces no eigen information.

lib/stages/hdf5FileWrite.cpp
geoffryan, line 1: SO MANY FIELDS

lib/utils/CHORDTelescope.cpp
```diff
@@ -692,6 +699,13 @@ EOP CHORDTelescope::get_EOP_at_UT1(int64_t t_ut1) const {
 
     {
         std::shared_lock lock(_eop_lock);
+
+        if (_eop_table.empty()) {
+            WARN("EOP table is empty, cannot interpolate EOP at time {:d} s + {:d} ns.",
```
geoffryan: "at  time" -> "at UT1 time"

tests/boost/test_timeUtil.cpp
```diff
@@ -87,7 +88,11 @@ void TimeData::setup() {
         f_exists(script),
         fmt::format("critical script {:s}/timeUtil.py was not found", TEST_SCRIPT_DIR));
 
-    std::string cmd = fmt::format("python3 {:s} isot", script);
+    // Prefer system kotekan_env python if available; else fallback to python3
+    std::string py_interpreter = "/opt/kotekan_env/bin/python";
+    if (access(py_interpreter.c_str(), X_OK) != 0)
+        py_interpreter = "python3";
+    std::string cmd = fmt::format("{:s} {:s} isot", py_interpreter, script);
```
geoffryan: I don't think this is the correct move here. The tests just need to be run from an environment that has these modules available, the same way the pytests are run.

tests/test_eigenvis.py
```diff
@@ -54,6 +54,7 @@
     },
     "num_dishes": 100,
     "gps_time": {"frame0_nano": 0},
+    "vis_layout": "FullUpperTri",
```
geoffryan: `vis_layout` in general should probably not be set globally. It is a per-frame thing.  The `FakeN2Buffer` class sets the layout to FullUpperTri by default already.

tests/test_n2_accumulate.py
```diff
@@ -127,6 +128,7 @@
         ],
     },
     "gps_time": {"frame0_nano": t_start_s * GIGA},
+    "vis_layout": "FullUpperTri",
```
geoffryan: Should not be here, this is a per-frame thing.

tests/test_n2_time_downsample.py
```diff
@@ -86,6 +86,7 @@
         "dish_inputs": [],
     },
     "gps_time": {"frame0_nano": t_start_inst_ns},
+    "vis_layout": "FullUpperTri",
```
geoffryan: Delete this, should not be here.  Handled by python/kotekan/runner.py

lib/metadata/Symbol.hpp
```diff
@@ -41,15 +41,14 @@ class Symbol {
 
     // Look up a string in the known symbols. If it is not known,
     // insert it.
-    const char* lookup_or_insert(const std::string_view& str);
+    const char* lookup_or_insert(const char* str);
```
geoffryan: Why this change? Feels less safe.

geoffryan (changes requested):
Overall looks good!  A few small changes requested. The pytest layout fix should be done in the `FakeN2Buffer` and `DumpN2Buffer` classes in `python/kotekan/runner.py`, not in the individual configs. 

A couple decisions to make: what exactly FATAL_ERROR should do? Should N2Accumulate accept num_ev != 0? (I vote no).

lib/metadata/N2Metadata.hpp
jbmertens: Depends on the type we decide for it or what we mean for it. I was following the UT1 type? May punt this change to when we actually populate it.

tests/test_eigenvis.py
jbmertens: I get errors without it.

jbmertens: Ah, per-buffer thing, though it's global to this particular test. Looks like I missed a "layout": ... in the runner.FakeN2Buffer.

lib/metadata/Symbol.hpp
jbmertens: Not needed any more, with updates to FATAL_ERROR

lib/stages/N2Accumulate.cpp
```diff
@@ -40,108 +40,117 @@ REGISTER_KOTEKAN_STAGE(N2Accumulate);
...
-    if (!(_n_fpga_samples_per_n2k_frame % _n_fpga_samples_per_n2k_correlation == 0)) {
-        FATAL_ERROR("samples_per_data_set ({:d}) is not a multiple of sub_integration_ntime ({:d})",
-                    _n_fpga_samples_per_n2k_frame, _n_fpga_samples_per_n2k_correlation);
-        std::abort();
-    }
+    // Ensure outgoing buffer is of type N2 and sized correctly
+    if (out_buf->buffer_type != "N2")
+        FATAL_ERROR("N2Accumulate out_buf ({:s}) is not of type N2.", out_buf->buffer_name);
+    size_t out_n2_frame_size =
+        N2FrameView::calculate_frame_size(_num_elements, 0, _N2_num_products); // Enforce 0 ev
```
geoffryan: Surprised this works, "_N2_num_products" isn't initialized until line ~150, later in the constructor.

geoffryan (changes requested):
One change to N2Accumulate constructor needed, and some Exception handling in Stage, and possibly in StageFactory and kotekanLogging to catch the constructors throwing.

geoffryan (approved):
LGTM!

lib/core/kotekanLogging.hpp
```diff
@@ -101,7 +106,38 @@
     } while (0)
 #endif // DEBUGGING
 
-// Use this for serious errors.  i.e. things that require the program to end.
+// Use this for fatal errors that need to exit immediately.
+// Prints an error message and immediately calls exit().
+#define EXIT_ERROR(m, ...)                                                                         \
+    do {                                                                                           \
+        ERROR(m, ##__VA_ARGS__);                                                                   \
+        std::exit();                                                                               \
```
andrerenard: Should this set a failure code?

lib/core/kotekanLogging.hpp
```diff
@@ -101,7 +106,38 @@
...
+    } while (0)
+#define EXIT_ERROR_NON_OO(m, ...)                                                                  \
+    do {                                                                                           \
+        ERROR_NON_OO(m, ##__VA_ARGS__);                                                            \
+        std::exit();                                                                               \
```
andrerenard: Same here.

lib/core/Stage.cpp
```diff
@@ -98,7 +98,15 @@ void Stage::start() {
         pid_t tid = syscall(SYS_gettid);
         register_tid(tid);
 #endif
-        main_thread_fn(std::ref(*this));
+        try {
+            main_thread_fn(std::ref(*this));
+        } catch (const FatalError& e) {
+            // FATAL_ERROR already called exit_kotekan; avoid terminating the thread abruptly.
```
andrerenard: Don't you mean avoid terminating the application abruptly?  We've already aborted the thread.

lib/core/Stage.cpp
```diff
@@ -98,7 +98,15 @@ void Stage::start() {
...
+            ERROR("Stage {:s} caught FatalError: {:s}", unique_name, e.what());
```
andrerenard:
```suggestion
            ERROR("Stage {:s} caught FatalError: {:s}, attempting controlled shutdown.", unique_name, e.what());
```

lib/core/Stage.cpp
```diff
@@ -98,7 +98,15 @@ void Stage::start() {
...
+        } catch (const std::exception& e) {
+            ERROR("Exception in stage {:s}: {:s}", unique_name, e.what());
+            throw;
```
andrerenard, line 111: Should we throw the exception we just got?

andrerenard (changes requested):

lib/core/Stage.cpp
jbmertens: `throw;` with no arguments should do just this.

andrerenard (approved):
Error handling changes look good, thanks for the updates.

PR #1366, opened 2025-11-22
correct sending of header for non-configracker setup, correct send/recv test input file

rhaas80:


jbmertens (approved):
Looks good except the mixed initializer syntax.

rhaas80:
> Looks good except the mixed initializer syntax.

yeah, clang does not like that one. There seems to be no way to use an initializer with the `.desc` syntax here. Oh well, I'll write 3 lines of code then.

Found a bug in my "fix" for `bufferRecv` ... no code without an explicit test case ought to be trusted.

PR #1368, opened 2025-11-25
Vis transpose writer 

jbmertens:
TODO:

 - [x] Ensure frequency ids map to an index in array output
        (Will expect everything between freq_id start and stop)
 - [x] Use HDF5 instead of zarr?
 - [x] Write config data
 -  ~Write gains (or reference to)~
 - [x] Write/note element indexes (maybe in config?)
        (Write a full map of indexes)
 - [x] Check consistency with CHIME format
 - [x] add CI test(s) (ongoing)
 - ~Track/write out digital gains?~


lib/stages/EigenVisIter.cpp
```diff
@@ -159,6 +159,13 @@ void EigenVisIter::main_thread() {
 
         // Create view to output frame
         in_buf->pass_metadata(input_frame_id, out_buf, output_frame_id);
+
+        // Ensure metadata reflects the number of eigenvectors so the frame layout matches the
+        // buffer's frame size (which is sized for the eigen data).
+        auto out_meta = get_N2_metadata(out_buf, output_frame_id);
+        if (out_meta)
+            out_meta->num_ev = _num_eigenvectors;
```
geoffryan, line 167: We need to check here that the output buffer is correctly sized for the write we're about to do, and FATAL_ERROR if not. Ideally this would be done in the constructor but I don't think it has all the necessary information as-is.  Just need to make sure the output frame size is consistent with the `num_elements` of the input, `num_ev` of the stage, and that its in the FullUpperTri layout.

lib/stages/EigenVisIter.cpp
geoffryan, line 1: Sorry, this isn't really this PR's job, but while you're modifying how this stage handles N2 frames...
The check on line 110, that the number of products is correct, should be replaced with a check that it's in the FullUpperTri layout.

lib/utils/CHORDTelescope.hpp
```diff
@@ -133,13 +133,17 @@ struct dishInputFields {
  * @conf    gps_host            string. The GPS server IP address.
  * @conf    gps_port            uint.   The port number on the GPS server.
  * @conf    gps_endpoint        string. The enpoint with the GPS time.
+ * @conf    sampling_rate_MHz   double. ADC Sampling Rate (default: 3.2 GHz for CHORD)
+ * @conf    fft_lenth           double. F-engine FFT length (default: 16384 for CHORD)
+ * @conf    nyquist_zone        uint8.  Nyquist Zone we're operating in (default: 1 for CHORD)
+ * @conf    max_output_freq_MHz double. Maximum output frequency for file writing, in MHz (default:
```
geoffryan: Maybe just "max_freq_MHz"?  "max_observed_freq_MHz"? (and similar for "min")

lib/utils/CHORDTelescope.hpp
```diff
@@ -473,29 +492,46 @@ class CHORDTelescope : public Telescope {
...
     /**
-     * @brief Get the frequency width of a given channel.  When the frequency spacing is constant,
-     *this is the equivalent to the spacing between frequency channels.
-     *
-     * @return  The width of the frequency channel in MHz.
+     * @brief Get the maximum frequency ID that will be output to files.
+     * @returns     The maximum frequency ID used in files.
      **/
-    double freq_width_MHz(freq_id_t freq_id) const override;
+    size_t max_output_freq_id() const;
```
geoffryan: This should return a "freq_id_t", not "size_t". (and possibly change name depending on refactor)

lib/utils/CHORDTelescope.hpp
```diff
@@ -473,29 +492,46 @@ class CHORDTelescope : public Telescope {
...
 
     /**
-     * @brief Get which Nyquist zone we are in.
+     * @brief Get the minimum frequency ID that will be output to files.
+     * @returns     The minimum frequency ID used in files.
+     **/
+    size_t min_output_freq_id() const;
```
geoffryan: return freq_id_t

lib/utils/CHORDTelescope.hpp
```diff
@@ -473,29 +492,46 @@ class CHORDTelescope : public Telescope {
...
+     * (corresponding to min_output_freq_id to max_output_freq_id).
+     * Not guarded against freq_id being outside output range.
      *
-     * @return  The Nyquist zone.
+     * @param  freq_id  The internal frequency ID.
+     *
+     * @returns     The frequency index written in files.
      **/
-    uint8_t nyquist_zone() const override;
+    size_t get_output_freq_idx(freq_id_t freq_id) const;
```
geoffryan: I think this should be moved to the file writer.

lib/utils/N2Util.hpp
```diff
@@ -78,6 +88,23 @@ inline uint32_t prod_index(uint32_t i, uint32_t j, uint32_t block, uint32_t N) {
...
 }
 
+/**
+ * @brief Correlator input type
+ */
+struct input_ctype {
+    input_ctype(uint16_t idx, std::string input_id) {
+        element_idx = idx;
+        std::memset(input_id_str, 0, 32);
+        input_id.copy(input_id_str, 32);
```
geoffryan: This should be 31 to keep the null-termination character.

lib/utils/N2Util.hpp
```diff
@@ -78,6 +88,23 @@ inline uint32_t prod_index(uint32_t i, uint32_t j, uint32_t block, uint32_t N) {
     return block * block * b_ix + (i % block) * block + (j % block);
 }
 
+/**
+ * @brief Correlator input type
+ */
+struct input_ctype {
+    input_ctype(uint16_t idx, std::string input_id) {
```
geoffryan, line 95: Should we check `input_id` is short enough to fit in 32 bytes?

lib/utils/N2Util.hpp
geoffryan, line 1: This might be over-thinking, but in the spirit of potentially flattening the data files in the future, I'm not sure we should keep these structs around? Maybe they should live in the viswriter as a CHIME compatibility layer?

tests/boost/test_CHORDTelescope.cpp
geoffryan, line 1: Add tests for the max/min freq_id functions.  Should test behaviour in nyquist zones 1 & 2, and have the correct behaviour (whatever you deem that is) when the given frequencies correspond exactly to a freq_id and when they do not.

config/ci-tests/standalone/test_pipeline.j2
```diff
@@ -169,38 +220,39 @@ host_pl_mask_exp_ringbuffer:
...
-    # type: const
-    # value: 34 # 1i # 34 # = 2 + 2i
-    type: random_signed_offset  # const_offset
+    type: random_signed_offset
     value: 120
     kotekan_stage: testDataGen
     out_buf: host_voltage_buffer
     name: "E"
-    array_shape: [ 2048, 16, 2, 64 ] # [ samples_per_data_set, num_local_freq, num_polarizations, num_dishes ]
+    array_shape: [ {{ samples_per_data_set }}, {{ num_local_freq }}, 2, {{ num_dishes }} ]
```
geoffryan, line 266: OH amazing!  Setting these manually was SO annoying. I should use jinja2 all the time.

lib/stages/hdf5N2Write.hpp
```diff
@@ -0,0 +1,265 @@
...
+#include <N2Util.hpp>
+#include <cassert>
+#include <filesystem>
+#include <highfive/H5File.hpp>
+#include <map>
+#include <memory>
+#include <optional>
+#include <string>
+#include <vector>
+#include <visUtil.hpp>
```
geoffryan: visUtil shouldn't be necessary if we're writing N2 files right? Only N2Util should be necessary.

geoffryan (commented):
Not a full review, some initial comments.

lib/stages/EigenVisIter.cpp
jbmertens: Agreed. This is actually added because it was missing in the output metadata. Maybe it was never finished / current CHIME pipelines just made assumptions and/or didn't use it?

lib/stages/hdf5N2Write.hpp
jbmertens: Yes. (Probably need to run iwyu for these files...)

lib/utils/CHORDTelescope.hpp
jbmertens: Hm, this is a different type of id/index though.
(Should freq_id_t be a size_t anyways?)

lib/utils/N2Util.hpp
jbmertens: Yeah, depends on what people downstream want 🤷

lib/utils/CHORDTelescope.hpp
geoffryan: Isn't this returning a freq_id?  Like if the maximum observed frequency has id 7500, this returns 7500?  In which case it should be the same type as a freq_id. (I have no opinion on what the underlying freq_id type actually is, we should just be consistent)

lib/stages/hdf5N2Write.hpp
```diff
@@ -0,0 +1,265 @@
...
+
+    // t-dependent metadata
+    std::vector<uint64_t> fpga_start_tick;         // (t)
+    std::vector<uint64_t> frame_length_fpga_ticks; // (t)
+    std::vector<int64_t> time_center_ut1;          // (t)
+    std::vector<int64_t> bin_ut1;                  // (t)
+    std::vector<double> bin_start_ERA_deg;         // (t)
+    std::vector<double> bin_end_ERA_deg;           // (t)
+    std::vector<int64_t> bin_start_LAST;           // (t)
+    std::vector<int64_t> bin_end_LAST;             // (t)
```
geoffryan: I don't think there's a world where these will be `int`s.  LAST is like ERA, it's a measure of angle and will be a `double`

lib/stages/hdf5N2Write.hpp
```diff
@@ -0,0 +1,265 @@
...
+ * @conf base_dir           String. Directory to write into
+ * @conf num_file_t         UInt. Number of time frames per file
+ * @conf late_frame_grace_seconds  UInt. Grace period in seconds for late frames (default: 60)
+ * @conf max_frames                Int.  Stop writing after this many frames (-1 = unlimited)
+ *
+ * @par Metrics
+ * @metric kotekan_N2write_write_time_seconds  Duration to write the last flush
+ * @metric kotekan_N2write_n_datasets          Number of datasets currently open
+ *
+ * @note User-level documentation lives in docs/sphinx/user/processes/hdf5N2Write.rst.
```
geoffryan, line 234: Don't see this doc in this PR?

lib/stages/hdf5N2Write.hpp
```diff
@@ -0,0 +1,265 @@
...
+        return num_file_t * f + t; // (f, t)
+    }
+};
+
+/**
+ * @class hdf5N2Write
+ * @brief Buffered-transpose writer: buffers sequential time frames and writes HDF5 files.
+ *
+ * This stage groups frames into UTC-midnight-aligned windows of length
+ * `file_seconds`, buffers a complete (num_file_f * file_nt) block per output file in
```
geoffryan: I think this has moved to a "fixed number of frames" setup, so this comment should be altered to describe the current behaviour.

lib/stages/hdf5N2Write.hpp
```diff
@@ -0,0 +1,265 @@
...
+    void _finalize_file(N2FileData& filedata);
+
+    /**
+     * @brief Finalize N2FileData objects that have been inactive for too long.
+     * @param files  Map of abs_idx-N2FileData objects to finalize
+     * @param exclude_abs_file_idx  Optional file index to skip while finalizing
+     */
+    void _grace_finalize_files(std::map<size_t, std::unique_ptr<N2FileData>>& files,
+                               const size_t* exclude_abs_file_idx = nullptr);
+
```
geoffryan, line 295: Doc comment for ` _finalfile_exists`?

lib/stages/hdf5N2Write.cpp
```diff
@@ -0,0 +1,891 @@
...
+#include <optional>
+#include <prometheusMetrics.hpp>
+#include <sstream>
+#include <string>
+#include <sys/stat.h>
+#include <type_traits>
+#include <unistd.h>
+#include <utility>
+#include <vector>
+#include <visUtil.hpp>
```
geoffryan: Should no longer be needed.

lib/stages/hdf5N2Write.cpp
```diff
@@ -0,0 +1,891 @@
...
+    // data structure attributes
+    _check_create_attribute(*file, "num_elements", fv.num_elements);
+    _check_create_attribute(*file, "num_prod", fv.num_prod);
+    _check_create_attribute(*file, "num_ev", fv.num_ev);
+    _check_create_attribute(*file, "num_freq",
+                            fv.nfreq); // telescope frequencies (not file freqs)
+    _check_create_attribute(*file, "vis_layout",
+                            std::string(fv.vis_layout == N2Layout::FullUpperTri
+                                            ? "FullUpperTri"
+                                            : "RedundantBaselineAvg"));
```
geoffryan: This will get troublesome to maintain if we add more layouts or change their names.  You should be able to just say something like  `json{fv.vis_layout}.dump()` instead?  (I should probably just add a to_str() method that does that on its own)

lib/stages/hdf5N2Write.cpp
```diff
@@ -0,0 +1,891 @@
...
+    _check_create_attribute(*file, "fpga_seq_length_nsec", telescope.seq_length_nsec());
+    _check_create_attribute(*file, "origin_itrs_lon_deg", telescope.get_origin_itrs_lon_deg());
+    _check_create_attribute(*file, "origin_itrs_lat_deg", telescope.get_origin_itrs_lat_deg());
+    _check_create_attribute(*file, "dish_coelev_deg", telescope.get_dish_coelev_deg());
+    _check_create_attribute(*file, "num_dishes", telescope.get_num_dishes());
+    _check_create_attribute(*file, "EOP_table_len", telescope.get_EOP_table_len());
+    _check_create_attribute(*file, "num_file_f",
+                            telescope.num_output_freq()); // just file frequencies
+
+    // Store EOP table ERA_deg and t_ut1 only
```
geoffryan: If we're storing the table (which we should) we need to store the whole table, should add the missing fields here.

lib/stages/hdf5N2Write.cpp
```diff
@@ -0,0 +1,891 @@
...
+    }
+
+    // Store dish input info
+    {
+        // Set up an object to receive the dish input info.
+        dishInputFields dish_inputs;
+        // Fill the object with the info.
+        telescope.fill_input_maps(dish_inputs);
+
+        _check_create_dataset(*file, "/index_map/grid_x_idx", {dish_inputs.grid_x_idx.size()},
```
geoffryan: This is of length `num_dishes` not `num_element`, so the axis name should probably be "dish"?  Alternately we could amend the `fill_input_maps` function to check num_polarizations and append extra copies of `grid_x_idx`, etc.

lib/stages/hdf5N2Write.cpp
```diff
@@ -0,0 +1,891 @@
...
+                              {"element"}, HighFive::create_datatype<int64_t>(), props_empty);
+        auto dataset_x = file->getDataSet("/index_map/grid_x_idx");
+        dataset_x.write(dish_inputs.grid_x_idx);
+
+        _check_create_dataset(*file, "/index_map/grid_y_idx", {dish_inputs.grid_y_idx.size()},
```
geoffryan: As above, dish vs element.

lib/stages/hdf5N2Write.cpp
```diff
@@ -0,0 +1,891 @@
...
+        auto dataset_pos = file->getDataSet("/index_map/feed_pos_disp_m");
+        dataset_pos.write(dish_inputs.feed_pos_disp_m);
+
+        _check_create_dataset(*file, "/index_map/coelev_disp_deg",
+                              {dish_inputs.coelev_disp_deg.size()}, {"element"},
+                              HighFive::create_datatype<double>(), props_empty);
+        auto dataset_coelev = file->getDataSet("/index_map/coelev_disp_deg");
+        dataset_coelev.write(dish_inputs.coelev_disp_deg);
+
+        _check_create_dataset(*file, "/index_map/type", {dish_inputs.type.size()}, {"element"},
```
geoffryan: as above element <-> dish

lib/stages/hdf5N2Write.cpp
```diff
@@ -0,0 +1,891 @@
...
+                          {"frequency", "time"}, HighFive::create_datatype<float>(), props_empty);
+    _check_create_dataset(*file, flags_group_prefix + "/frac_rfi", {num_file_f, num_file_t_},
+                          {"frequency", "time"}, HighFive::create_datatype<float>(), props_empty);
+
+    _check_create_dataset(*file, "/fpga_start_tick", {num_file_t_}, {"time"},
+                          HighFive::create_datatype<uint64_t>(), props_empty);
+    _check_create_dataset(*file, "/frame_length_fpga_ticks", {num_file_t_}, {"time"},
+                          HighFive::create_datatype<uint64_t>(), props_empty);
+
+    _check_create_dataset(*file, "/time_center_ut1", {num_file_t_}, {"time"},
```
geoffryan: add "_ns" to name for units.

lib/stages/hdf5N2Write.cpp
```diff
@@ -0,0 +1,891 @@
...
+                          HighFive::create_datatype<int64_t>(), props_empty);
+    _check_create_dataset(*file, "/bin_ut1", {num_file_t_}, {"time"},
```
geoffryan: add "_ns" for units

lib/stages/hdf5N2Write.cpp
```diff
@@ -0,0 +1,891 @@
...
+    size_t check_idx = idx_ft(f_index, t_index);
+    if (added_ft[check_idx] != 0) {
+        ERROR_NON_OO("N2FileData: duplicate frame insertion at (f={}, t={})", f_index, t_index);
+        return false;
+    }
+
+    // Structural data consistency checks
+    // TODO: time ints might be off by a bit? Use appx. comparison instead?
+    if (fv.vis.size() != N2FrameView::get_num_prod(fv.num_elements, N2Layout::FullUpperTri)
+        || fv.weight.size() != N2FrameView::get_num_prod(fv.num_elements, N2Layout::FullUpperTri)
```
geoffryan: are we requiring frames be in the full layout here?  I think you could replace `N2Layout::FullUpperTri` with just `fv.vis_layout`?  Or add a check specifically requiring the layout being full.

lib/stages/hdf5N2Write.cpp
```diff
@@ -0,0 +1,891 @@
...
+        || fv.eval.size() != fv.num_ev || fv.evec.size() != fv.num_ev * fv.num_elements
+        || fv.gain.size() != fv.num_elements || fv.flags.size() != fv.num_elements
+        || fv.num_elements != num_elements || fv.num_prod != num_prod || fv.num_ev != num_ev
+        || fv.frame_length_fpga_ticks == 0
+        || (fpga_start_tick[t_index] > 0 && fpga_start_tick[t_index] != fv.fpga_start_tick)
+        || (frame_length_fpga_ticks[t_index] > 0
+            && frame_length_fpga_ticks[t_index] != fv.frame_length_fpga_ticks)
+        || (time_center_ut1[t_index] > 0
+            && time_center_ut1[t_index] != fv.time_center_eop.t_ut1)      // TODO: relax a bit?
```
geoffryan: Should not need to relax the UT1 checks.

lib/stages/hdf5N2Write.cpp
```diff
@@ -0,0 +1,891 @@
...
+        || (bin_ut1[t_index] > 0 && bin_ut1[t_index] != fv.bin_eop.t_ut1) // TODO: relax a bit?
+        || (bin_start_ERA_deg[t_index] < 0) || (bin_end_ERA_deg[t_index] < 0)
```
geoffryan: If you're checking for invalid values the other bad range is ERA > 360.

lib/stages/hdf5N2Write.cpp
```diff
@@ -0,0 +1,891 @@
...
+
+    // Mark (f, t) as added
+    size_t si = idx_ft(f_index, t_index);
+    added_ft[si] = 1;
+    ++added_count; // increment total number of frames added
+
+    return true;
+}
+
+std::optional<std::string> N2FileData::_get_final_filename() {
```
geoffryan, line 577: This function will not do what you expect and I don't think is quite the right thing to do regardless.

1. I think (and this is also what the CHIME spec does) we should label files by the start time of the acquistion, not the middle of the first frame (as bin_ut1 does). Then users are guaranteed the file only contains data from the file name time and after, instead of there being an awkward pipeline-dependent half-frame of data from before the labelled time.

2. Even if the data has been binned in UT1, it makes sense to label the files with the UTC of their acquisition.  That will be easier to compare to other logs, filesystem things, weather events, etc.

3. If we do want to label in UT1, this is not the way to do it.  Our internal UT1 epoch is roughly January 1, 2000, (precisely: 2451545 jd(ut1)) so these values are not compatible with gmtime or anything like that. Nor would we want them to be, UNIX time is deeply unserious.  If you want to get a hh/mm/ss from a ut1 value, you'll have to account for the epoch change.

lib/stages/hdf5N2Write.cpp
```diff
@@ -0,0 +1,891 @@
...
+            file_it = files.erase(file_it);
+        } else {
+            ++file_it;
+        }
+    }
+}
+
+bool hdf5N2Write::_finalfile_exists(std::uint64_t abs_file_idx,
+                                    const std::string& search_dir) const {
+    const std::string prefix = "vis_" + std::to_string(abs_file_idx) + "_";
```
geoffryan: Should this use a fixed width string for the abs_file_idx?  8 digits or so?

lib/stages/hdf5N2Write.cpp
```diff
@@ -0,0 +1,891 @@
...
+                       const size_t compression_level_, const bool use_bitshuffle_,
+                       const std::string base_dir_) :
+    num_elements(fv.num_elements), num_prod(fv.num_prod), num_ev(fv.num_ev),
+    num_file_f(Telescope::instance().cast<CHORDTelescope>().num_output_freq()),
+    num_file_t(num_file_t_), file_mode(file_mode_), blocksize_f(blocksize_f_),
+    blocksize_p(blocksize_p_), blocksize_t(blocksize_t_), compression(compression_),
+    compression_level(compression_level_), use_bitshuffle(use_bitshuffle_),
+    open_wall_s(open_wall_s_), last_update_wall_s(open_wall_s_), abs_file_idx(abs_file_idx_),
+    base_dir(std::move(base_dir_)),
+    partial_filepath(base_dir + "/.partial/" + "vis_" + std::to_string(abs_file_idx_) + ".h5"),
```
geoffryan, line 439: Should this use a minimum width string for abs_file_idx?  8 digits?

lib/utils/CHORDTelescope.hpp
```diff
@@ -573,30 +600,55 @@ class CHORDTelescope : public Telescope {
...
+    /// time0_ns is a UNIX timestamp, in nanoseconds. It does not include
+    /// leap seconds.
+    // TODO: Other fields like these should probably be const too
+    bool gps_enabled = false;
+    uint64_t time0_ns = 0;
+
+    const double _sampling_rate_MHz;
+    const uint64_t _fft_length;
+    const uint8_t _nyquist_zone;
+    // Set the physical frequency of id=0, and the spacing, taking into account aliasing of each
```
geoffryan: Should this comment be here? "Set the physical frequency..." It think it was copied out from the set_sampling_params function

lib/utils/CHORDTelescope.hpp
```diff
@@ -508,29 +544,20 @@ class CHORDTelescope : public Telescope {
 
 protected:
     /**
-     * @brief Set the internal parameters `dt_ns`, `freq0_MHz`, `df_MHz`, `nfreq_total`, and
-     * `ny_zone` which set the basic time and frequency sampling behaviour.  Reads the
-     * `sampling_rate_MHz`, `fft_length`, and `nyquist_zone` Config fields.
-     * @param config    The config.
-     * @param path      This telescope object's path.
-     */
-    void set_sampling_params(const kotekan::Config& config, const std::string& path);
```
geoffryan: There is enough important logic that happens in this function I think I'd prefer it remain, instead of getting lost in an initializer list.  Easier to document and inspect.

lib/utils/CHORDTelescope.cpp
```diff
@@ -31,21 +31,37 @@ using kotekan::connectionInstance;
...
+    _gps_endpoint(config.get_default<std::string>(path, "gps_endpoint", "/get-frame0-time")),
+    // Time and frequency sampling parameters
+    _sampling_rate_MHz(config.get_default<double>(path, "sampling_rate_MHz", 3.2e3)),
+    _fft_length(config.get_default<uint64_t>(path, "fft_length", 16384)),
+    _nyquist_zone(config.get_default<uint8_t>(path, "nyquist_zone", 1)),
+    _dt_ns(uint64_t(double(_fft_length) / (_sampling_rate_MHz * 1.0e6) * 1.0e9)),
+    _freq0_MHz((_nyquist_zone / 2) * _sampling_rate_MHz),
+    _df_MHz((_nyquist_zone % 2 == 1 ? 1 : -1) * _sampling_rate_MHz
+            / _fft_length), // Odd zones count up from freq0, even zones count down.
+    _nfreq_total(_fft_length / 2),
```
geoffryan: I think I'd prefer this (the dt_ns, freq0, df initialization) stay in a function, unless there is a clear benefit for them being const.  The math here is simple but important and in a function it is easier to document, more obvious for users, etc.  We only ever interact with the Telescope via const reference, so there are very limited mechanisms where these parameters could change.

lib/utils/CHORDTelescope.cpp
```diff
@@ -774,35 +774,59 @@ const struct dishInfo& CHORDTelescope::get_dish_at_idx(int64_t idx) const {
...
+    return _nfreq_total;
+}
+
+size_t CHORDTelescope::max_output_freq_id() const {
+    return (_max_output_freq_MHz - _freq0_MHz) / _df_MHz;
+}
+
+size_t CHORDTelescope::min_output_freq_id() const {
+    return (_min_output_freq_MHz - _freq0_MHz) / _df_MHz;
+}
```
geoffryan: Be careful of nyquist zone things, df_MHz can be negative.

geoffryan (changes requested):
A few minor changes.

lib/stages/hdf5N2Write.cpp
jbmertens: Ah, I thought of this as a more "casual", approximate value, needed just for a description and to avoid name conflicts.
Can definitely use the acquisition start time though!

geoffryan: Yeah, I'd search for the minimum `fpga_start_tick` and then just run `Telescope::instance().to_time()` on it to get the instrument time as a timespec. That's something you can give to gmtime() or whatever and its equal to UTC most of the time.

lib/stages/EigenVisIter.cpp
jbmertens: Punting this to a future/dedicated PR.

lib/stages/EigenVisIter.cpp
jbmertens: Punting this to a future/dedicated PR.

lib/utils/N2Util.hpp
jbmertens: (I agree, but am leaving this decision til later.)

config/ci-tests/standalone/test_pipeline.j2
```diff
@@ -1,6 +1,66 @@
...
+# CHORD telescope sampling setup (matches CHORDTelescope defaults)
+{% set sampling_rate_MHz = 3.2e3 %}
+{% set fft_length = 16384 %}
+{% set nyquist_zone = 1 %}
+{% set df_MHz = (sampling_rate_MHz / fft_length) * (1 if (nyquist_zone % 2) == 1 else -1) %}
+{% set freq0_MHz = (nyquist_zone // 2) * sampling_rate_MHz %}
+
+# Science band narrowed to just the simulated freq_ids above
+{% set science_min_freq_MHz = freq0_MHz + df_MHz * (freq_ids_list | first) %}
+{% set science_max_freq_MHz = freq0_MHz + df_MHz * (freq_ids_list | last) %}
```
geoffryan: For what its worth, its probably fine for this test to hardcode nyquist_zone = 1.  If you want to let it be free, you might want to change the min/max freq logic to make sure they correspond to the right ids. (Looks here like min_freq_MHz is always the smallest ID)

config/ci-tests/standalone/test_pipeline.j2
```diff
@@ -59,6 +119,9 @@ telescope:
     [151.0, 0.0, 0.0],
     [0.0, 204.0, 0.0],
```
geoffryan: `dish_positions` has been replaced by `dish_inputs`.  It will not be read, does not have to be here.

config/ci-tests/standalone/test_pipeline.j2
```diff
@@ -323,33 +420,18 @@ n2_accumulate:
...
-buffer_recv_n2:
-    log_level: debug
-    kotekan_stage: bufferRecv
-    buf: n2_recv_buffer
-    listen_port: 11025
 
 # Add eigen calculation
 
 eigencalc:
+  num_ev: 2
```
geoffryan: For what its worth, this could still be a global, so you don't need to set it (and the frame values) manually

config/ci-tests/standalone/test_pipeline.j2
```diff
@@ -129,16 +192,50 @@ host_counts_buffer:
...
+n2_recv_buffer_1:
+    kotekan_buffer: N2
+    num_ev: 2
+    vis_layout: "FullUpperTri"
+    num_frames: buffer_depth * num_local_freq
+    metadata_pool: n2_pool
+
+n2_recv_buffer_2:
+    kotekan_buffer: N2
+    num_ev: 2
```
geoffryan: could be global, or set from a global

config/ci-tests/standalone/test_pipeline.j2
```diff
@@ -129,16 +192,50 @@ host_counts_buffer:
...
+    vis_layout: "FullUpperTri"
+    num_frames: buffer_depth * num_local_freq
+    metadata_pool: n2_pool
+
+n2_merge_buffer:
     kotekan_buffer: N2
+    num_ev: 2
```
geoffryan: could be global, or set from a global

config/ci-tests/standalone/test_pipeline.j2
```diff
@@ -129,16 +192,50 @@ host_counts_buffer:
...
+n2_split_2:
+    kotekan_buffer: N2
+    num_ev: 2
+    vis_layout: "FullUpperTri"
+    num_frames: buffer_depth * num_local_freq
+    metadata_pool: n2_pool
+
+n2_recv_buffer_1:
+    kotekan_buffer: N2
+    num_ev: 2
```
geoffryan: could be global, or set from a global

config/ci-tests/standalone/test_pipeline.j2
```diff
@@ -129,16 +192,50 @@ host_counts_buffer:
...
+n2_split_1:
+    kotekan_buffer: N2
+    num_ev: 2
+    vis_layout: "FullUpperTri"
+    num_frames: buffer_depth * num_local_freq
+    metadata_pool: n2_pool
+
+n2_split_2:
+    kotekan_buffer: N2
+    num_ev: 2
```
geoffryan: could be global, or set from a global

config/ci-tests/standalone/test_pipeline.j2
```diff
@@ -129,16 +192,50 @@ host_counts_buffer:
...
     kotekan_buffer: N2
+    num_ev: 2
+    vis_layout: "FullUpperTri"
     num_frames: buffer_depth * num_local_freq
     metadata_pool: n2_pool
 
-n2_eigen_buffer:
+n2_split_1:
+    kotekan_buffer: N2
+    num_ev: 2
```
geoffryan: could be global, or set from a global

config/ci-tests/standalone/test_pipeline.j2
```diff
@@ -129,16 +192,50 @@ host_counts_buffer:
...
     kotekan_buffer: N2
+    num_ev: 0
+    vis_layout: "FullUpperTri"
     num_frames: buffer_depth * num_local_freq
     metadata_pool: n2_pool
 
-n2_recv_buffer:
+n2_eigen_buffer:
     kotekan_buffer: N2
+    num_ev: 2
```
geoffryan: could be global, or set from a global

config/ci-tests/standalone/test_pipeline.j2
```diff
@@ -14,36 +74,36 @@ type: config
...
+sub_integration_ntime: {{ sub_integration_ntime }}
+samples_per_data_set: {{ samples_per_data_set }}
 rfi_downsampling_factor: 256
 num_times: samples_per_data_set  # New N2K needs this.
 num_polarizations: 2
-num_dishes: 64
-num_elements: num_polarizations * num_dishes
-num_local_freq: 16
-
-num_ev: 4
```
geoffryan, line 36: Don't need to drop this, just need to set the ev-less frames with a local `num_ev: 0`.  Or could rename this parameter to something like `num_ev_to_calculate` and set the others from there so you only need to change it one place but are still explicit.

docs/sphinx/user/processes/hdf5N2Write.rst
```diff
@@ -0,0 +1,110 @@
...
+hdf5N2Write
+***********
+
+Buffered writer for CHORD/CHIME N2 visibilities.  Frames arrive from an
+``N2Buffer`` with ``N2Metadata``; the stage groups them by
+``abs_time_idx / num_file_t``, buffers a full ``nfreq × nt`` block in memory,
+and writes HDF5 files in one shot.  Missing frames stay zero-filled with
+``frames_added`` showing which (f, t) pairs were present and ``frac_lost=1`` for
+those gaps.  Work happens in ``<base_dir>/.partial/vis_<abs_idx>.h5`` and is
+renamed to ``vis_<abs_idx>_<YYYYMMDDThhmmss>_<nsec>.h5`` using the earliest
```
geoffryan, line 13: This file naming is different from the CHIME format.  I think that's fine, I prefer this, but a future upgrade might introduce an option for consistent CHIME naming.

docs/sphinx/user/processes/hdf5N2Write.rst
geoffryan, line 1: Hell yes docs.

lib/metadata/N2Metadata.hpp
```diff
@@ -64,8 +61,8 @@ struct N2MetadataFormat {
     // bin start/end info for convenience
     double bin_start_ERA_deg; /// Earth Rotation Angle at start of bin
     double bin_end_ERA_deg;   /// Earth Rotation Angle at end of bin
-    int64_t bin_start_LAST;   /// local apparent sidereal time (nanoseconds) at start of bin
-    int64_t bin_end_LAST;     /// local apparent sidereal time (nanoseconds) at end of bin
+    double bin_start_LAST;    /// local apparent sidereal time (nanoseconds) at start of bin
```
geoffryan, line 64: thanks!

lib/stages/hdf5N2Write.hpp
```diff
@@ -0,0 +1,307 @@
...
+        return num_file_t * (num_elements * f + i) + t; // (f, i, t)
+    }
+    /// Get the index for a (f, t) pair
+    inline size_t idx_ft(size_t f, size_t t) const {
+        return num_file_t * f + t; // (f, t)
+    }
+};
+
+/**
+ * @class hdf5N2Write
```
geoffryan, line 191: Add a sentence here about file names, partial and finalized.

lib/stages/hdf5N2Write.cpp
```diff
@@ -0,0 +1,1057 @@
...
+                fpga_start_tick_min = fpga_start_tick[t];
+            }
+        }
+    }
+    if (!fpga_start_tick_min.has_value())
+        return std::nullopt;
+
+    timespec earliest_fpga_tick_time = Telescope::instance().to_time(fpga_start_tick_min.value());
+
+    // Construct final filename based on earliest_ut1_ns
```
geoffryan: comment out of date (not using ut1)

lib/stages/hdf5N2Write.cpp
```diff
@@ -0,0 +1,1057 @@
...
+    size_t si = idx_ft(f_index, t_index);
+    added_ft[si] = 1;
+    ++added_count; // increment total number of frames added
+
+    return AddFrameStatus::Success;
+}
+
+std::optional<std::string> N2FileData::_get_final_filename() {
+
+    // Get the earliest (non-zero) time in the bin ut1 array
```
geoffryan: comment out of date (not using ut1)

geoffryan (approved):
A handful of small comments, mostly doc comments and cleaning up the standalone yaml a bit.  LGTM

config/ci-tests/standalone/test_pipeline.j2
jbmertens: It's different in different buffers. 
EigenVisIter could/should perhaps infer this from the buffers?

config/ci-tests/standalone/test_pipeline.j2
jbmertens: Leaving it per-stage, just being explicit for testing.

PR #1370, opened 2025-12-04
Rewrite FRB quantizing kernels

eschnett:


geoffryan (commented):
Most of the changes look fine, although I'm not in a place to judge the correctness of the new kernel.  A test so we trust this is correct (and don't break it in the future) would be good here.

lib/cuda/cudaQuantize.hpp
geoffryan, line 1: The documentation comment here should be updated with the required buffer shapes and config parameters (missing num_times, num_frequencies, num_times.  num_chunks is out of date).

lib/cuda/cudaQuantizeKernel.cu
geoffryan, line 1: Is this file still needed?  Are we keeping it to retain past behaviour?

geoffryan (changes requested):
The changes look reasonable, the docs need to be updated, and I see there's a cpu implementation!  It looks correct as far as I can tell, but we should have a CI test that compares the CPU and GPU outputs to ensure we don't regress.

lib/cuda/cudaQuantize.hpp
eschnett: Done.

lib/cuda/cudaQuantizeKernel.cu
eschnett: Removed.

eschnett:
@rhaas80 This is your CHIME 8-bit FRB beam quantizer! I am checking that the GPU and CPU implementations have the same behaviour. You might want to check that the CPU version does what it's supposed to do.

lib/cuda/testQuantizeKernel8.cpp
```diff
@@ -0,0 +1,267 @@
...
+                // Calculate expected result
+                //
+                // When comparing to the expected result, two things can go wrong:
+                // 1. The scale can be zero, or can be close to zero (`scale_may_collapse`)
+                // 2. The input can contain large numbers, or infinities, or not-a-numbers
+                //    (`may_overflow`)
+                //
+                // We need to handle both cases, and we need to be lenient when comparing when this
+                // happens. For example, there might be an overflow on the GPU, but not in the CPU
+                // implementation. This should not be flagged as error.
```
rhaas80: seems dicey. This way a simple way to pass all tests is to always produce an overflow on the GPU.

lib/cuda/testQuantizeKernel8.cpp
```diff
@@ -0,0 +1,267 @@
...
+                case 4:
+                    return 0.5f * x + 0.5f * x * x * x;
+                case 5:
+                    return 0.5f * x + 0.5f * x * x * x + 10 * (time % 23 == 0);
+                case 6:
+                    return 10000 * time; // large but not infinity
+                case 7:
+                    return 10000 * time; // some values are infinity
+                case 8:
+                    return time / 10000; // small, but 1/x is less than infinity
```
rhaas80: this is 0 until t>=10000, isn't it? integer division.

lib/cuda/testQuantizeKernel8.cpp
```diff
@@ -0,0 +1,267 @@
...
+                case 9:
+                    return time / 100000; // small, and some 1/x are infinity
```
rhaas80: this zero until t>=10000, isn't it? integer division.

lib/cuda/testQuantizeKernel8.cpp
```diff
@@ -0,0 +1,267 @@
...
+        };
+
+
+        ////////////////////////////////////////////////////////////////////////////////
+
+        // Set input
+        for (int beam = 0; beam < nbeams; ++beam) {
+            for (int freq = 0; freq < nfreqs; ++freq) {
+                for (int time = 0; time < ntimes; ++time) {
+                    const float x = f(beam, time % chunk_size);
```
rhaas80: some dependence on `freq` would be good to avoid the (unlikely) case of the output getting the frequency index wrong (or always reading from the `freq=0` input)

lib/cuda/testQuantizeKernel8.cpp
```diff
@@ -0,0 +1,267 @@
...
+                //
+                // We need to handle both cases, and we need to be lenient when comparing when this
+                // happens. For example, there might be an overflow on the GPU, but not in the CPU
+                // implementation. This should not be flagged as error.
+                float minval = +FLT_MAX, maxval = -FLT_MAX;
+                bool may_overflow = false;
+                for (int time = 0; time < chunk_size; ++time) {
+                    const float x = f(beam, time);
+                    minval = isnan(x) ? 0.0f / 0.0f : fmin(minval, x);
+                    maxval = isnan(x) ? 0.0f / 0.0f : fmax(maxval, x);
```
rhaas80: since `fmin` and `fmax` are special wrt to how NaN is handled, does this only set `minval` / `maxval` to `NaN` if the last input is a `NaN` (or at least the last ones)?

lib/cuda/testQuantizeKernel8.cpp
```diff
@@ -0,0 +1,267 @@
...
+                    // Allow the respective float16 calculations to overflow
+                    may_overflow |= !isfinite(x) || fabs(x) >= 2048.0f;
```
rhaas80: is 2048f a strict limit or a guess?

lib/cuda/testQuantizeKernel8.cpp
```diff
@@ -0,0 +1,267 @@
...
+                }
+                may_overflow |= !isfinite(minval) || fabs(minval) >= 2048.0f || !isfinite(maxval)
+                                || fabs(maxval) >= 2048.0f;
+                // Allow the scale to be inaccurate when all inputs are close together
+                const bool scale_may_collapse = fabs(maxval - minval) < epsilon;
+                const int uint8_range = 254; // avoid 0 and 255
+                float expected_scale = (maxval - minval) / uint8_range;
+                float expected_offset = minval - 0.5f * expected_scale;
```
rhaas80: `- 0.5f * expected_scale;`? I'd have expected something like `scale = (maxval - minval)/2` and then `offset = (maxval + minval)/2`. I guess I will need to see and compare to the old implementation.

config/fengine/include/frb_chime.j2
```diff
@@ -131,7 +130,7 @@ run_frb23:
         out_buffers:
             host_frb2_beams_buffer: host_frb2_beams_buffer
             host_frb3_beams_buffer: host_frb3_beams_buffer
-            host_frb3_beams_meanstd_buffer: host_frb3_beams_meanstd_buffer
+            host_frb3_beams_offsetscale_buffer: host_frb3_beams_offsetscale_buffer
```
rhaas80, line 134: this gives me infinite recursions when I try (and @andrerenard  says this is as it should be :-) )

config/fengine/include/frb_pathfinder.j2
```diff
@@ -361,7 +361,7 @@ run_frb23:
         out_buffers:
             host_frb2_beams_buffer: host_frb2_beams_buffer
             host_frb3_beams_buffer: host_frb3_beams_buffer
-            host_frb3_beams_meanstd_buffer: host_frb3_beams_meanstd_buffer
+            host_frb3_beams_offsetscale_buffer: host_frb3_beams_offsetscale_buffer
```
rhaas80, line 364: same comment about infinite recursion (and a couple other lines are similar)

lib/cuda/cudaQuantize4.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+    // These are the exact array sizes supported by the Cuda code
+    static constexpr int CHUNK_SIZE = 256;
+    static constexpr int FRAME_SIZE = 32;
+
+private:
+    const int _num_beams;
+    const int _num_frequencies;
+    const int _num_times;
+    const int64_t _num_chunks;
```
rhaas80, line 46: why `int64_t` here but just `int` above? `_num_beams * _num_frequencies * _num_times` can overflow 32bit I guess?

lib/cuda/cudaQuantize4.cpp
```diff
@@ -0,0 +1,172 @@
...
+    NDArrayBuffer<float16_t, 4> offsetscale_buffer;
+};
+
+REGISTER_CUDA_COMMAND(cudaQuantize4);
+
+cudaQuantize4::cudaQuantize4(Config& config, const std::string& unique_name,
+                             bufferContainer& host_buffers, cudaDeviceInterface& device, int inst) :
+    cudaCommand(config, unique_name, host_buffers, device, inst),
+    //
+    _num_beams(config.get<std::int64_t>(unique_name, "num_beams")),
```
rhaas80, line 69: the variable is an int, I would try and match variable and type of template

lib/cuda/cudaQuantize4.cpp
```diff
@@ -0,0 +1,172 @@
...
+    _num_frequencies(config.get<std::int64_t>(unique_name, "num_frequencies")),
```
rhaas80, line 70: variable is int

lib/cuda/cudaQuantize4.cpp
```diff
@@ -0,0 +1,172 @@
...
+    _num_times(config.get<std::int64_t>(unique_name, "num_times")),
```
rhaas80, line 71: variable is int

lib/cuda/cudaQuantize4.cpp
```diff
@@ -0,0 +1,172 @@
...
+        const std::array<std::ptrdiff_t, 4> offsetscale_lengths{_num_beams, _num_frequencies,
+                                                                _num_times / CHUNK_SIZE, 2};
+        const std::array<std::string, 4> offsetscale_dimnames{"R", "Fbar", "Ttilde256",
+                                                              "offset/scale"};
+        return NDArrayBuffer<float16_t, 4>(_gpu_mem_beams_offsetscale, "frb3_beams_offsetscale",
+                                           offsetscale_lengths, offsetscale_dimnames, *this);
+    }())
+//
+{
+    if (_num_chunks % FRAME_SIZE)
```
rhaas80, line 110: `!= 0` given that the assert above is `== 0`?

lib/cuda/cudaQuantize4.cpp
```diff
@@ -0,0 +1,172 @@
...
+    const auto& input_ndarray = input_buffer.get_ndarray();
+    auto& offsetscale_ndarray = offsetscale_buffer.get_ndarray();
+    auto& beam_ndarray = beam_buffer.get_ndarray();
+
+    const float16_t* const input = input_ndarray.data();
+    float16_t* const outputf = offsetscale_ndarray.data();
+    kotekan::int4x2_t* const outputi = beam_ndarray.data();
+    const int input_size1 = input_ndarray.extent(2);
+    const int input_size2 = input_ndarray.extent(1);
+    const int input_size3 = input_ndarray.extent(0);
```
rhaas80:
```
   const int input_size1 = input_ndarray.extent(2);
   const int input_size2 = input_ndarray.extent(1);
   const int input_size3 = input_ndarray.extent(0);
```

the inverted order is b/c of row vs column major array indexing?

lib/cuda/cudaQuantize8.cpp
```diff
@@ -0,0 +1,164 @@
...
+    cudaStream_t stream = device.getStream(cuda_stream_id);
+
+    gpu_quantize8(input, outputf, outputi, input_size1, input_size2, input_size3, input_stride2,
+                  input_stride3, outputf_size1, outputf_size2, outputf_size3, outputf_stride2,
+                  outputf_stride3, outputi_size1, outputi_size2, outputi_size3, outputi_stride2,
+                  outputi_stride3, stream);
+    CHECK_CUDA_ERROR(cudaGetLastError());
+
+    return record_end_event();
+}
```
rhaas80, line 169: all the same comments as for `cudaQuantize4.cpp` I guess

lib/cuda/cudaQuantizeKernel4.cu
```diff
@@ -0,0 +1,405 @@
...
+        // Avoid non-finite numbers
+        using std::isfinite;
+        if (!isfinite(__half2float(offseth)) || !isfinite(__half2float(scaleh))) {
+            // Offset or scale are non-finite
+            offseth = __float2half_rn(0.0f);
+            scaleh = __float2half_rn(0.0f);
+            inv_offseth = __float2half_rn(inv_offset);
+            inv_scaleh = __float2half_rn(inv_scale);
+        } else if (!isfinite(__half2float(inv_offseth)) || !isfinite(__half2float(inv_scaleh))) {
+            // Offset and scale are fine, but the inverse scale is non-finite. This means the scale
```
rhaas80, line 243: comment does not match code (`inv_offseth` could also be non-finite)

lib/cuda/cudaQuantizeKernel8.cu
```diff
@@ -0,0 +1,393 @@
...
+}
+
+////////////////////////////////////////////////////////////////////////////////
+
+// Kernels (called by the drivers below)
+
+// Quantize a "chunk". This is a CPU implementation that is
+// straightforward and thus "obviously correct". If the GPU version
+// produces a different result, then the GPU version is wrong.
+void cpu_quantize8_chunk(const __half* __restrict__ const input, __half* __restrict__ const outputf,
```
rhaas80, line 75: for CHIME `outputf` should be float to match https://github.com/CHIMEFRB/ch_frb_io/blob/1c283cda329c16cd04cf79d9d520a2076f748c9d/L0_L1_packet.hpp#L99:

```

    float32     scale[nbeam * nfreq_coarse];   // byte offset (24 + 2*nbeam + 2*nfreq_coarse)
    float32     offset[nbeam * nfreq_coarse];  // byte offset (24 + 2*nbeam + 2*nfreq_coarse + 2*nbeam*nfreq_coarse)
```

rhaas80: that is unless the final type conversion is done in the network sending stage, which is also fine.

lib/cuda/cudaQuantizeKernel8.cu
```diff
@@ -0,0 +1,393 @@
...
+    constexpr float outf_min = outi_min - 0.5f;
+    constexpr float outf_max = outi_max + 0.5f;
+    // We want this mapping:
+    //     minval => outf_min
+    //     maxval => outf_max
+    const float outf_range = outf_max - outf_min;
+    const float in_range = maxval - minval;
+
+    float scale = in_range / outf_range;
+    float offset = minval - outf_min * scale;
```
rhaas80, line 103: ok, that answers my previous question.

lib/cuda/cudaQuantizeKernel8.cu
```diff
@@ -0,0 +1,393 @@
...
+        // Avoid non-finite numbers
+        using std::isfinite;
+        if (!isfinite(__half2float(offseth)) || !isfinite(__half2float(scaleh))) {
+            // Offset or scale are non-finite
+            offseth = __float2half_rn(0.0f);
+            scaleh = __float2half_rn(0.0f);
+            inv_offseth = __float2half_rn(inv_offset);
+            inv_scaleh = __float2half_rn(inv_scale);
+        } else if (!isfinite(__half2float(inv_offseth)) || !isfinite(__half2float(inv_scaleh))) {
+            // Offset and scale are fine, but the inverse scale is non-finite. This means the scale
```
rhaas80, line 237: same comment as for cudaQuantizeKernel8

lib/cuda/testQuantizeKernel8.cpp
eschnett: No, overflows are only allowed when the numbers get large on the CPU. Otherwise overflows are flagged as errors. See how `may_overflow` is calculated.

We could add a new flags for tests that may never overflow.

lib/cuda/testQuantizeKernel4.cpp
```diff
@@ -0,0 +1,279 @@
...
+                case 4:
+                    return 0.5f * x + 0.5f * x * x * x;
+                case 5:
+                    return 0.5f * x + 0.5f * x * x * x + 10 * (time % 23 == 0);
+                case 6:
+                    return 10000 * time; // large but not infinity
+                case 7:
+                    return 10000 * time; // some values are infinity
+                case 8:
+                    return time / 10000; // small, but 1/x is less than infinity
```
rhaas80: integer division

lib/cuda/testQuantizeKernel4.cpp
```diff
@@ -0,0 +1,279 @@
...
+        };
+
+
+        ////////////////////////////////////////////////////////////////////////////////
+
+        // Set input
+        for (int beam = 0; beam < nbeams; ++beam) {
+            for (int freq = 0; freq < nfreqs; ++freq) {
+                for (int time = 0; time < ntimes; ++time) {
+                    const float x = f(beam, time % chunk_size);
```
rhaas80: could depend on freq for better testing of array accesses

rhaas80 (changes requested):
haven't tested the CPU implementation yet, but the CPU to GPU test has bugs as far as I can tell

lib/cuda/testQuantizeKernel8.cpp
eschnett: Done.

lib/cuda/testQuantizeKernel8.cpp
eschnett: D'oh.

lib/cuda/testQuantizeKernel8.cpp
eschnett: D'oh.

lib/cuda/testQuantizeKernel8.cpp
eschnett: D'oh.

lib/cuda/testQuantizeKernel8.cpp
eschnett: 2049 is the smallest integer that is not representable as float16. Apart from that this is just a guess.

lib/cuda/testQuantizeKernel4.cpp
eschnett: Corrected.

lib/cuda/testQuantizeKernel4.cpp
eschnett: Done.

lib/cuda/cudaQuantizeKernel8.cu
eschnett: I think this would be a good choice. The internal calculations happen in float16 anyway, so converting on the GPU would just increase transfer bandwidth.

rhaas80 (approved):
the identified bugs were fixed, my nagging comments were ignored :-) LGTM.

eschnett:
@geoffryan Can you re-review this PR, and approve it if you are happy?

geoffryan (approved):
LGTM!

lib/stages/calcFRB2Weights.cpp
```diff
@@ -20,8 +20,28 @@ class calcFRB2Weights : public kotekan::Stage {
...
 
     // FRB1 beamformer setup
-    const int num_dishes_x = config.get<int>(unique_name, "num_dishes_x");
-    const int num_dishes_y = config.get<int>(unique_name, "num_dishes_y");
+
+    // The directions we call "x" and "y" are really just describing
+    // the in-memory layout of the dishes (antennae) in the dish
+    // array. For CHORD, the physical x-direction (east-west) runs
+    // fastest, but for CHIME, the y-direction (north-south) runs
+    // fastest. This is necessary to get an efficient FRB1 beamformer
```
rhaas80, line 28: naturally :-) 

At least they both go east-west and not west-east I assume?

eschnett: CHORD does. Not 100% sure about CHIME.

rhaas80: I talked with people here. So far I am told that CHIME defines the positions of its feeds in terms of "Easting" (the x coordinate) and "Northing" (the y coordinate). Starting from the Weternmost-Southernmost corner of the array. Consecutive channel ids (which is how the feeds are orderd after the xpose2048 kernel) move North along a cylinder, first for the first polarization (east-west), then the second polarization (north-south), the to the next cylinder.

So given a channel id the x coordinate (Easting) is x0 + (id % 256) dx (dx positive ie moving East) and the y coordinate (Northing) is y0 + (id / 512) dy (dy positive ie moving North).

So for CHIME the direction which changes most quickly when changing feeds is "y" increasing towards the North.

Just in case this is at all helping reduce confusion :-)

PR #1372, opened 2025-12-04
Support new CHORD pathfinder dish layout

eschnett:


geoffryan (approved):
LGTM!  Assuming the auto-generated code is ok ;)

PR #1373, opened 2025-12-05
Factor CHORDTelescope data into objects

jbmertens:
Main changes:

 - Add "science" frequency range, relevant range for processing/output
 - Introduce structs for frequency, gps, and dish positioning parameters
 - make type handling more consistent/standard
 - const data const in the CHORDTelescope object
 - add tests for frequency and dish object population

Misc cleaning:
 - remove old fatal_error
 - change correlation frame name ([fix non-fatal test errors](https://github.com/kotekan/kotekan/actions/runs/19933193607/job/57150701301#step:8:416))
 - gpu ci compile on self hosted ([oom on github runners](https://github.com/kotekan/kotekan/actions/runs/19954199968/job/57225137767))
 - remove Astron correlator (dependence removed in #1371)


lib/metadata/chordMetadata.hpp
```diff
@@ -101,8 +102,8 @@ class chordMetadata : public metadataObject {
     // Dish layout
     int ndishes;                                  // number of dishes
     int n_dish_locations_ew, n_dish_locations_ns; // number of possible dish locations
-    int* dish_index; // [non-owning pointer] dish index for a possible dish location, or -1
-    int get_dish_index(int dish_loc_ew, int dish_loc_ns) const {
+    dish_index_t* dish_index; // [non-owning pointer] dish index for a possible dish location, or -1
+    dish_index_t get_dish_index(int dish_loc_ew, int dish_loc_ns) const {
```
geoffryan, line 106: For what its worth, these functions will be removed soon.  But right now, they are called in a lot of the auto-generated kernels and the result is stored in an `int`.

lib/stages/FEngine.cpp
geoffryan, line 1: Bunch of changes here, they seem innocuous, but a few data types are changing and I'm not sure how that interacts with the Julia-side of things.  Would be good to run the fengine tests and ensure they still work.

lib/stages/asdfFileRead.cpp
```diff
@@ -378,11 +378,12 @@ class asdfFileRead : public kotekan::Stage {
...
-                        new int[meta->n_dish_locations_ns * meta->n_dish_locations_ew];
+                        new dish_index_t[meta->n_dish_locations_ns * meta->n_dish_locations_ew];
                     const auto data = dish_index->get_data();
-                    assert(data->nbytes()
-                           == sizeof(int) * meta->n_dish_locations_ns * meta->n_dish_locations_ew);
-                    std::memcpy(meta->dish_index, data->ptr(), data->nbytes());
+                    const size_t bytes = sizeof(dish_index_t) * meta->n_dish_locations_ns
+                                         * meta->n_dish_locations_ew;
+                    assert(data->nbytes() == bytes);
+                    std::memcpy(meta->dish_index, data->ptr(), bytes);
```
geoffryan: feels better to leave this last argument as `data->nbytes()`?

lib/utils/CHORDTelescope.hpp
```diff
@@ -135,111 +159,264 @@ class dishGrid {
...
     }
 };
 
+/**
+ * @brief Struct containing dish parameters.
+ * @details This struct contains all the necessary parameters to describe the dish
+ * positions and orientations in a CHORD-like telescope. Generally, this is constructed
+ * from configuration parameters using the from_config() static method.
+ **/
+struct DishParams {
```
geoffryan: `DishParams` is not a good name for this.  `SiteParams` or `GeographicParams` would be better.  A lot of this (the site longitude and latitude, itrs <-> topocentric transformations) have nothing to do with the specific architecture of chord.

lib/utils/CHORDTelescope.hpp
```diff
@@ -135,111 +159,264 @@ class dishGrid {
...
+                                 size_t fft_length) noexcept;
+};
+
+/**
+ * @brief   Struct containing "GPS" data fields.
+ * @details This struct contains all the necessary parameters to describe the GPS
+ * time configuration in a CHORD-like telescope. Generally, this is constructed
+ * from configuration parameters using the from_config() static method.
+ **/
+struct GPSParams {
```
geoffryan: A better name would be `GPSTimeParams` (otherwise could be confused with where geographic information should be stored)

lib/utils/Telescope.hpp
```diff
@@ -144,7 +145,7 @@ class Telescope : public kotekan::kotekanLogging {
      *
      * @return  The number of frequencies on a stream.
      **/
-    virtual uint32_t num_freq_per_stream() const = 0;
```
geoffryan, line 147: Shouldn't these be the same type as `freq_id_t`?  That's what they'll be compared to.

lib/utils/Telescope.hpp
```diff
@@ -154,7 +155,7 @@ class Telescope : public kotekan::kotekanLogging {
      *
      * @return  The total number of frequency channels.
      **/
-    virtual uint32_t num_freq() const = 0;
+    virtual size_t num_freq() const = 0;
```
geoffryan, line 158: ditto.  Should be the same type as `freq_id_t` perhaps?

lib/utils/CHORDTelescope.hpp
```diff
@@ -135,111 +159,264 @@ class dishGrid {
...
+
+    /**
+     * @brief Build from kotekan::Config
+     *
+     * @conf   sampling_rate_MHz   double. ADC Sampling Rate (default: 3.2 GHz for CHORD)
+     * @conf   fft_length          size_t. F-engine FFT length (default: 16384 for CHORD)
+     * @conf   nyquist_zone        nyquist_zone_t.  Nyquist Zone we're operating in (default: 1 for
+     *CHORD)
+     * @conf   max_science_freq_MHz double. Maximum frequency used for science, in MHz (default:
+     *1500.0). Will be clamped to be inside the physical band.
```
geoffryan: This comment makes it sound like the `max_freq_MHz` will be clamped to be inside the physical band, but actually valid id's will be clamped to be inside `max_freq_MHz`.   Could you expand this comment to make it a little more clear?

lib/utils/CHORDTelescope.hpp
```diff
@@ -135,111 +159,264 @@ class dishGrid {
...
+     * @conf   min_science_freq_MHz double. Minimum frequency used for science, in MHz (default:
```
geoffryan, line 322: Ditto for comment clamping clarity.

lib/utils/CHORDTelescope.hpp
geoffryan: AH sorry I hadn't seen the initialization yet.  Maybe clarify that this will be clamped to possible values given the fpga parameters? and that the valid science ID's will then be strictly inside the (clamped) bounds.

lib/utils/CHORDTelescope.cpp
```diff
@@ -91,135 +69,345 @@ CHORDTelescope::CHORDTelescope(const kotekan::Config& config, const std::string&
...
+        config.get<std::string>(path, "updatable_config"),
+        std::bind(&CHORDTelescope::receive_eop_updates, this, _1));
+
+    restServer& rest_server = restServer::instance();
+
+    rest_server.register_get_callback(path + "/time0_ns",
+                                      std::bind(&CHORDTelescope::send_time0_ns, this, _1));
+    rest_server.register_get_callback(path + "/eop_table",
+                                      std::bind(&CHORDTelescope::send_eop_table, this, _1));
+}
```
geoffryan, line 406: WOW that's a short constructor.

lib/utils/CHORDTelescope.cpp
```diff
@@ -236,9 +424,10 @@ bool CHORDTelescope::receive_eop_updates(nlohmann::json& json) {
         }
 
         if (tmp_eop_table.empty()) {
-            FATAL_ERROR_NON_OO(
+            ERROR_NON_OO(
```
geoffryan: The telescope inherits from Logging, so I think it can use regular errors?

geoffryan (changes requested):
Looks good!  I think just:
- a couple naming changes (DishParams and GPSParams)
- double-checking if num_freq() and similar functions should return `size_t` or `freq_id_t` (or if just everything should be `u64` :p )
- double checking fengine still works after these changes.

lib/utils/Telescope.hpp
jbmertens: This is a count of things, rather than a logical id (frequency ids are not a reliable size/counter/index), so using different types made sense to me here.

geoffryan: It's a count but its also the maximum for some logical ids, same with `num_freq()` below.  So it may be better/simpler for it to just be compatible?  There's code like:  `for(freq_id_t f = 0; f < Telescope::instance().num_freq(); f++)`

jbmertens: Yeah, I do see that in a few chime-related places. But even for chime, the IDs aren't reliable as an index. I was wondering about renaming num_freq something like num_fpga_freq?

lib/stages/FEngine.cpp
jbmertens: Agreed, though the fengine code doesn't seem to run as-is on the chord branch. I'll aim to correct any issues this causes in #1295 .

lib/utils/Telescope.hpp
jbmertens: Going to leave this as-is for now, but will fix if this causes any issues.

geoffryan (approved):
LGTM (noting the fengine simulator appears to have already not been working, this should be addressed in a future PR)

PR #1375, opened 2025-12-11
Rhaas/update02

rhaas80:


config/ci-tests/cpu_batch/test_lostSamplesToPLMask.yaml
```diff
@@ -0,0 +1,98 @@
...
+    second_buf: out_pl_mask_buffer
+    num_frames_to_test: buffer_depth
+    trigger_exit_on_pass: false
+    check_metadata: true
+
+# dump to disk to inspect manually
+writers:
+    base_dir: test_lostSamplesToPLMask
+    prefix_hostname: false
+    max_frames: buffer_depth
```
jbmertens, line 77: Is this exit racy with more than one writer? If so, consider exiting with monitorBuffer.

lib/metadata/NDArray.cpp
```diff
@@ -81,4 +81,30 @@ std::shared_ptr<GenericNDArray> GenericNDArray::create(const DataType value_data
...
+        return false;
+    if (this->get_extents() != other.get_extents())
+        return false;
+    if (this->get_dimnames() != other.get_dimnames())
+        return false;
+    if (this->get_strides() != other.get_strides())
+        return false;
+
+    // curerntly all NDArrays have "simple" strides, ie there are no gaps in
+    // memory, os a bulk memory compare will do
```
jbmertens: Typo. Maybe worth issuing a warning for non-simple strides, as a future reminder

config/ci-tests/cpu_batch/test_lostSamplesToPLMask.yaml
rhaas80: I did not experience any issues with a race condition. Could be that there's one. Closest in spirit to this would be the f-engine which has one `max_frames` per hdf5FileWrite for the voltage buffers and none for the other output files. 

It's only for debugging though, and currently skipped. How would I set this up with monitorBuffer? I have only seen it to implement a timeout (which likely results in an error) in the N2Accumulate test code.

lib/metadata/NDArray.cpp
rhaas80: true.

rhaas80: the "simple-ness" is hard-coded in ndarray btw, ie it alwasy sets `stride[0]` to `1`.

config/ci-tests/cpu_batch/test_lostSamplesToPLMask.yaml
jbmertens: With max_frames, exit_kotekan() could be called while other stages are still writing, so the number of frames written isn't guaranteed to be consistent, at least to my understanding. Although this is probably ok just for tests, monitorBuffer may be a smidge more reliable if one of the writers is hanging? The idea is to generate a finite number of frames, and monitor (each) output buffer(s) for the expected number This pattern is in test_send_recv.yaml.

lib/stages/lostSamplesToPLMask.cpp
```diff
@@ -0,0 +1,234 @@
...
+        if (buf->num_frames != lost_samples_bufs.at(0)->num_frames)
+            ERROR("Input buffers have different number of frames: {:d} != {:d}", buf->num_frames,
+                  lost_samples_bufs.at(0)->num_frames);
+    }
+    const int num_freq_bins = int(lost_samples_bufs.size());
+
+    if (pl_mask_buf->frame_size
+        != lost_samples_bufs.at(0)->frame_size / PL_MASK_DOWNSAMPLING_FACTOR * NUM_DISHES
+               / PL_MASK_DISHES_PER_BIN * NUM_POLARIZATIONS / BITS_PER_BYTE * num_freq_bins)
+        ERROR("Unexpected frames sizes for pl_mask {:d} and lost_samples {:d}",
```
jbmertens: Should this be fatal?

config/ci-tests/cpu_batch/test_lostSamplesToPLMask.yaml
rhaas80: > With max_frames, exit_kotekan() could be called while other stages are still writing, so the number of frames written isn't guaranteed to be consistent, at least to my understanding. Although this is probably ok just for tests, monitorBuffer may be a smidge more reliable if one of the writers is hanging? The idea is to generate a finite number of frames, and monitor (each) output buffer(s) for the expected number This pattern is in test_send_recv.yaml.

This may not be the case. @eschnett may comment on this, but there's the global `waiting_for_max_frames` counter in `./lib/stages/waitingForMaxFrames.hpp` which each `hdf5FileWrite` instance increments in its constructor and decrements once it has reached its target number of frames. Only once that counter has reached 0 do they trigger kotekan termination.

Using `bufferMonitor` (which I have just tried) is a bit tricky since not only do I only want to terminate once all `hdf5FileWrite` have written their requested number of frames, I also do not want to write more frames, and there's (currently) on way to stop the producer produce more frames than the `hdf5FileWrite` are being asked to write.

So I think I won't need to modify the test yaml file. I have added code to check for simple strided-ness though.

lib/metadata/NDArray.cpp
rhaas80: Added a check in https://github.com/kotekan/kotekan/pull/1375/changes/b54024e3c4810bacb552d0cd7013491a28577554#diff-9526ed033276868bf38c2ae98013d25713f2dbb4f6f2d20a0ebe6bbfc96e6281R100

config/ci-tests/cpu_batch/test_lostSamplesToPLMask.yaml
jbmertens: Yes, looking closer, the config here looks safe. Iirc the issue that bit me was having writers both with and without max_frames.
The monitorBuffer stop is not ideal here without more changes, but does give a way to stop CI tests without needing to write anything.

geoffryan: the `waiting_for_max_frames` functionality is the right thing to do here I think.  @rhaas80 is correct, it should not cause a race condition, the global `waiting_for_max_frames` is updated atomically.

lib/testing/testLostSamplesToPLMask.cpp
```diff
@@ -0,0 +1,228 @@
...
+                            pl_mask_frame[pl_idx / BITS_PER_BYTE] |= (!lost)
+                                                                     << (pl_idx % BITS_PER_BYTE);
+                            pl_idx += 1;
+                        }
+
+        pl_mask_buf->allocate_new_metadata_object(frame_id);
+        auto pl_mask_meta = get_chord_metadata(pl_mask_buf, frame_id);
+
+        // physics metadata
+        // TODO: add more that dpdk adds
```
jbmertens, line 126: Should these TODOs be done before merging?

lib/stages/lostSamplesToPLMask.cpp
rhaas80: yes, I am used to `ERROR` being fatal. How likely is `FATAL_ERROR` to remain, given https://github.com/kotekan/kotekan/pull/1373/changes#diff-a8fd13e2c8140acf96872bed3f5b390bc3447b61636d9a5e86cf02ab6e0f1ca9L167 ?

Namely  git hash [17aa174f](https://github.com/kotekan/kotekan/commits/17aa174f6f6f07f2c2df61554068f668b350c137) "Factor CHORDTelescope data into objects (#1373)" of [kotekan](https://github.com/kotekan/kotekan) seems to remove `FATAL_ERROR_F` from `error.h`.

Hmm, it remains in `lib/core/kotekanLogging.hpp` though.

lib/testing/testLostSamplesToPLMask.cpp
rhaas80: Maybe but it works without so far, so I am in no rush. If I find something that fails, I will add them. This is only the test generator, so not crucial for actual functionality.

lib/stages/lostSamplesToPLMask.cpp
rhaas80: switched over to `FATAL_ERROR`.

jbmertens: That should've been it's own PR, or there should be another one for consistency. Per discussion about inconsistent exits, ERROR just logs an error. FATAL_ERROR throws, EXIT_ERROR hard exits.
_F functions were ~legacy C macros, I think should no longer be used and could be updated.

lib/testing/testLostSamplesToPLMask.cpp
```diff
@@ -0,0 +1,228 @@
...
+    pl_mask_buf->register_producer(unique_name);
+
+    // Register as producer for all lost samples buffers
+    json ls_bufs = config.get_value(unique_name, "lost_samples_buffers");
+    for (json::iterator it = ls_bufs.begin(); it != ls_bufs.end(); ++it) {
+        Buffer* buf = buffer_container.get_buffer(it.value());
+        lost_samples_bufs.push_back(buf);
+        buf->register_producer(unique_name);
+        if (buf->frame_size != lost_samples_bufs.at(0)->frame_size)
+            ERROR("Lost samples buffers have different frame sizes: {:d} != {:d}", buf->frame_size,
```
jbmertens: Should these other errors be fatal as well?

lib/stages/lostSamplesToPLMask.cpp
```diff
@@ -0,0 +1,234 @@
...
+            FATAL_ERROR("Input buffers have different frame sizes: {:d} != {:d}", buf->frame_size,
+                        lost_samples_bufs.at(0)->frame_size);
+        if (buf->num_frames != lost_samples_bufs.at(0)->num_frames)
+            FATAL_ERROR("Input buffers have different number of frames: {:d} != {:d}",
+                        buf->num_frames, lost_samples_bufs.at(0)->num_frames);
+    }
+    const int num_freq_bins = int(lost_samples_bufs.size());
+
+    if (pl_mask_buf->frame_size
+        != lost_samples_bufs.at(0)->frame_size / PL_MASK_DOWNSAMPLING_FACTOR * NUM_DISHES
```
jbmertens, line 64: This is hardcoded for chime, which is probably fine, but just checking in case there is a scenario this will be used that doesn't follow this. (Outriggers?)

jbmertens (approved):

lib/stages/lostSamplesToPLMask.cpp
rhaas80: hmm it actually should not be hard-coded here (though I would argue that hard-coding this stuff in `testLostSamplesToPLMask` wold be ok for the most part). The simplest solution is obviously to add a `num_dishes` configuration option, but I dislike having the user pass in basic information about the telescope.

`CHORDTelescope` has a `get_num_dishes()` member function, which unfortunately `ICETelescope` lacks (or any analogous function).

No idea if the Outriggers have a different physical layout.

For now, I would leave it as is and once we have a nice `Telescope` base class that hopefully gets some of the changes now in `CHORDTelescope` then I would update this code, just in case.

lib/testing/testLostSamplesToPLMask.cpp
rhaas80: yes

PR #1376, opened 2025-12-12
hdf5 cmake summary fix

jbmertens:
Only print HDF5 was found when it was found.


geoffryan (approved):
LGTM once tests pass!

PR #1377, opened 2025-12-13
writer and point source updates

geoffryan:
- Updates the `waiting_for_max_frames` mechanic in writers, value is only decremented if it was incremented first.
- Writers unregister as consumers on exit (unless they're the last consumer)
- PointSource visTestPattern has noise and frequency-dependent beam size.


lib/stages/asdfFileWrite.cpp
```diff
@@ -461,10 +461,16 @@ class asdfFileWrite : public kotekan::Stage {
                 break;
             }
         } // for
-
-        if (--waiting_for_max_frames == 0) {
-            WARN("Shutting down Kotekan");
-            exit_kotekan(CLEAN_EXIT);
+    
+        // Not quite thread-safe
+        if (buffer->get_num_consumers() > 1)
```
jbmertens: Good addition I think, but Is the destructor a better place for this?

lib/stages/hdf5N2Write.cpp
```diff
@@ -739,6 +742,9 @@ hdf5N2Write::hdf5N2Write(kotekan::Config& config, const std::string& unique_name
     if (buffer->buffer_type != "N2") {
         FATAL_ERROR("Input buffer must be a N2-type buffer.");
     }
+
+    if (max_frames >= 0)
+        ++waiting_for_max_frames;
```
jbmertens, line 751: Should we also issue a warning if max_frames is less than the expected number in a file?

lib/testing/FakeVisPattern.cpp
```diff
@@ -474,9 +475,16 @@ PointSourceVisPattern::PointSourceVisPattern(kotekan::Config& config, const std:
     stokes_Q = config.get_default<double>(path, "stokes_Q", 0.0);
     stokes_U = config.get_default<double>(path, "stokes_U", 0.0);
     stokes_V = config.get_default<double>(path, "stokes_V", 0.0);
+
     noise_var = config.get_default<double>(path, "noise_var", 0.01);
+    beam_fwhm_300MHz_deg = config.get_default<double>(path, "beam_FWHM_300MHz_deg", 10.0);
```
jbmertens, line 480: Looks good, let's also add some docs/comments in the header on config inputs, esp. noise and source properties.
(E.g. spectral index is for source vs white(?) noise, how are amplitudes related.)

jbmertens (approved):
Just some minor comments.

lib/stages/hdf5N2Write.cpp
geoffryan: I don't think so... could be a good way of testing what a file looks like that is not fully filled?

lib/stages/asdfFileWrite.cpp
geoffryan: Destructor is called too late, want this to let the rest of the pipeline run. Ended up editing the `unregister_consumer()` function on `Buffer` to take an optional flag.  If set, it doesn't unregister the last consumer. This happens inside the lock and should be thread safe.

lib/stages/hdf5N2Write.cpp
jbmertens: Just to be user-friendly, i.e. something that reminds a config writer what to expect.
I think the tests already cover partly-full writes sufficiently.

PR #1379, opened 2025-12-14
Correct F-Engine simulator

eschnett:


cmake/Toolchain.cmake
```diff
@@ -143,6 +143,10 @@ if(CMAKE_BUILD_TYPE MATCHES Debug OR CMAKE_BUILD_TYPE MATCHES Test)
     kmsg_status("Asserts enabled")
 endif()
 
+# Do not test for unused parameters
+# (The Julia header files define some functions with unused parameters.)
+add_compile_options($<$<OR:$<COMPILE_LANGUAGE:C>,$<COMPILE_LANGUAGE:CXX>>:-Wno-unused-parameter>)
```
jbmertens: If this is specific to Julia, can we try something more targeted? E.g. mark Julia's headers as system includes. Something like:
`set_property(DIRECTORY APPEND PROPERTY SYSTEM_INCLUDE_DIRECTORIES ${JULIA_INCLUDE_DIRS})`
In the FindJulia module, or similar (I haven't tested that this works).

lib/stages/FEngine.cpp
```diff
@@ -441,7 +429,8 @@ FEngine::~FEngine() {
 
 void FEngine::main_thread() {
     static bool stale = false;
-    assert(!stale);
+    if (stale)
```
jbmertens: Is this meant to ensure only one FEngine stage exists in a pipeline?
If so, could we issue a helpful error here in addition to the abort()?

jbmertens: Should it also be an std::atomic_bool?

lib/core/juliaManager.cpp
```diff
@@ -168,4 +170,21 @@ std::any juliaCallAny(const std::function<std::any()>& fun) {
...
+
+        // Call showerror(io, exc)
+        jl_function_t* showerror = jl_get_function(jl_base_module, "showerror");
+        jl_call2(showerror, jl_stderr_obj(), exc);
+        // jl_flush_cstdio();
+        jl_print_backtrace();
+
+        std::fflush(stdout);
+        std::fflush(stderr);
+        std::abort();
```
jbmertens, line 189: I'm not familiar with possible julia exceptions -- do we *always* want to abort?
Consider throwing here to give other stages/threads a chance to clean up.

lib/stages/FEngine.cpp
eschnett: This should never fail if Kotekan works correctly. `assert` would be the right choice, but if the `assert` is optimized away then we get a warning about an unused variable, and we treat this as error. Hence I'm using `if` and `abort` instead of `assert`. No warning is necessary.

lib/core/juliaManager.cpp
eschnett: We don't need to clean up. The usual exceptions come e.g. from syntax errors in the Julia module.

cmake/Toolchain.cmake
eschnett: I tried this
```
macro(_kotekan_detect_julia OUT_FOUND OUT_REASON)
    set(${OUT_FOUND} OFF)
    set(${OUT_REASON} "executable not found")

    find_program(_julia_exe julia)
    if(NOT _julia_exe)
        return()
    endif()

    execute_process(
        COMMAND "${_julia_exe}" --startup-file=no --version
        RESULT_VARIABLE _julia_rv
        OUTPUT_QUIET
        ERROR_QUIET)
    if(NOT _julia_rv EQUAL 0)
        set(${OUT_REASON} "executable not runnable")
        return()
    endif()

    find_package(Julia)
    if(Julia_FOUND)
        set(${OUT_FOUND} ON)
        set(${OUT_REASON} "found")
    else()
        set(${OUT_REASON} "not found")
    endif()

    #TODO
    if(Julia_FOUND)
        set_property(DIRECTORY APPEND PROPERTY SYSTEM_INCLUDE_DIRECTORIES ${JULIA_INCLUDE_DIRS})
    endif()
endmacro()
```
and this does not work. I do not know cmake well enough to debug this – can you help me?

lib/stages/FEngine.cpp
jbmertens: What is this intended to guard against though? The resulting behavior could be an unexpected exit for users/devs when multiple stages are declared are in configs, tests, or if it's called in daemon mode, so let's document intent at least. A user-facing warning seems like a good option.

cmake/Toolchain.cmake
jbmertens: The #pragmas look like a good solution.

lib/core/juliaManager.cpp
```diff
@@ -1,14 +1,19 @@
 #include "juliaManager.hpp"
 
-#include "errors.h"        // for INFO_F, DEBUG_F
-#include "julia_fasttls.h" // for JULIA_DEFINE_FAST_TLS
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-parameter"
```
rhaas80, line 4: with these, are the CMake options still required?

lib/core/juliaManager.cpp
```diff
@@ -168,4 +173,21 @@ std::any juliaCallAny(const std::function<std::any()>& fun) {
...
+void juliaHandleExceptions() {
+    jl_value_t* exc = jl_exception_occurred();
+    if (exc) {
+        ERROR_F("Caught Julia exception");
+
+        // Call showerror(io, exc)
+        jl_function_t* showerror = jl_get_function(jl_base_module, "showerror");
+        jl_call2(showerror, jl_stderr_obj(), exc);
+        // jl_flush_cstdio();
+        jl_print_backtrace();
```
rhaas80, line 185: that's just a Julia backtrace or also a backtrace in the C++ code up to `main()`? Admittedly right now there's exactly one Julia code so we kind of know which C++ code called Julia.

lib/stages/FEngine.cpp
```diff
@@ -414,19 +417,7 @@ FEngine::FEngine(kotekan::Config& config, const std::string& unique_name,
...
-                // Get Base.showerror
-                jl_function_t* showerror = jl_get_function(jl_base_module, "showerror");
-
-                // Call showerror(io, exc)
-                jl_call2(showerror, jl_stderr_obj(), exc);
-
-                // jl_flush_cstdio()
-                // jl_print_backtrace()
-            }
+            kotekan::juliaHandleExceptions();
```
rhaas80: Hmm, in this case the should be called "juliaHandlePossibleExceptions"? Since it also covers the case where no exception happened?

lib/stages/FEngine.cpp
rhaas80: it's still a race even with `std::atomic<bool>`. Two threads can enter the `if(stale)` check at the same time both thinking they are the first. You need a mutex here (to mimic what an `omp critical` would do) to protect both the read of `stale` an its reset against the race.

lib/stages/FEngine.cpp
```diff
@@ -515,8 +509,7 @@ void FEngine::main_thread() {
             assert(iargc == nargs);
             FEngine_setup = jl_call(setup, args, nargs);
             JL_GC_POP();
-            if (jl_exception_occurred())
-                FATAL_ERROR("Julia exception:\n{:s}", jl_typeof_str(jl_exception_occurred()));
+            kotekan::juliaHandleExceptions();
```
rhaas80: same comment as above.

lib/stages/FEngine.cpp
```diff
@@ -563,10 +556,9 @@ void FEngine::main_thread() {
                 args[2] = jl_box_int64(num_dishes);
                 args[3] = FEngine_setup;
                 jl_value_t* const res = jl_call(set_dish_positions, args, nargs);
-                if (jl_exception_occurred())
-                    FATAL_ERROR("Julia exception:\n{:s}", jl_typeof_str(jl_exception_occurred()));
-                assert(res);
                 JL_GC_POP();
+                kotekan::juliaHandleExceptions();
```
rhaas80: same as above

lib/stages/FEngine.cpp
```diff
@@ -782,10 +774,9 @@ void FEngine::main_thread() {
                 args[2] = jl_box_int64(bb_num_beams);
                 args[3] = FEngine_setup;
                 jl_value_t* const res = jl_call(set_bb_beam_positions, args, nargs);
-                if (jl_exception_occurred())
-                    FATAL_ERROR("Julia exception:\n{:s}", jl_typeof_str(jl_exception_occurred()));
-                assert(res);
                 JL_GC_POP();
+                kotekan::juliaHandleExceptions();
```
rhaas80: same as above

lib/stages/FEngine.cpp
```diff
@@ -876,10 +867,9 @@ void FEngine::main_thread() {
                 args[5] = jl_box_int64(num_frequencies);
                 args[6] = FEngine_setup;
                 jl_value_t* const res = jl_call(set_A, args, nargs);
-                if (jl_exception_occurred())
-                    FATAL_ERROR("Julia exception:\n{:s}", jl_typeof_str(jl_exception_occurred()));
-                assert(res);
                 JL_GC_POP();
+                kotekan::juliaHandleExceptions();
```
rhaas80: same as above

lib/stages/FEngine.cpp
```diff
@@ -1123,11 +1113,9 @@ void FEngine::main_thread() {
                     args[5] = jl_box_int64(num_local_channels * U);
                     args[6] = jl_box_int64(W1_frame_index + 1);
                     jl_value_t* const res = jl_call(set_W1, args, nargs);
-                    if (jl_exception_occurred())
-                        FATAL_ERROR("Julia exception:\n{:s}",
-                                    jl_typeof_str(jl_exception_occurred()));
-                    assert(res);
                     JL_GC_POP();
+                    kotekan::juliaHandleExceptions();
```
rhaas80: same as above

lib/stages/FEngine.cpp
```diff
@@ -1451,11 +1439,9 @@ void FEngine::main_thread() {
                             args[6] = FEngine_setup;
                             args[7] = jl_box_int64(E_frame_index % num_frames + 1);
                             jl_value_t* const res = jl_call(set_E, args, nargs);
-                            if (jl_exception_occurred())
-                                FATAL_ERROR("Julia exception:\n{:s}",
-                                            jl_typeof_str(jl_exception_occurred()));
-                            assert(res);
                             JL_GC_POP();
+                            kotekan::juliaHandleExceptions();
```
rhaas80: same as above

lib/stages/FEngine.cpp
```diff
@@ -1740,10 +1726,9 @@ void FEngine::main_thread() {
                     args[5] = jl_box_int64(bb_num_beams);
                     args[6] = jl_box_int64(J_frame_index + 1);
                     jl_value_t* const res = jl_call(set_J, args, nargs);
-                    if (jl_exception_occurred())
-                        FATAL_ERROR("Julia exception:\n{:s}", jl_typeof_str(jl_exception_occurred()));
-                    assert(res);
                     JL_GC_POP();
+                    kotekan::juliaHandleExceptions();
```
rhaas80: same as above

lib/stages/FEngine.cpp
```diff
@@ -1826,10 +1811,9 @@ void FEngine::main_thread() {
                         args[5] = jl_box_int64(num_frequencies * U);
                         args[6] = jl_box_int64(I1_frame_index + 1);
                         jl_value_t* const res = jl_call(set_I, args, nargs);
-                        if (jl_exception_occurred())
-                            FATAL_ERROR("Julia exception:\n{:s}", jl_typeof_str(jl_exception_occurred()));
-                        assert(res);
                         JL_GC_POP();
+                        kotekan::juliaHandleExceptions();
```
rhaas80: same as above

rhaas80 (dismissed):

lib/stages/FEngine.cpp
eschnett: I'll remove the test.

lib/core/juliaManager.cpp
eschnett: No.

lib/core/juliaManager.cpp
eschnett: Just the Julia code. The point is to help debug Julia code.

lib/stages/FEngine.cpp
eschnett: It handles all the Julia exception that there are.

rhaas80: It also handles the "there is no exception" case though, which at least to me, was surprising.

geoffryan (approved):
Exception handling function has a good name, LGTM

PR #1381, opened 2025-12-15
Updates for system installs

jbmertens:
Just some library linking and tests for system installs.

### DIFF
--- a/config/CMakeLists.txt
+++ b/config/CMakeLists.txt
@@ -1,9 +1,5 @@
-# Installs all the .yaml files to /etc/kotekan/
-
-file(
-    GLOB configFiles
-    RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}"
-    "${CMAKE_CURRENT_SOURCE_DIR}/*.yaml")
-
-install(DIRECTORY DESTINATION /etc/kotekan/)
-install(FILES ${configFiles} DESTINATION /etc/kotekan/)
+# Install the entire config tree (YAMLs, Jinja templates, examples, tests) to /etc/kotekan/.
+install(
+    DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/"
+    DESTINATION /etc/kotekan
+    PATTERN "CMakeLists.txt" EXCLUDE)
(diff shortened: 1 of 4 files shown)


config/CMakeLists.txt
```diff
@@ -1,9 +1,5 @@
-# Installs all the .yaml files to /etc/kotekan/
-
-file(
-    GLOB configFiles
-    RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}"
-    "${CMAKE_CURRENT_SOURCE_DIR}/*.yaml")
-
-install(DIRECTORY DESTINATION /etc/kotekan/)
-install(FILES ${configFiles} DESTINATION /etc/kotekan/)
+# Install the entire config tree (YAMLs, Jinja templates, examples, tests) to /etc/kotekan/.
```
andrerenard, line 1: Hmm, I wonder if we should be more selective here?  There is a lot of stuff here that won't get used in production type runs where we'd do an install.

external/n2k/CMakeLists.txt
```diff
@@ -82,6 +82,9 @@ target_link_libraries(n2k gputils)
 target_include_directories(n2k PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/include")
 target_include_directories(n2k PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}")
 
+# Install the shared library so installed kotekan can find libn2k at runtime.
```
andrerenard, line 85: Do we also need Kendrick's gpuutils?

kotekan/CMakeLists.txt
```diff
@@ -33,6 +33,18 @@ else()
     message(FATAL_ERROR "System type ${CMAKE_SYSTEM_NAME} not supported")
 endif()
 
+if(Julia_FOUND)
```
andrerenard, line 36: I don't think there is any harm in having this, but we don't actually use Julia for any production operations of kotekan.

andrerenard (approved):

kotekan/CMakeLists.txt
jbmertens: Indeed. Though, Julia is on by default, so will be in release builds unless we change that.

external/n2k/CMakeLists.txt
jbmertens: This seems to be sufficient - I get no warnings or errors when running an installed kotekan with gpu configs.

config/CMakeLists.txt
jbmertens: I'm not opposed to including tests in actual installs, but maybe we should be more selective, yeah. A "release" directory? Will ask for opinions tomorrow but merge for now.

PR #1382, opened 2025-12-15
changes to ringbuffer and cuda RFI code to attach to dpdk

rhaas80:
DPDK does not always start its sequences with 0 since it may have dropped network packets trying to synchronize threads, so `sample0_offset` may not start out as `0`. Similarly dpdk currently does not set any upsampling factors for the raw input data, which trips up the first kurtosis kernel.

The changes to the RingBuffer code should be fine (and match what `cudaCopyNToRingBuffer` does), the change to `cudaRFIS012` is a bit more questionable, since it somehow just kicks the can down the road and if one ever had a situation where another cuda kernel that expects upsampling factors present is called first after dpdk then it would also fail.

A "cleaner" solution may be to have dpdk set the upsampling factors to 1 itself, even though this makes it set "unused" metadata (there's no upsampling yet at that point in the pipeline).

### DIFF
--- a/lib/cuda/cudaCopyFromRingbuffer.hpp
+++ b/lib/cuda/cudaCopyFromRingbuffer.hpp
@@ -52,6 +52,7 @@ private:
     size_t _ring_buffer_size;
 
     size_t input_cursor;
+    int64_t initial_sample0_offset; // sample0_offset of first frame, used for consitency checking
 
     /// GPU side memory name for the time-stream input
     std::string _gpu_mem_input;
--- a/lib/cuda/cudaCopyToRingbuffer.cpp
+++ b/lib/cuda/cudaCopyToRingbuffer.cpp
@@ -28,7 +28,7 @@ cudaCopyToRingbuffer::cudaCopyToRingbuffer(Config& config, const std::string& un
                                            cudaDeviceInterface& device, int instance_num) :
     cudaCommand(config, unique_name, host_buffers, device, instance_num, no_cuda_command_state,
                 "cudaCopyToRingbuffer", ""),
-    output_cursor(0) {
+    output_cursor(0), initial_sample0_offset(-1) {
     _input_size = config.get<size_t>(unique_name, "input_size");
     _ring_buffer_size = config.get<size_t>(unique_name, "ring_buffer_size");
     _gpu_mem_output = config.get<std::string>(unique_name, "gpu_mem_output");
@@ -156,7 +156,10 @@ cudaEvent_t cudaCopyToRingbuffer::execute(cudaPipelineState& pipestate,
         meta->set_sample0_offset(meta->get_sample0_offset()
                                  - output_cursor * meta->get_offset_downsampling()
                                        / meta->sample_bytes());
-        assert(meta->get_sample0_offset() == 0);
+        if (initial_sample0_offset == -1)
+            initial_sample0_offset = meta->get_sample0_offset();
+        else
+            assert(meta->get_sample0_offset() == initial_sample0_offset);
         assert(meta->dims > 0);
         assert(in_buffer->frame_size % meta->sample_bytes() == 0);
         assert(meta->dim[0] == std::ptrdiff_t(in_buffer->frame_size / meta->sample_bytes()));
--- a/lib/cuda/cudaRFIS012.cpp
+++ b/lib/cuda/cudaRFIS012.cpp
@@ -191,8 +191,16 @@ cudaEvent_t cudaRFIS012::execute(cudaPipelineState& /*pipestate*/,
 
     rfi_S012.set_metadata(voltage.get_metadata());
     const auto& rfi_S012_meta = rfi_S012.get_metadata();
-    auto freq_upchan_factor = rfi_S012_meta->get_freq_upchan_factor();
-    auto time_downsampling_fpga = rfi_S012_meta->get_time_downsampling_fpga();
+    std::vector<int> freq_upchan_factor;
+    if (rfi_S012_meta->has_freq_upchan_factor())
+        freq_upchan_factor = rfi_S012_meta->get_freq_upchan_factor();
+    else
+        freq_upchan_factor.assign(rfi_S012_meta->get_nfreq(), 1);
+    std::vector<int> time_downsampling_fpga;
+    if (rfi_S012_meta->has_time_downsampling_fpga())
+        time_downsampling_fpga = rfi_S012_meta->get_time_downsampling_fpga();
+    else
+        time_downsampling_fpga.assign(rfi_S012_meta->get_nfreq(), 1);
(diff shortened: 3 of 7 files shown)


geoffryan:
upsampling = 1 should be the default value of upsampling factors (denotes that no upsampling has taken place), it makes sense (to me) for dpdk to set that.

andrerenard:
Just a note on what DPDK is doing.   The offset is not non-zero because of lost packets/synchronization, but because the F-engine's clock starts at an FPGA seq number of 0, and then increases continuously.  So unless the X-engine is started before that first packet is generated, the offset will always be non-zero.  However the process of synchronizing the threads does insure that frames are always generate starting at a multiple of the `alignment` value.   Even if some packets are missing when it performs the alignment, the packets are aligned in the frames and missing values are zeroed.   Additionally in each buffer the first frame has the same offset within a given system (or it shuts down, which is the process the process I need to make a little more robust to random startup delays in the system...). 

Having said all this, I think the way that the seq number is tracked (important for timing keeping) might have changed in the ring buffer.  I need to take another look at that, since there were plans to change that after I wrote that CHIME code.   If the time offset (seq number) is stored differently now, then setting the frame offset to zero at startup would be fine, since DPDK requires all frames be aligned to the same edge (even with packet losses).

lib/cuda/cudaCopyFromRingbuffer.hpp
```diff
@@ -52,6 +52,7 @@ class cudaCopyFromRingbuffer : public cudaCommand {
     size_t _ring_buffer_size;
 
     size_t input_cursor;
+    int64_t initial_sample0_offset; // sample0_offset of first frame, used for consitency checking
```
eschnett, line 55: typo "consitency"

lib/cuda/cudaCopyToRingbuffer.cpp
```diff
@@ -156,7 +156,10 @@ cudaEvent_t cudaCopyToRingbuffer::execute(cudaPipelineState& pipestate,
         meta->set_sample0_offset(meta->get_sample0_offset()
                                  - output_cursor * meta->get_offset_downsampling()
                                        / meta->sample_bytes());
-        assert(meta->get_sample0_offset() == 0);
+        if (initial_sample0_offset == -1)
+            initial_sample0_offset = meta->get_sample0_offset();
+        else
+            assert(meta->get_sample0_offset() == initial_sample0_offset);
```
eschnett, line 162: For regular buffers, `sample0_offset` changes with every frame. It isn't supposed to stay the same.

lib/cuda/cudaRFIS012.cpp
```diff
@@ -191,8 +191,16 @@ cudaEvent_t cudaRFIS012::execute(cudaPipelineState& /*pipestate*/,
 
     rfi_S012.set_metadata(voltage.get_metadata());
     const auto& rfi_S012_meta = rfi_S012.get_metadata();
-    auto freq_upchan_factor = rfi_S012_meta->get_freq_upchan_factor();
-    auto time_downsampling_fpga = rfi_S012_meta->get_time_downsampling_fpga();
+    std::vector<int> freq_upchan_factor;
+    if (rfi_S012_meta->has_freq_upchan_factor())
```
eschnett, line 195: There should always be a frequency upchannelization factor. It would always be 1 for this kernel, but it needs to be there.

lib/cuda/cudaRFIS012.cpp
```diff
@@ -191,8 +191,16 @@ cudaEvent_t cudaRFIS012::execute(cudaPipelineState& /*pipestate*/,
...
+        freq_upchan_factor = rfi_S012_meta->get_freq_upchan_factor();
+    else
+        freq_upchan_factor.assign(rfi_S012_meta->get_nfreq(), 1);
+    std::vector<int> time_downsampling_fpga;
+    if (rfi_S012_meta->has_time_downsampling_fpga())
```
eschnett, line 200: Same here, this attribute should always be there.

eschnett (changes requested):
.

lib/cuda/cudaCopyToRingbuffer.cpp
rhaas80: If I recall correctly then `meta` here is the `chordMetadata` associated with the RingBuffer. Do those count as regular buffers? 

This seems likely to all be updated soon given that conversation in the kotekan / CHORD call just now. Plus at least the change to the SK102 kernel will be made superfluous by changes in dpdk. I''ll pause this pull request and re-open once the updates have settled in place.

lib/cuda/cudaCopyFromRingbuffer.hpp
rhaas80: thanks. Will need to see how to enable (Canadian, if I am doing so anyway) spellchecking in my editor.

lib/cuda/cudaRFIS012.cpp
rhaas80: We will need to decide on a list of required attributes. Right now it seems that these ones need to be set:

```
coarse_freq
freq_upchan_factor
half_fpga_sample0
time_downsampling_fpga
offset_downsampling
```

in addition to the ones already being set by dpdk's IceBoardShuffle handler.

lib/cuda/cudaCopyToRingbuffer.cpp
eschnett: Yes, a ring buffer has the same metadata as a regular buffer, except that `sample0_offset` is not yet.

lib/cuda/cudaRFIS012.cpp
eschnett: Good idea. Maybe it's time to define a new metadata class? I was going to clean up `chordMetadata` by removing all cruft, but I now realize that I don't know what is cruft and what is essential for some other parts of Kotekan I haven't seen.

lib/cuda/cudaCopyToRingbuffer.cpp
rhaas80: right now `sample0_offset` of the ring-buffers remains constant though. Only updated by `cudaCopy...` routines and those set it once and never update. It's the offset of the first data item ever written into the buffer. So the ring-buffer's cursor is an offset counting from that origin.

lib/cuda/cudaRFIS012.cpp
rhaas80: I would avoid having CHORMetadataN with N incremented each time we conceive of an improvement. Instead I would suggest to go the internet-rfc way and first find an implementation that works, then standardize it. Removing unused items would certainly be a good first step for this.

If this ends up being essentially a flat struct again (and no dictionary in the backend) then so be it.

Right now it would be good if the various components (stages) agreed on what to do with the current metatada.

rhaas80:
Superseded by #1386 and #1384

PR #1384, opened 2025-12-17
Handle `fpga seq num` correctly

eschnett:


lib/cuda/cudaCopyToRingbuffer.cpp
```diff
@@ -151,6 +151,16 @@ cudaEvent_t cudaCopyToRingbuffer::execute(cudaPipelineState& pipestate,
...
         meta = tmp;
+
+        // TODO: Don't set the metadata every time (to, hopefully, the same values). Set them once
+        // in the beginning only.
+        assert(output_cursor % meta->sample_bytes() == 0);
+        meta->set_fpga_seq_num(meta->get_fpga_seq_num()
+                               - meta->get_time_downsampling_fpga().at(0)
+                                     * (output_cursor / meta->sample_bytes()));
+        // This should hold most of the time. It will be wrong when we start Kotean with a non-zero
+        // `fpga_seq_num`.
```
rhaas80: we do. This will fail, pretty much every single time I start kotekan on the test node since the synchronization that dpdk does will mean that the very first `fpga_seq_num` passed to kotekan will not be zero (in fact it's rather large for me).

lib/cuda/cudaCopyFromRingbuffer.cpp
```diff
@@ -126,9 +126,15 @@ cudaEvent_t cudaCopyFromRingbuffer::execute(cudaPipelineState& pipestate,
     auto tmp = std::make_shared<chordMetadata>();
     tmp->deepCopy(meta);
     meta = tmp;
+    assert(meta->get_fpga_seq_num() == 0);
     assert(meta->get_sample0_offset() == 0);
     assert(input_cursor % meta->sample_bytes() == 0);
-    meta->set_sample0_offset(meta->get_sample0_offset() + input_cursor / meta->sample_bytes());
+    meta->set_fpga_seq_num(meta->get_fpga_seq_num()
+                           + meta->get_time_downsampling_fpga().at(0)
```
geoffryan: `time_downsampling_fpga` is frequency-dependent, but `fpga_seq_num` is not.  Should the factor here be a new scalar variable like `fpga_seq_stride` or something similar?

lib/cuda/cudaCopyToRingbuffer.cpp
```diff
@@ -151,6 +151,16 @@ cudaEvent_t cudaCopyToRingbuffer::execute(cudaPipelineState& pipestate,
         auto tmp = std::make_shared<chordMetadata>();
         tmp->deepCopy(meta);
         meta = tmp;
+
+        // TODO: Don't set the metadata every time (to, hopefully, the same values). Set them once
+        // in the beginning only.
+        assert(output_cursor % meta->sample_bytes() == 0);
+        meta->set_fpga_seq_num(meta->get_fpga_seq_num()
+                               - meta->get_time_downsampling_fpga().at(0)
```
geoffryan: same as above

lib/stages/hdf5FileRead.cpp
```diff
@@ -156,8 +156,12 @@ class hdf5FileRead : public kotekan::Stage {
                 }
                 meta->offset = 0;
 
+                if (dataset.hasAttribute("fpga_seq_num"))
+                    meta->set_fpga_seq_num(
+                        dataset.getAttribute("fpga_seq_num").read<std::int64_t>());
                 if (dataset.hasAttribute("sample0_offset"))
-                    meta->set_sample0_offset(dataset.getAttribute("sample0_offset").read<int>());
+                    meta->set_sample0_offset(
+                        dataset.getAttribute("sample0_offset").read<std::int64_t>());
```
rhaas80: 👍

lib/cuda/cudaCorrelator.cpp
```diff
@@ -120,45 +120,45 @@ cudaEvent_t cudaCorrelator::execute(cudaPipelineState&, const std::vector<cudaEv
...
     const std::shared_ptr<const chordMetadata> rfi_meta = rfi_RFImask.get_metadata();
-    const std::shared_ptr<chordMetadata> out_meta = n2k_correlation.get_metadata();
+    const std::shared_ptr<chordMetadata> n2k_corr_meta = n2k_correlation.get_metadata();
 
     // The input ringbuffers do not contain time-dependent data,
     // so we must reconstruct it here. (fpga_seq_num and sample0_offset)
-    // TODO: do this automatically in `NDArrayRingBuffer`
+    n2k_corr_meta->set_fpga_seq_num(voltage_meta->get_fpga_seq_num()
+                                    + voltage.get_read_valid().begin()
+                                          * voltage_meta->get_time_downsampling_fpga().at(0));
```
geoffryan: same as above

lib/cuda/cudaPL1bitCorrelator.cpp
```diff
@@ -198,36 +198,33 @@ cudaEvent_t cudaPL1bitCorrelator::execute(cudaPipelineState& /*pipestate*/,
...
-
-    // The RFI mask has a fast time index of size 1024, so has an apparent
-    // time downsampling of 1024. To get the needed outgoing fpga_seq_num,
-    // we need to undo that.
-    out_meta->set_fpga_seq_num(1024 * rfi_RFImask.get_read_valid().begin());
-    out_meta->set_sample0_offset(
-        div_noremainder(out_meta->get_fpga_seq_num(), n2k_sub_integration_ntime));
+    n2k_counts_meta->set_fpga_seq_num(rfi_meta->get_fpga_seq_num()
+                                      + rfi_RFImask.get_read_valid().begin()
+                                            * rfi_meta->get_time_downsampling_fpga().at(0));
```
geoffryan: same as above

rhaas80 (changes requested):
Nothing wrong is being added, but a bug persists, namely the assumption that `fpga_seq_num` starts out as 0, which is not true.

lib/cuda/cudaCopyFromRingbuffer.cpp
```diff
@@ -126,9 +126,15 @@ cudaEvent_t cudaCopyFromRingbuffer::execute(cudaPipelineState& pipestate,
...
+                                 * (input_cursor / meta->sample_bytes()));
+    meta->set_sample0_offset(meta->get_sample0_offset()
+                             + meta->get_time_downsampling_fpga().at(0)
```
geoffryan: Should this include `offset_downsampling`?

lib/testing/gpuSimulateN2kCorr.cpp
```diff
@@ -285,6 +285,7 @@ void gpuSimulateN2kCorr::main_thread() {
         /* test that things are consistent */
         meta_out->check_frame_desc(output_buf->get_frame_desc());
 
+        // This looks inconsistent
         meta_out->set_fpga_seq_num(meta_in->get_fpga_seq_num());
         meta_out->set_sample0_offset(meta_in->get_sample0_offset() / _sub_integration_ntime);
```
rhaas80: [edited]

It's almost consistent, since: `T_actual = sample0_offset / time_downsampling_fpga` (at least at index T=0) so to get the same `T_actual` when changing `t_downsampling_fpga` by a factor of `_sub_integration_ntime` this needs to be compensated though I would think using a `*` as well if a `*` is used for `time_downsampling_fpga`.

rhaas80: may resolve iself though: https://chime-experiment.slack.com/archives/CBXE5SMNK/p1766013754753409

lib/cuda/cudaCopyFromRingbuffer.cpp
rhaas80: this is not done quite consistently, sometimes it does `fpga_seq_num = time_downsamplng_fpga * (input_cursor / sample_bytes)` like here, and sometimes it misses the `time_downsampling_fpga` like in https://github.com/kotekan/kotekan/pull/1384/changes#diff-31ced5ba1ee248fe32f1668ff1ffa2fc6abaf6752420acd4bb05754974a71e4dR233

lib/testing/gpuSimulateN2kCorr.cpp
geoffryan: Can't see that slack link, but I can say when I wrote that code it was necessary for sample0_offset to explicitly have the downsampling factor, it was treated differently than an fpga tick.  This stage simply replicated what `cudaCorrelator` output, which was governed by the checks in copying data off the GPU.    However, this seems to no longer be the case!  Not sure when that changed.  

Likely a moot point since sample0_offset is being removed :)

lib/cuda/cudaCopyFromRingbuffer.cpp
eschnett: Yes. I changed this.

lib/cuda/cudaCopyFromRingbuffer.cpp
eschnett: That attribute has been removed.

lib/cuda/cudaCopyToRingbuffer.cpp
eschnett: Fixed.

lib/testing/gpuSimulateN2kCorr.cpp
eschnett: I went through all (?) the code setting sequence numbers and downsampling factors. I think it's correct now.

lib/cuda/cudaCopyFromRingbuffer.cpp
eschnett: I think I fixed all of these.

lib/cuda/cudaCopyNToRingbuffer.cpp
```diff
@@ -184,8 +181,8 @@ cudaEvent_t cudaCopyNToRingbuffer::execute(cudaPipelineState& /*pipestate*/,
                     "cudaCopyNToRingbuffer: input buffer has no chordMetadata");
             // Set the frequency for each of the input buffers
             coarse_freq.at(i) = meta_in->get_coarse_freq()[0];
-            // Check that the sample0 matches for all input buffers
-            assert(meta_ring->get_sample0_offset() == meta_in->get_sample0_offset());
+            // Check that the seq_num and sample0 matches for all input buffers
```
rhaas80, line 184: sample0 no longer exists

lib/cuda/cudaFRBBeamReformer.cpp
```diff
@@ -227,9 +221,10 @@ cudaEvent_t cudaFRBBeamReformer::execute(cudaPipelineState&, const std::vector<c
         DEBUG("Set output metadata: array shape {:s}, array type {:s}",
               out_meta->get_dimensions_string(), out_meta->get_type_string());
 
-        // Since we do not use a ring buffer we need to set `meta->sample0_offset`
+        // Since we do not use a ring buffer we need to set `meta->fpga_seq_num`
         assert(input_cursor % in_meta->sample_bytes() == 0);
-        out_meta->set_sample0_offset(div_noremainder(input_cursor, in_meta->sample_bytes()));
+        out_meta->set_fpga_seq_num(in_meta->get_fpga_seq_num()
+                                   + div_noremainder(input_cursor, in_meta->sample_bytes()));
```
rhaas80, line 227: this ignores `time_downsampling_fpga` even in the input.

lib/cuda/cudaRFIS012.cpp
```diff
@@ -189,19 +189,11 @@ cudaEvent_t cudaRFIS012::execute(cudaPipelineState& /*pipestate*/,
...
-    assert(time_downsampling_fpga.size() == static_cast<size_t>(rfi_S012_meta->get_nfreq()));
-    assert(rfi_S012_meta->get_nfreq() >= 0);
-    for (int freq = 0; freq < rfi_S012_meta->get_nfreq(); ++freq) {
-        freq_upchan_factor[freq] *= rfi_downsampling_factor;
-        time_downsampling_fpga[freq] *= rfi_downsampling_factor;
-    }
-    rfi_S012_meta->set_freq_upchan_factor(freq_upchan_factor);
-    rfi_S012_meta->set_time_downsampling_fpga(time_downsampling_fpga);
+    rfi_S012_meta->set_time_downsampling_fpga(rfi_S012_meta->get_time_downsampling_fpga()
+                                              * rfi_downsampling_factor);
```
rhaas80, line 196: this removes the update of ` freq_upchan_factor` is that intentional?

lib/cuda/cudaRFIS012bar.cpp
```diff
@@ -155,19 +155,11 @@ cudaEvent_t cudaRFIS012bar::execute(cudaPipelineState& /*pipestate*/,
...
-    assert(time_downsampling_fpga.size() == static_cast<size_t>(rfi_S012bar_meta->get_nfreq()));
-    assert(rfi_S012bar_meta->get_nfreq() >= 0);
-    for (int freq = 0; freq < rfi_S012bar_meta->get_nfreq(); ++freq) {
-        freq_upchan_factor[freq] *= rfi_second_downsampling_factor;
-        time_downsampling_fpga[freq] *= rfi_second_downsampling_factor;
-    }
-    rfi_S012bar_meta->set_freq_upchan_factor(freq_upchan_factor);
-    rfi_S012bar_meta->set_time_downsampling_fpga(time_downsampling_fpga);
+    rfi_S012bar_meta->set_time_downsampling_fpga(rfi_S012bar_meta->get_time_downsampling_fpga()
+                                                 * rfi_second_downsampling_factor);
```
rhaas80, line 162: removes ` freq_upchan_factor`, intentional?

lib/cuda/cudaRFISKtilde.cpp
```diff
@@ -191,15 +191,10 @@ cudaEvent_t cudaRFISKtilde::execute(cudaPipelineState& /*pipestate*/,
...
-    std::vector<int64_t> half_fpga_sample0(rfi_meta->get_nfreq());
-    std::vector<int> time_downsampling_fpga(rfi_meta->get_nfreq());
-    for (int freq = 0; freq < rfi_meta->get_nfreq(); ++freq) {
-        half_fpga_sample0[freq] = 128 * 8;
-        time_downsampling_fpga[freq] = 128 * 8;
-    }
-    rfi_meta->set_half_fpga_sample0(half_fpga_sample0);
-    rfi_meta->set_time_downsampling_fpga(time_downsampling_fpga);
+    rfi_meta->set_time_downsampling_fpga(rfi_meta->get_time_downsampling_fpga()
+                                         * div_noremainder(128 * 8, rfi_downsampling_factor));
```
rhaas80, line 197: this seems odd. Is it really the case that fro `rfi_downsampling_factor` of 1024 there would not be any time downsampling in this step (and there would be maximum downsampling for `rfi_downsampling_factor` of 1)?

lib/stages/lostSamplesToPLMask.cpp
```diff
@@ -162,25 +162,15 @@ void lostSamplesToPLMask::main_thread() {
...
 
-                const std::vector<int64_t> half_fpga_sample0(
-                    PL_MASK_FREQS_PER_BIN, PL_MASK_DOWNSAMPLING_FACTOR * PL_MASK_HILO_SPLIT / 2);
-                pl_mask_meta->set_half_fpga_sample0(half_fpga_sample0);
-
-                const std::vector<int> time_downsampling_fpga(
-                    PL_MASK_FREQS_PER_BIN, PL_MASK_DOWNSAMPLING_FACTOR * PL_MASK_HILO_SPLIT);
-                pl_mask_meta->set_time_downsampling_fpga(time_downsampling_fpga);
+                pl_mask_meta->set_time_downsampling_fpga(PL_MASK_DOWNSAMPLING_FACTOR
+                                                         * PL_MASK_HILO_SPLIT);
```
rhaas80, line 170: does not take possible downsampling of lostSamples into account (which is always 1 in the current dpdk code)

lib/stages/lostSamplesToPLMask.cpp
```diff
@@ -197,21 +187,8 @@ void lostSamplesToPLMask::main_thread() {
...
-
-                const std::vector<int> time_downsampling_fpga(
-                    PL_MASK_FREQS_PER_BIN, PL_MASK_DOWNSAMPLING_FACTOR * PL_MASK_HILO_SPLIT);
-                auto pl_mask_time_downsampling_fpga = pl_mask_meta->get_time_downsampling_fpga();
-                pl_mask_time_downsampling_fpga.insert(pl_mask_time_downsampling_fpga.end(),
-                                                      time_downsampling_fpga.begin(),
-                                                      time_downsampling_fpga.end());
-                pl_mask_meta->set_time_downsampling_fpga(pl_mask_time_downsampling_fpga);
+                pl_mask_meta->set_time_downsampling_fpga(PL_MASK_DOWNSAMPLING_FACTOR
+                                                         * PL_MASK_HILO_SPLIT / 2);
```
rhaas80, line 191: wrong value, sets the same scalar twice.

lib/testing/testN2kGen.cpp
```diff
@@ -240,26 +233,20 @@ void testN2kGen::set_counts_metadata(const std::shared_ptr<chordMetadata>& meta,
     meta->set_array_dimension(4, count_blocksize, "D8Plo2");
     meta->set_strides_simple();
 
+    // This looks inconsistent
```
rhaas80, line 236: does it still do so?

rhaas80 (changes requested):
one actual bug I think, setting `time_downsampling_fpga` in `lostSamplesToPLMask`

lib/cuda/cudaRFIS012.cpp
rhaas80: looks like not changing it is correct, since the RFI computation does not actually change the frequency channels.

lib/stages/lostSamplesToPLMask.cpp
rhaas80: fixed in 529245476

lib/stages/lostSamplesToPLMask.cpp
rhaas80: fixed in 529245476

lib/cuda/cudaRFIS012bar.cpp
rhaas80: removal would seem correct since RFI does not change frequency channels.

lib/cuda/cudaFRBBeamReformer.cpp
rhaas80: fixed in d61f67dba

lib/cuda/cudaCopyNToRingbuffer.cpp
rhaas80: fixed in ac8f6413

lib/cuda/cudaRFISKtilde.cpp
rhaas80: looks like it, yes, since rfimask undoes `rfi_downsampling-factor`

rhaas80:
Additional changes in https://github.com/kotekan/kotekan/pull/1386

eschnett:
Superseded by https://github.com/kotekan/kotekan/pull/1386

PR #1385, opened 2025-12-28
Add frame desc for N2 buffer types

jbmertens:
Main changes:
 - Add a generic FrameDesc class both GenericNDArray and N2FrameDesc inherit from
 - N2 frame descs are set by the buffer factory. NDArray still follow producers setting frame descriptions
 - Move structural data from N2 metadata and frameview to frame desc
 - Some validation for stages that work with N2 buffers.
 - EigenVisIter/eigenVis stages reverted for compatibility with CHIME; EigenN2Iter added to support N2 buffer type; update tests
 
 Tangential edits:
 - Add digital (fpga) gains to N2 output
 - Product indexing update
 - Placeholder (i.e. incorrect) N2Subset stage


lib/core/buffer.hpp
```diff
@@ -536,25 +537,30 @@ class Buffer : public GenericBuffer {
      * @param[in] dimnames Array axis labels in the D dimensions
      */
     template<typename T, std::size_t D>
-    void allocate_new_frame_desc(kotekan::Symbol quantity_name,
-                                 const std::array<std::ptrdiff_t, D>& extents,
-                                 const std::array<kotekan::Symbol, D>& dimnames) {
+    void allocate_ndarray_frame_desc(kotekan::Symbol quantity_name,
```
rhaas80, line 540: do we really anticipate anything that is not an NDArray? Note that this will also affect all the readers / writers that want to use this information to describe the data in the files.

lib/core/buffer.hpp
```diff
@@ -536,25 +537,30 @@ class Buffer : public GenericBuffer {
...
+                                     const std::array<std::ptrdiff_t, D>& extents,
+                                     const std::array<kotekan::Symbol, D>& dimnames) {
         buffer_lock lock(mutex);
         if (!frames_desc)
             frames_desc =
                 std::make_shared<kotekan::NDArray<T, D>>(quantity_name, extents, dimnames, nullptr);
         else {
 #if 0
```
rhaas80: oha, this should obviously not have been commented out. no idea why this slipped through in 5faca89a7

lib/core/buffer.hpp
```diff
@@ -599,10 +610,37 @@ class Buffer : public GenericBuffer {
      * @return The NDArray data structure describing the array
      */
     std::shared_ptr<const kotekan::GenericNDArray> get_frame_desc() {
```
rhaas80: I'd also rename this is to match the name in `allocate_ndarray_frame_desc`.

lib/core/buffer.hpp
```diff
@@ -599,10 +610,37 @@ class Buffer : public GenericBuffer {
...
-        // TODO: use a get/set pair instead?
+        return std::dynamic_pointer_cast<const kotekan::GenericNDArray>(frames_desc);
+    }
+
+    /**
+     * @brief provides read access to the frame description
+     * @return The data structure describing the frame
+     */
+    std::shared_ptr<const kotekan::FrameDesc> get_frame_description() {
```
rhaas80, line 625: highly confusion with `get_frame_desc` above.

lib/core/buffer.hpp
```diff
@@ -599,10 +610,37 @@ class Buffer : public GenericBuffer {
...
         return frames_desc;
     }
 
+    /**
+     * @brief Sets the frame description
+     */
+    void set_frame_desc(std::shared_ptr<kotekan::FrameDesc> new_desc) {
+        buffer_lock lock(mutex);
+
+        if (new_desc->get_byte_size() > aligned_frame_size) {
```
rhaas80: shouldn't it be exactly equal?

lib/core/buffer.hpp
```diff
@@ -599,10 +610,37 @@ class Buffer : public GenericBuffer {
...
+     */
+    std::shared_ptr<const kotekan::FrameDesc> get_frame_description() {
         return frames_desc;
     }
 
+    /**
+     * @brief Sets the frame description
+     */
+    void set_frame_desc(std::shared_ptr<kotekan::FrameDesc> new_desc) {
+        buffer_lock lock(mutex);
```
rhaas80, line 634: this lock protects `aligned_frame_size`?

lib/metadata/N2FrameDesc.cpp
```diff
@@ -0,0 +1,131 @@
...
+    field_sizes.push_back({N2Field::eval, sizeof(float) * num_ev_in});
+    field_sizes.push_back({N2Field::evec, sizeof(N2::cfloat) * num_ev_in * num_elements_in});
+    field_sizes.push_back({N2Field::emethod, sizeof(N2EigenMethod) * 1});
+    field_sizes.push_back({N2Field::erms, sizeof(float) * 1});
+    field_sizes.push_back({N2Field::gain, sizeof(N2::cfloat) * num_elements_in});
+
+    std::map<N2Field, std::pair<size_t, size_t>> frame_layout;
+    size_t offset = 0;
+    for (const auto& field : field_sizes) {
+        frame_layout[field.first] = std::make_pair(offset, offset + field.second);
```
rhaas80: this seems to assume that aggregate members of different types pack tightly. Is this ensured when the frame is populated? I would guess so, since the sizes of the fields seem to be dynamic, so a `struct`  is not used but just a bytes buffer?

lib/metadata/N2FrameDesc.cpp
```diff
@@ -0,0 +1,131 @@
...
+    field_sizes.push_back({N2Field::vis, sizeof(N2::cfloat) * num_prod_in});
+    field_sizes.push_back({N2Field::weight, sizeof(float) * num_prod_in});
+    field_sizes.push_back({N2Field::flags, sizeof(float) * num_elements_in});
+    field_sizes.push_back({N2Field::eval, sizeof(float) * num_ev_in});
+    field_sizes.push_back({N2Field::evec, sizeof(N2::cfloat) * num_ev_in * num_elements_in});
+    field_sizes.push_back({N2Field::emethod, sizeof(N2EigenMethod) * 1});
+    field_sizes.push_back({N2Field::erms, sizeof(float) * 1});
+    field_sizes.push_back({N2Field::gain, sizeof(N2::cfloat) * num_elements_in});
+
+    std::map<N2Field, std::pair<size_t, size_t>> frame_layout;
```
rhaas80: instead of a pair maybe use a simple struct with named members? 

```
struct n2field_member_t {
  size_t begin, end;
}
```

or so?

lib/metadata/N2FrameDesc.cpp
```diff
@@ -0,0 +1,131 @@
...
+        frame_layout[field.first] = std::make_pair(offset, offset + field.second);
+        offset += field.second;
+    }
+    return frame_layout;
+}
+
+size_t N2FrameDesc::calculate_frame_size(uint32_t num_elements_in, uint32_t num_ev_in,
+                                         size_t num_prod_in) {
+    auto layout = get_frame_layout(num_elements_in, num_ev_in, num_prod_in);
+    return layout[N2Field::gain].second;
```
rhaas80: this hard-codes knowledge that "gain" is the last field member. Better to define an actual variable for that, or eg add 0 size field "end".

lib/metadata/N2FrameDesc.cpp
```diff
@@ -0,0 +1,131 @@
...
+}
+
+size_t N2FrameDesc::calculate_frame_size(kotekan::Config& config, const std::string& unique_name) {
+    const int num_elements_in = config.get<int>(unique_name, "num_elements");
```
rhaas80, line 84: doesn't this get config information each this function is called (which may be often)? To me it would seem better to pull config information only once and store it in member variables (like a stage would do).

rhaas80: if this is a just a helper for clients, and does not actually need access to a N2FrameDesc object, then I'd make it a static member function so that it can be called without a (dummy?) object.

lib/metadata/N2FrameDesc.cpp
```diff
@@ -0,0 +1,131 @@
...
+    const int num_ev_in = config.get<int>(unique_name, "num_ev");
+    const N2Layout n2_layout = config.get<N2Layout>(unique_name, "n2_layout");
+    const uint64_t num_prod_in = get_num_prod(num_elements_in, n2_layout);
+    return calculate_frame_size(num_elements_in, num_ev_in, num_prod_in);
+}
+
+void N2FrameDesc::fill_prod_maps_FullUpperTri(std::vector<N2::prod_ctype>& prods,
+                                              uint32_t num_elements_in) {
```
rhaas80: make a static member?

lib/metadata/N2FrameDesc.cpp
```diff
@@ -0,0 +1,131 @@
...
+    for (uint16_t i = 0; i < num_elements_in; i++) {
+        for (uint16_t j = i; j < num_elements_in; j++) {
+            prods[p] = {.input_a = i, .input_b = j};
+            p++;
+        }
+    }
+}
+
+void N2FrameDesc::fill_prod_maps_Autocorrelations(std::vector<N2::prod_ctype>& prods,
+                                                  uint32_t num_elements_in) {
```
rhaas80: make a static member (if called by client)?

lib/metadata/N2FrameDesc.cpp
```diff
@@ -0,0 +1,131 @@
...
+    prods.resize(num_prod_in);
+    for (uint16_t i = 0; i < num_elements_in; i++) {
+        prods[i] = {.input_a = i, .input_b = i};
+    }
+}
+
+void N2FrameDesc::fill_prod_maps(std::vector<N2::prod_ctype>& prods) const {
+    switch (n2_layout) {
+        case N2Layout::FullUpperTri:
+            fill_prod_maps_FullUpperTri(prods, num_elements);
```
rhaas80: hmm, so thy are called from a proper member function and passed a member variable. Why pass in `num_elements` which will shadow a member variable? Will there ever be a different value passed in?

lib/metadata/N2FrameDesc.hpp
```diff
@@ -0,0 +1,135 @@
...
+        return num_ev;
+    }
+    uint32_t get_num_products() const {
+        return num_products;
+    }
+    N2Layout get_n2_layout() const {
+        return n2_layout;
+    }
+
+    // Static helpers (previously in frame view)
```
rhaas80, line 79: ah, so they were static already. My bad.

lib/metadata/N2FrameDesc.hpp
```diff
@@ -0,0 +1,135 @@
...
+     * @brief Calculate the size of the frame.
+     */
+    static size_t calculate_frame_size(uint32_t num_elements_in, uint32_t num_ev_in,
+                                       size_t num_prod_in);
+    static size_t calculate_frame_size(kotekan::Config& config, const std::string& unique_name);
+
+    /**
+     * @brief The layout of data/fields within the frame.
+     *
+     * @return A map of the field to the { start, end } of the field in the frame.
```
rhaas80: `begin`, `end` might be more familiar to C++

lib/metadata/N2Layout.hpp
```diff
@@ -0,0 +1,19 @@
...
+
+#include <cstdint> // for int32_t
+#include <string>  // for string
+
+enum class N2Layout : int32_t { FullUpperTri = 0, RedundantBaselineAvg = 1, Autocorrelations = 2 };
+
+void to_json(nlohmann::json& j, const N2Layout& t);
+void from_json(const nlohmann::json& j, N2Layout& t);
+
+inline std::string N2Layout_to_string(N2Layout l) {
```
rhaas80, line 21: personally I'd have the strings here and also a `from_string` function then use those in the json conversions rather than the other way around.

lib/metadata/N2Metadata.cpp
```diff
@@ -1,15 +1,14 @@
...
 #include "timeUtil.hpp" // for EOP()
 
 #include <string.h> // for size_t, memset
 
 REGISTER_TYPE_WITH_FACTORY(metadataObject, N2Metadata);
 N2Metadata::N2Metadata() :
-    N2MetadataFormat{
-        0, 0, 0, 0, N2Layout::FullUpperTri, 0, 0.0, 0, eop_null, eop_null, 0, 0, 0, 0, 0,
-        0, 0, 0, 0} {
+    N2MetadataFormat{0, 0, 0.0, 0, eop_null, eop_null, 0, 0, 0, 0, 0, 0, 0, 0, 0} {
```
rhaas80: a pretty incomprehensible set of arguments :-)

rhaas80: this is different from a (assumed to exist) default constructor for N2MetadataFormat?

lib/metadata/N2Metadata.cpp
```diff
@@ -93,6 +82,15 @@ nlohmann::json N2Metadata::to_json() {
     return rtn;
 }
 
+void N2Metadata::check_frame_desc(
```
rhaas80, line 77: this does not check the actual values (like some other `check_FOO` functions do), only the type. It also only warns, which is strange since one would expect that receiving the wrong sort of shared pointer will lead to a failure soon after.

lib/metadata/NDArray.hpp
```diff
@@ -80,7 +81,7 @@ make_arrays_from_pairs(std::initializer_list<std::pair<T, U>> values) {
 // nor the rank are known at compile time. This is convenient e.g.
 // when reading data from a file, but it prevents efficient operations
 // on array elements.
-class GenericNDArray {
+class GenericNDArray : public FrameDesc {
```
rhaas80, line 84: somewhat strange inheritance here. Admittedly I abused GenericNDArray when using it for frame description and indeed having a separate frame ndarray description and ndarry data class would be nice (which this is not quite)

lib/stages/eigenVis.cpp
```diff
@@ -39,12 +44,10 @@ eigenVis::eigenVis(Config& config, const std::string& unique_name,
 
     input_buffer = get_buffer("in_buf");
     input_buffer->register_consumer(unique_name);
-
     output_buffer = get_buffer("out_buf");
     output_buffer->register_producer(unique_name);
-
+    num_eigenvectors = config.get<uint32_t>(unique_name, "num_ev");
```
rhaas80, line 49: this is now a user input compared to having been obtained (safer?) from the previous stage's output. Is this required? It seem like an extra source of mistakes to me.

lib/stages/eigenVis.cpp
```diff
@@ -150,25 +171,29 @@ void eigenVis::main_thread() {
         }
 
         nside = (int32_t)num_elements;
+        nev = (int32_t)num_eigenvectors;
```
rhaas80, line 174: since `nev` and `nside` never change (or do they?) why not use `num_elements` and `num_eigenvectors` directly instead of indirecting access through new variables making it harder to follow the flow of information by having to keep track of more information? is the reason for their existence only that one would like to have the cast in only one place? If they are kept and they are const then I'd move their declaration here as well and add a const.

lib/utils/N2FrameView.cpp
```diff
@@ -7,18 +7,30 @@
 #include <complex>  // for complex
 #include <cstring>  // for memset, size_t
 
+namespace {
+std::shared_ptr<const kotekan::N2FrameDesc>
+validate_desc(std::shared_ptr<const kotekan::FrameDesc> desc) {
```
rhaas80: same functionality but different name than a previous `check_FOO` function. I would try and keep the names consistent as much as possible.

tests/boost/test_utils.hpp
```diff
@@ -301,10 +302,10 @@ make_writer_config(const std::string& unique_name, const std::string& in_buf,
...
     BOOST_REQUIRE(meta);
 
-    const size_t num_prod = N2FrameView::get_num_prod(num_input, N2Layout::FullUpperTri);
-    meta->num_elements = num_input;
-    meta->num_prod = num_prod;
-    meta->num_ev = num_ev;
+    const size_t num_prod = kotekan::N2FrameDesc::get_num_prod(num_input, N2Layout::FullUpperTri);
+    // meta->num_elements = num_input;
+    // meta->num_prod = num_prod;
+    // meta->num_ev = num_ev;
```
rhaas80: remove commented out lines

rhaas80 (commented):
Nothing obviously wrong, but I have lots of questions and comments.

lib/core/buffer.hpp
rhaas80: hmm, apparently it makes tests fail: https://github.com/kotekan/kotekan/pull/1387/checks though at least one test is a lint failure, and the other one shows up as a Python error. Will try again.

rhaas80: ok, now it passes, but given that the only change was running `clang-format-18` on `buffer.hpp` the failure in https://github.com/kotekan/kotekan/actions/runs/20680243163/job/59373486720?pr=1387#step:8:174 may be some sort of not fully reproducible behaviour again. Python's error (in `test_eigen[file]` so may actually be of interest here) was:

```
>           assert input_frame.erms == output_frame.erms
E           ValueError: The truth value of an empty array is ambiguous. Use `array.size > 0` to check that an array is not empty.
```

this should always error out (needs `np.all(...)` I guess).

lib/core/buffer.hpp
jbmertens: Yes, sort of the crux of this PR is... How do we handle visibility data? I'm definitely open to discussion/changes.

I do think there is a solid argument for passing around multiple arrays at once, in the same frame. E.g. data + data derivatives. This could be done with multiple buffers, but then one runs into (arguably) more overhead and synchronization issues. We do need multiple arrays stored in the same file anyways, so we will necessarily have specialized file writers and readers.

Yes, technically all frames are just a 1D array of bits. But it is nice to have a more useful description. Blocked, stacked, or masked visibility matrices can be represented as a higher (or lower) dimension array with a block index, or a raw 1D data array, but in this case calling it an "NDArray" misleading, and loses information about the data layout that can be made explicit.

lib/core/buffer.hpp
jbmertens: Thanks, ~marking this as resolved~ will update after #1387 is merged

lib/core/buffer.hpp
jbmertens: And `frames_desc` itself (multiple producers may attempt to set it).

lib/metadata/N2FrameDesc.cpp
jbmertens: Sure, the struct is more readable.

lib/metadata/N2FrameDesc.cpp
jbmertens: This is the idea, yes, assuming the N2FrameView (or older visFrameView) wrappers are used.

lib/metadata/N2Layout.hpp
jbmertens: I think the intent behind this was to take advantage of nlohmann::json's enum conversion, rather than define our own. So this is backwards for that reason.

lib/metadata/N2Metadata.cpp
jbmertens: Just some legacy code... sure, worth adding defaults though. (Adding as default member initializers.)

lib/metadata/NDArray.hpp
jbmertens: Open to suggestions for a change here - was aiming for something minimal that "just worked".

lib/stages/eigenVis.cpp
jbmertens: Yes - these changes are due to a reversion of this stage to the state it was in for CHIME. (Rather than digging through and making sure I'm cleaning consistently, I'd like to leave this as-was for the moment - I'm adding an EigenN2 stage for CHORD/N2 type buffers in another PR in place of these changes.)

lib/stages/eigenVis.cpp
jbmertens: Similar to above, this is just due to a reversion of this stage to the state it was in for CHIME. Aiming for consistency rather than cleaning in this file.

lib/utils/N2FrameView.cpp
jbmertens: Changed to `validate_desc_type`, since that was the intent rather than checking the members.

lib/metadata/N2Metadata.cpp
jbmertens: Changed to an error.

For N2, structural data is no longer carried in the metadata (which I thought was the intent of the frame_desc, anyways?), so it isn't checked. I've added few sanity checks.

lib/metadata/FrameDesc.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+class FrameDesc {
+public:
+    virtual ~FrameDesc() = default;
+
+    // A stored name for the quantity represented by this frame description
+    virtual Symbol get_quantity_name() const = 0;
+
+    // Output the array metadata, useful for logging or debugging
+    virtual void output_metadata(std::ostream& os) const = 0;
```
andrerenard: This might get confusing with the per-frame time variable metadata (e.g. chordMetadata)

lib/stages/N2Subset.hpp
```diff
@@ -0,0 +1,100 @@
...
+#include <stddef.h> // for size_t
+#include <string>   // for string
+#include <tuple>    // for tuple
+#include <utility>  // for pair
+#include <vector>   // for vector
+
+
+/**
+ * @class N2Subset
+ * @brief ``kotekan::Stage`` that extracts a subset of the products.
```
andrerenard, line 25: Add TODO that this isn't working yet?

lib/stages/countCheck.cpp
```diff
@@ -1,7 +1,9 @@
 #include "countCheck.hpp"
```
andrerenard, line 1: Should we consider making this VisCountCheck and N2CountCheck?  or something that works generically on any buffer type with fpga_seq metadata.

lib/stages/hdf5N2Write.cpp
```diff
@@ -141,6 +141,90 @@ void N2FileData::_check_create_dataset(HighFive::File& file, const std::string&
...
+        if (entry.is_directory()) {
+            dir_list.push_back(entry.path());
+        }
+    }
+    // Sort directories by name (timestamp)
+    std::sort(dir_list.begin(), dir_list.end());
+    // Assume the most recent directory was used
+    // (TODO: use API instead)
+    if (dir_list.empty()) {
+        ERROR_NON_OO("No baseband directories found in {}, cannot read digital gains.",
```
andrerenard:
```suggestion
        ERROR_NON_OO("No gain directories found in {}, cannot read digital gains.",
```

lib/stages/hdf5N2Write.cpp
```diff
@@ -141,6 +141,90 @@ void N2FileData::_check_create_dataset(HighFive::File& file, const std::string&
...
+std::optional<N2FileData::DigitalGains> N2FileData::_get_digital_gains() const {
+
+    // If no gains directory specified, skip digital gains
+    if (gains_base_directory.empty()) {
+        DEBUG_NON_OO("No gains_base_directory specified, skipping digital gains.");
+        return std::nullopt;
+    }
+
+    // Search for *most recent* time in directories names as "YYYYMMDDTHHMMSSZ_baseband" :
+    std::filesystem::path base_path(gains_base_directory);
```
andrerenard, line 154: We might want to store the gain directory name, so we know where this came from. 

Also this might be addressed when we make this an API, but should we have some method of validating that these gains are _actualy_ what the F-engine is using?  e.g. query fpga_master to see what gains it's running with.  (Since this isn't ready yet, add a big todo and/or ticket.)

lib/stages/hdf5N2Write.cpp
```diff
@@ -141,6 +141,90 @@ void N2FileData::_check_create_dataset(HighFive::File& file, const std::string&
...
+        ERROR_NON_OO("No baseband directories found in {}, cannot read digital gains.",
+                     gains_base_directory);
+        return std::nullopt;
+    }
+    std::filesystem::path latest_dir = dir_list.back();
+
+    // Gains file should be named "gains.hdf5" in this directory
+    std::filesystem::path gains_path = latest_dir / "gains.hdf5";
+    if (!std::filesystem::exists(gains_path)) {
+        ERROR_NON_OO("Digital gains file {} does not exist.", gains_path.string());
```
andrerenard: I would be tempted to make this (and other file system type errors) fatal, just to make sure people know that something has gone wrong - and we don't end up with partly filled data files.

lib/stages/rawFileRead.cpp
```diff
@@ -37,6 +37,13 @@ rawFileRead::rawFileRead(Config& config, const std::string& unique_name,
 
     buf = get_buffer("buf");
     buf->register_producer(unique_name);
+
+    if (buf->buffer_type == "N2") {
```
andrerenard: Does this issue impact other formats?  Should this become an "include" vs. "exclude" check?

lib/utils/Telescope.cpp
```diff
@@ -22,6 +22,8 @@ const Telescope& Telescope::instance(const kotekan::Config& config) {
 
     auto telescope_name = config.get_default<std::string>("/telescope", "name", "ICETelescope");
     if (!FACTORY(Telescope)::exists(telescope_name)) {
+        if (telescope_name == "fake")
+            WARN_NON_OO("Did you build with -DWITH_TESTS=ON?");
```
andrerenard: You could check the preprocessor variable.

andrerenard (changes requested):

lib/stages/rawFileRead.cpp
jbmertens: The issue is just the frame description not being read/written.
For N2 data, this isn't an issue now, since the description is set in the bufferFactory.
But for NDArrays, this stage will no longer work, per frame descs.

andrerenard (approved):

PR #1389, opened 2026-01-07
Output additional metadata to file

eschnett:
Output telescope name, sampling time scale, whether we have GPS time, and the telescope time for the first time sample.

### DIFF
--- a/lib/stages/asdfFileWrite.cpp
+++ b/lib/stages/asdfFileWrite.cpp
@@ -1,45 +1,45 @@
-#include "Config.hpp"          // for Config
-#include "NDArray.hpp"         // for GenericNDArray
-#include "asdfFiles.hpp"       // for beautify_buffer_name, chord2asdf, chord_metadata_version
-#include "buffer.hpp"          // for Buffer
-#include "bufferContainer.hpp" // for bufferContainer
-#include "kotekanLogging.hpp"  // for DEBUG, FATAL_ERROR, ERROR, INFO, WARN
-#include "metadata.hpp"        // for metadataObject
-
-#include "fmt.hpp"      // for compile_string_to_view, join
-#include "gsl-lite.hpp" // for span
-
-#include <DataType.hpp>            // for type_to_string, type_total_bytes
-#include <N2FrameView.hpp>         // for N2FrameView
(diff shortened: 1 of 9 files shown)


lib/stages/asdfFileWrite.cpp
```diff
@@ -1,45 +1,45 @@
...
-#include <errors.h>                // for exit_kotekan, ReturnCode
-#include <functional>              // for function
-#include <iomanip>                 // for operator<<, setfill, setw
-#include <memory>                  // for shared_ptr, __shared_ptr_access, make_shared, allocator
-#include <optional>                // for optional
-#include <prometheusMetrics.hpp>   // for Metrics, Gauge
-#include <sstream>                 // for basic_ostream, operator<<, basic_ostringstream, ostri...
-#include <string>                  // for basic_string, char_traits, string, operator<<
-#include <sys/stat.h>              // for mkdir
+#include <Config.hpp>       // for Config
```
jbmertens, line 1: Seems not to make a difference, but isn't `#include "file.hpp"` generally preferred for local includes?

jbmertens:
The change of `#include ""` to `#include <>` could use some reasoning or should be reverted, but otherwise LGTM.

It would be nice to have an explicit test covering the writers.

lib/stages/asdfFileWrite.cpp
eschnett: the `"file"` syntax looks in the current directory, then in the include path. `<file>` only looks in the include path.

i guess it's a coding style issue. what do we want to use?

jbmertens: Hm, iwyu is our linter for includes, but it doesn't enforce `""` vs `<>`, it just groups them into two different alpha lists.

Most of the code uses `""` for project includes, and `<>` for system, although cmake sets up include paths properly regardless. I'd prefer keeping that separation for consistency, e.g. I was wondering if there was a substantive change here.

But up to you whether to revert.

jbmertens (approved):

eschnett:
I will convert the include file syntax to a common scheme (`<...>` for externals, `"..."` for internals) in a future PR.

PR #1390, opened 2026-01-08
Update Julia to 1.12.4

eschnett:


jbmertens (approved):
Looks good to me, as long as tests are passing.

rhaas80:
stabilizing the Julia version in use (for code generation) would be good

jbmertens:
We had discussed keeping the LTS version, but I assume this was updated because a new feature was needed?

Is generated code different between versions?

PR #1391, opened 2026-01-09
Poison buffers

eschnett:


lib/cuda/NDArrayBuffer.hpp
```diff
@@ -223,20 +223,17 @@ class NDArrayBuffer : public kotekan::kotekanLogging {
 
     // Poison an NDArray buffer
     void set_to_poison([[maybe_unused]] const std::uint8_t poison_value) {
```
rhaas80: no longer maybe_unused

lib/cuda/NDArrayRingBuffer.hpp
```diff
@@ -467,7 +467,6 @@ class NDArrayRingBuffer : public kotekan::kotekanLogging {
     void set_to_poison([[maybe_unused]] const std::uint8_t poison_value,
```
rhaas80: no longer maybe_unused

lib/cuda/NDArrayBuffer.hpp
```diff
@@ -223,20 +223,17 @@ class NDArrayBuffer : public kotekan::kotekanLogging {
...
         assert(buffer_device_ptr);
         const cudaStream_t cuda_stream =
             cuda_command.get_device().getStream(cuda_command.get_cuda_stream_id());
         CHECK_CUDA_ERROR(
             cudaMemsetAsync(buffer_device_ptr, poison_value, buffer_length, cuda_stream));
-#endif
     }
 
     // Check an NDArray buffer for poison
     void check_for_poison([[maybe_unused]] const std::uint8_t poison_value) {
```
rhaas80: no longer maybe_unused

rhaas80 (changes requested):
looks ok, but leaves a couple of `maybe_unused` attributes behind that are no longer true (harmless admittedly), since the variable is now always used.

rhaas80 (approved):

PR #1392, opened 2026-01-12
Use consistent `#include` style

eschnett:


jbmertens (approved):
LGTM - although it would be nice to have a linter instead of tuning this by hand.
We can merge this in, but can also consider using clang-tidy for this.

PR #1394, opened 2026-01-16
bunch of minor fixes all over kotekan

rhaas80:
none of these should affect any science

### DIFF
--- a/lib/stages/lostSamplesToPLMask.cpp
+++ b/lib/stages/lostSamplesToPLMask.cpp
@@ -78,7 +78,7 @@ void lostSamplesToPLMask::main_thread() {
         uint8_t* pl_mask_frame =
             pl_mask_buf->wait_for_empty_frame(unique_name, pl_mask_buf_frame_id);
         if (pl_mask_frame == nullptr)
-            break;
+            return;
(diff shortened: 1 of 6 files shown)


lib/stages/lostSamplesToPLMask.cpp
```diff
@@ -87,7 +87,7 @@ void lostSamplesToPLMask::main_thread() {
             uint8_t* flag_frame =
                 lost_samples_buf->wait_for_full_frame(unique_name, lost_samples_buf_frame_id);
             if (flag_frame == nullptr)
-                break;
+                return;
```
jbmertens, line 90: I'm not familiar with this stage, just checking that this should indeed be a return and not break (i.e. code at the end of this inner loop should not run).

jbmertens (approved):
LGTM

lib/stages/lostSamplesToPLMask.cpp
rhaas80: this inner loop gets a number of frames to combine them in the outer loop. So if we are being shut down (`nullptr` from `wait_for_full_frame`) then we want to terminate both loop and since we did not collect enough frames to an outer loop iteration there is output frame would could provide. 

This all happens during shutdown only, so nothing should matter since we cannot pass any more frames down the pipeline anyway.

PR #1395, opened 2026-01-20
remove DOS style line endings

rhaas80:
this will cause conflicts far and wide.

```
sed -i $'s/\r$//
```

to fix your own files.

```
git filter-branch -f --tree-filter 'find . -name .git -prune -or -type f -print | while read i ; do file $i | grep -q CRL && sed -i $'\''s/\r$//'\'' $i ; done' origin/chord..HEAD
```

to filter your history from origin/chord to the current branch's head (will be very slow though).


jbmertens (approved):
Or `git add --renormalize` .
Thankfully git(hub) diff has an ignore whitespace feature. In which I see no changes here.

PR #1397, opened 2026-01-21
n2k: remove full mod calculation from index computation

rhaas80:


rhaas80:
This does not quite remove all `/` and `%` uses since the calculation below

```
// Unpack "flattened" index n to 3-d index (f,t,m).
```

uses sizes known only at run-time that are not powers of 2 (contains a factor of 3 due to the "S" index).

external/n2k/src_lib/SkKernel.cu
```diff
@@ -100,29 +100,9 @@ __global__ void sk_kernel(
...
-    };
-
-    const auto ringbuffer_sk_single_feed = [=](float& lval) -> float& {
-        const int out_sk_single_feed_min_in_elems = sk_single_feed_Tmin * F * 3 * S;
-        const int out_sk_single_feed_size_in_elems = sk_single_feed_Tsize * F * 3 * S;
-        return *(out_sk_single_feed
-                 + (&lval - out_sk_single_feed + out_sk_single_feed_min_in_elems)
-                       % out_sk_single_feed_size_in_elems);
-    };
+    assert(S012_Tsize & (S012_Tsize - 1) == 0 && "S012_Tsize must be a power of two");
```
eschnett, line 103: Is there a proper test for this condition? By default, asserts are removed in production builds. Can you check whether there is a proper check further up, and if not, e.g. call `std::abort` instead?

eschnett (changes requested):
I don't understand most of the changes here. They seem complicated, and it would be easy to make a mistake.

How did you test this?

Would it be possible to have a version of the code that runs both code paths (before and after your changes) at the same time to catch differences?

In the version with `%`, it's straightforward to see that the result is always inside the ring buffer. I don't see this in the new version which uses pointer addition. Could you add respective assert statements which check that the indexing variables and respective pointers are in bounds? The asserts can be expensive since they won't be there in production.

external/n2k/src_lib/SkKernel.cu
rhaas80: I'll put in debugging code. 

The change is actually rather simple, but github diff viewer makes it look complicated. 

* remove the `ringerbuffer_foo` functions
* when computing `in_ix` which is the only time the "time" axis actually comes into play, since "time" is also the axis that wraps, replace "t" by "t % tsize".

rhaas80: E.g. a change is:

```
long in_ix = (t1*long(F) + f1) * (3*S);
```

to

```
long in_ix_S012 = (((S012_Tmin + t1) & S012_Tsize_mask) * long(F) + f1) * (3*S); 
```

that is offset the `t1` time index in the current set of elements to process (runs from `0` to `T`) by the `Tmin` value to get absolute time, then wrap around to Tsize.

This the happens 3 times (or so) since the various arrays can have different Tsize and Tmin values.

rhaas80: > Is there a proper test for this condition? By default, asserts are removed in production builds. Can you check whether there is a proper check further up, and if not, e.g. call `std::abort` instead?

This runs on the GPU, so sure I could call `abort`. I'd rather however have some more meaningful check on the CPU side and output a useful error message.

rhaas80: > > Is there a proper test for this condition? By default, asserts are removed in production builds. Can you check whether there is a proper check further up, and if not, e.g. call `std::abort` instead?
> 
> This runs on the GPU, so sure I could call `abort`. I'd rather however have some more meaningful check on the CPU side and output a useful error message.

Done in fcb6b554

external/n2k/src_lib/SkKernel.cu
```diff
@@ -100,46 +100,64 @@ __global__ void sk_kernel(
...
-        auto f = idx / rfimask_fstride;        // frequency
-        auto t32 = idx % rfimask_fstride;      // time / 32
-        auto t = t32 * 32;                     // time
-        auto t1024hi = t / rfimask_time_split; // coarse time
-        auto t1024lo = t % rfimask_time_split; // fine time
-        auto thi = t1024hi;                    // first array index
-        auto tlo = t1024lo / 32;               // third array index
-        auto thi_rb = (thi + rfimask_T1024min) & (rfimask_T1024size - 1);
-        auto idx2 = tlo + f * (rfimask_time_split / 32) + thi_rb * (rfimask_time_split / 32) * F;
+        ptrdiff_t idx = &lval - out_rfimask;        // index
```
rhaas80, line 151: other than making the type explicit, there is no change here (ie `auto` was a `ptrdiff_t` anyway).

eschnett (approved):

rhaas80:
Naturally, yesterday I ran for a long time on the test node and actually got a cuda launcher error due to "integer overflow". So this may need more fixes or may be a different kernel (but I did not run with the sanitizer so do not even know which kernel is the one triggering).

lib/cuda/cudaRFIS012.cpp
```diff
@@ -206,7 +206,8 @@ cudaEvent_t cudaRFIS012::execute(cudaPipelineState& /*pipestate*/,
 
     const std::ptrdiff_t Tsize = voltage.get_ndarray().extent(0);
     assert(Tsize % rfi_downsampling_factor == 0);
-    const std::ptrdiff_t Tmin = voltage.get_read_valid().begin();
+    // Tmin wraps around into actual array index to avoid overflows
+    const std::ptrdiff_t Tmin = voltage.get_read_valid().begin() % Tsize;
```
rhaas80, line 210: @eschnett like so?

eschnett: yes!

PR #1398, opened 2026-01-22
n2k: Simplify cmake file

eschnett:
Written with the help of ChatGPT.


jbmertens (approved):
Is there a functional change here? It looks like it should be equivalent, but is easier to read, so LGTM once linting passes.

eschnett:
No functional change intended.

PR #1399, opened 2026-01-22
Handle frequency metadata consistently

eschnett:
Handle frequency metadata consistently:
- Ensure that `coarse_freq` and `freq_upchan_factor` are set consistently and propagated consistently
- New metadata `freq_upchan_index` describing the "fine" index within an upchannelized channel
- Set ringbuffer metadata only once at startup
- Correct race conditions when setting metadata in kernels that combine outputs into a single buffer
- Correct a few bugs here and there
- Add respective tests
- Regenerate kernels


lib/cuda/cudaCopyNToRingbuffer.cpp
```diff
@@ -211,6 +219,22 @@ cudaEvent_t cudaCopyNToRingbuffer::execute(cudaPipelineState& /*pipestate*/,
                   meta_in->get_coarse_freq().at(0));
             throw std::runtime_error("cudaCopyNToRingbuffer: metadata frequency mismatch");
         }
+        if (coarse_freq.at(i) != meta_in->get_freq_upchan_factor().at(0)) {
```
rhaas80: should this not be:

```
if (freq_upchan_factor.at(i) != meta_in->get_freq_upchan_factor().at(0)) {
```

instead?

julia/kernels/chimefrb_template.cxx
```diff
@@ -332,36 +332,122 @@ cudaEvent_t cuda{{{kernel_name}}}::execute(cudaPipelineState& /*pipestate*/, con
...
+        I_meta->set_freq_upchan_factor(I_freq_upchan_factor);
+
+        const auto Ebar_freq_upchan_index = Ebar_meta->get_freq_upchan_index();
+        assert(Ebar_freq_upchan_index.size() == static_cast<std::size_t>(Ebar_nfreq));
+        std::vector<int> I_freq_upchan_index;
+        if (!I_has_metadata)
+          I_freq_upchan_index.resize(I_meta->dim[I_rank - 1 - I_index_Fbar], -1);
+        else
+          I_freq_upchan_index = I_meta->get_freq_upchan_index();
+        for (int freq = 0; freq < I_nfreq; ++freq)
```
rhaas80: couldn't one assign the whole container at once instead of manually looping over it's components? Or use `std::copy` if the containers hold different types (int vis int64_t or so)?

julia/kernels/frb_template.cxx
```diff
@@ -340,58 +341,133 @@ cudaEvent_t cuda{{{kernel_name}}}::execute(cudaPipelineState& /*pipestate*/, con
...
-                        assert(metadata->get_dimension_name(d) == dimname[d]);
-                        // The ring buffer direction is special
-                        if (d > 0)
-                            assert(metadata->dim[d] == int(ndarray.extent(d)));
-                        assert(metadata->stride[d] == ndarray.stride(d));
+    // Since we use a ring buffer we need to set the metadata only once
+    static bool did_set_metadata = false;
+    static std::mutex set_metadata_mutex;
+    if (instance_num == 0 && !did_set_metadata) {
+        did_set_metadata = true;
```
rhaas80, line 354: this (and the same code in chimefrb) only works if there is only 1 such stage-or-cuda-command instantiated, otherwise the `static` variables would be shared among them all.

julia/kernels/frb_template.cxx
```diff
@@ -340,58 +341,133 @@ cudaEvent_t cuda{{{kernel_name}}}::execute(cudaPipelineState& /*pipestate*/, con
...
+
+        // Our output buffer I has multiple producers.
+        // They must serialize setting their part of the metadata.
+        std::lock_guard<std::mutex> lock(set_metadata_mutex);
```
rhaas80: this is odd, the mutex is visible only inside of this function, and the way the `did_set_metadata` is used will not guarantee that metadata is actually set before used if multiple threads can enter here before the thread with `instance_num == 0` finishes, doesn't it?

julia/kernels/frb_template.cxx
```diff
@@ -340,58 +341,133 @@ cudaEvent_t cuda{{{kernel_name}}}::execute(cudaPipelineState& /*pipestate*/, con
...
+        const auto Ebar_freq_upchan_factor = Ebar_meta->get_freq_upchan_factor();
+        assert(Ebar_freq_upchan_factor.size() == static_cast<std::size_t>(Ebar_nfreq));
+        std::vector<int> I_freq_upchan_factor;
+        if (!I_has_metadata)
+          I_freq_upchan_factor.resize(I_meta->dim[I_rank - 1 - I_index_Fbar], -1);
+        else
+          I_freq_upchan_factor = I_meta->get_freq_upchan_factor();
+        for (int freq = 0; freq < I_nfreq; ++freq)
+          assert(static_cast<std::size_t>(Fbar_out_min + freq) < I_freq_upchan_factor.size()),
+            assert(static_cast<std::size_t>(Fbar_in_min) + freq < Ebar_freq_upchan_factor.size()),
```
rhaas80: is this really a comma operator to do:

```
for()
  assert(something), assert(somthing else), a = b;
```
? 

In that case I would have thought a regular `{}` block and multiple statements would be easier to read.

julia/kernels/upchan_template.cxx
```diff
@@ -328,32 +328,75 @@ cudaEvent_t cuda{{{kernel_name}}}::execute(cudaPipelineState& /*pipestate*/, con
...
-            {{/isoutput}}
-            {{#isoutput}}
-                {{{name}}}_buffer.set_metadata(E_buffer.get_metadata());
-            {{/isoutput}}
-        {{/hasbuffer}}
-    {{/kernel_arguments}}
+    // Since we use a ring buffer we need to set the metadata only once
+    static bool did_set_metadata = false;
+    if (instance_num == 0 && !did_set_metadata) {
+        did_set_metadata = true;
```
rhaas80, line 334: same comment about multiple instances of this command as for the frb stuff

lib/cuda/cudaQuantizeKernel8.cu
```diff
@@ -0,0 +1,114 @@
...
+    assert(chunk_size % 4 == 0);
+
+    float sum = 0, sum2 = 0;
+    for (int i = 0; i < chunk_size; i += 2) {
+        const __half2 x = ((const __half2*)input)[i / 2];
+        sum += hsum2(x);
+        sum2 += hsum2(__hmul2(x, x));
+    }
+    const float mean = sum / chunk_size;
+    const float scale = fmax(1.0f, sqrt(scale / (chunk_size - 1)) / 2.3f);
```
rhaas80: isn't `scale` still unset in the call to `sqrt` here?

lib/cuda/cudaQuantizeKernel8.cu
```diff
@@ -0,0 +1,114 @@
...
+
+    const __half2 meanh_scaleh = __floats2half2_rn(mean, scale);
+    ((__half2*)outputf)[0] = meanh_scaleh
+
+        const __half meanh = __low2half(meanh_scaleh);
```
rhaas80: strange indentation

lib/cuda/cudaQuantizeKernel8.cu
```diff
@@ -0,0 +1,114 @@
...
+    for (int i = 0; i < chunk_size; i += 2) {
+        const __half2 x = ((const __half2*)input)[i / 2];
+        sum += hsum2(x);
+        sum2 += hsum2(__hmul2(x, x));
+    }
+    const float mean = sum / chunk_size;
+    const float scale = fmax(1.0f, sqrt(scale / (chunk_size - 1)) / 2.3f);
+
+    const __half2 meanh_scaleh = __floats2half2_rn(mean, scale);
+    ((__half2*)outputf)[0] = meanh_scaleh
```
rhaas80: isn't this missing a `;`?

lib/cuda/cudaQuantizeKernel8.cu
```diff
@@ -0,0 +1,114 @@
...
+    assert(outputf_stride3 >= 0);
+    assert(outputi_size1 >= 0);
+    assert(outputi_size2 >= 0);
+    assert(outputi_size3 >= 0);
+    assert(outputi_stride2 >= 0);
+    assert(outputi_stride3 >= 0);
+    assert(chunk_size > 0);
+
+    assert(chunk_size % 4 == 0);
+    assert(ninputs1 % chunk_size == 0);
```
rhaas80: I don't see a `ninputs1` argument to the function. Is this supposed to be `input_size1`?

lib/cuda/cudaQuantizeKernel8.cu
```diff
@@ -0,0 +1,114 @@
...
+
+    assert(outputf_size1 == 2 * input_size1 / chunk_size);
+    assert(outputf_size2 == input_size2);
+    assert(outputf_size3 == input_size3);
+    assert(outputi_size1 == input_size1);
+    assert(outputi_size2 == input_size2);
+    assert(outputi_size3 == input_size3);
+
+    const int nchunks = ninputs1 / chunk_size;
```
rhaas80: variable name?

lib/metadata/jsonMetadata.hpp
```diff
@@ -20,7 +20,7 @@ namespace jsonMetadata {
 // these must match chimeMetadata and choordMetadata for now
 // in namespace for C++ const
 #define MAX_NUM_BEAMS 20
-const int CHORD_META_MAX_FREQ = 1024;
+const int CHORD_META_MAX_FREQ = 4096;
```
rhaas80, line 23: I probably should not have duplicated this definition

lib/stages/gdalFileRead.cpp
```diff
@@ -195,6 +195,19 @@ class gdalFileRead : public kotekan::Stage {
                 }
             }
 
+            {
+                const auto freq_upchan_index = group->GetAttribute("freq_upchan_index");
+                assert((meta->get_nfreq() >= 0) == bool(freq_upchan_index));
```
rhaas80: a strange condition (and cast), but I assume this is fine?

rhaas80 (changes requested):
some bugs I think

lib/cuda/cudaCopyNToRingbuffer.cpp
eschnett: Thanks.

julia/kernels/chimefrb_template.cxx
eschnett: This combines frequencies from several kernels into the same `I` buffer. `std::copy` would work, but would need to take the offsets into account ...

... is what the code originally did, because it had the same structure for the CHORD and CHIME kernels. Now I see that the CHIME code is simplified because all frequencies have the same upchannelization factor, and the frequency-combining logic collapsed into nothing.

julia/kernels/frb_template.cxx
eschnett: Yes. I think this conditions holds because different upchannelization factors are run by different kernels which are different stages with different source codes. I'll change the code anyways.

julia/kernels/frb_template.cxx
eschnett: Yes, the mutex needs to be a single global variable shared by all kernels.

julia/kernels/frb_template.cxx
eschnett: That's just me debugging, should have removed the asserts. They create nicer error messages than `std::vector::at`, which just raises an exception that eats the stack trace and doesn't create any error messages. Will remove the asserts.

julia/kernels/upchan_template.cxx
eschnett: Will fix.

lib/cuda/cudaQuantizeKernel8.cu
eschnett: That file was committed accidentally.

lib/stages/gdalFileRead.cpp
eschnett: That's probably a left-over from switching to JSON metadata. The function `get_nfreq()` used to be a field `nfreq` that is negative when there are no frequency metadata. Now `get_nfreq()` throws an exception when there are no frequency metadata, an one has to use `has_nfreq()` to check whether they exist. I'll fix the conditions.

Yes, the `... == bool(...)` is intended. I want to say "frequency metadata exist if and only if the `freq_upchan_index` exists". Otherwise I'd have an if statement with two asserts:
```
if (meta->has_nfreq())
    assert(freq_upchan_index);
else
    assert(!freq_upchan_index);
```
or a really complicated assert:
```
assert((meta->has_nfreq() && freq_upchan_index) || (! meta->has_nfreq() && !freq_upchan_index);
```
where I really just want to check that both conditions are true or both are false, i.e. they are equal.

julia/kernels/frb_template.cxx
rhaas80: I think the stack trace gets eaten by the `catch { throw;}` construct in `lib/core/Stage.cpp` .

rhaas80 (approved):
looks good to me.

eschnett:
@rhaas80 I fixed the problems you found. Can you have a look again?

PR #1400, opened 2026-01-23
New 100G network code for use with CRS boards and switch based corner turns

andrerenard:


build/.gitignore
jbmertens, line 1: Probably this should stay?

.gitignore
```diff
@@ -13,6 +13,7 @@
 /.vscode/
 /nbproject/
 /cmake-build*/
+build/
```
jbmertens: I think we want to keep the build directory.

config/chord_pathfinder.j2
```diff
@@ -0,0 +1,530 @@
...
+# n2k sub-integrations to combine into each output N2 frame by N2Accumulate.
+# Choose a multiple of integrations_per_frame so n2k_frames_per_output is an integer and generators run long enough.
+# 238 ~= 10 seconds. 
+{% set num_n2k_samples_to_accumulate = 238 %} 
+
+# correlation frames needed to form one output N2 frame
+{% set n2k_frames_per_output = num_n2k_samples_to_accumulate // integrations_per_frame %}
+
+
+##########################################
```
jbmertens: Comment here is no longer correct

jbmertens (commented):
I have not read through the dpdk-specific additions, but can if it needs more eyes.

.gitignore
andrerenard: Yeah that's a mistake, will fix.

andrerenard:
Note I have not done I cleanup pass of this yet.   There's a bunch of small things that need to be fixed/cleaned up before someone spends too much time looking this over.

external/ksgpu/CMakeLists.txt
jbmertens, line 1: Just noting, these changes should be reverted in favor of
https://github.com/kotekan/kotekan/pull/1462

lib/core/kotekanMode.cpp
```diff
@@ -19,15 +19,58 @@
...
+/**
+ * @brief Return the first non-loopback IPv4 address on this host as a
+ * dotted-quad string, or empty string if none found.
+ *
+ * Used to give this kotekan instance a globally unique self-identity in
+ * ConfigTracker when REST is bound to 0.0.0.0. Falling back to 127.0.0.1
+ * causes every node to self-label identically and collide when peers
+ * exchange configs over REST.
+ */
+std::string first_non_loopback_ipv4() {
```
andrerenard, line 46: @jbmertens   This could get us in some degree of trouble.   There are likely to be many IPs on each node once we add the backend networks.  Also by binding to only one IP we are going to prevent the REST server for replying on anything other than that IP.  Which is not the expected behaviour for binding to `0.0.0.0`.   Can we identify the host in the config tracker by hostname instead?

jbmertens: Yes, this is a bit annoying, it needs to be made more robust. I didn't want to rely on DNS or self-reported hostname. A slightly better fix is for the config tracker to avoid sending the local config in both the list of configs it tracks and the self-reported one.

lib/dpdk/dpdkCore.cpp
```diff
@@ -396,13 +503,14 @@ int dpdkCore::lcore_rx(void* args) {
     // TODO Figure out why this sleep is need.  It seems like the when starting the E810
     // ports, there is a delay between when they are started in DPDK and when they become
     // ready.  There might be a function to check their readness state we could query?
-    sleep(40);
+    sleep(10);
     while (!core->stop_thread) {
-        for (uint32_t i = 0; i < num_local_ports; ++i) {
+        for (uint32_t i = 0; i < ports.size(); ++i) {
```
rhaas80: if one is worried about the overhead of a virtual function call below then here one may consider:

```
for(uint32_t i = 0, i_end = ports.size(); i < i_end; ++i)
```

to make very sure the compiler is aware that the upper end of the loop is not changing.

lib/dpdk/dpdkCore.cpp
```diff
@@ -396,13 +503,14 @@ int dpdkCore::lcore_rx(void* args) {
...
             uint32_t port = ports[i];
 
+            // INFO_NON_OO("Polling port {:d} for packets", port);
             const uint16_t num_rx = rte_eth_rx_burst(port, 0, mbufs, burst_size);
-
+            // INFO_NON_OO("Port {:d} received {:d} packets", port, num_rx);
```
rhaas80: remove commented out out code

lib/dpdk/dpdkCore.cpp
```diff
@@ -412,7 +520,8 @@ int dpdkCore::lcore_rx(void* args) {
                     goto exit_lcore;
                 }
 
-                rte_pktmbuf_free(mbufs[j]);
+                // TODO, free the mbuf in the handler.
```
rhaas80: does it now?

lib/metadata/chordMetadata.hpp
```diff
@@ -285,6 +285,22 @@ class chordMetadata : public metadataObject {
...
+
+    bool has_stream_ids() const {
+        std::lock_guard<std::mutex> lock(this->lock);
+        return metadata.contains(jsonMetadata::STREAM_IDS);
+    }
+
+    std::vector<uint32_t> get_stream_ids() const {
+        std::lock_guard<std::mutex> lock(this->lock);
+        return metadata.at(jsonMetadata::STREAM_IDS).template get<std::vector<uint32_t>>();
+    }
```
rhaas80, line 302: ah, they are coming back. Soon this will be just like production CHIME :-)

lib/stages/EigenN2Iter.cpp
```diff
@@ -217,22 +217,40 @@ void EigenN2Iter::main_thread() {
 
                 failed_buf->mark_frame_full(unique_name, failed_frame_id++);
             }
+
+            // Pass through vis data even when eigen computation fails
```
rhaas80: how does this interact with the option to pass failed eigen computations down a different buffer?

rhaas80: or is this to make things actually not stall if their is a failure and the passing to a different buffer is disabled?

lib/stages/PLMaskExpandedToCompact.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+        const uint64_t* exp_ptr = (const uint64_t*)in_frame;
+        uint64_t* comp_ptr = (uint64_t*)out_frame;
+
+        for (size_t t128 = 0; t128 < T_div_128; t128++) {
+            for (size_t f4 = 0; f4 < F_compact; f4++) {
+                for (size_t e = 0; e < E_div_8; e++) {
+
+                    // Start with all bits set (all valid)
+                    uint64_t compact_val = 0xffffffffffffffffULL;
```
rhaas80: one could write `~0` to avoid issues with having miscounted the `f`

lib/stages/TransposeBasebandArray.cpp
```diff
@@ -0,0 +1,451 @@
...
+    // All requirements are met
+    use_avx512_fast_path = true;
+    INFO("TransposeBasebandArray: AVX512 fast path enabled");
+#else
+    INFO("TransposeBasebandArray: AVX512 not available, using scalar path");
+#endif
+
+    // Set the output buffer frame ndarray shape
+
+    // Confusingly the array name is "E" for electic field
```
rhaas80: "electRic"

lib/stages/hdf5N2Write.cpp
```diff
@@ -52,6 +55,100 @@
 
 using namespace HighFive;
 
+/// Compute a hash of a file's contents for change detection.
+static size_t hash_file_contents(const std::string& path) {
+    std::ifstream ifs(path, std::ios::binary);
+    if (!ifs)
+        return 0;
+    std::string contents((std::istreambuf_iterator<char>(ifs)), std::istreambuf_iterator<char>());
```
rhaas80, line 62: so presumably not intended for use with a multi-hundred-GB output file?

lib/stages/hdf5N2Write.cpp
```diff
@@ -52,6 +55,100 @@
...
+    const char* path_c = evhttp_uri_get_path(uri);
+    std::string path = (path_c && path_c[0] != '\0') ? path_c : "/";
+    const char* query_c = evhttp_uri_get_query(uri);
+    if (query_c && query_c[0] != '\0') {
+        path += "?";
+        path += query_c;
+    }
+    evhttp_uri_free(uri);
+
+    INFO_NON_OO("hdf5N2Write: downloading gains file from http://{}:{}{}", host, port, path);
```
rhaas80: why a "gains" file. What about "losses"? :)

lib/stages/hdf5N2Write.cpp
```diff
@@ -52,6 +55,100 @@
...
+                           dest_path);
+        return;
+    }
+    ofs.write(reply.second.data(), static_cast<std::streamsize>(reply.second.size()));
+    if (!ofs) {
+        FATAL_ERROR_NON_OO("hdf5N2Write: failed to write {} bytes to '{}'",
+                           reply.second.size(), dest_path);
+        return;
+    }
+    ofs.close();
```
rhaas80, line 128: needs to error check here as well (in case not all data was flushed by write). 

> Effectively calls [rdbuf()->close()](https://en.cppreference.com/cpp/io/basic_filebuf/close). If an error occurs during operation, setstate(failbit) is called.

https://en.cppreference.com/cpp/io/basic_ofstream/close

lib/stages/hdf5N2Write.cpp
```diff
@@ -52,6 +55,100 @@
...
+    INFO_NON_OO("hdf5N2Write: wrote {} bytes to '{}'", reply.second.size(), dest_path);
+}
+
+/// Generate an acquisition base directory path: base_dir/acq_YYYYMMDD_HHMMSS_NNNNNNNNN
+static std::string make_acq_base_dir(const std::string& base_dir) {
+    // Strip trailing slashes
```
rhaas80, line 168: why? The OS does not mind them.

lib/stages/hdf5N2Write.cpp
```diff
@@ -52,6 +55,100 @@
...
+        FATAL_ERROR_NON_OO("hdf5N2Write: failed to write {} bytes to '{}'",
+                           reply.second.size(), dest_path);
+        return;
+    }
+    ofs.close();
+    INFO_NON_OO("hdf5N2Write: wrote {} bytes to '{}'", reply.second.size(), dest_path);
+}
+
+/// Generate an acquisition base directory path: base_dir/acq_YYYYMMDD_HHMMSS_NNNNNNNNN
+static std::string make_acq_base_dir(const std::string& base_dir) {
```
rhaas80: misleading name: does not make a directory (in the sense `mkdir` does), just constructs a pth (like the doxygen comment indicates).

lib/stages/hdf5N2Write.cpp
```diff
@@ -501,25 +511,45 @@ std::unique_ptr<HighFive::File> N2FileData::_open_or_create_file(const std::stri
...
-
-                _check_create_dataset(*file, "/digital_gains/gains_lin",
-                                      {gains_data->gains_lin.size()}, {"input"},
-                                      HighFive::create_datatype<uint16_t>(), props_empty);
-
-                _check_create_dataset(*file, "/digital_gains/gains_log",
-                                      {gains_data->gains_log.size()}, {"input"},
-                                      HighFive::create_datatype<uint16_t>(), props_empty);
+                try {
+                    HighFive::File src(baseband_gain_file, HighFive::File::ReadOnly);
```
rhaas80, line 555: could still fail if file is not accessible, or corrupted? Or does HighFive call `throw` in that case?

lib/stages/hdf5N2Write.cpp
```diff
@@ -501,25 +511,45 @@ std::unique_ptr<HighFive::File> N2FileData::_open_or_create_file(const std::stri
...
+
+                    // Copy entire gains file root group as /digital_gains
+                    herr_t err = H5Ocopy(src.getId(), ".", file->getId(), "digital_gains",
+                                         H5P_DEFAULT, H5P_DEFAULT);
+                    if (err < 0) {
+                        FATAL_ERROR_NON_OO("H5Ocopy failed copying gains file {} into output.",
+                                           baseband_gain_file);
```
rhaas80, line 562: well I guess this would catch a bad input file as well.

lib/stages/hdf5N2Write.hpp
```diff
@@ -92,12 +87,14 @@ class N2FileData {
...
     const std::string
-        gains_base_directory; // Base directory for gains. If empty/absent, gains are not written.
+        baseband_gain_file;             // Path to gains HDF5 file. If empty, gains are not written.
+    const int baseband_gain_update_idx; // update_time index (-1 = latest)
     const std::string partial_filepath; // working on-disk location
     const N2Layout n2_layout;           // visibility (N2) layout
 
     double last_update_wall_s;               // last frame receipt
     std::unique_ptr<HighFive::File> h5_file; // Working on-disk HDF5 file handle
+    size_t gains_file_hash = 0; // hash of gains file contents at copy time (0 = no gains)
```
rhaas80: why a `size_t` and not a integer type of defined bit-width? Hashes of 0 are valid I would think (and potentially valuable).

lib/stages/hdf5N2Write.cpp
```diff
@@ -1109,6 +1130,15 @@ void hdf5N2Write::main_thread() {
             const char* const msg = strerror(errno);
             FATAL_ERROR("Could not create directory \"{:s}\":\n{:s}", partial_dir.c_str(), msg);
         }
+
+        // If a URL was provided instead of a local path, fetch the gains file
+        // once into the partial dir and point _baseband_gain_file at the cached
+        // copy. From here on the existing local-file code path is used.
+        if (!_baseband_gain_url.empty()) {
+            const std::string cached = partial_dir + "/baseband_gains.h5";
```
rhaas80, line 1191: are conflicts that may arise if there are multiple independent instances of `hdf5N2write` running on the same node anticipated?

lib/stages/hdf5N2Write.cpp
jbmertens: Just for the gains, a few MB.

lib/stages/hdf5N2Write.cpp
jbmertens: Talks to fpga_master to get the digital gains. Could clarify this

lib/stages/hdf5N2Write.cpp
jbmertens: Not necessary, just looks a bit more tidy when printed.

lib/stages/hdf5N2Write.cpp
jbmertens: Both stages would also have to be writing to the same directory (very unlikely given the acq_*_NANOSECS path), but yes we can guard against that.

lib/stages/hdf5N2Write.hpp
jbmertens: std::hash returns a size_t, so yes it should be std::optional<size_t>.

lib/stages/EigenN2Iter.cpp
jbmertens: Ah, yes thanks I forgot about this, added it for development. I'll revert some of this, but would still like to set the emethod and zeros. Please have a look again and see if the behavior is ok for CHIME.

lib/stages/hdf5N2Write.cpp
jbmertens: The base_dir should now be unique in the config.

lib/dpdk/dpdkCore.cpp
andrerenard: Yep good call!

lib/dpdk/dpdkCore.cpp
andrerenard: Done.

lib/dpdk/dpdkCore.cpp
andrerenard: Ah, actually this is an tricky difference between the CHIME and CHORD designs...   I've pushed a bigger fix to this.

lib/metadata/chordMetadata.hpp
andrerenard: Although confusing they have a different meaning/use here for the CHORD F-engine.  I may have to some detangling of the ICE vs CRS logic at some point.

lib/stages/PLMaskExpandedToCompact.cpp
andrerenard: Done

andrerenard:
I think I've addressed all comments in my code.   This does change the CHIME packet capture logic a little.  With some optimizations to buffer shapes and sizes I've been able to get to nearly true 0% packet loss in the chime n2 configuration after startup.  

We shouldn't merge until I've been able to recheck for errors in the CHORD code path, but the Pathfinder is down at the moment, so testing will likely have to wait for next week.

jbmertens:
Good by me to merge then too, just needs a final lint pass.

jbmertens (approved):

PR #1402, opened 2026-01-26
Rhaas/n2 accumulate threads

rhaas80:
adds some speed optimization and multi-threading to N2Accumulate. Required for CHIME where 3 cores are needed to keep CPU load to 66% for each of them and not fall behind.

### DIFF
--- a/lib/stages/N2Accumulate.cpp
+++ b/lib/stages/N2Accumulate.cpp
@@ -152,8 +152,6 @@ N2Accumulate::N2Accumulate(Config& config, const std::string& unique_name,
     // _num_freq_per_n2k_frame (accumulate the full, blocked matrix x frequencies from the GPU)
     _vis = std::vector<int32_t>(2 * _num_freq_per_n2k_frame * _n2k_correlation_num_products,
                                 0); // vis with complex as 2 ints
-    _vis_even = std::vector<int32_t>(2 * _num_freq_per_n2k_frame * _n2k_correlation_num_products,
-                                     0); // store even vis matrix for weights calculation
     _weights = std::vector<float>(_num_freq_per_n2k_frame * _n2k_correlation_num_products,
                                   0.0f); // real-valued weights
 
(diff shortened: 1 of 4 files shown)


lib/stages/N2Accumulate.cpp
```diff
@@ -299,6 +299,13 @@ void N2Accumulate::main_thread() {
 
         // Accumulate each visibility sample in the in_frame
         // t_outer
+
+        // vis_even may be an odd frame since fpga_seq_num does not have to
+        // start from 0, in which case the very first visibilty accumulation
+        // will have an incorrect weight computed. The only alternative would
+        // be to skip until we reach the first even frame or to tie even / odd
```
geoffryan: To replicate the behaviour of CHIME's visAccumulate we'd want to skip here until we get to an even sample.

geoffryan (approved):
Overall behaviour looks good to me!  You bring up a good point about the initial fpga_seq_num not necessarily being even.  

In CHIME, the downstream timeDownsample stage requires starting on an even sample, so we'd want to enforce that here (otherwise I think timeDownsample will skip all frames and never output).  

That can be a different PR, or here if you want.  As-is N2Accumulate isn't enforcing this, so it isn't a regression, but should be fixed soon.

rhaas80:
Grr. Had a bug. `int32_t const* vis_even` has to be declared outside of the main thread loop (otherwise it gets reset each loop iteration).

geoffryan:
Ack.  Good catch.  I need to expand the test coverage in `test_n2_accumulate.py`, it should catch things like that.

geoffryan (approved):
LGTM (once checks pass)

lib/stages/N2Accumulate.cpp
```diff
@@ -343,13 +354,13 @@ void N2Accumulate::main_thread() {
             // Potential optimization: copying vis_even is only really
             // necessary if we've started accumulating a new frame
```
jbmertens: Should clean up this comment too.

rhaas80: that's still there? Ups.

rhaas80:
Alright, round two of reviews, please. This time with a bug fixed setting up the `vis_even` pointer and functionality to skip whole input frames until we have synched up with a nice started point (even and a multiple of the number of visframes we want to accumulate over).

geoffryan (approved):
LGTM!

PR #1403, opened 2026-01-26
Telescope Reorganization

geoffryan:
Reorganizes the *Telescope objects.
- Moves EOP and related functionality (REST endpoints) into the generic Telescope
- Moves `time0_ns` GET endpoint to generic Telescope
- Adds `to_time_ns` virtual function to generic Telescope

### DIFF
--- a/config/tests/testCHORDTelescope.yaml
+++ b/config/tests/testCHORDTelescope.yaml
@@ -71,6 +71,8 @@ gps_time:
   frame0_nano: 50 * nsec_in_jyear
 
 num_dishes: 32
+num_dishes_x: 16
+num_dishes_y: 16
 
 # Add a consumer stage that prints the output buffer to screen
 screen_dump:
--- a/config/tests/testCHORDTelescope.yaml
+++ b/config/tests/testCHORDTelescope.yaml
@@ -49,7 +49,7 @@ telescope:
   dish_vert_axis: [0.0, 0.0, 1.0]
   dish_coelev_deg: 0.0
 
-  updatable_config: "/earth_rotation_data"
+  eop_updatable_config: "/earth_rotation_data"
 
   dish_separation_x_m: 6.3
   dish_separation_y_m: 8.5
--- a/lib/utils/CHIMETelescope.cpp
+++ b/lib/utils/CHIMETelescope.cpp
@@ -17,8 +17,7 @@
 REGISTER_TELESCOPE(CHIMETelescope, "CHIMETelescope");
 
 CHIMETelescope::CHIMETelescope(const kotekan::Config& config, const std::string& path) :
-    ICETelescope(path,
-                 config.get<std::string>(path, "log_level"),
+    ICETelescope(path, config.get<std::string>(path, "log_level"),
                  config.get_default<std::string>(path, "updatable_config", "")) {
     INFO("Building CHIMETelescope");
 
--- a/lib/utils/CHIMETelescope.cpp
+++ b/lib/utils/CHIMETelescope.cpp
@@ -18,7 +18,7 @@ REGISTER_TELESCOPE(CHIMETelescope, "CHIMETelescope");
 
 CHIMETelescope::CHIMETelescope(const kotekan::Config& config, const std::string& path) :
     ICETelescope(path, config.get<std::string>(path, "log_level"),
-                 config.get_default<std::string>(path, "updatable_config", "")) {
+                 config.get_default<std::string>(path, "eop_updatable_config", "")) {
     INFO("Building CHIMETelescope");
 
     // This is always 1 for CHIME
--- a/lib/utils/CHIMETelescope.cpp
+++ b/lib/utils/CHIMETelescope.cpp
@@ -17,7 +17,10 @@
 REGISTER_TELESCOPE(CHIMETelescope, "CHIMETelescope");
 
 CHIMETelescope::CHIMETelescope(const kotekan::Config& config, const std::string& path) :
-    ICETelescope(config.get<std::string>(path, "log_level")) {
+    ICETelescope(path,
+                 config.get<std::string>(path, "log_level"),
+                 config.get_default<std::string>(path, "updatable_config", "")) {
+    INFO("Building CHIMETelescope");
 
     // This is always 1 for CHIME
     _num_freq_per_stream = 1;
--- a/lib/utils/CHORDTelescope.cpp
+++ b/lib/utils/CHORDTelescope.cpp
@@ -378,7 +378,7 @@ void GPSTimeParams::set_gps_time_params_from_remote(GPSTimeParams& gps, const st
 
 CHORDTelescope::CHORDTelescope(const kotekan::Config& config, const std::string& path) :
     Telescope(path, config.get<std::string>(path, "log_level"),
-              config.get_default<std::string>(path, "updatable_config", "")),
+              config.get_default<std::string>(path, "eop_updatable_config", "")),
     // Frequency sampling parameters
     _freq_params(FreqParams::from_config(config, path)),
     // GPS time configuration parameters
--- a/lib/utils/CHORDTelescope.cpp
+++ b/lib/utils/CHORDTelescope.cpp
@@ -1056,14 +1056,6 @@ EOP CHORDTelescope::build_EOP_from_update(int64_t time_ns, double delta_ut1_inst
     return eop;
 }
 
-bool EOP_comp_time(const EOP& eop1, const EOP& eop2) {
-    return eop1.t_inst < eop2.t_inst;
-}
-
-bool EOP_comp_ut1(const EOP& eop1, const EOP& eop2) {
-    return eop1.t_ut1 < eop2.t_ut1;
-}
-
 void to_json(nlohmann::json& j, const EOP& m) {
     assert(j.empty());
 
--- a/lib/utils/CHORDTelescope.cpp
+++ b/lib/utils/CHORDTelescope.cpp
@@ -389,7 +389,9 @@ CHORDTelescope::CHORDTelescope(const kotekan::Config& config, const std::string&
     INFO("Building CHORDTelescope");
 }
 
-CHORDTelescope::~CHORDTelescope() {}
+CHORDTelescope::~CHORDTelescope() {
+    INFO("Removing CHORDTelescope");
+}
 
 timespec CHORDTelescope::to_time(uint64_t seq) const {
     return nanosec_i64_to_timespec(to_time_ns(seq));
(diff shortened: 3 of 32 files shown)


config/tests/testCHORDTelescope.yaml
```diff
@@ -49,7 +49,7 @@ telescope:
   dish_vert_axis: [0.0, 0.0, 1.0]
   dish_coelev_deg: 0.0
 
-  updatable_config: "/earth_rotation_data"
+  eop_updatable_config: "/earth_rotation_data"
```
rhaas80, line 52: will this ever be changed? If not, could be hard-coded and reduces the number of parameters one has to provide.

config/tests/testCHORDTelescope.yaml
```diff
@@ -71,6 +71,8 @@ gps_time:
   frame0_nano: 50 * nsec_in_jyear
 
 num_dishes: 32
+num_dishes_x: 16
+num_dishes_y: 16
```
rhaas80, line 75: this is not a full array I guess? `num_dishes` being 32 does not match well with `num_dishes_x = num_dishes_y = 16`.

lib/utils/CHIMETelescope.cpp
```diff
@@ -17,7 +17,9 @@
 REGISTER_TELESCOPE(CHIMETelescope, "CHIMETelescope");
 
 CHIMETelescope::CHIMETelescope(const kotekan::Config& config, const std::string& path) :
-    ICETelescope(config.get<std::string>(path, "log_level")) {
+    ICETelescope(path, config.get<std::string>(path, "log_level"),
+                 config.get_default<std::string>(path, "eop_updatable_config", "")) {
```
rhaas80, line 21: since the `Telescope` constructor is called before `ICETelescope`'s constructor, can't it get it's config options in its own constructor?

lib/utils/CHIMETelescope.cpp
```diff
@@ -17,7 +17,9 @@
 REGISTER_TELESCOPE(CHIMETelescope, "CHIMETelescope");
```
rhaas80, line 17: isn't this missing the `_to_ns` function?

lib/utils/CHORDTelescope.cpp
```diff
@@ -382,88 +377,20 @@ void GPSTimeParams::set_gps_time_params_from_remote(GPSTimeParams& gps, const st
...
-        config.get<std::string>(path, "updatable_config"),
-        std::bind(&CHORDTelescope::receive_eop_updates, this, _1));
-
-    restServer& rest_server = restServer::instance();
-
-    rest_server.register_get_callback(path + "/time0_ns",
-                                      std::bind(&CHORDTelescope::send_time0_ns, this, _1));
-    rest_server.register_get_callback(path + "/eop_table",
-                                      std::bind(&CHORDTelescope::send_eop_table, this, _1));
+    INFO("Building CHORDTelescope");
```
rhaas80: is this leftover debug or do we really want "Building CHORDTelescope" show up in regular production log files?

lib/utils/CHORDTelescope.cpp
```diff
@@ -382,88 +377,20 @@ void GPSTimeParams::set_gps_time_params_from_remote(GPSTimeParams& gps, const st
...
-        reply["eop_table"] = _eop_table;
-    }
-    conn.send_json_reply(reply);
-}
-
-void CHORDTelescope::send_time0_ns(connectionInstance& conn) {
-    nlohmann::json reply;
-    reply["time0_ns"] = _gps_time_params.time0_ns;
-    conn.send_json_reply(reply);
+    INFO("Removing CHORDTelescope");
```
rhaas80: same as for the constructor

lib/utils/ICETelescope.cpp
```diff
@@ -170,6 +172,11 @@ timespec ICETelescope::to_time(uint64_t seq) const {
     return {(time_t)(time_ns / GIGA), (long)(time_ns % GIGA)};
 }
 
+int64_t ICETelescope::to_time_ns(uint64_t seq) const {
+    int64_t time_ns = time0_ns + seq * dt_ns;
```
rhaas80, line 176: ah, so CHIME inherits from here, ok.

lib/utils/Telescope.cpp
```diff
@@ -1,13 +1,52 @@
...
 #include <stdexcept> // for invalid_argument
 
+using kotekan::connectionInstance;
+using kotekan::restServer;
+
+#define GIGA 1'000'000'000L
 
-Telescope::Telescope(const std::string& log_level) {
+Telescope::Telescope(const std::string& tel_path, const std::string& log_level,
+                     const std::string& updatable_config_path) : _unique_name(tel_path) {
```
rhaas80: `updatable_config_path` could be more descriptive since it's really the `updatable_config_path` for the EOP value and not something else that the telescope could allow to be updated.

lib/utils/Telescope.cpp
```diff
@@ -1,13 +1,52 @@
...
     set_log_level(log_level);
     set_log_prefix("/telescope");
+
+    INFO("Building Telescope");
```
rhaas80: do we want this INFO output in production?

lib/utils/Telescope.cpp
```diff
@@ -1,13 +1,52 @@
...
+
+    // Set up callbacks for updating EOP and sending time0_ns
+    using namespace std::placeholders;
+
+    // Subscribe to config updates if updatable_config is set.
+    if (updatable_config_path.length() > 0) {
+        INFO("Subscribing {:s} to updatable config.", updatable_config_path);
```
rhaas80: DEBUG?

lib/utils/Telescope.cpp
```diff
@@ -1,13 +1,52 @@
...
+        kotekan::configUpdater::instance().subscribe(
+            updatable_config_path, std::bind(&Telescope::receive_eop_updates, this, _1));
+    }
+
+    INFO("Adding telescope REST GET endpoints");
+    restServer& rest_server = restServer::instance();
+
+    rest_server.register_get_callback(tel_path + "/time0_ns",
```
rhaas80, line 38: is this specific to the EOP stuff? Right now the CHIMETelescope already has a callback / way-of setting GPS time (and this seems similar)? So there may now be duplicate functionality in (some) Telescopes

lib/utils/Telescope.cpp
```diff
@@ -1,13 +1,52 @@
...
+    rest_server.register_get_callback(tel_path + "/eop_table",
+                                      std::bind(&Telescope::send_eop_table, this, _1));
+}
+
+Telescope::~Telescope() {
+    // Must manually remove the GET callbacks
+    restServer& rest_server = restServer::instance();
+    rest_server.remove_get_callback(_unique_name + "/time0_ns");
+    rest_server.remove_get_callback(_unique_name + "/eop_table");
+    INFO_NON_OO("/telescope: Removed REST GET endpoints");
```
rhaas80: this uses NON_OO while above regular INFO was used.

lib/utils/Telescope.cpp
```diff
@@ -20,6 +59,8 @@ const Telescope& Telescope::instance() {
 
 const Telescope& Telescope::instance(const kotekan::Config& config) {
 
+    // This defaults to ICETelescope because the "Telescope" is virtual
+    // and is not registered with the Factory.
```
rhaas80, line 63: inconvenient

lib/utils/Telescope.cpp
```diff
@@ -54,3 +95,241 @@ timespec Telescope::seq_length() const {
     auto dt_ns = seq_length_nsec();
     return {(time_t)(dt_ns / 1000000000), (long)(dt_ns % 1000000000)};
 }
+
+bool Telescope::receive_eop_updates(nlohmann::json& json) {
+    try {
+        // Fill a temporary table with the updated values.
+        std::vector<EOP> tmp_eop_table;
+        for (const auto& elem : json.at("earth_orientation_parameter_table")) {
+            INFO("Telescope EOP update: {:s}", elem.dump());
```
rhaas80, line 104: do we want this INFO output in production? Or DEBUG?

lib/utils/Telescope.cpp
```diff
@@ -54,3 +95,241 @@ timespec Telescope::seq_length() const {
...
+            .ERA_deg = era,
+            .xp_as = xp_as,
+            .yp_as = yp_as};
+
+    return eop;
+}
+
+size_t Telescope::get_EOP_table_len() const {
+    std::shared_lock lock(_eop_lock);
+    return _eop_table.size();
```
rhaas80: nohting stops this len from becoming invalid in between being queried here and once the value is actually used to do something (allocate a sort of buffer?), does it?

lib/utils/Telescope.hpp
```diff
@@ -180,6 +189,11 @@ class Telescope : public kotekan::kotekanLogging {
      **/
     virtual timespec to_time(uint64_t seq) const = 0;
 
+    /**
+     * @brief   Return the time corresponding to the given fpga sequence number as an int64_t.
```
rhaas80: "time in ns"

lib/utils/Telescope.hpp
```diff
@@ -217,16 +231,106 @@ class Telescope : public kotekan::kotekanLogging {
...
+     * @brief   Return the number of entries in the EOP table.
+     **/
+    size_t get_EOP_table_len() const;
+
+    /**
+     * @brief   Return the EOP table entry at an index.
+     *
+     * @param   i   Index of desired EOP entry, 0 <= i < EOP_table_len
+     **/
+    EOP get_EOP_at_table_idx(uint64_t i) const;
```
rhaas80: I'd argue for a `size_t` instead of `uint64_t` to reserve the bit-wide types for science data and the C++ standard types for counters etc.

lib/utils/Telescope.hpp
```diff
@@ -180,6 +189,11 @@ class Telescope : public kotekan::kotekanLogging {
      **/
     virtual timespec to_time(uint64_t seq) const = 0;
 
+    /**
+     * @brief   Return the time corresponding to the given fpga sequence number as an int64_t.
+     *          Uses the epoch of time0_ns.
```
rhaas80: I'd spell the epoch out here, or otherwise use doxygen command to make sure `time0_ns` become a link if one produces docs.

lib/utils/Telescope.hpp
```diff
@@ -217,16 +231,106 @@ class Telescope : public kotekan::kotekanLogging {
...
+     * @brief   Callback to send time0_ns value
+     *
+     * @param   conn    Kotekan connection.
+     */
+    void send_time0_ns(kotekan::connectionInstance& conn);
+
+    /**
+     * @brief   Build a single EOP struct from config values
+     *
+     * @param   t_ns    Instrument time in nanoseconds.
```
rhaas80, line 354: this defines the epoch? Since it says "instrument time" I guess "yes".

lib/utils/Telescope.hpp
```diff
@@ -217,16 +231,106 @@ class Telescope : public kotekan::kotekanLogging {
...
+     * @param   delta_ut1_inst  Diff between UT1 and Instrument time in seconds
```
rhaas80, line 355: why not combine with the parameter above? Or are both needed?

config/tests/testCHORDTelescope.yaml
geoffryan: This is the standard pattern for `updatable_config` fields, to give a pointer to where the field actually lives. A little verbose, but since it points to a location outside the Telescope's tree, I'd rather not hard code it in and lock other parts of the config to assumptions I make here.

config/tests/testCHORDTelescope.yaml
geoffryan: `num_dishes` will almost never be `num_dishes_x * num_dishes_y`.  The latter parameters specify the maximum grid size in each dimension, num_dishes is the actual number of dishes we're processing data for.   CHORD and CHORD Pathfinder will have gaps and unfilled rows, for instance.

lib/utils/CHIMETelescope.cpp
geoffryan: This is a bit of a kludge, was just following what the Telescope/ICE*/CHIME* chain was already doing.   ICETelescope sets up a forwarding constructor to intercept these calls and pass them straight to Telescope, BUT it also has its own ICETelescope(config, path) constructor for making an ICETelescope directly.  The issue here is CHIMETelescope does not want to call the ICETelescope(config, path) constructor at all, but its not allowed to call the Telescope constructor directly, so it has to bypass by a function with a different signature.  

I could make the Telescope constructor something like Telescope(path, config) instead, but that also seems like a kludge.  Really, the Telescope can never be instantiated on its own, so it not having direct config access and having this (string, string, string) constructor sort of marks that... open to refactoring suggestions, didn't want to mess with ICE/CHIME too much.

lib/utils/CHIMETelescope.cpp
geoffryan: Nope, you found it below. ~(although I should probably re-implement it here since CHIMETelescope has re-implements its GPS time functionality)~ CHIMETelescope only implements a couple functions, leaves all its time handling (other than in the constructor) to ICETelescope.

lib/utils/CHORDTelescope.cpp
geoffryan: Leftover debug.

lib/utils/Telescope.cpp
geoffryan: renamed `eop_updatable_config_path`

lib/utils/Telescope.cpp
geoffryan: now DEBUG

lib/utils/Telescope.cpp
geoffryan: Probably erring on verbosity but this triggers a chain of calls that I think warrants a note that its happening.

lib/utils/Telescope.cpp
geoffryan: This is different functionality, it allows an external actor to ask the Telescope what its configured time0 is.  The GPS time setters are for setting this value initially.

geoffryan: It's needed for the EOP update script to compute instrument times properly.

lib/utils/Telescope.cpp
geoffryan: correct.

lib/utils/Telescope.cpp
geoffryan: I think this is ok as INFO, it should only occur ~once a day and the table will only have a few entries.

lib/utils/Telescope.cpp
geoffryan: It does not.  Replaced this and `get_EOP_at_table_idx()` with a function that just returns a copy of the current table.

lib/utils/Telescope.hpp
geoffryan: Better solution, remove the function altogether ;)

lib/utils/Telescope.hpp
geoffryan: I mean, in leaving this general, we're in principle allowing derived classes to set their own epochs.  This comment is also a little out-of-date because `time0_ns` doesn't even exist in the Telescope, that's an ICE/CHIME/CHORD-specific field.

geoffryan: Updated to say "Returns the Instrument time..." which I think is the correct thing to do here.

lib/utils/Telescope.hpp
geoffryan: Yes, clarified this in a few places (and there's a longer writeup in `timeUtil.hpp`).  Instrument time is a UNIX time unless we are close to a leap second occurring, or one has occurred since instrument startup.  Really Instrument time is the value of the UNIX time at instrument start PLUS the TAI time that has elapsed since startup.

lib/utils/Telescope.hpp
geoffryan: Both are needed.  We need to have values for instrument and UT1 time at the same instant. Two pieces of info.

lib/utils/Telescope.cpp
geoffryan: Uses INFO now.

jbmertens (commented):
LGTM, but I will let Roland approve per his questions/requests.

lib/utils/CHIMETelescope.cpp
rhaas80: it's passed a `kotekan::Config` via `Telescope::instance()` though so could pull this out on its own rather than the sub-classes having to know which option `Telescope` needs for its functionality. Anyway, I'll take a look and maybe come up with "nice-up" pull request afterwards. This pull request implements needed right now.

config/tests/testCHORDTelescope.yaml
rhaas80: so this become the absolute path of the REST endpoint ie. `http://<ip>:<port>/earth_rotation_data` and not eg `http://<ip>:<port>/${unique_name}/earth_rotation_data`? Fine with me then, I guess it means that the endpoint stays constant even if the layout of the input yaml file is reorganized, which is definitely a plus. :+1:

rhaas80:
> LGTM, but I will let Roland approve per his questions/requests.

meh I can always submit a fix pull request afterwards :-)

rhaas80 (approved):

config/tests/testCHORDTelescope.yaml
geoffryan: Correct!

PR #1404, opened 2026-01-27
N2 subset

jbmertens:
Add support for different N2 subset types: add (fix) N2Subset stage, add product determination logic to N2FrameDesc.

(And unrelated pipeline viewer changes.)


lib/core/bufferFactory.cpp
```diff
@@ -90,15 +89,7 @@ GenericBuffer* bufferFactory::new_buffer(const string& type_name, const string&
...
-        // Read N2 parameters from config
-        const uint32_t num_elements = config.get<uint32_t>(location, "num_elements");
-        const uint32_t num_ev = config.get<uint32_t>(location, "num_ev");
-        const N2Layout n2_layout = config.get<N2Layout>(location, "n2_layout");
-        const uint32_t num_prod = N2FrameDesc::get_num_prod(num_elements, n2_layout);
-
-        // Create the frame descriptor
-        n2_frame_desc =
-            std::make_shared<kotekan::N2FrameDesc>(num_elements, num_ev, num_prod, n2_layout);
+        n2_frame_desc = N2FrameDesc::from_config(config, location);
```
geoffryan: Minor, but it feels weird for `n2_frame_desc` to be declared outside this if/else block.  Should this instead be a generic frame_desc (called `config_frame_desc` or something) which is then set on the buffer below?

This would make it a bit less clunky for other buffer types to add their own config'd frame_desc in the future.

lib/stages/N2Subset.cpp
```diff
@@ -72,10 +34,73 @@ N2Subset::N2Subset(Config& config, const std::string& unique_name,
...
+    if (!out_desc) {
+        FATAL_ERROR("N2Subset: out_buf does not have an N2FrameDesc");
+    }
+
+    // Validate num_elements and num_ev match (required for copying non-product fields)
+    if (in_desc->get_num_elements() != out_desc->get_num_elements()) {
+        FATAL_ERROR("N2Subset: num_elements mismatch: in_buf has {:d}, out_buf has {:d}",
+                    in_desc->get_num_elements(), out_desc->get_num_elements());
+    }
+    if (in_desc->get_num_ev() != out_desc->get_num_ev()) {
```
geoffryan, line 69: Potential future feature: allow `num_ev_out < num_ev_in` so this stage could also serve as a way to drop extraneous ev's?  Don't know if anyone would want this, but it would be easy to do here.

lib/metadata/N2FrameDesc.hpp
```diff
@@ -76,17 +89,95 @@ class N2FrameDesc : public FrameDesc {
...
+     * @brief Get the explicit product list for this frame descriptor.
+     *
+     * For FullUpperTri and Autocorrelations layouts, this returns an empty vector
+     * (products are computed on-the-fly by fill_prod_maps).
+     * For subset layouts (InputANDMasked, InputORMasked, GeneralSubset), this returns
+     * the explicit list of products stored in this descriptor.
+     *
+     * @return Reference to the product list (may be empty for computed layouts)
+     */
+    const std::vector<N2::prod_ctype>& get_product_list() const {
```
geoffryan, line 99: There is some overlap in behaviour/logic between `get_product_list` and `fill_prod_map` that might be nice to de-tangle.  Perhaps the "calculating the prod list" for FullUpperTri and Autocorrelations could happen here, and fill_prod_map would just do a copy?  Or, fill prod map could be replaced with a function that just gives a shared_ptr to the internal product list that *all* N2FrameDescs would have?

One thing to avoid would be unnecessary construction of a product list (which can be ~10^6 elements long) in functions that are called more than once.  I'm thinking specifically of the `calculate_frame_size(config, path)` function which creates a new N2FrameDesc to get its size... we wouldn't want that to *also* build a 10^6 long array just to throw it away every time the size is requested.

geoffryan (commented):
Looks good!  A couple suggestions to think about, happy to approve if you think this is fine as-is.

lib/stages/N2Subset.cpp
jbmertens: Agreed, though will punt on this until later.

lib/metadata/N2FrameDesc.hpp
jbmertens: Hm - agreed.

The `calculate_frame_size(config, path)` variant isn't actually used now that the bufferFactory doesn't call it, so I can remove that altogether.

geoffryan (approved):
LGTM!

PR #1407, opened 2026-02-04
n2k: Correct ring buffer checks

eschnett:
C++ operator precedence. Enough said.


jbmertens (approved):
LGTM!

rhaas80:
oh great. Bitwise operations are lower in precendence order than comparisons. Very odd. Thanks for catching this.

PR #1409, opened 2026-02-06
Correct FRB beamformer

eschnett:
zero is eight, which is negative eight.


eschnett:
I set unused dish locations to the bit pattern `0x00`, which is wrong. They need to be set to `0x88`, which corresponds to the value `0`.

julia/kernels/frb.jl
```diff
@@ -590,7 +590,7 @@ function write_Fsh2!(emitter)
         delete!(layout, Dish(:dish, W, RF1))
         emitter.environment[:Freg1′] = layout
         real_dish_value = Symbol(:Freg1_dish, string(W * i))
-        zero_dish_value = zero(Int4x8)
+        zero_dish_value = chimify(zero(Int4x8))
```
jbmertens, line 593: Is this correct for CHORD?

jbmertens (approved):

julia/kernels/frb.jl
eschnett: Yes, we're using the same encoding in CHORD as in CHIME – 8-bit offset-encoded with the real part in the high bits.

PR #1410, opened 2026-02-10
config updates: dish info, smaller subset, and gpu rfi

jbmertens:


lib/stages/TransposeBasebandArray.cpp
```diff
@@ -197,7 +197,7 @@ void TransposeBasebandArray::transpose_block_avx512(const uint8_t* in, uint8_t*
         // Gather next 64 bytes (elements 64-127 from e_long 8-15)
         __m512i data_hi = _mm512_i64gather_epi64(indices_hi, (const long long*)base, 1);
 
-        if (unlikely(check_for_zero_nibbles)) {
```
andrerenard, line 200: Why remove this?

lib/utils/CHORDTelescope.cpp
```diff
@@ -186,11 +186,13 @@ void GeographicParams::set_dish_info(const kotekan::Config& config, const std::s
         // Catch inconsistencies
         assert(dish_info.idx == dish);
         dish_index_t& dish_index = dish_grid.dish_index(dish_info.grid_x_idx, dish_info.grid_y_idx);
-        // Catch inconsistencies
+        // Not catching inconsistencies -- Pathfinder is wired bad
```
andrerenard: I don't love having commented out code, easy to forget about.   Can we make this a config option that we have to force disable in the PF configs?

lib/stages/hdf5N2Write.hpp
```diff
@@ -68,8 +68,8 @@ class N2FileData {
 public:
     enum FileMode { CHORD, CHIME };
     struct DigitalGains {
-        std::vector<std::uint16_t> gains_lin;
-        std::vector<std::uint16_t> gains_log;
+        std::vector<std::vector<std::uint16_t>> gains_lin;
```
andrerenard, line 71: Are the axis of these vectors recorded some place?  Perhaps include a comment here?

lib/stages/hdf5FileWrite.cpp
```diff
@@ -81,6 +81,10 @@ class hdf5FileWrite : public kotekan::Stage {
     const bool skip_writing = config.get_default<bool>(unique_name, "skip_writing", false);
     const bool create_single_file =
         config.get_default<bool>(unique_name, "create_single_file", false);
+    const int64_t write_x_frames =
+        config.get_default<int64_t>(unique_name, "write_x_frames", -1);
+    const int64_t per_y_frames =
+        config.get_default<int64_t>(unique_name, "per_y_frames", -1);
```
andrerenard, line 95: Add some documentation to the header for how these work?

julia/Project.toml
```diff
@@ -1,24 +1,50 @@
 name = "Kotekan"
 uuid = "4b3ecac4-cfa8-46ac-ae66-cdd0e0e0ae48"
-version = "0.1.0"
 authors = ["Erik Schnetter <schnetter@gmail.com>"]
+version = "0.1.0"
 
 [deps]
+ASDF = "686f71d1-807d-59a4-a860-28280ea06d7b"
```
andrerenard: Just checking these changes should be coming from here, and not some other branch heading (or already in) in `chord`?

config/chord_pathfinder_rfi.yaml.part
```diff
@@ -0,0 +1,274 @@
+## This is just a yaml snippet for reference.
+## Eventually something like this should used for RFI flagging.
```
andrerenard, line 2:
```suggestion
## Eventually something like this should used for RFI flagging.
## Delete once included with one of the production configs. 
```

config/chord_pathfinder_recv.j2
```diff
@@ -0,0 +1,281 @@
...
+
+buffer_recv_subset_n2_1:
+    kotekan_stage: bufferRecv
+    buf: n2_recv_subset_buffer_1
+    listen_port: 11025
+
+# buffer_recv_subset_n2_2:
+#     kotekan_stage: bufferRecv
+#     buf: n2_recv_subset_buffer_2
+#     listen_port: 11026
```
andrerenard: Why do we have these extra commented out receivers?   This code has the option to use worker threads, so we shouldn't need this for performance reasons.

config/chord_pathfinder_1numa.j2
```diff
@@ -373,10 +402,23 @@ transpose_voltage_1:
     cpu_affinity: [3]
     timesamples_per_frame: samples_per_data_set
     in_buf: network_input_buffer_0
-    pl_mask_buf: host_pl_mask_buffer
+    pl_mask_buf: host_pl_mask_exp_buffer
     out_buf: host_voltage_buffer
     frame_mode: odd
 
+# Generate fake bad feed mask filled with ones for now (all feeds good)
+generate_bf_mask:
```
andrerenard: I think this isn't actually used for this base config right?

config/chord_pathfinder_1numa.j2
```diff
@@ -63,20 +68,23 @@ log_level: INFO
 
 numa_node: 0
 
-freq_ids: {{ freq_ids_list | list }}
```
andrerenard, line 66: Glad to see this removed!

config/tests/point_source_pathfinder.yaml
```diff
@@ -0,0 +1,122 @@
+---
```
andrerenard, line 1: Short comment on what this does?

andrerenard (changes requested):
Just tiny changes/comments.

lib/stages/TransposeBasebandArray.cpp
jbmertens: Wasn't compiling for me with that.

andrerenard: Ah yeah, we can remove it.

lib/stages/hdf5N2Write.hpp
jbmertens: Future item to do.

config/tests/point_source_pathfinder.yaml
jbmertens: Future item to do.

geoffryan: done.

lib/stages/hdf5N2Write.hpp
geoffryan: `fpga_master` produces different fields than this, but is still being worked on.  This is fine for now but can be updated shortly once we have fpga_master running

jbmertens (commented):
Comments

andrerenard (approved):

PR #1413, opened 2026-02-11
cleanup and dataset handling for N2Accumulate to use within CHIME

rhaas80:
should not affect CHORD use at all, just adds some optional configuration option and removes all CHORDTelescope files in favor of just Telescope and EOP, which is what is actually used.


geoffryan (approved):
LGTM!

PR #1414, opened 2026-02-13
Correct xpose2048 kernel

eschnett:
- Traverse only the valid region of the ring buffer, not the whole ring buffer (d'oh!)
- Flag invalid inputs (bit pattern 0x00), aborting if found
- Require newer CUDASIMDTypes which supports this
- Add option to build in debug mode to catch integer overflows
- Clean up template for C++ code
- Allow checking input buffers for poison, not just output buffers
- The 16 separate CHIME input buffers have 1 frequency each, not 0


rhaas80:
> * Traverse only the valid region of the ring buffer, not the whole ring buffer (d'oh!)
> 
>     * Flag invalid inputs (bit pattern 0x00), aborting if found
> 
>     * Require newer CUDASIMDTypes which supports this
> 
>     * Add option to build in debug mode to catch integer overflows
> 
>     * Clean up template for C++ code
> 
>     * Allow checking input buffers for poison, not just output buffers
> 
>     * The 16 separate CHIME input buffers have 1 frequency each, not 0

let me take a look

julia/kernels/xpose2048_template.cxx
```diff
@@ -378,10 +375,10 @@ cudaEvent_t cuda{{{kernel_name}}}::execute(cudaPipelineState& /*pipestate*/, con
 
     // Pass time spans to kernel
     // The kernel will wrap the upper bounds to make them fit into the ringbuffer
-    Tin_min_arg = 16384 * mod(Tin_min, Tin_ringbuf);
-    Tin_max_arg = 16384 * (mod(Tin_min, Tin_ringbuf) + Tin_length);
```
rhaas80, line 382: this is the actual bugfix I think? I am surprised this worked before.

julia/kernels/xpose2048.jl
```diff
@@ -57,12 +57,13 @@ const Time = Index{Physics,TimeTag} # time
 # Setup
 
 @assert D * P == 2048
-@assert T % 16384 == 0
 
-const T1 = 32
+const T16384 = 16384
+const Tin = idiv(T, T16384)
```
rhaas80, line 62: ah, this goes along with the removal of 16384 in the template C++ code I guess?

rhaas80 (approved):
I do not see anything obviously wrong. I cannot claim to understand the Julia code (Julia syntax wise or what it actually does).

rhaas80:
hmm, when running causes issues with cudaCopyNToRingbuffer being unhappy about inconsistent time values. Will test with just an FEhgine setup or at least something with less custom code of mine.

Was indeed caused by code of my own that still had the "E" array as 2 dimensional ("T" and "E" dimensions). Changing to 3 dimensional (with an "F" axis of size 1) removes the error message.

However dpdk also produces "E" frames with just "T" and "E" dimensions so will trigger the same issue I suspect, unless updated. Updating is trivial, but just to be sure: this is intentional?

julia/kernels/xpose2048_template.cxx
eschnett: `Tin` should probably be called `T16384`. The CHIME input buffers have the wrong layout and come in "chunks" of 16384 times. `Tin` counts these chunks (we always process 1 at a time at the moment). Previously I multiplied by 16384 here to "simplify the code", but this confused me, so I'm not doing that any more. I now index the arrays properly on the GPU.

julia/kernels/xpose2048.jl
eschnett: Yes.

eschnett:
Yes, this is intentional. These arrays contains 1 frequency, and should be declared as 3d arrays with a length of 1 in the frequency direction. (This doesn't change the layout.) It fixes the way metadata are handled – it's difficult to attach frequency metadata (e.g. coarse frequency index etc.) to an array that doesn't have any frequencies. I briefly thought about adding a special case, but I think that properly stating that these input arrays have a frequency axis (even if with length 1) is better.

dpdk doesn't need to change its data, just the metadata it produces. one of the axes (the outermost) needs to be the frequency axis so that we can glue the arrays together.

rhaas80:
> Yes, this is intentional. These arrays contains 1 frequency, and should be declared as 3d arrays with a length of 1 in the frequency direction. (This doesn't change the layout.) It fixes the way metadata are handled – it's difficult to attach frequency metadata (e.g. coarse frequency index etc.) to an array that doesn't have any frequencies. I briefly thought about adding a special case, but I think that properly stating that these input arrays have a frequency axis (even if with length 1) is better.
> 
> dpdk doesn't need to change its data, just the metadata it produces. one of the axes (the outermost) needs to be the frequency axis so that we can glue the arrays together.

The outermost? The changes here make "F" the middle axis:

https://github.com/kotekan/kotekan/pull/1414/files#diff-4e38d0f8edcde5111294044bdb2a587edc7287b634b22db47cffc4774392a9deR163

In any case a change to dpdk is very simple since it only affects how it presents its metadata, yes.

lib/cuda/cudaCopyNToRingbuffer.cpp
```diff
@@ -156,12 +156,14 @@ cudaEvent_t cudaCopyNToRingbuffer::execute(cudaPipelineState& /*pipestate*/,
...
         // Set the shape of the array
+        assert(meta_in0->dims == 3);
         meta_ring->dims = 4;
         meta_ring->set_array_dimension(0, _gpu_buffer_depth, "Thi16384");
         meta_ring->set_array_dimension(1, in_buffers.size(), "F");
-        assert(meta_in0->dim[0] == 16384);
+        assert(meta_in0->dim[0] == 16384); // T
         meta_ring->set_array_dimension(2, meta_in0->dim[0], "Tlo16384");
-        meta_ring->set_array_dimension(3, meta_in0->dim[1], "E");
+        assert(meta_in0->dim[1] == 1); // F
```
rhaas80: middle axis

eschnett: Maybe it should be the outermost axis? I might change this.

rhaas80: it is all just window dressing to some extent. Since these are not ringbuffers it does not really matter if the outermost axis is a "time" axis or not.

eschnett: Fixed.

PR #1418, opened 2026-02-24
chime n2 input file and converter stages

rhaas80:
This introduces a converter stage from N2FrameView to CHIME's VisFrameView expected by the science receivers. Adds a trial set of input files (quite kludgy right now since they borrow from the fengine tests).

### DIFF
--- a/config/chime_upgrade_n2.j2
+++ b/config/chime_upgrade_n2.j2
@@ -1,0 +1,2897 @@
+##########################################
+#
+# chime_upgrade_n2
+#
+# Pipeline for running the n2k correlator, accumulate onto a visibility matrix
+# and compute eigen-decomposition.
+#
+# Currently only uses one NUMA domain, but it shouldn't be hard to extend to
+# the second NUMA domain.
+#
+# Author: Andre Renard, Roland Haas, Erik Schnetter
+#
+##########################################
+---
+########################
+#### Global options ####
+########################
+type: config
+# Logging level can be one of:
+# OFF, ERROR, WARN, INFO, DEBUG, DEBUG2 (case insensitive)
+# Note DEBUG and DEBUG2 require a build with (-DCMAKE_BUILD_TYPE=Debug)
+log_level: info
+num_polarizations: 2
+num_dishes: 1024
+num_elements: num_polarizations * num_dishes
+num_components: 2 # Re/Im
+num_freq_host_buf: 1
+{% set num_freq_gpu_buf = 16 %}
+num_freq_gpu_buf: {{ num_freq_gpu_buf }}
+num_ev_to_calculate: 4            # from n2 test case
+samples_per_data_set: 16384
+sample_period_ns: 2560
+network_frame_arrival_period: 16384 * sample_period_ns / 1e9
+baseband_buffer_depth: 3*272 # 3*272 = ~33 seconds after accounting for active frames
+# see where / if / how to use host_buffer_depth and gpu_buffer_depth
+host_buffer_depth: 12
+gpu_buffer_depth: 4
+num_gpus: 2
+cpu_affinity: [0, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19]
+
+# Constants
+sizeof_float: 4
+sizeof_float32: 4
+sizeof_int8: 1
+sizeof_int: 4
+sizeof_int32: 4
+sizeof_uint64: 8
+sizeof_short: 2
+
+use_hugepages: true
+
+# Telescope
+telescope:
+  name: CHIMETelescope
+  require_gps: true
+  query_gps: true # GPS time is provided via coco
+  query_frequency_map: true # Frequency map is provided via coco
+  require_frequency_map: true # Make sure we have the full frequency map
+  allow_default_frequency_map: false # Don't fall back on default map
+  eop_updatable_config: /earth_rotation_data
+
+earth_rotation_data:
+  kotekan_update_endpoint: json
+  earth_orientation_parameter_table:
+    - {time_inst_ns: 0, delta_UT1_inst: 0.0, x_pm: 0.0, y_pm: 0.0}
+
+dataset_manager:
+  use_dataset_broker: False
+
+main_pool:
+  kotekan_metadata_pool: chordMetadata
+  num_metadata_objects: 30 * host_buffer_depth + 5 * baseband_buffer_depth
+
+# Baseband metadata pool
+baseband_pool:
+  kotekan_metadata_pool: BasebandMetadata
+  num_metadata_objects: 30 * host_buffer_depth * 32
+
+n2_pool:
+    kotekan_metadata_pool: N2Metadata
+    num_metadata_objects: num_freq_gpu_buf * host_buffer_depth
+
+vis_pool:
+    kotekan_metadata_pool: VisMetadata
+    num_metadata_objects: num_freq_gpu_buf * gpu_buffer_depth # same as n2_pool
+
+
+################################################################################
+# Packet intake
+
+# Buffers
+
+host_voltage_buffers:
+  num_frames: baseband_buffer_depth
+  frame_size: samples_per_data_set * num_elements * num_freq_host_buf
+  metadata_pool: main_pool
+  numa_0:
+    numa_node: 0
+{%- for id in range(num_freq_gpu_buf) %}
+    host_voltage_buffer_{{ id }}:
+      kotekan_buffer: standard
+{%- endfor %}
+host_voltage_ringbuffer:
+  kotekan_buffer: ring
+  ring_buffer_size: gpu_buffer_depth * samples_per_data_set * num_elements * num_freq_gpu_buf
+  metadata_pool: main_pool
+
+{%- for id in range(num_freq_gpu_buf // 4) %}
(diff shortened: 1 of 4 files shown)


lib/stages/N2FrameToVisFrame.hpp
```diff
@@ -0,0 +1,69 @@
...
+    void main_thread() override;
+
+private:
+    void register_base_dataset_states(std::string& instrument_name,
+                                      std::vector<std::pair<uint32_t, freq_ctype>>& freqs,
+                                      std::vector<input_ctype>& inputs,
+                                      std::vector<prod_ctype>& prods);
+
+private:
+    datasetManager& dm = datasetManager::instance();
```
tristpinsm: would it be better if the actual instance was fetched in the `.cpp` file? I'm not sure I understand how the order of operations would work here, and I don't know if the dataset manager is certain to have been instantiated from the config at this point.

lib/stages/N2FrameToVisFrame.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include <string>   // for string
+#include <vector>   // for vector
+
+/**
+ * @brief Merges and expands CHIME lost_samples buffers (1 bytes per sample for 4
+ * frequencies) into a CHORD style package loss mask (1 bit per downsampled
+ * time, per 4 frequencies, per polarization, per dish).
+ *
+ * @par Buffers
+ * @buffer pl_mask_buf Kotekan buffer for package loss mask
```
tristpinsm: doesn't seem like this is the right docstring

lib/stages/N2FrameToVisFrame.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+            first_time = false;
+        }
+        vis_frame.dataset_id = ds_id_vis;
+
+        vis_frame.fpga_seq_length = n2_frame.frame_length_fpga_ticks;
+
+        // items set in visAccumulate::main_thread
+        vis_frame.fpga_seq_total =
+            n2_frame.n_valid_fpga_ticks; // RH: do I need to add rfi_total to this?
```
tristpinsm: no, `fpga_seq_total` counts the number of FPGA frames that were accumulated. `rfi_total` counts the number of frames that were excised due to RFI.

This way we can track the fraction of frames lost, and of those the faction that were lost to RFI. This is what is eventually [saved in the archive files](https://github.com/kotekan/kotekan/blob/64ab9a7016724750b777ac953d48797f3dc9e542/lib/stages/VisTranspose.cpp#L282).

config/chime_upgrade_n2.j2
```diff
@@ -0,0 +1,680 @@
...
+sizeof_float: 4
+sizeof_float32: 4
+sizeof_int8: 1
+sizeof_int: 4
+sizeof_int32: 4
+sizeof_uint64: 8
+sizeof_short: 2
+
+# ADC
+adc_frequency: 1.6e+9           # [Hz]
```
tristpinsm: this is for CHORD presumably?

config/chime_upgrade_n2.j2
```diff
@@ -0,0 +1,680 @@
...
+    out_buf: vis_10s_buffer
+    packet_loss_is_scalar: True
+    num_threads: 3
+    # round_to_even(10 / ( n2k_samples_per_data_set * frame_arrival_period ) * n2k_num_subintegrations)
+    # round_to_even(10s / ( 16384 * 2.56us ) * 1) = 238
+    num_n2k_samples_to_accumulate: 238
+    # these recap values from cudaCorrelator
+    num_freq_per_n2k_frame: num_frequencies
+    samples_per_data_set: n2k_samples_per_data_set
+    sub_integration_ntime: n2k_sub_integration_ntime
```
tristpinsm, line 788: is it me or are these `n2k_*` parameters not defined here?

tristpinsm (commented):
Good work tracking the path through all the stages in the old code! It does end up being more compact now...

As far as I can tell it is all there, with a few comments. It will be good to see the result of testing.

lib/stages/N2FrameToVisFrame.hpp
rhaas80: this is shorthand in newer version of C++ for setting the variable in (well outside of) the constructor. So is the same as:

```
N2FrameToVisFrame::N2FrameToVisFrame(/* ... */) : dm(datasetManager::instance()) {
  // constructor code
}
```

it needs to be a reference since one cannot copy (or destroy) a `dataSetManager` instance.

Almost certainly I copied this code from somewhere else since I normally do tend to use the older method of initializing things in the constructor.

I'll take a look to see which one gives more uniform code.

lib/stages/N2FrameToVisFrame.hpp
rhaas80: right, I have to update all these comments and docstrings.

lib/stages/N2FrameToVisFrame.cpp
rhaas80: very good. I'll remove the comment. Thanks.

config/chime_upgrade_n2.j2
rhaas80: the same is set for CHIME as well in the FEngine test files where I copied this from. What should it be for CHIME? 800MHz? 

It is only only used to compute the length of a single FPGA sample (after FFT), so as long as 2.56us is the correct length for a sample in the baseband data then at least the results are fine, even if the input is wrong.

Hmm yes, the overview paper says "CHIME ADC sampling cadence of 800 MHz".

tristpinsm: yeah for CHIME the ADC samples at 800MHz and the FFT effectively integrates for 2048 frames (resulting in 1024 frequencies) :shrug:

config/chime_upgrade_n2.j2
rhaas80: Possibly. They could also be in one of the include files. I think I will want to get rid of depending on those as well. They seem to start becoming a moving target once more.

In `fengine/include/n2k_chime.j2`.

config/chime_upgrade_n2.j2
rhaas80: now of course I wonder if my desired 10s integration time is only 5s. Though that should not do a whole lot to the result (and I can always re-sum afterwards).

lib/stages/N2FrameToVisFrame.hpp
rhaas80: updated

lib/stages/N2FrameToVisFrame.hpp
rhaas80: moved to constructor

lib/stages/N2FrameToVisFrame.cpp
rhaas80: removed

config/chime_upgrade_n2.j2
```diff
@@ -0,0 +1,2691 @@
...
+#
+# chime_upgrade_network.yaml
+#
+# Pipeline for testing the CHIME upgrade F-engine packet capture and
+# transfer to the GPU.
+#
+# Currently only uses one NUMA domain, but it shouldn't be hard to extend to
+# the second NUMA domain.
+#
+# Author: Andre Renard
```
ljgray: Doesn't reflect this config

config/chime_upgrade_n2.j2
```diff
@@ -0,0 +1,2691 @@
...
+---
+########################
+#### Global options ####
+########################
+type: config
+# Logging level can be one of:
+# OFF, ERROR, WARN, INFO, DEBUG, DEBUG2 (case insensitive)
+# Note DEBUG and DEBUG2 require a build with (-DCMAKE_BUILD_TYPE=Debug)
+log_level: INFO
+num_elements: 2048
```
ljgray: Is there a reason to define `num_elements` independently of `num_dishes` and `num_polarisations`?

config/chime_upgrade_n2.j2
rhaas80: removed the adc frequency and instead introduced a sample_period variable

config/chime_upgrade_n2.j2
rhaas80: updated

config/chime_upgrade_n2.j2
rhaas80: they are now explicit since I expanded the includes: https://github.com/kotekan/kotekan/blob/cca411c723765603e55888ec21cca5f024dc43fa/config/chime_upgrade_n2.j2#L644

config/chime_upgrade_n2.j2
rhaas80: well, it avoids infinite recursions in things like:

```
num_elements: num_dishes * num_polarizations

some stage:
  num_dishes: num_elements / num_polarizations
```

I have however changed things so that `num_dishes` and `num_polarizations` are the defined variables, even though CHIME does not really have "dishes" and a split in dishes and polarizations is also not quite what the data layout uses (splits into cylinder, polarization, antennas).

In any case it avoid duplicate information right now.

rhaas80:
now this needs a review

ljgray (approved):

PR #1419, opened 2026-02-24
chime bb dump input file

rhaas80:
only a new input file

### DIFF
--- a/config/chime_upgrade_bb.j2
+++ b/config/chime_upgrade_bb.j2
@@ -1,14 +1,14 @@
 ##########################################
 #
-# chime_upgrade_network.yaml
+# chime_upgrade_bb.yaml
 #
-# Pipeline for testing the CHIME upgrade F-engine packet capture and
-# transfer to the GPU.
+# Pipeline for capturing and dumping frames from the CHIME upgrade F-engine
+# packet capture code.
 #
 # Currently only uses one NUMA domain, but it shouldn't be hard to extend to
 # the second NUMA domain.
 #
-# Author: Andre Renard
+# Author: Andre Renard, Roland Haas
 #
 ##########################################
 ---
@@ -23,8 +23,10 @@ log_level: info
 num_elements: 2048
 num_freq_host_buf: 1
 num_freq_gpu_buf: 16
-samples_per_frame: 16384
-buffer_depth: 3*272 # == baseband_buffer_depth
+samples_per_data_set: 16384
+sample_period: 2.56e-6
+baseband_buffer_depth: 3*272 # 3*272 = ~33 seconds after accounting for active frames
+host_buffer_depth: 12
 gpu_buffer_depth: 4
 num_gpus: 2
 cpu_affinity: [0, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19]
@@ -47,11 +49,16 @@ telescope:
 
 main_pool:
   kotekan_metadata_pool: chordMetadata
-  num_metadata_objects: 30 * buffer_depth
+  num_metadata_objects: 30 * host_buffer_depth + 5 * baseband_buffer_depth
+
+# Baseband metadata pool
+baseband_pool:
+  kotekan_metadata_pool: BasebandMetadata
+  num_metadata_objects: 30 * host_buffer_depth * 32
 
 host_voltage_buffers:
-  num_frames: buffer_depth
-  frame_size: samples_per_frame * num_elements * num_freq_host_buf
+  num_frames: baseband_buffer_depth
+  frame_size: samples_per_data_set * num_elements * num_freq_host_buf
   metadata_pool: main_pool
   numa_0:
     numa_node: 0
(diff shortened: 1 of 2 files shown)


config/chime_upgrade_network.j2
```diff
@@ -27,7 +27,7 @@ samples_per_frame: 16384
 buffer_depth: 32
 gpu_buffer_depth: 4
 num_gpus: 2
-cpu_affinity: [0, 10, 11, 12, 13, 14, 15, 16, 16, 18, 19]
+cpu_affinity: [0, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19]
```
ljgray, line 30: I assume this was just a typo in the old config?

rhaas80: so do I, @andrerenard was not saying though :-)

rhaas80: what I have checked though is that `16` is not one of the cores reserved for dpdk (those are `4,5,6,7,36,37,38,39`)

config/chime_upgrade_bb.j2
```diff
@@ -0,0 +1,270 @@
+##########################################
+#
+# chime_upgrade_network.yaml
```
rhaas80: wrong file name mentioned

config/chime_upgrade_bb.j2
```diff
@@ -0,0 +1,270 @@
...
+#
+# Pipeline for testing the CHIME upgrade F-engine packet capture and
+# transfer to the GPU.
+#
+# Currently only uses one NUMA domain, but it shouldn't be hard to extend to
+# the second NUMA domain.
+#
+# Author: Andre Renard
```
rhaas80: should add Roland, to correctly assign praise or blame

config/chime_upgrade_bb.j2
```diff
@@ -0,0 +1,270 @@
...
+use_hugepages: true
+
+# Telescope
+telescope:
+  name: CHIMETelescope
+  require_gps: true
+  query_gps: true # GPS time is provided via coco
+  query_frequency_map: true # Frequency map is provided via coco
+  require_frequency_map: true # Make sure we have the full frequency map
+  allow_default_frequency_map: false # Don't fall back on default map
```
rhaas80, line 48: is missing `earth_rotation_data` settings which will result in a warning at runtime if they were accessed

config/chime_upgrade_bb.j2
```diff
@@ -0,0 +1,270 @@
...
+{%- for id in range(16) %}
+        - host_voltage_buffer_{{ id }}
+{%- endfor %}
+      signal_buf: gpu_voltage_ringbuffer
+      gpu_mem_output: blocked_voltage_buffer
+
+# RH: merged in CHIME science pipeline stuff
+samples_per_data_set: samples_per_frame # 1/3 of CHIME number
+num_data_sets: 1
+baseband_buffer_depth: 3*272 # 3*272 = ~33 seconds after accounting for active frames
```
rhaas80: TODO: double check the 33s claim

config/chime_upgrade_bb.j2
```diff
@@ -0,0 +1,270 @@
...
+  kotekan_stage: bufferMerge
+  timeout: 0.1
+  in_bufs:
+{%- for id in range(16) %}
+    - baseband_output_buffer_{{ id }}
+{%- endfor %}
+  out_buf: baseband_output_buffer_merged
+
+write_baseband:
+  root_path: /home/rhaas/CHIME/kotekan/
```
rhaas80: should not have my user name in it

config/chime_upgrade_bb.j2
```diff
@@ -0,0 +1,270 @@
+##########################################
+#
+# chime_upgrade_network.yaml
+#
+# Pipeline for testing the CHIME upgrade F-engine packet capture and
+# transfer to the GPU.
+#
+# Currently only uses one NUMA domain, but it shouldn't be hard to extend to
+# the second NUMA domain.
```
ljgray: I suppose this description is still technically correct but incomplete

config/chime_upgrade_bb.j2
```diff
@@ -0,0 +1,270 @@
...
+  num_metadata_objects: 30 * baseband_buffer_depth * 96
+
+# Baseband capture and dump system
+baseband_output_buffers:
+  frame_size: samples_per_data_set * num_elements * num_freq_host_buf * num_data_sets / 32
+  metadata_pool: baseband_pool
+{%- for id in range(16) %}
+  baseband_output_buffer_{{ id }}:
+    kotekan_buffer: standard
+    num_frames: 3 * 11 * 32 # baseband_buffer_depth
```
ljgray: This is different from the `baseband_buffer_depth` defined on line 210

config/chime_upgrade_bb.j2
rhaas80: fixed

config/chime_upgrade_bb.j2
rhaas80: done

config/chime_upgrade_bb.j2
rhaas80: assuming 16 frames are "in flight" I get 33.5s. Good enough.

config/chime_upgrade_bb.j2
rhaas80: updated lines 5 and 6. Kept 8 and 9 since they are still true.

config/chime_upgrade_bb.j2
rhaas80: hmm, in particular its larger. Matches the situation in `chime_science_run_gpu.yaml` though:

```
baseband_buffer_depth: 272 # 272 = ~33 seconds after accounting for active frames
[...]
num_frames: 11 * 96 # baseband_buffer_depth
```

with the difference (32 vs 96 being due to CHIME production's frames being 3 times larger than CHIME upgrades).

The actual baseband buffers are the network buffers, the input to `basebandReadout`:

```
baseband:
  max_dump_samples: 300000 # 768ms
  num_frames_buffer: baseband_buffer_depth - 10
{%- for id in range(16) %}
  baseband{{ id }}:
    kotekan_stage: basebandReadout
    in_buf: host_voltage_buffer_{{ id }}
    out_buf: baseband_output_buffer_{{ id }}
{%- endfor %}
```

of which there are indeed "baseband_buffer_depth" many. There's some confusion on my part though on the size of `main_pool` which is (way) larger than for CHIME production right now (since I set it to `30*basband_buffer_depth` while the old one would only have been `5*baseband_buffer_depth`). 

I should however adjust `baseband_pool` accordingly so that it reads:

```
baseband_pool:
  kotekan_metadata_pool: BasebandMetadata
  num_metadata_objects: 30 * baseband_buffer_depth * 32
```

looking at the baseband dump code (https://github.com/kotekan/kotekan/blob/97bcfba7f30e40793c050640b71c1276337ee7a1/lib/stages/basebandReadout.cpp#L412) the baseband dump frames will contain sufficient samples to fill them up:


```
const int out_frame_samples = out_buf->frame_size / _num_elements;
```


so the code here tries to keep the size of a baseband dump frame the same as before.

In any case, this indeed needs some cleanup.

config/chime_upgrade_bb.j2
rhaas80: is not required for baseband dumps.

config/chime_upgrade_bb.j2
rhaas80: fixed

config/chime_upgrade_bb.j2
rhaas80: I think I cleaned those up now. Running a test first. Seems to be better now.

jbmertens (approved):
🚀

ljgray (approved):

PR #1420, opened 2026-02-24
iceBoardShuffle: match FEngine metadata

rhaas80:


eschnett (approved):

rhaas80:
apparently I noticed a bug and added checks in  4161cc08c and d1e6eee3c but never fixed the actual bug in 96e8e0e70. Thus now we have https://github.com/kotekan/kotekan/pull/1426

PR #1408, opened 2026-02-05
N2Accumulate Upgrades

geoffryan:
Numerous upgrades to the N2Accumulator and related stages.

Todo:
- [x]  ERA bins
- [x]  Fringestopping
- [x]  Second Stage RFI Filtering
- [x]  Improved Weights
- [x]  Accurate Packet Loss & RFI Accounting 
- [x]  local ERA calculation
- [x]  Better CI coverage (N2Accumulate unit test) (punting the fringestopping test)
- [x]  Performance testing
- [x]  Prometheus Metrics (punted to later PR)
- [x]  Radiometer Chi2 Metric (plumbed, implementation punted)

PR #1422, opened 2026-02-24
misc fixes and updates

rhaas80:
* 10f74d3a - EigenN2Iter: correct use of optional buffer for failures
* e3eebba8 - chime_upgrade_n2: add buffer for failed eigenvalues
* b3c376d0 - bufferQuiet: utility stage to hold frames until no new ones arrive anymore
* bcc65fc3 - ringbuffer: output number of producers / consumers
* 9a0f0e6c - buffer: output number of producers / consumers
* 94c70381 - givenDataGen: create a kotekan buffer with data given as config inputs
* a120aac9 - EigenN2Iter: catch eigenvalue calculation errors from blaze

none of these are actually required by anything. Some are "nice to have", some are "would this make sense"?

### DIFF
--- a/config/chime_upgrade_n2.j2
+++ b/config/chime_upgrade_n2.j2
@@ -741,9 +741,22 @@ eig_10s_buffer:
     num_frames: gpu_buffer_depth * num_freq_gpu_buf
     metadata_pool: n2_pool
 
+failed_10s_buffer:
+    kotekan_buffer: N2
+    num_ev: 0
+    n2_layout: "FullUpperTri"
+    num_frames: gpu_buffer_depth * num_freq_gpu_buf
+    metadata_pool: n2_pool
+
 vis_buffer:
     kotekan_buffer: vis
     num_ev: num_ev_to_calculate
+    num_frames: gpu_buffer_depth * num_freq_gpu_buf
+    metadata_pool: vis_pool
+
+failed_vis_buffer:
+    kotekan_buffer: vis
+    num_ev: 0
     num_frames: gpu_buffer_depth * num_freq_gpu_buf
     metadata_pool: vis_pool
 
@@ -796,6 +809,7 @@ eigencalc:
     num_ev: num_ev_to_calculate
     in_buf: vis_10s_buffer
     out_buf: eig_10s_buffer
+    failed_buf: failed_10s_buffer
     # Convergence config from chime science run
     tol_eval: 0.00001
     tol_evec: 0.0001
@@ -808,6 +822,11 @@ vis_convert:
     kotekan_stage: n2FrameToVisFrame
     n2_buf: eig_10s_buffer
     vis_buf: vis_buffer
+
+failed_vis_convert:
+    kotekan_stage: n2FrameToVisFrame
+    n2_buf: failed_10s_buffer
+    vis_buf: failed_vis_buffer
 
 ################################################################################
 # output to receivers
@@ -841,6 +860,12 @@ write_vis:
     root_path: vis_data
     file_type: raw
     in_buf: vis_buffer
+
+write_failed_vis:
+    kotekan_stage: VisWriter
+    root_path: failed_eig
+    file_type: raw
+    in_buf: failed_vis_buffer
 
 # Information for input reordering, packed as
 #   (adc_id, chan_id, correlator_input)
--- a/config/chime_upgrade_n2.j2
+++ b/config/chime_upgrade_n2.j2
@@ -274,6 +274,8 @@ host_pl_expanded_mask_ringbuffer:
 # glue code to handle lost frames formats
 lostSamplesToPLMask:
     kotekan_stage: lostSamplesToPLMask
+    # num_dishes: num_dishes # set via a global
+    # num_polarizations: num_polarizations # set via a global
     pl_mask_buf: host_pl_mask_buffer
     lost_samples_buffers:
 {%- for id in range(num_freq_gpu_buf // 4) %}
--- a/lib/stages/EigenN2Iter.cpp
+++ b/lib/stages/EigenN2Iter.cpp
@@ -9,7 +9,7 @@
 #include "StageFactory.hpp"      // for REGISTER_KOTEKAN_STAGE, StageMakerTemplate
 #include "buffer.hpp"            // for allocate_new_metadata_object, mark_frame_empty, mark_fr...
 #include "kotekanLogging.hpp"    // for DEBUG
-#include "prometheusMetrics.hpp" // for Gauge, Metrics, MetricFamily
+#include "prometheusMetrics.hpp" // for Counter, Gauge, Metrics, MetricFamily
 
 #include "fmt.hpp"      // for format, fmt
 #include "gsl-lite.hpp" // for span
@@ -41,6 +41,7 @@ EigenN2Iter::EigenN2Iter(Config& config, const std::string& unique_name,
     Stage(config, unique_name, buffer_container, std::bind(&EigenN2Iter::main_thread, this)),
 
     in_buf(get_buffer("in_buf")), out_buf(get_buffer("out_buf")),
+    failed_buf(config.exists(unique_name, "failed_buf") ? get_buffer("failed_buf") : nullptr),
 
     _num_eigenvectors(config.get<size_t>(unique_name, "num_ev")),
 
@@ -67,7 +68,9 @@ EigenN2Iter::EigenN2Iter(Config& config, const std::string& unique_name,
     eigenvalue_convergence_metric(Metrics::instance().add_gauge(
         "kotekan_eigenN2iter_eigenvalue_convergence", unique_name, {"freq_id"})),
     eigenvector_convergence_metric(Metrics::instance().add_gauge(
-        "kotekan_eigenN2iter_eigenvector_convergence", unique_name, {"freq_id"})) {
+        "kotekan_eigenN2iter_eigenvector_convergence", unique_name, {"freq_id"})),
+    num_failed_eigencalc(
+        Metrics::instance().add_counter("kotekan_eigenN2iter_num_failed_eigencalc", unique_name)) {
 
     in_buf->register_consumer(unique_name);
     out_buf->register_producer(unique_name);
@@ -96,6 +99,34 @@ EigenN2Iter::EigenN2Iter(Config& config, const std::string& unique_name,
                     out_desc->get_num_ev(), _num_eigenvectors);
     }
 
+    if (failed_buf) {
+        failed_buf->register_producer(unique_name);
+        // this replicates some of the tests on in_buf to not rely on previous tests
+        if (in_buf->buffer_type != "N2" || failed_buf->buffer_type != "N2")
+            FATAL_ERROR("EigenN2Iter stage requires 'N2' buffers as input and failed.");
+
+        // Validate that input and failed buffers have N2 frame descriptors set
+        auto in_desc =
+            std::dynamic_pointer_cast<const kotekan::N2FrameDesc>(in_buf->get_frame_description());
+        auto failed_desc = std::dynamic_pointer_cast<const kotekan::N2FrameDesc>(
+            failed_buf->get_frame_description());
+        if (!in_desc || !failed_desc) {
+            FATAL_ERROR("EigenN2Iter: Input and failed buffers must have N2FrameDesc set");
+        }
+        // Validate num_elements and layout match
+        if (in_desc->get_num_elements() != failed_desc->get_num_elements()) {
+            FATAL_ERROR("EigenN2Iter: Input and failed buffer num_elements must match");
+        }
+        if (in_desc->get_n2_layout() != failed_desc->get_n2_layout()) {
+            FATAL_ERROR("EigenN2Iter: Input and failed buffer n2_layout must match");
+        }
+        // Validate failed buffer has correct num_ev for this stage
+        if (failed_desc->get_num_ev() != 0) {
+            FATAL_ERROR("EigenN2Iter: Failed buffer num_ev ({:d}) not 0",
+                        failed_desc->get_num_ev());
+        }
+    }
+
     if (_num_ev_conv > _num_eigenvectors)
         FATAL_ERROR("The `num_ev_conv` config parameter ({:d}) must be less than or equal to "
                     "`num_ev` ({:d}).",
@@ -114,6 +145,9 @@ void EigenN2Iter::main_thread() {
 
     frameID input_frame_id(in_buf);
     frameID output_frame_id(out_buf);
+    // need to pass a valid buffer to avoid a segfault, even if failed_frame_id
+    // is not used afterwards
+    frameID failed_frame_id(failed_buf != nullptr ? failed_buf : out_buf);
 
     DynamicHermitian<float> mask;
     uint32_t num_elements = 0;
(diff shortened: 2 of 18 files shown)


config/chime_upgrade_n2.j2
```diff
@@ -0,0 +1,703 @@
+##########################################
+#
+# chime_upgrade_network.yaml
```
jbmertens: Assuming this comment needs to be updated.

config/chime_upgrade_n2.j2
```diff
@@ -0,0 +1,703 @@
...
+#
+# Pipeline for testing the CHIME upgrade F-engine packet capture and
```
jbmertens: If this is in fact just for testing, how about moving this to the tests/ directory?

config/chime_upgrade_n2.j2
```diff
@@ -0,0 +1,703 @@
...
+zero_samples:
+  duplicate_ls_buffer: False
+{%- for id in range(16) %}
+  zero_{{ id }}:
+    kotekan_stage: zeroSamples
+    lost_samples_buf: lost_samples_buffer_{{ id // 4 }}
+    out_buf: host_voltage_buffer_{{ id }}
+{%- endfor %}
+
+# replaced Andre's version by Erik's
```
jbmertens: Remove this commented section?

config/chime_upgrade_n2.j2
```diff
@@ -0,0 +1,703 @@
...
+        1776, 1777, 1778, 1779, 1780, 1781, 1782, 1783, 1784, 1785, 1786, 1787, 1788, 1789, 1790, 1791, 1744, 1745, 1746, 1747, 1748, 1749, 1750, 1751, 1752, 1753, 1754, 1755, 1756, 1757, 1759, 1758, 1712, 1713, 1714, 1715, 1716, 1717, 1718, 1719, 1720, 1721, 1722, 1723, 1724, 1725, 1726, 1727, 1680, 1681, 1682, 1683, 1684, 1685, 1686, 1687, 1688, 1689, 1690, 1691, 1692, 1693, 1694, 1695,
+        2032, 2033, 2034, 2035, 2036, 2037, 2038, 2039, 2040, 2041, 2042, 2043, 2044, 2045, 2046, 2047, 2000, 2001, 2002, 2003, 2004, 2005, 2006, 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015, 1968, 1969, 1970, 1971, 1972, 1973, 1974, 1975, 1976, 1977, 1978, 1979, 1980, 1981, 1982, 1983, 1936, 1937, 1938, 1939, 1940, 1941, 1942, 1943, 1944, 1945, 1946, 1947, 1948, 1949, 1950, 1951,
+    ]
+
+debug_dump_voltage:
+    prefix_hostname: false
+    prefix_host_rank: true
+    base_dir: data/debug
+#{%- for id in range(4) %}
```
jbmertens: Wondering if commented lines here and below should be cleaned up.

lib/stages/bufferQuiet.cpp
```diff
@@ -0,0 +1,123 @@
...
+                         bufferContainer& buffer_container) :
+    Stage(config, unique_name, buffer_container, std::bind(&bufferQuiet::main_thread, this)) {
+
+    in_buf = get_buffer("in_buf");
+    in_buf->register_consumer(unique_name);
+
+    out_buf = get_buffer("out_buf");
+    out_buf->register_producer(unique_name);
+
+    // Buffer sizes must match exactly
```
jbmertens: Frame, not buffer sizes, right?
Use the frame_desc for this?

lib/stages/bufferQuiet.cpp
```diff
@@ -0,0 +1,123 @@
...
+    frameID in_frame_release_id(in_buf);
+    frameID out_frame_id(out_buf);
+    int in_frame_hold_ctr = 0;
+
+    uint8_t* output_frame = nullptr;
+
+    while (!stop_thread) {
+
+        timespec now;
+        clock_gettime(CLOCK_REALTIME, &now);
```
jbmertens: Channeling my inner Geoff, consider std::chrono::steady_clock here.

lib/stages/bufferQuiet.cpp
```diff
@@ -0,0 +1,123 @@
...
+            DEBUG("Got new input frame {:d}", in_frame_id);
+            in_frame_hold_ctr += 1;
+            in_frame_id += 1;
+        }
+
+        int frames_to_release = -1;
+        if (reason == 1) { // timeout, release all frames
+            frames_to_release = in_frame_hold_ctr;
+        } else if (in_frame_hold_ctr + _num_reserve_frames
+                   >= out_buf->num_frames) { // emergency release a frame since buffer is full
```
jbmertens, line 74: note this behavior in the docstring.

lib/stages/bufferQuiet.cpp
```diff
@@ -0,0 +1,123 @@
...
+        }
+        while (frames_to_release-- > 0) {
+
+            // Get a new output frame
+            output_frame = out_buf->wait_for_empty_frame(unique_name, out_frame_id);
+            if (output_frame == nullptr)
+                return;
+
+            if (_copy_frame || in_buf->get_num_consumers() > 1) {
+                out_buf->allocate_new_metadata_object(out_frame_id);
```
jbmertens, line 91: What about frame_desc's?

lib/stages/bufferQuiet.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include "Config.hpp"          // for Config
+#include "Stage.hpp"           // for Stage
+#include "buffer.hpp"          // for Buffer
+#include "bufferContainer.hpp" // for bufferContainer
+
+#include <string> // for string
+
+
+/**
+ * @brief Waits for a quiet period before releasing  of frames of the buffer.
```
jbmertens: Note here we are waiting for the input buffer to be quiet.

lib/stages/EigenN2Iter.cpp
```diff
@@ -96,6 +97,32 @@ EigenN2Iter::EigenN2Iter(Config& config, const std::string& unique_name,
                     out_desc->get_num_ev(), _num_eigenvectors);
     }
 
+    if (failed_buf) {
+        failed_buf->register_producer(unique_name);
+        if (in_buf->buffer_type != "N2" || failed_buf->buffer_type != "N2")
```
jbmertens, line 105: Some of these checks are repeats of above, right?

lib/stages/EigenN2Iter.cpp
```diff
@@ -156,9 +186,28 @@ void EigenN2Iter::main_thread() {
...
 
         // Perform the actual eigen-decomposition
-        std::tie(eigpair, stats) =
-            eigen_masked_subspace(vis, mask, _num_eigenvectors, _tol_eval, _tol_evec,
-                                  _max_iterations, _num_ev_conv, _krylov, _subspace);
+        try {
+            std::tie(eigpair, stats) =
+                eigen_masked_subspace(vis, mask, _num_eigenvectors, _tol_eval, _tol_evec,
+                                      _max_iterations, _num_ev_conv, _krylov, _subspace);
+        } catch (const std::runtime_error& e) {
```
jbmertens, line 197: Can we add a Prometheus metric to track this?

jbmertens (changes requested):
Need to look at N2FrameToVisFrame still, but passing along some comments for now.

lib/stages/bufferQuiet.cpp
rhaas80: there's not `frame_desc` for the output buffer yet. I do (I hope) ensure that the output `frame_desc` is the same as the input (by copying it), the buffer's frame size is given in the input file so is something that a user can get wrong. 

I will take a look.

lib/stages/bufferQuiet.cpp
rhaas80: that is a very good question. I better make sure I actually set them.

lib/stages/bufferQuiet.hpp
rhaas80: true

lib/stages/EigenN2Iter.cpp
rhaas80: they are, yes. Harmless though. In copies the tests for `out_buf` but for `failed_buf` and thus tries to similarity of code rather than minimal lines of codes.

lib/stages/EigenN2Iter.cpp
rhaas80: likely, I have never used those though. Looking at other stages "this seems simple" (famous last words).

lib/stages/bufferQuiet.cpp
rhaas80: there's `CLOCK_MONOTONIC` https://docs.redhat.com/en/documentation/red_hat_enterprise_linux_for_real_time/7/html/reference_guide/sect-posix_clocks

lib/stages/bufferQuiet.cpp
jbmertens: Ah, right, using the frame_desc here is only relevant for N2 frames which have descriptions allocated by the bufferFactory. This works as a more generic check.

But, then maybe it's also worth checking that the input and output buffer types are the same.

lib/stages/EigenN2Iter.cpp
jbmertens: Ok, makes sense. The error message(s) could reflect this.

lib/stages/EigenN2Iter.cpp
jbmertens: It hopefully is simple 😬 
More useful in production I think, not so much for development.

lib/stages/N2FrameToVisFrame.cpp
```diff
@@ -0,0 +1,214 @@
...
+                                     bufferContainer& buffer_container) :
+    Stage(config, unique_name, buffer_container, std::bind(&n2FrameToVisFrame::main_thread, this)) {
+
+    n2_buf = get_buffer("n2_buf");
+    n2_buf->register_consumer(unique_name);
+
+    vis_buf = get_buffer("vis_buf");
+    vis_buf->register_producer(unique_name);
+
+    // TODO: add any sanity checks required
```
jbmertens: Flagging in case you want to add these before merging.

lib/stages/N2FrameToVisFrame.cpp
```diff
@@ -0,0 +1,214 @@
...
+        n2_buf->mark_frame_empty(unique_name, n2_buf_frame_id++);
+    }
+}
+
+// from visAccumulate
+void n2FrameToVisFrame::register_base_dataset_states(
+    std::string& instrument_name, std::vector<std::pair<uint32_t, freq_ctype>>& freqs,
+    std::vector<input_ctype>& inputs, std::vector<prod_ctype>& prods) {
+    // weight calculation is hardcoded, so is the weight type name
+    // TODO: check if this is still true
```
jbmertens, line 209: The weight calculation should be getting options in #1408 .

lib/stages/N2FrameToVisFrame.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include "bufferContainer.hpp" // for bufferContainer
+#include "datasetManager.hpp"
+#include "version.h"
+#include "visUtil.hpp"
+
+#include <stdint.h> // for int32_t, uint32_t, uint8_t
+#include <string>   // for string
+#include <vector>   // for vector
+
+/**
```
jbmertens, line 16: Docstring needs an update.

jbmertens (commented):
More comments.

config/chime_upgrade_n2.j2
rhaas80: fixed

config/chime_upgrade_n2.j2
rhaas80: done

config/chime_upgrade_n2.j2
rhaas80: well, "testing" as in stuff we want to do before we tear down current production system. Not as in "a CI test".

config/chime_upgrade_n2.j2
rhaas80: this has been rewritten quite a bit. Should be cleaner now (in https://github.com/kotekan/kotekan/pull/1418 )

config/chime_upgrade_n2.j2
jbmertens: There's the `ci-tests` directory for CI tests, and `tests` for other misc tests. But yeah there isn't really a pattern to what configs go where 🤷

lib/stages/bufferQuiet.cpp
rhaas80: I added:

```
                // this just copies the shared pointer, and is only valid since
                // a frame description must never change
                out_buf->set_frame_desc(in_buf->get_frame_description());
```

I note inconsistent naming of the `set_frame_desc` and `get_frame_description` functions.

lib/stages/bufferQuiet.cpp
rhaas80: added

lib/stages/bufferQuiet.cpp
rhaas80: changed to:

```
/**
 * @brief Waits for a quiet period or imminent buffer overflow before releasing
 * of frames of the buffer.
 ```

lib/stages/bufferQuiet.cpp
rhaas80: There's currently 9 instances of `clock_gettime(CLOCK_` and 10 instance of `std::chrono::steady_clock::now()` so the C++ construct wins the popularity contest (though I bet it just calls the POSIX one under the hood).

However I need a `timespec` afterwards so `clock_gettime` is more convenient (due to the existence of functions to do arithmetic on). So I changed to `CLOCK_MONOTONIC`.

lib/stages/EigenN2Iter.cpp
rhaas80: I added a comment and fixed some error messages.

lib/stages/EigenN2Iter.cpp
jbmertens: (This can also wait until it is decided it would be useful to have in production)

rhaas80: done

lib/stages/N2FrameToVisFrame.cpp
rhaas80: updated the comment. There are, unfortunately, no other sanity checks (that I can think of).

lib/stages/N2FrameToVisFrame.hpp
rhaas80: updated

lib/stages/N2FrameToVisFrame.cpp
rhaas80: ok

jbmertens (approved):

rhaas80:
@jbmertens thanks for the review. Since I sneakily rebased this branch onto my rhaas/chime_n2 and rhaas/parse_input_reorder branches I'll wait for their review before pushing.

PR #1423, opened 2026-02-25
Replicate correlator counts from CHIME lost_samples buffer

ljgray:
Creates the `counts` buffer required by `N2Accumulate` using the CHIME lost_samples and rfi masks.

I've _assumed_ that the masking convention is 0 == bad, 1 == good (based on a couple of comments I've found), but if this isn't the case then it's a simple fix.

Probably some missing metadata

### DIFF
--- a/lib/stages/lostSamplesToN2Counts.cpp
+++ b/lib/stages/lostSamplesToN2Counts.cpp
@@ -1,0 +1,240 @@
+#include "lostSamplesToN2Counts.hpp"
+
+#include "Config.hpp" // for Config
+#include "DataType.hpp"
+#include "N2Util.hpp"          // for frameID
+#include "StageFactory.hpp"    // for REGISTER_KOTEKAN_STAGE
+#include "buffer.hpp"          // for Buffer
+#include "bufferContainer.hpp" // for bufferContainer
+#include "chordMetadata.hpp"   // for get_chord_metadata, chordMetadata
+
+#include "json.hpp" // for basic_json, json, iter_impl
+
+#include <memory>
+#include <string>
+#include <sys/types.h>
+
+using kotekan::bufferContainer;
+using kotekan::Config;
+using kotekan::Stage;
+using nlohmann::json;
+
+
+REGISTER_KOTEKAN_STAGE(lostSamplesToN2Counts);
+
+// Random helpers for clarity
+#define BITS_PER_BYTE 8
+// Counts ALWAYS have tile size of 8
+#define COUNTS_BLOCK_SIZE 8
+#define COUNTS_ELEMENT_DOWNSAMPLE 8
+
+
+lostSamplesToN2Counts::lostSamplesToN2Counts(Config& config, const std::string& unique_name,
+                                             bufferContainer& buffer_container) :
+    Stage(config, unique_name, buffer_container,
+          std::bind(&lostSamplesToN2Counts::main_thread, this)),
+    num_elements(config.get<size_t>(unique_name, "num_elements")),
+    num_n2k_freq(config.get<size_t>(unique_name, "num_n2k_freq")),
+    samples_per_data_set(config.get<size_t>(unique_name, "samples_per_data_set")),
+    sub_integration_ntime(config.get<size_t>(unique_name, "sub_integration_ntime")) {
+
+    // Set up for n2 counts buffer. One buffer containing all
+    // frequencies on this node
+    n2k_counts_buf = get_buffer("n2k_counts_buf");
+    n2k_counts_buf->register_producer(unique_name);
+
+    // Register as a consumer for the rfi mask buffer, which should
+    // also be a single buffer for all frequencies
+    rfi_mask_buf = get_buffer("rfi_mask_buf");
+    rfi_mask_buf->register_consumer(unique_name);
+
+    // Register as a consumer for all the lost samples buffers
+    json ls_in_bufs = config.get_value(unique_name, "lost_samples_buffers");
+
+    for (json::iterator it = ls_in_bufs.begin(); it != ls_in_bufs.end(); ++it) {
+        Buffer* buf = buffer_container.get_buffer(it.value());
+
+        lost_samples_buffers.push_back(buf);
+        buf->register_consumer(unique_name);
+
+        if (buf->frame_size != lost_samples_buffers.at(0)->frame_size)
+            FATAL_ERROR("Input lost_samples buffers have different frame sizes: {:d} != {:d}",
+                        buf->frame_size, lost_samples_buffers.at(0)->frame_size);
+
+        if (buf->num_frames != lost_samples_buffers.at(0)->num_frames)
+            FATAL_ERROR("Input lost_samples buffers have different number of frames: {:d} != {:d}",
+                        buf->num_frames, lost_samples_buffers.at(0)->num_frames);
+    }
+
+    // Check counts size against both inputs
+    // lost_samples->frame_size == samples_per_data_set
+    // rfi_mask->frame_size == nfreq * samples_per_data_set / 8
+    // counts->frame_size == nfreq * subintegrations * _counts_stride * sizeof(int32_t)
+    if (samples_per_data_set != lost_samples_buffers.at(0)->frame_size)
+        FATAL_ERROR("lost_samples_buf does not have the expected number of samples. Expected {:d}, "
+                    "got {:d}",
+                    samples_per_data_set, lost_samples_buffers.at(0)->frame_size);
+
+    size_t _nbufs = lost_samples_buffers.size();
+    // Can't have a fractional number of frequencies
+    // rfi_mask is a bitmap, so frame size is reduced by a factor of 8
+    if ((BITS_PER_BYTE * rfi_mask_buf->frame_size) % samples_per_data_set != 0)
+        FATAL_ERROR("rfi_mask_buf and lost_samples_buf frame sizes are incompatable. rfi_mask: "
+                    "{:d}; lost_samples: {:d}",
+                    rfi_mask_buf->frame_size, samples_per_data_set);
+
+    if (num_n2k_freq % _nbufs != 0)
+        FATAL_ERROR(
+            "Number of lost_samples buffers ({:d}) does not evenly divide total frequencies ({:d})",
+            _nbufs, num_n2k_freq);
+
+    // Check rfi make frame size against expectation
+    size_t _expected_rfi_frame_size = num_n2k_freq * samples_per_data_set / BITS_PER_BYTE;
+    if (rfi_mask_buf->frame_size != _expected_rfi_frame_size)
+        FATAL_ERROR("Unexpected frame size for rfi_mask {:d} - expected {:d}",
+                    rfi_mask_buf->frame_size, _expected_rfi_frame_size);
+
+    // This is a bit of a misnomer - the lost_samples buffer does not _include_
+    // multiple frequencies, but information may be shared by multiple frequencies
+    _num_freq_per_lost_samples_buffer = num_n2k_freq / _nbufs;
+    // Number of subintegrations - i.e., time dimension of counts
+    // Integers, so no remainder
+    _num_subintegrations = samples_per_data_set / sub_integration_ntime;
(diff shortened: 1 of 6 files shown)


lib/stages/lostSamplesToN2Counts.cpp
```diff
@@ -0,0 +1,235 @@
...
+using kotekan::Stage;
+using nlohmann::json;
+
+
+REGISTER_KOTEKAN_STAGE(lostSamplesToN2Counts);
+
+// Random helpers for clarity
+#define BITS_PER_BYTE 8
+// CHIME-specific
+#define NUM_ELEMENTS 2048
```
geoffryan: I think it would be good to keep num_elements configurable at runtime? Unless the 2048 is also similarly hard coded into the lost samples buffer.

lib/stages/lostSamplesToN2Counts.cpp
```diff
@@ -0,0 +1,235 @@
...
+#define COUNTS_STRIDE (COUNTS_NTILES * COUNTS_BLOCK_SIZE * COUNTS_BLOCK_SIZE)
+
+
+lostSamplesToN2Counts::lostSamplesToN2Counts(Config& config, const std::string& unique_name,
+                                             bufferContainer& buffer_container) :
+    Stage(config, unique_name, buffer_container,
+          std::bind(&lostSamplesToN2Counts::main_thread, this)),
+    num_n2k_freq(config.get<size_t>(unique_name, "num_n2k_freq")),
+    time_samples_per_subintegration(
+        config.get<size_t>(unique_name, "time_samples_per_subintegration")) {
```
geoffryan: At least in the config's I've seen we've been calling this "sub_integration_ntime", but if this is the Chime name for this parameter that's totally ok.

lib/stages/lostSamplesToN2Counts.cpp
```diff
@@ -0,0 +1,235 @@
...
+            }
+            // Release the current lost_samples frame
+            lost_samples_buf->mark_frame_empty(unique_name, lost_samples_frame_id);
+            lost_samples_frame_id++;
+        }
+
+        // Set the frame description
+        // The buffer name and axis names are the same as those set in
+        // cudaPL1butCorrelator
+        n2k_counts_buf->allocate_ndarray_frame_desc<kotekan::GetType_t<kotekan::int32>, 5>(
```
geoffryan: Allocating the frame_desc should happen in the constructor.

lib/stages/lostSamplesToN2Counts.cpp
```diff
@@ -0,0 +1,235 @@
...
+    // Integers, so no remainder
+    _num_subintegrations = _num_time_samples / time_samples_per_subintegration;
+
+    // Check counts frame size against expectations. counts has type int32
+    size_t _expected_counts_frame_size =
+        _num_subintegrations * num_n2k_freq * COUNTS_STRIDE * sizeof(int32_t);
+
+    if (n2k_counts_buf->frame_size != _expected_counts_frame_size)
+        FATAL_ERROR("Unexpected frame sizes for n2_counts {:d} - expected {:d}",
+                    n2k_counts_buf->frame_size, _expected_counts_frame_size);
```
geoffryan, line 115: You should allocate the `ndarray_frame_desc` for the output buffer here in the stage constructor.

lib/stages/lostSamplesToN2Counts.cpp
```diff
@@ -0,0 +1,235 @@
...
+            FATAL_ERROR("Input lost_samples buffers have different number of frames: {:d} != {:d}",
+                        buf->num_frames, lost_samples_buffers.at(0)->num_frames);
+    }
+
+    // Check counts size against both inputs
+    // lost_samples->frame_size == samples_per_frame
+    // rfi_mask->frame_size == nfreq * samples_per_frame / 8
+    // counts->frame_size == nfreq * subintegrations * COUNTS_STRIDE * sizeof(int32_t)
+    size_t _nbufs = lost_samples_buffers.size();
+    _num_time_samples = lost_samples_buffers.at(0)->frame_size;
```
geoffryan: Would it be worth grabbing the `samples_per_data_set` field from the config and checking it is equal to _num_time_samples?

lib/stages/lostSamplesToN2Counts.cpp
```diff
@@ -0,0 +1,235 @@
...
+REGISTER_KOTEKAN_STAGE(lostSamplesToN2Counts);
+
+// Random helpers for clarity
+#define BITS_PER_BYTE 8
+// CHIME-specific
+#define NUM_ELEMENTS 2048
+// Counts ALWAYS have tile size of 8
+#define COUNTS_BLOCK_SIZE 8
+#define COUNTS_ELEMENT_DOWNSAMPLE 8
+#define COUNTS_NTILES                                                                              \
```
geoffryan: Is this missing a factor of 1/2? n2k_counts only keeps the lower triangular part of the blocked matrix.

lib/stages/lostSamplesToN2Counts.cpp
```diff
@@ -0,0 +1,235 @@
...
+                    size_t t0 = tau * time_samples_per_subintegration;
+                    // Accumulate for this subintegration
+                    int32_t _sum = 0;
+                    for (size_t ts = t0; ts < t0 + time_samples_per_subintegration; ++ts) {
+                        // base offset + coarse_time * stride + fine_time
+                        size_t tsr =
+                            ridx_f + (ts / _rfi_t_lo) * _rfi_t_hi_stride + (ts % _rfi_t_lo);
+                        // Convert bit index in a uint1 buffer to byte index in a uint8 buffer
+                        tsr /= BITS_PER_BYTE;
+                        // Bit stream packing corresponds to big endian bit layout
```
geoffryan: I'm not sure the bits are in big endian... at least I've been assuming the opposite!  Although, practically, the number of samples in a sub integration has always been a multiple of 8 so they'd all get summed anyways...

geoffryan (changes requested):
Few easy comments.  Bigger requested change is to add a unit test that runs in CI which verifies the correctness of the stage.

lib/stages/lostSamplesToN2Counts.cpp
ljgray: I pulled this from the design doc, but I think there's a similar comment there about a missing factor of ½. I'll go back and actually work through it

lib/stages/lostSamplesToN2Counts.cpp
ljgray: My comment is just a bit nonsensical - what I'm assuming here is that the bits are packed in order, so when it's viewed as `uint8_t` the left-most is the first sample received and right-most is the last sample received of the 8.

lib/stages/lostSamplesToN2Counts.cpp
ljgray: Yeah looks like there's a missing 1/2

lib/stages/lostSamplesToN2Counts.cpp
ljgray: The argument for defining it this way is that CHIME is the only system that produces lost_samples buffers instead of the PL_mask, and lost_samples has no concept of elements (just a timestream), so there shouldn't ever be a non-chime use for this stage. I'm happy to make it a config parameter if you'd prefer, though!

lib/stages/lostSamplesToN2Counts.cpp
ljgray: I've changed this comment to be more clear, but I'll keep this unresolved for now to make sure I'm not misunderstanding something about the rfi buffer format

lib/stages/lostSamplesToN2Counts.cpp
rhaas80: Re: "unless the 2048 is similarly hard coded". Of course it is :-) https://github.com/kotekan/kotekan/pull/1375/files#r2615942251

geoffryan: I'd prefer for these both to be read from the config or otherwise (num_elements is always available in the config), but following the linked discussion, if we'd prefer to wait for this to be available in the telescope, hard coded is fine.

rhaas80: keep in mind that this is very CHIME specific. None of the newer telescopes have a `lost_samples` ~~input~~ output from dpdk, do they?

geoffryan: Not that I know of, although I don't know what HIRAX's plan is.  Keeping it a config parameter just leaves us the option of testing it on smaller arrays, and its a config parameter everywhere else in kotekan.  Again, just a suggestion, won't hold up merging over this.  Just want to see a correctness test in CI and then we can merge this in.

rhaas80: hmm, I thought I heard that HIRAX was using CSR boards (same so CHORD), but then their [website](https://hirax.ukzn.ac.za/) says "will share much of the back-end technology, including FPGA-based ICE boards and the GPU correlator". Well, we'll see.

eschnett: Hirax (mixed case!) is using ICE boards, same as CHIME.

ljgray: I've just made this a config parameter

geoffryan (approved):
Once you have all tests passing, looks good to merge!

lib/stages/testDataGen.cpp
```diff
@@ -458,7 +467,7 @@ void testDataGen::main_thread() {
                 temp_output = ((new_real << 4) & 0xF0) + (new_imaginary & 0x0F);
                 frame[j] = temp_output;
             } else {
-                ERROR("unexpected type: {s}", type);
+                ERROR("unexpected type: {:s}", type);
```
rhaas80, line 501: more those are around :-(

lib/stages/testDataGen.cpp
```diff
@@ -306,6 +312,9 @@ void testDataGen::main_thread() {
         } else if (type == "square") {
             if (chordmeta)
                 chordmeta->type = kotekan::int4x2;
+        } else if (type == "onehot") {
+            if (chordmeta)
+                chordmeta->type = kotekan::int64;
         } else {
             ERROR("unexpected type: {s}", type);
```
rhaas80: also has the bad format `{s}`

lib/stages/testDataGen.cpp
```diff
@@ -285,6 +285,9 @@ void testDataGen::main_thread() {
             if (chordmeta)
                 chordmeta->type = kotekan::float16;
 #endif
+        } else if (type == "random") {
+            if (chordmeta)
+                chordmeta->type = kotekan::uint4x2;
```
rhaas80, line 303: did you add these?

lib/stages/testDataGen.cpp
ljgray: This is actually the one I was _going_ to fix because it's where the error was coming from...

lib/stages/testDataGen.cpp
ljgray: Yes, although I wasn't entirely sure about what type to use. Either way, the issue was that `ramp`, which I was trying to use, was not included as an option here and ended up hitting the ERROR in `else`, triggering the format error.

ljgray: Given that a runtime error is also thrown, it seems like this intends for all available type options to be checked.

ljgray:
@geoffryan Just re-pinging for review due to the additions to `testDataGen`.

lib/stages/testDataGen.cpp
```diff
@@ -306,8 +312,11 @@ void testDataGen::main_thread() {
         } else if (type == "square") {
             if (chordmeta)
                 chordmeta->type = kotekan::int4x2;
+        } else if (type == "onehot") {
+            if (chordmeta)
+                chordmeta->type = kotekan::int64;
```
geoffryan: I think `onehot` should be a `kotekan::uint8`. It's accessing the frame via the uint8 pointer.

geoffryan (approved):
Just address the `onehot` type, then LGTM!

PR #1421, opened 2026-02-24
add stage to parse input reorder table for chime_xpos2048

rhaas80:
This parses the yaml file member that is also used by (say) the FRB baseband archiver, so that there is a single source of truth for this.

PR #1424, opened 2026-02-25
Remove F-Engine simulator from Kotekan

eschnett:
See https://github.com/eschnett/FEngine.jl for its new, shiny version. Having a separate Julia package makes it easier to run and maintain.

- Create C++ stages that fill various auxiliary buffers (scatter indices, phases, weights)
- Add a new HDF5 file reader that understands the output format of the Julia F-Engine
- Clean up some parameter names
- Regenerate kernels


config/fengine/include/input_chime.j2
```diff
@@ -0,0 +1,20 @@
+# -*-yaml-*-
+
+# CHORD pathfinder read initial conditions
```
rhaas80: CHIME read initial conditions

config/fengine/include/input_chime.j2
```diff
@@ -0,0 +1,20 @@
...
+################################################################################
+
+# Stages
+
+{% set frequency_channels = [1024, 1088, 1152, 1216, 1280, 1344, 1408, 1472, 1536, 1600, 1664, 1728, 1792, 1856, 1920, 1984,] %}
+
+read_data:
+{% for id in range(16) %}   # num_frequencies = 16
+    read_voltage_{{ id }}:
+        kotekan_stage: hdf5FileReadSingleFile
```
rhaas80, line 14: if this is the special FEngine HDF5 reader, then I would try and reflect that in its name

lib/core/Stage.cpp
```diff
@@ -105,10 +105,10 @@ void Stage::start() {
...
             ERROR("Stage {:s} caught FatalError: {:s}, attempting controlled shutdown.",
                   unique_name, e.what());
-        } catch (const std::exception& e) {
-            ERROR("Exception in stage {:s}: {:s}", unique_name, e.what());
-            // throw will throw the same exception.
-            throw;
+        // } catch (const std::exception& e) {
+        //     ERROR("Exception in stage {:s}: {:s}", unique_name, e.what());
+        //     // throw will throw the same exception.
+        //     throw;
```
rhaas80: while I approve of this change (and have done the same in my code), this seems outside of the scope of this pull request :)

lib/cuda/cudaFRBBeamReformer.cpp
```diff
@@ -49,12 +49,12 @@ cudaFRBBeamReformer::cudaFRBBeamReformer(Config& config, const std::string& uniq
...
     // Input and output buffer names
     _gpu_mem_beamgrid = config.get<std::string>(unique_name, "gpu_mem_beamgrid");
-    _gpu_mem_phase = config.get<std::string>(unique_name, "gpu_mem_phase");
+    _gpu_mem_weights = config.get<std::string>(unique_name, "gpu_mem_weights");
     _gpu_mem_beamout = config.get<std::string>(unique_name, "gpu_mem_beamout");
 
     // Calculate buffer sizes (in bytes)
     beamgrid_len = sizeof(float16_t) * num_input_beams * _max_num_local_freq * _Td;
-    phase_len = sizeof(float16_t) * num_input_beams * _num_beams * _num_local_freq;
+    weights_len = sizeof(float16_t) * num_input_beams * _num_beams * _num_local_freq;
```
rhaas80: strange naming convention. I'd not have expected the `sizeof(float16_t)` for a variable called "length" (though maybe would have for "size"). Anyhow, this is not a new change, so I am too late complaining

lib/metadata/chordMetadata.cpp
```diff
@@ -106,6 +106,7 @@ void chordMetadata::check_frame_desc(
 
 void chordMetadata::set_from_frame_desc(
     const std::shared_ptr<const kotekan::GenericNDArray>& frame_desc) {
+    strncpy(this->name, frame_desc->get_quantity_name().get_c_string(), sizeof this->name);
```
rhaas80: Dangerous in that `strncpy` does not actually guarantee that the resulting string is `NUL` terminated.

A better solution may be d3f7d5a3b62a0b018dfb22d3189b992f4a0ccb95 (which I had forgotten to include in my latest "minor changes" pull request :-( ).

lib/stages/UpchannelizationSchedule.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+std::map<int, std::set<int>> UpchannelizationSchedule::make_upchan_factors_to_channels() const {
+    std::map<int, std::set<int>> upchan_factors_to_channels;
+    // for (const int upchan_factor : upchan_factors) {
+    //     std::ostringstream key;
+    //     key << "upchan_U" << upchan_factor << "_channels";
+    //     const auto channels = config.get<std::vector<int>>(unique_name, key.str());
+    //     upchan_factors_to_channels[upchan_factor] = std::set<int>(channels.begin(),
+    //     channels.end());
+    // }
```
rhaas80, line 33: remove commented out (debug?) code.

lib/stages/UpchannelizationSchedule.cpp
```diff
@@ -0,0 +1,205 @@
...
+    for (const int upchan_factor : upchan_factors) {
+        std::ostringstream key;
+        key << "upchan_channel_ranges/" << upchan_factor << "";
```
rhaas80, line 36: what does the `<< ""` at the end do?

lib/stages/UpchannelizationSchedule.cpp
```diff
@@ -0,0 +1,205 @@
...
+        const auto& range = config.get<std::vector<int>>(unique_name, key.str());
```
rhaas80, line 37: maybe use `get<nlohmann::json>` and then use it to dig down to the data instead? That seems to be how this is handled in other parts of kotekan.

lib/stages/UpchannelizationSchedule.cpp
```diff
@@ -0,0 +1,205 @@
...
+std::map<int, int> UpchannelizationSchedule::make_frequency_channels_to_indices() const {
+    std::map<int, int> channels_to_indices;
+    for (std::size_t index = 0; index < frequency_channels.size(); ++index)
+        channels_to_indices[frequency_channels.at(index)] = index;
+    return channels_to_indices;
+}
+
+std::set<int> UpchannelizationSchedule::make_upchan_factors() const {
+    const auto upchan_factors = config.get<std::vector<int>>(unique_name, "upchan_factors");
+    return std::set<int>(upchan_factors.begin(), upchan_factors.end());
```
rhaas80, line 22: should duplicates be reported as errors? They are probably typos.

lib/stages/UpchannelizationSchedule.cpp
```diff
@@ -0,0 +1,205 @@
...
+        key << "upchan_channel_ranges/" << upchan_factor << "";
+        const auto& range = config.get<std::vector<int>>(unique_name, key.str());
+        assert(range.size() == 2);
+        std::set<int> channels;
+        for (int channel = range.at(0); channel < range.at(1); ++channel) {
+            // Only include the local channels
+            if (has_frequency_channel(channel))
+                channels.insert(channel);
+        }
+        upchan_factors_to_channels[upchan_factor] = channels;
```
rhaas80, line 45: do we care about duplicates being silently replaced with the last copy encountered?

lib/stages/UpchannelizationSchedule.cpp
```diff
@@ -0,0 +1,205 @@
...
+    for (const int upchan_factor : upchan_factors) {
+        const auto& channels = get_upchan_channels(upchan_factor);
+        for (const int channel : channels)
+            upchan_channels_to_factors[channel].insert(upchan_factor);
+    }
+    return upchan_channels_to_factors;
+}
+
+void UpchannelizationSchedule::output_statistics() const {
+    const auto& chord_telescope = Telescope::instance().cast<CHORDTelescope>();
```
rhaas80, line 61: is this CHORD specific? Otherwise using just `Telescope` would be better (CHIME is derived from `ICETelescope` for example)

lib/stages/UpchannelizationSchedule.cpp
```diff
@@ -0,0 +1,205 @@
...
+        }
+    }
+
+    return true;
+}
+
+UpchannelizationSchedule::UpchannelizationSchedule(kotekan::Config& config) :
+    kotekan::kotekanLogging(),
+    //
+    unique_name(""), config(config),
```
rhaas80, line 168: a `unique_name` of "" seems strange. Is it used anywhere?

lib/stages/UpchannelizationSchedule.cpp
```diff
@@ -0,0 +1,205 @@
...
+    upchan_factors_to_channels(make_upchan_factors_to_channels()),
+    upchan_channels_to_factors(make_upchan_channels_to_factors())
+//
+{
+    output_statistics();
+    assert(invariant());
+}
+
+const UpchannelizationSchedule& UpchannelizationSchedule::instance(kotekan::Config& config) {
+    static const UpchannelizationSchedule the_instance(config);
```
rhaas80, line 184: are we concerned about `instance(config)` being called more than once with different `config` values (the first one would "win")?

lib/stages/calcBBPhase.cpp
```diff
@@ -0,0 +1,263 @@
+#include "CHORDTelescope.hpp"
```
rhaas80, line 1: specific to CHORD?

lib/stages/calcBBPhase.cpp
```diff
@@ -0,0 +1,263 @@
...
+                  return const_cast<kotekan::Stage&>(stage).main_thread();
+              }),
+        bb_beam_positions_buffer(get_buffer("bb_beam_positions")), A_buffer(get_buffer("bb_phase")),
+        s_buffer(get_buffer("bb_shift"))
+    //
+    {
+        assert(bb_beam_positions_buffer);
+        assert(A_buffer);
+        assert(s_buffer);
+        bb_beam_positions_buffer->register_producer(unique_name);
```
rhaas80, line 63: check that size of frame is as expected?

lib/stages/calcBBPhase.cpp
```diff
@@ -0,0 +1,263 @@
...
+        const float dish_separation_x = chord_telescope.get_dish_separation_x_m();
+        const float dish_separation_y = chord_telescope.get_dish_separation_y_m();
+        const auto& dish_grid = chord_telescope.get_dish_grid();
+        assert(std::ptrdiff_t(chord_telescope.get_num_dishes()) == num_dishes);
+        std::vector<float> dish_loc_x(num_dishes, -1), dish_loc_y(num_dishes, -1),
+            dish_loc_z(num_dishes, -1);
+        assert(std::ptrdiff_t(dish_grid.get_dish_indices().size()) >= num_dishes);
+        for (const auto dish_index : dish_grid.get_dish_indices()) {
+            if (dish_index >= 0) {
+                const dishInfo& dish_info = chord_telescope.get_dish_at_idx(dish_index);
```
rhaas80, line 91: ok yes, this is CHORD specific

lib/stages/calcBBPhase.cpp
```diff
@@ -0,0 +1,263 @@
...
+        std::int32_t* const s_frame = static_cast<std::int32_t*>(
+            static_cast<void*>(s_buffer->wait_for_empty_frame(unique_name, frame_id)));
+        if (!s_frame)
+            return;
+
+        // Check buffer sizes
+        assert(std::ptrdiff_t(bb_beam_positions_buffer->frame_size)
+               == bb_beam_positions_frame_size);
+        assert(std::ptrdiff_t(A_buffer->frame_size) == A_frame_size);
+        assert(std::ptrdiff_t(s_buffer->frame_size) == s_frame_size);
```
rhaas80, line 135: could be done in constructor (so errors out earlier)

lib/stages/calcBBPhase.cpp
```diff
@@ -0,0 +1,263 @@
...
+                                                   / sizeof *bb_beam_positions_frame));
+                    bb_beam_positions_frame[idx + 0] = theta_x;
+                    bb_beam_positions_frame[idx + 1] = theta_y;
+                }
+            }
+        }
+
+        // Set A
+        {
+            const float c0 = 299792458.0f; // speed of light in vacuum [m/s]
```
rhaas80, line 193: a magic number. admittedly fixed by definition, at least up to typos.

lib/stages/calcBBPhase.cpp
```diff
@@ -0,0 +1,263 @@
...
+            const std::ptrdiff_t str_dish = 1;
```
rhaas80, line 194: shouldn't there be a "components" looking at line 146? Or is this component of the complex number and we are writing both at once?

lib/stages/calcFRB2Weights.cpp
```diff
@@ -0,0 +1,296 @@
...
+                  return const_cast<kotekan::Stage&>(stage).main_thread();
+              }),
+        frb2_beam_positions_buffer(get_buffer("frb2_beam_positions")),
+        W2_buffer(get_buffer("frb2_weights"))
+    //
+    {
+        assert(frb2_beam_positions_buffer);
+        assert(W2_buffer);
+        frb2_beam_positions_buffer->register_producer(unique_name);
+        W2_buffer->register_producer(unique_name);
```
rhaas80, line 59: can check size here already

lib/stages/calcFRB2Weights.cpp
```diff
@@ -0,0 +1,296 @@
...
+            //   cos(dra)~1 nhat_x ~ cos(zd) * dra
+            // (but here we don't use the small-angle approx)
+
+            // Start timer
+            DEBUG("Calculating FRB2 beam weights...");
+            const double t0 = current_time();
+
+            using std::cos, std::sin;
+
+            const float c0 = 299792458.0f; // speed of light in vacuum [m/s]
```
rhaas80, line 226: magic number

lib/testing/inventBFMask.cpp
```diff
@@ -0,0 +1,96 @@
+#include "Config.hpp"            // for Config
```
rhaas80, line 1: would't testDataGen produce this as well?

lib/testing/inventPLMask.cpp
```diff
@@ -0,0 +1,115 @@
+#include "Config.hpp"            // for Config
```
rhaas80, line 1: doable with testDataGen?

lib/stages/setScatterIndices.cpp
```diff
@@ -0,0 +1,95 @@
...
+        // Set buffer
+        const std::ptrdiff_t str_dish = 1;
+        const std::ptrdiff_t str_polr = str_dish * num_dishes;
+        for (int polr = 0; polr < num_polarizations; ++polr) {
+            for (int dish = 0; dish < num_dishes; ++dish) {
+                const std::ptrdiff_t idx = str_dish * dish + str_polr * polr;
+                assert(idx >= 0
+                       && idx < std::ptrdiff_t(scatter_indices_buffer->frame_size
+                                               / sizeof *scatter_indices_frame));
+                scatter_indices_frame[idx] = scatter_indices.at(idx);
```
rhaas80, line 80: special case of 94c703817c4625f6959eeb2a3570490763d082ac but more importantly this scatters the mapping information across two places in the yaml file. 

Two places b/c the `baseband_archiver.py` script used for FRB actually parses CHIME's input yaml file to get this mapping. So it would be better to keep a single source of truth and use something like https://github.com/kotekan/kotekan/pull/1421 (since this is only used for CHIME anyway, is it?)

lib/stages/setUpchanGain.cpp
```diff
@@ -0,0 +1,116 @@
...
+        // Wait for buffer
+        DEBUG("[{:s}/{:d}] Waiting for buffer...", upchan_gain_buffer->buffer_name, frame_index);
+        float16_t* const upchan_gain_frame = static_cast<float16_t*>(
+            static_cast<void*>(upchan_gain_buffer->wait_for_empty_frame(unique_name, frame_id)));
+        if (!upchan_gain_frame)
+            return;
+
+        // Check buffer size
+        assert(upchan_gain_buffer->frame_size
+               == sizeof(float16_t) * upchan_max_num_channels * upchan_factor);
```
rhaas80: can be done in constructor

lib/stages/calcFRB2Weights.cpp
eschnett: `#include <PhysicalConstants/CODATA2022.h>`

config/fengine/include/input_chime.j2
eschnett: Done.

lib/core/Stage.cpp
jbmertens: This provides additional logging, right? Why remove it?

config/fengine/include/input_chime.j2
eschnett: The file format is generic. It is currently the same as the regular Kotekan format. However, as the Kotekan format evolves and metadata change, we may or may not want to update this reader because regenerating data is expensive.

The main difference is that the file is kept open, and that its time samples are read in chunks, and that times/frequencies can be transposed if needed for CHIME. I started implementing these features into the regular HDF5 file reader and this turned into a mess, with most complications coming from handling all the possible options. (Read all the data or only a hyperslab? When should the file be opened? Does it need to be closed? Should dishes and polarizations be combined? Should frequencies and times be transposed? Does the file even have dishes, or polarizations, or frequencies, or times? How is the dish layout being described? Is there even a dish layout? Which attributes need or can be present? ...)

In the end I decided that writing a new stage isn't difficult, and that it's simpler and shorter to have two separate stages.

lib/core/Stage.cpp
eschnett: I've removed that change again.

rhaas80: it breaks gdb since the stack trace no longer leads to the actual source of the exception.

rhaas80 (approved):

lib/core/Stage.cpp
jbmertens: gdb can break on the original throw though, `catch throw`, right? 
But we could also adjust behavior depending on debug/test/release builds.

rhaas80: that is not what I experienced, unfortunately:

```
haengie2: ~/tmp$ nl except.cc
     1  #include <exception>
     2  #include <iostream>

     3  void trigger(void) {
     4    throw std::runtime_error("From trigger");
     5  }

     6  int main(void) {
     7    try {
     8      trigger();
     9    } catch (const std::exception& e) {
    10      std::cerr << "caught exception" << std::endl;
    11      throw;
    12    }

    13    return 0;
    14  }
haengie2: ~/tmp$ gdb -ex run -ex bt ./a.out
[...]
Reading symbols from ./a.out...
Starting program: /home/rhaas/tmp/a.out
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/x86_64-linux-gnu/libthread_db.so.1".
caught exception
terminate called after throwing an instance of 'std::runtime_error'
  what():  From trigger

Program received signal SIGABRT, Aborted.
__pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0)
    at ./nptl/pthread_kill.c:44
? warning: 44   ./nptl/pthread_kill.c: No such file or directory
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6,
    no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
#1  0x00007ffff7aa147f in __pthread_kill_internal (threadid=<optimized out>, signo=6)
    at ./nptl/pthread_kill.c:89
#2  0x00007ffff7a4a942 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007ffff7a324ac in __GI_abort () at ./stdlib/abort.c:77
#4  0x00007ffff7ca6aa4 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff7cb915a in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff7ca662e in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007ffff7ca6645 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#8  0x0000555555555280 in main () at except.cc:13

File: lib/stages/calcBBPhase.cpp [hunk 26]
```diff
@@ -0,0 +1,265 @@
...
+#include <cassert>
+#include <cstdint>
+#include <string>
+#include <unistd.h>
+#include <vector>
+
+
+class calcBBPhase : public kotekan::Stage {
+    // Telescope setup
+    const int num_components = config.get<int>(unique_name, "num_components");
```
jbmertens, line 19: Should we grab these quantities below from the telescope object (as is possible)?

lib/stages/calcBBPhase.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+        // Set metadata
+        bb_beam_positions_buffer->allocate_ndarray_frame_desc<float, 2>(
+            "bb_beam_positions", {bb_num_beams, 2}, {"B", "X/Y"});
+        bb_beam_positions_buffer->allocate_new_metadata_object(frame_id);
+        const auto& bb_beam_positions_meta =
+            get_chord_metadata(bb_beam_positions_buffer->get_metadata(frame_id));
+        bb_beam_positions_meta->set_from_frame_desc(
+            bb_beam_positions_buffer->get_ndarray_frame_desc());
+        bb_beam_positions_meta->set_fpga_seq_num(0);           // ???
```
jbmertens: Checking up on  // ??? 
I guess this can be changed later if needed.

lib/stages/calcFRB2Weights.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+#include <cassert>
+#include <cstdint>
+#include <string>
+#include <unistd.h>
+#include <vector>
+
+class calcFRB2Weights : public kotekan::Stage {
+    // Telescope setup
+    const int num_dishes = config.get<int>(unique_name, "num_dishes");
```
jbmertens, line 20: Similar here. Maybe we should grab this from the telescope object.

lib/stages/inventBFMask.cpp
```diff
@@ -0,0 +1,96 @@
...
+        // Set metadata
+        buffer->allocate_ndarray_frame_desc<std::int8_t, 2>(
+            "bf_mask", {num_polarizations, num_dishes}, {"P", "D"});
+        buffer->allocate_new_metadata_object(frame_id);
+        const auto& meta = get_chord_metadata(buffer->get_metadata(frame_id));
+        meta->set_from_frame_desc(buffer->get_ndarray_frame_desc());
+        meta->set_fpga_seq_num(0);           // ???
+        meta->set_time_downsampling_fpga(1); // ???
+
+        // We should probably set this...
```
jbmertens: Checking in on this.

lib/stages/inventPLMask.cpp
```diff
@@ -0,0 +1,115 @@
...
+                {div_noremainder(num_times, 2 * 64), div_noremainder(num_frequencies, 4),
+                 num_polarizations, div_noremainder(num_dishes, 8), 64 / 8},
+                {"T2hi64", "F4", "P", "D8", "T2lo64"});
+            buffer->allocate_new_metadata_object(frame_id);
+            const auto& meta = get_chord_metadata(buffer->get_metadata(frame_id));
+            meta->set_from_frame_desc(buffer->get_ndarray_frame_desc());
+            meta->set_fpga_seq_num(0);                // ???
+            meta->set_time_downsampling_fpga(2 * 64); // ???
+
+            // We should probably set this...
```
jbmertens: Flagging this too.
(Grab from telescope?)

lib/testing/inventPLMask.cpp
```diff
@@ -0,0 +1,115 @@
...
+                        for (int dish = 0; dish < num_dishes; dish += 8) {
+                            const int idx4 = time / 2 % 64 / 8;
+                            const int idx3 = dish / 8;
+                            const int idx2 = polr;
+                            const int idx1 = freq / 4;
+                            const int idx0 = time / 2 / 64;
+                            const std::ptrdiff_t idx =
+                                str4 * idx4 + str3 * idx3 + str2 * idx2 + str1 * idx1 + str0 * idx0;
+                            assert(idx >= 0 && idx < std::ptrdiff_t(buffer->frame_size));
+                            frame[idx] = 0xff; // no packets were lost
```
jbmertens, line 101: Do we want other options?
Otherwise why not just use the testDataGen stage?

lib/stages/setFRB1Phase.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include "kotekanLogging.hpp"
+
+#include <cassert>
+#include <cstdint>
+#include <string>
+#include <unistd.h>
+#include <vector>
+
+class setFRB1Phase : public kotekan::Stage {
+    // Telescope layout
```
jbmertens, line 16: Consider grabbing from the telescope object.

lib/stages/setScatterIndices.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include "kotekanLogging.hpp"
+
+#include <cassert>
+#include <cstdint>
+#include <string>
+#include <unistd.h>
+#include <vector>
+
+class setScatterIndices : public kotekan::Stage {
+    // Telescope layout
```
jbmertens, line 16: Consider grabbing from the telescope object.

jbmertens (commented):
Few additional comments.

lib/testing/inventBFMask.cpp
geoffryan: I believe so.  If we'd like a separate stage just for this though, it should be in `lib/testing` not `lib/stages/`

lib/testing/inventPLMask.cpp
geoffryan: I believe so.  If we'd like a separate stage just for this though, it should be in `lib/testing` not `lib/stages/`

lib/cuda/cudaFRBBeamReformer.cpp
eschnett: Fixed.

lib/metadata/chordMetadata.cpp
eschnett: All the metadata string fields are like that. (They used to be part of a fixed binary format.) We want to get rid of them anyway – I'm just waiting for a `frame_desc` infrastructure for GPU arrays.

eschnett: These files have a maximum length, and they are documented to not be NUL terminated if the strings have this maximum length. So, `snprintf` will just waste one byte.

If we fix this we should just go to `std::string`.

lib/testing/inventBFMask.cpp
eschnett: Maybe? I don't know how flexible this stage has become.

lib/stages/setScatterIndices.cpp
eschnett: The scatter indices are currently in one place only, namely the place where the transpose kernel is called. If there's a better way we should switch things over and remove this stage.

lib/stages/setUpchanGain.cpp
eschnett: Done.

lib/stages/calcBBPhase.cpp
eschnett: There are many stages which look at these parameters. These are essentially global parameters at the moment. Yes, we should change this to use the telescope object, but maybe not in this PR?

lib/stages/calcBBPhase.cpp
eschnett: I'm setting this to 0 because we're calculating the baseband beam phases only once. Once we recalculate them periodically this needs to change.

eschnett: But, more importantly: This buffer does not contain any time samples, so these parameters shouldn't be necessary! However, some other stages (copy between host and gpu) currently require them.

lib/stages/inventBFMask.cpp
eschnett: I'm unsure how important our list of dish position in the metadata is at the moment. This should all go away (replaced by the telescope).

lib/stages/calcBBPhase.cpp
jbmertens: Sounds good.

lib/stages/inventPLMask.cpp
eschnett: Thanks. This is actually a bug!

lib/testing/inventPLMask.cpp
jbmertens: (Discussed above)

lib/stages/inventPLMask.cpp
eschnett: Well, not the dish thingy, but the lines above you flagged – the sequence number is wrong.

eschnett: Fixed.

lib/testing/inventBFMask.cpp
eschnett: I want to keep these (inventing BF and PL masks) as separate stages so that we can add respective features to test with bad feeds and packet loss. I think extending testDataGen for this would be too specific. I'll move them the `testing`.

lib/stages/setScatterIndices.cpp
eschnett: Yes, #1421 looks good. The current code is just copied from the now-removed `FEngine.cpp`, so that's not a new feature.

lib/testing/inventPLMask.cpp
eschnett: Moved to `lib/testing`.

lib/stages/calcFRB2Weights.cpp
eschnett: Same as above: The current code is copied from `FEngine.cpp`. All the `num_dishes` etc. should go away at some point. Same with `num_frequencies`, although there doesn't seem to be a global object for handling frequencies yet.

lib/metadata/chordMetadata.cpp
rhaas80: I am not convinced all use of `name` and `dim_name` etc is fully aware of it possibly not being `NUL` terminated, though a quick grep only digs up one bug (and it might be mine):

```
193 ERROR("name={} type={} dims={} laststride={}", meta->name,
194       type_to_string(frame_desc->get_value_datatype()),
195       fmt::join(extents, ", ")
196       frame_desc->get_stride(frame_desc->get_rank() - 1));
```

in `lib/stages/asdfFileWrite.cpp`. Thankfully often the names are compile time constants, which I believe results in a compile time warning if the field is too small.

Anyhow, I agree, that the other fields are similar and hopefully will not be necessary for much longer.

PR #1425, opened 2026-02-25
Fix Indexing Bug In RFI Kernels

geoffryan:
- Adds unit tests for some of the `n2k` RFI kernels.
- Fixes indexing bugs in `cudaRFIS012`, `cudaRFIS012tilde`.

### DIFF
--- a/config/ci-tests/gpu_batch/verify_cuda_rfi_s012.yaml
+++ b/config/ci-tests/gpu_batch/verify_cuda_rfi_s012.yaml
@@ -1,8 +1,8 @@
 ##########################################
 #
-# verify_cuda_n2k.yaml
+# verify_cuda_rfi_s012.yaml
 #
-# Config to test Kendrick's cuda RFI S012 kernel
+# Config to test the cuda RFI S012 kernel
 #
 ##########################################
 ---
--- a/config/ci-tests/gpu_batch/verify_cuda_rfi_s012.yaml
+++ b/config/ci-tests/gpu_batch/verify_cuda_rfi_s012.yaml
@@ -14,7 +14,7 @@ log_level: INFO
 
 num_gpus: 1
 buffer_depth: 4
-cpu_affinity: [2,]
+cpu_affinity: [0,]
 sizeof_float: 4
 sizeof_int: 4
 sizeof_long: 8
--- a/config/ci-tests/gpu_batch/verify_cuda_rfi_s012.yaml
+++ b/config/ci-tests/gpu_batch/verify_cuda_rfi_s012.yaml
@@ -14,23 +14,20 @@ log_level: INFO
 
 num_gpus: 1
 buffer_depth: 4
-cpu_affinity: [2,3,4,5,8,9,10,11]
+cpu_affinity: [2,]
 sizeof_float: 4
-frame_arrival_period: 5.12e-6 * samples_per_data_set
 sizeof_int: 4
 sizeof_long: 8
 
 # For faster CHORD testing, we're using...
-samples_per_data_set: 128
-# sub_integration_ntime: 8192  # ~40 ms for chord
-# samples_per_data_set: 8192   # 
-num_times: samples_per_data_set
+samples_per_data_set: 1024
+frame_arrival_period: 5.12e-6 * samples_per_data_set
 num_polarizations: 2
 num_dishes: 64
 num_elements: num_polarizations * num_dishes
-num_local_freq: 4
+num_local_freq: 8
 
-rfi_downsampling_factor: 128
+rfi_downsampling_factor: 256
 
 # Pool
 main_pool:
@@ -87,23 +84,19 @@ host_rfi_s012_ringbuffer:
 
 gen_data_voltage:
     type: random_signed_offset
-    type: const_offset
-    value: (1 + 8) * 16 + (0 + 8)
     kotekan_stage: testDataGen
     out_buf: raw_voltage_buffer
     name: "E"
-    array_shape: [ 128, 4, 2, 64 ]
+    array_shape: [ 1024, 8, 2, 64 ]
     dim_name: ["T", "F", "P", "D" ]
     num_frames: 2*buffer_depth  # Must be at least what check_data expects
 
 gen_pl_mask:
-  # type: const1x8 
     type: random1x8 
-    value: 255
     kotekan_stage: testDataGen
     out_buf: host_pl_mask_buffer
     name: "pl_mask"
-    array_shape: [ 1, 1, 2, 8, 8 ]  # = num_times/128, num_freq/4, 2, num_dishes/8, 8
+    array_shape: [ 8, 2, 2, 8, 8 ]  # = num_times/128, num_freq/4, 2, num_dishes/8, 8
     dim_name: ["T2hi64", "F4", "P", "D8", "T2lo64" ]
     num_frames: 2*buffer_depth  # Must be at least what check_data expects
     meta_time_downsample_factor: 128  # The number of time samples represented by each coarse time index.
@@ -187,17 +180,8 @@ cpu_rfi_s012:
     in_plmask_buf: host_pl_mask_buffer
     out_buf: simulated_rfi_s012_buffer
 
-# eat_cpu_rfi_s012:
-#   kotekan_stage: dropAllFrames
-#   in_buf: simulated_rfi_s012_buffer
-#   max_frames: 2 * buffer_depth
-# 
-# eat_rfi_s012:
-#   kotekan_stage: dropAllFrames
-#   in_buf: host_rfi_s012_buffer
-#   max_frames: 2 * buffer_depth
 
-check_correlation_data:
+check_rfi_data:
   kotekan_stage: testDataCheckUlong
   first_buf: host_rfi_s012_buffer
   second_buf: simulated_rfi_s012_buffer
--- a/config/ci-tests/gpu_batch/verify_cuda_rfi_s012.yaml
+++ b/config/ci-tests/gpu_batch/verify_cuda_rfi_s012.yaml
@@ -1,0 +1,206 @@
+##########################################
+#
+# verify_cuda_n2k.yaml
+#
+# Config to test Kendrick's cuda RFI S012 kernel
+#
+##########################################
+---
+type: config
+# Logging level can be one of:
+# OFF, ERROR, WARN, INFO, DEBUG, DEBUG2 (case insensitive)
+# Note DEBUG and DEBUG2 require a build with (-DCMAKE_BUILD_TYPE=Debug)
+log_level: INFO
+
+num_gpus: 1
+buffer_depth: 4
+cpu_affinity: [2,3,4,5,8,9,10,11]
+sizeof_float: 4
+frame_arrival_period: 5.12e-6 * samples_per_data_set
+sizeof_int: 4
+sizeof_long: 8
+
+# For faster CHORD testing, we're using...
+samples_per_data_set: 128
+# sub_integration_ntime: 8192  # ~40 ms for chord
+# samples_per_data_set: 8192   # 
+num_times: samples_per_data_set
+num_polarizations: 2
+num_dishes: 64
+num_elements: num_polarizations * num_dishes
+num_local_freq: 4
+
+rfi_downsampling_factor: 128
+
+# Pool
+main_pool:
+    kotekan_metadata_pool: chordMetadata
+    num_metadata_objects: 20 * buffer_depth
+
+# Buffers
+raw_voltage_buffer:
+    kotekan_buffer: standard
+    num_frames: buffer_depth
+    frame_size: samples_per_data_set * num_local_freq * num_elements
+    metadata_pool: main_pool
+
+host_voltage_buffer:
+    kotekan_buffer: standard
+    num_frames: buffer_depth
+    frame_size: samples_per_data_set * num_local_freq * num_elements
+    metadata_pool: main_pool
+
+host_pl_mask_buffer:
+    kotekan_buffer: standard
+    num_frames: buffer_depth
+    frame_size: ((samples_per_data_set / 2) * (num_local_freq / 4) * (num_elements / 8)) / 8
+    metadata_pool: main_pool
+
+host_rfi_s012_buffer:
+    kotekan_buffer: standard
+    num_frames: buffer_depth
+    frame_size: (samples_per_data_set / rfi_downsampling_factor) * num_local_freq * 3 * num_elements * sizeof_long
+    metadata_pool: main_pool
+
+simulated_rfi_s012_buffer:
+    kotekan_buffer: standard
+    num_frames: buffer_depth
+    frame_size: (samples_per_data_set / rfi_downsampling_factor) * num_local_freq * 3 * num_elements * sizeof_long
+    metadata_pool: main_pool
+
+
+# Ringbuffer signalling buffer
+host_voltage_ringbuffer:
+    kotekan_buffer: ring
+    ring_buffer_size: buffer_depth * samples_per_data_set * num_local_freq * num_elements 
+    metadata_pool: main_pool
+
+host_pl_mask_ringbuffer:
+    kotekan_buffer: ring
+    ring_buffer_size: buffer_depth * ((samples_per_data_set / 2) * (num_local_freq / 4) * (num_elements / 8)) / 8
+    metadata_pool: main_pool
+
+host_rfi_s012_ringbuffer:
+    kotekan_buffer: ring
+    ring_buffer_size: buffer_depth * (samples_per_data_set / rfi_downsampling_factor) * num_local_freq * 3 * num_elements * sizeof_long
+    metadata_pool: main_pool
+
+gen_data_voltage:
+    type: random_signed_offset
+    type: const_offset
+    value: (1 + 8) * 16 + (0 + 8)
+    kotekan_stage: testDataGen
+    out_buf: raw_voltage_buffer
+    name: "E"
+    array_shape: [ 128, 4, 2, 64 ]
+    dim_name: ["T", "F", "P", "D" ]
+    num_frames: 2*buffer_depth  # Must be at least what check_data expects
+
+gen_pl_mask:
+  # type: const1x8 
+    type: random1x8 
+    value: 255
+    kotekan_stage: testDataGen
+    out_buf: host_pl_mask_buffer
+    name: "pl_mask"
+    array_shape: [ 1, 1, 2, 8, 8 ]  # = num_times/128, num_freq/4, 2, num_dishes/8, 8
+    dim_name: ["T2hi64", "F4", "P", "D8", "T2lo64" ]
+    num_frames: 2*buffer_depth  # Must be at least what check_data expects
+    meta_time_downsample_factor: 128  # The number of time samples represented by each coarse time index.
+
+applyPL:
+  kotekan_stage: applyGenPL
+  voltage_in_buf: raw_voltage_buffer
+  voltage_out_buf: host_voltage_buffer
+  plmask_in_buf: host_pl_mask_buffer
+
+run_send_voltage:
+    gpu_0:
+        kotekan_stage: cudaProcess
+        gpu_id: 0
+        in_buffers:
+            host_voltage_buffer_in: host_voltage_buffer
+        out_buffers:
+            host_voltage_ringbuffer_out: host_voltage_ringbuffer
+        commands:
+        - name: cudaCopyToRingbuffer
+          input_size: samples_per_data_set * num_local_freq * num_elements
+          ring_buffer_size: buffer_depth * input_size
+          in_buf: host_voltage_buffer_in
+          signal_buf: host_voltage_ringbuffer_out
+          gpu_mem_output: voltage_buffer
+
+run_send_pl_mask:
+    gpu_0:
+        kotekan_stage: cudaProcess
+        gpu_id: 0
+        in_buffers:
+            host_pl_mask_buffer_in: host_pl_mask_buffer
+        out_buffers:
+            host_pl_mask_ringbuffer_out: host_pl_mask_ringbuffer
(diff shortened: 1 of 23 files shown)


rhaas80:
to be sure (since there is temporal overlap in work on this) this is different from https://github.com/kotekan/kotekan/pull/1416 (which is admittedly not touching that particular kernel I think).

geoffryan:
> to be sure (since there is temporal overlap in work on this) this is different from #1416 (which is admittedly not touching that particular kernel I think).

This is indeed a different bug.  the `s0_kernel.cu` and `s12_kernel.cu` read their input at ~`input_ptr + Tmin` but write their output to `output_ptr` directly. the `cudaRFIS012.cpp` wrapper assumed the kernels also applied the offset to the output, so data was only being written to the beginning of the output buffer.

rhaas80:
clean up history a bit?

lib/cuda/cudaRFIS012.cpp
```diff
@@ -211,19 +211,25 @@ cudaEvent_t cudaRFIS012::execute(cudaPipelineState& /*pipestate*/,
...
+                          Tmin, Tsize, num_frequencies, num_dishes * num_polarizations,
+                          rfi_downsampling_factor, F_stride, device.getStream(cuda_stream_id));
 #ifdef DEBUGGING
     CHECK_CUDA_ERROR(cudaStreamSynchronize(device.getStream(cuda_stream_id)));
 #endif
-    n2k::launch_s12_kernel((ulong*)(rfi_S012_memory + S_stride), (const uint8_t*)voltage_memory, T,
-                           Tmin, Tsize, num_frequencies, num_dishes * num_polarizations,
-                           rfi_downsampling_factor, F_stride, offset_encoded,
-                           device.getStream(cuda_stream_id));
+    n2k::launch_s12_kernel((ulong*)(rfi_S012_memory + S_stride + Trfi_offset),
```
rhaas80, line 249: this makes `rfi_S012_memory` different from the other inputs and outputs. To have the kernel treat all inputs and outputs the same, would it make sense to instead modify the kernel to do the offsetting the same way does for its other ringbuffer inputs / outputs?

config/ci-tests/gpu_batch/verify_cuda_rfi_s012tilde.yaml
```diff
@@ -0,0 +1,164 @@
...
+---
+type: config
+# Logging level can be one of:
+# OFF, ERROR, WARN, INFO, DEBUG, DEBUG2 (case insensitive)
+# Note DEBUG and DEBUG2 require a build with (-DCMAKE_BUILD_TYPE=Debug)
+log_level: INFO
+
+num_gpus: 1
+buffer_depth: 4
+cpu_affinity: [2,]
```
rhaas80: are we ever in danger of running on a 2 core node (eg one of github's runners may only offer 2 cores)? What happens in that case (is the setting ignored or are we then running on 0 cores?)

lib/stages/CMakeLists.txt
```diff
@@ -100,6 +100,7 @@ add_library(
     N2Accumulate.cpp
     N2TimeDownsample.cpp
     N2FringeStop.cpp
+    dropAllFrames.cpp
```
rhaas80, line 103: I also wrote such a stage ;-) Isn't this what kotekan does anyway if a buffer has no consumer? Ie such a stage is not needed?

lib/testing/applyGenPL.cpp
```diff
@@ -129,7 +133,7 @@ void applyGenPL::main_thread() {
                     if (pl)
                         output[v_idx] = input[v_idx];
                     else
-                        output[v_idx] = 8; // = 0 in offset-encoding
+                        output[v_idx] = 16 * 8 + 8; // = 0 in offset-encoding
```
rhaas80: so this is a `int4x2` really? `0x88` is what is used in other places (which is indeed `16*8+8`, yes).

lib/testing/gpuSimulateRFIS012.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+    if (_samples_per_data_set % _rfi_downsampling_factor != 0) {
+        FATAL_ERROR("samples_per_data_set must be a multiple of rfi_downsampling_factor");
+    }
+    assert(_samples_per_data_set % _rfi_downsampling_factor == 0);
+
+    int64_t nt = _samples_per_data_set / _rfi_downsampling_factor;
+    int64_t nf = _num_local_freq;
+    int64_t ne = _num_elements;
+    out_rfis012_buf->allocate_ndarray_frame_desc<uint64_t, 5>("S012", {nt, nf, 3, 2, ne / 2},
```
rhaas80: strictly speaking you will want to `assert(ne % 2 == 0)`.

lib/testing/gpuSimulateRFIS012.cpp
```diff
@@ -0,0 +1,211 @@
...
+    _samples_per_data_set = config.get<int64_t>(unique_name, "samples_per_data_set");
+    _rfi_downsampling_factor = config.get<int64_t>(unique_name, "rfi_downsampling_factor");
+
+    in_voltage_buf = get_buffer("in_voltage_buf");
+    in_voltage_buf->register_consumer(unique_name);
+    in_plmask_buf = get_buffer("in_plmask_buf");
+    in_plmask_buf->register_consumer(unique_name);
+
+    out_rfis012_buf = get_buffer("out_buf");
+    out_rfis012_buf->register_producer(unique_name);
```
rhaas80, line 47: check `frame_size` against inferred size based on input values?

lib/testing/gpuSimulateRFIS012.cpp
```diff
@@ -0,0 +1,211 @@
...
+        if (pl_mask == nullptr)
+            break;
+        uint8_t* voltage =
+            (uint8_t*)in_voltage_buf->wait_for_full_frame(unique_name, in_voltage_frame_id);
+        if (voltage == nullptr)
+            break;
+        uint64_t* rfi_s012 =
+            (uint64_t*)out_rfis012_buf->wait_for_empty_frame(unique_name, out_rfis012_frame_id);
+        if (rfi_s012 == nullptr)
+            break;
```
rhaas80, line 104: can check `ndarray_frame_desk` of  input buffers against expected description.

lib/testing/gpuSimulateRFIS012.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+        // Start with a copy
+        meta_out->deepCopy(meta_in);
+
+        meta_out->set_from_frame_desc(out_rfis012_buf->get_ndarray_frame_desc());
+
+        // test that things are consistent
+        meta_out->check_frame_desc(out_rfis012_buf->get_ndarray_frame_desc());
+
+        // Set non-NDArray things.
```
rhaas80, line 221: the `deepCopy` should have already set `fpga_seq_num`, did it not do so?

lib/testing/gpuSimulateRFIS012.hpp
```diff
@@ -0,0 +1,83 @@
...
+private:
+    Buffer* in_plmask_buf;
+    Buffer* in_voltage_buf;
+    Buffer* out_rfis012_buf;
+
+    // Config options
+    int64_t _num_local_freq;
+    int64_t _num_elements;
+    int64_t _samples_per_data_set;
+    int64_t _rfi_downsampling_factor;
```
rhaas80: make `const` and initialize outside of constructor function body?

eschnett:
History doesn't matter much, it's all squashed in the end anyway.

lib/cuda/cudaRFIS012.cpp
eschnett: If the ringbuffer indexing never wraps around then it's much simpler to do the indexing here (in Kotekan) and leave the n2k kernels unchanged. If they can wrap around then the indexing needs to happen in n2k.

I think we can control the buffer and chunk sizes and downsampling factors, and we can do the indexing here. I don't recall why I started doing the indexing in the n2k kernels. (That was a major headache.)

lib/stages/CMakeLists.txt
eschnett: Indeed, I never needed such a stage.

lib/cuda/cudaRFIS012.cpp
rhaas80: ghost points and adjacent stages with input / output requirements that do not have small least-common-multiples of each other  were mentioned

lib/stages/CMakeLists.txt
geoffryan: This stage also counts consumed frames and triggers an exit when max_frames has been hit, is nice for some tests.

rhaas80: The writer stages (`hdf5WriteFile` for example) can also be (ab)used for that if one sets `skip_writing` to true and sets a `max_frames` value.

Does not count (visibly) the frames it has dropped though (it counts internally of course to be able to exit when `max_frames` has been reached).

It's an abuse admittedly, but works nicely even if there are multiple stages, all of which should process N frames (since the writers coordinate and do not exit kotekan until all of them have reached their `max_frames` value).

config/ci-tests/gpu_batch/verify_cuda_rfi_s012tilde.yaml
geoffryan: Ah, I don't think this has been a problem yet, but I can remove that line.

lib/testing/gpuSimulateRFIS012.cpp
geoffryan: Just grab the num_pol and num_dish now instead.

lib/testing/gpuSimulateRFIS012.cpp
geoffryan: It does, this is redundant I suppose, I just like being very explicit with the seq num.

lib/testing/gpuSimulateRFIS012.cpp
geoffryan: I think just checking the size (as I now do) is sufficient for this stage.

geoffryan:
Dealt with comments, moved initial time offset logic to RFI wrapper classes for RFIS012 and RFIS012tilde.

config/ci-tests/gpu_batch/verify_cuda_rfi_s012.yaml
```diff
@@ -0,0 +1,190 @@
...
+    dim_name: ["T", "F", "P", "D" ]
+    num_frames: 2*buffer_depth  # Must be at least what check_data expects
+
+gen_pl_mask:
+    type: random1x8 
+    kotekan_stage: testDataGen
+    out_buf: host_pl_mask_buffer
+    name: "pl_mask"
+    array_shape: [ 8, 2, 2, 8, 8 ]  # = num_times/128, num_freq/4, 2, num_dishes/8, 8
+    dim_name: ["T2hi64", "F4", "P", "D8", "T2lo64" ]
```
rhaas80, line 100: I'd have expected `T2lo64` to always have a size of `64`.

lib/cuda/cudaRFIS012.cpp
```diff
@@ -210,20 +216,40 @@ cudaEvent_t cudaRFIS012::execute(cudaPipelineState& /*pipestate*/,
     const std::ptrdiff_t Tmin = voltage.get_read_valid().begin() % Tsize;
     assert(Tmin % 128 == 0);
     const std::ptrdiff_t T = voltage.get_read_valid().size();
+    if (Tmin + T > Tsize) {
+        FATAL_ERROR("Chunk starting at Tmin={:d} of size T={:d} runs past end of ringbuffer {:s} "
+                    "of size Tsize={:d}",
+                    Tmin, T, voltage.get_buffer_name(), Tsize);
+    }
+    assert(Tmin + T <= Tsize);
+    const std::ptrdiff_t T_offset = Tmin * voltage.get_ndarray().stride(0);
```
rhaas80, line 225: should this really be `voltage.get_ndarray().stride(0)` and not eg `rfi_S012.get_ndarray().stride(0)` given that the base-name of the variable is `T` which seems to be used for `rfi_S012` times?

lib/cuda/cudaRFIS012.cpp
```diff
@@ -210,20 +216,40 @@ cudaEvent_t cudaRFIS012::execute(cudaPipelineState& /*pipestate*/,
...
+    const std::ptrdiff_t F_stride = rfi_S012.get_ndarray().stride(1);
+    const std::ptrdiff_t S_stride = rfi_S012.get_ndarray().stride(2);
     constexpr bool offset_encoded = true;
-    n2k::launch_s0_kernel((ulong*)rfi_S012_memory, (const ulong*)pl_mask_memory, T, Tmin, Tsize,
-                          num_frequencies, num_dishes * num_polarizations, rfi_downsampling_factor,
-                          F_stride, device.getStream(cuda_stream_id));
+
+    // Current offset into the rfi_S012 buffer
+    // n2k does not apply this itself
+    const std::ptrdiff_t Trfi_offset = Tmin / rfi_downsampling_factor * T_stride;
```
rhaas80, line 233: (I know voltage is never downsampled). Should this not be something like `Tmin * voltage_downsampling_factor / rfi_downsampling_factor * T_stride`? where `Tmin * voltage_downsampling_factor` recovers the global, non-downsampled time sample number?

lib/cuda/cudaRFIS012.cpp
```diff
@@ -210,20 +216,40 @@ cudaEvent_t cudaRFIS012::execute(cudaPipelineState& /*pipestate*/,
...
+
+    const auto& pl_mask_meta = pl_mask.get_metadata();
+    const std::ptrdiff_t Tpl_stride = pl_mask.get_ndarray().stride(0);
+    const std::ptrdiff_t Tpl_offset =
+        Tmin / pl_mask_meta->get_time_downsampling_fpga() * Tpl_stride;
```
rhaas80: same as above

lib/cuda/cudaRFIS012bar.cpp
```diff
@@ -175,22 +183,34 @@ cudaEvent_t cudaRFIS012bar::execute(cudaPipelineState& /*pipestate*/,
...
     const std::ptrdiff_t Trfibar_min = rfi_S012bar.get_write_valid().begin() % Trfibar_size;
     const std::ptrdiff_t Trfibar = rfi_S012bar.get_write_valid().size();
     DEBUG("Trfibar_size={:d} Trfibar_min={:d} Trfibar={:d}", Trfibar_size, Trfibar_min, Trfibar);
 
-    ulong* const Sout = rfi_S012bar_memory;
-    const ulong* const Sin = rfi_S012_memory;
+    // Time offsets to access S012 buffers
+    const std::ptrdiff_t Trfi_offset = Trfi_min * rfi_S012.get_ndarray().stride(0);
+    const std::ptrdiff_t Trfibar_offset =
+        Trfi_min / rfi_second_downsampling_factor * rfi_S012bar.get_ndarray().stride(0);
```
rhaas80, line 202: ok, this one is a relative downsmapling wrt to its (single) input, so fine.

lib/stages/dropAllFrames.cpp
```diff
@@ -0,0 +1,77 @@
+#include "dropAllFrames.hpp"
```
rhaas80, line 1: I would still maintain that such a stage is not required.

rhaas80: mabye move to testing (since it's extra feature compared to having no consumer at all is stopping after N frames)

lib/stages/dropAllFrames.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+        if (max_frames >= 0 && num_frames >= max_frames) {
+            INFO("Ate the requested number of frames ({:d}) - exiting", num_frames);
+            break;
+        }
+    }
+
+    if (max_frames >= 0) {
+        // Unregister to allow the pipeline to continue, unless I'm the last
+        // consumer on this buffer.
```
rhaas80, line 67: seems a bit strange. Does this mean "unregister unless I am the last consumer"?

lib/testing/gpuSimulateRFIS012.cpp
```diff
@@ -0,0 +1,234 @@
...
+using kotekan::Stage;
+using N2::frameID;
+
+REGISTER_KOTEKAN_STAGE(gpuSimulateRFIS012);
+
+gpuSimulateRFIS012::gpuSimulateRFIS012(Config& config, const std::string& unique_name,
+                                       bufferContainer& buffer_container) :
+    Stage(config, unique_name, buffer_container, std::bind(&gpuSimulateRFIS012::main_thread, this)),
+    _num_polarizations(config.get<int64_t>(unique_name, "num_polarizations")),
+    _num_dishes(config.get<int64_t>(unique_name, "num_dishes")),
```
rhaas80, line 34: ambitious number of dishes supported :-)

lib/testing/gpuSimulateRFIS012.hpp
```diff
@@ -0,0 +1,85 @@
...
+ *          equivalently [samples_per_data_set, num_local_freq,
+ *          num_polarizations, num_dishes]
+ *      @buffer_metadata chordMetadata
+ *
+ * @buffer out_rfis012_buf The output RFI S012 buffer
+ *      @buffer_format uint64_t
+ *      @buffer_shape [samples_per_data_set / rfi_downsampling_factor, num_local_freq,
+ *          3, num_polarizations, num_dishes] or equivalently
+ *          [samples_per_data_set / rfi_downsampling_factor, num_local_freq * 3 * num_elements]
+ *      @buffer_metadata chordMetadata. time_downsampling_fpga = rfi_downsampling_factor
```
rhaas80, line 56: the code actually sets it to `input-time-downsampling-fpga * rdi_downsampling_factor`. Admittedly ``input-time-downsampling-fpga` is always going to be 1.

rhaas80 (approved):
looks very good. The CPU implementations for the CI tests are great! They will help avoid many future bugs!

lib/stages/dropAllFrames.cpp
jbmertens: We do have monitorBuffer already, which can produce a similar behavior with the benefit of a less racy shutdown.

rhaas80: the shutdown should be fine, since it taps into the same global variable that the file writers use. I had not verified how this was done in `dropAllFrames.cpp` before. So long as kotekan initializes all stages before the first buffer gets send, there should be no race (if buffers do start to flow the moment the first stage is initialized then, in theory, one instance of `dropAllFrames` (or a file writer for that matter) could see its `N` frames and trigger an exit before a second stage is even initialized. Since this functionality os used for testing only (it is, is it?) the risk seems acceptable. :-)

config/ci-tests/gpu_batch/verify_cuda_rfi_s012.yaml
geoffryan: It indexes a single bit, but the array shape here is in bytes.

rhaas80: ah, right. I change my expectation then.

lib/cuda/cudaRFIS012.cpp
geoffryan: It's the other way around, `T` is the base (voltage) time.  `Trfi` is the `rfi_S012` times.

lib/cuda/cudaRFIS012.cpp
geoffryan: `rfi_downsampling_factor` is treated as the relative downsampling factor between the voltage and rfi buffers (not the absolute downsampling factor of the rfi buffer), so this is the correct thing do.

lib/cuda/cudaRFIS012.cpp
geoffryan: Ah yes here that is the right thing to do.

lib/stages/dropAllFrames.cpp
geoffryan: ~Moved to `lib/testing/`~   This doesn't like being moved to `lib/testing` since the atomic max_frames counter lives in `lib/stages`.  Having stages reach between these directories smells bad, so I'll leave this in `lib/stages`.

I've found it be a useful stage in tests when I'm really trying to isolate a particular stage, and still have test configs exit simply.

lib/testing/gpuSimulateRFIS012.cpp
geoffryan: Gotta dream big!

lib/testing/gpuSimulateRFIS012.hpp
geoffryan: I think this is still an accurate description, the voltage buffer contains `samples_per_data_set` entries and the RFI buffer contains `samples_per_data_set / rfi_downsampling_factor`.    Here if we wanted to entertain downsampled voltages that just means "samples_per_data_set" refers to the downsampled values in the voltage buffer, not the raw fpga counts.

lib/cuda/cudaRFIS012.cpp
rhaas80: consistent variable naming is for dummies anyway :-)

lib/stages/dropAllFrames.cpp
geoffryan: Yep!  This is in all of the waiting-for-max-frames blocks.   It helps avoid a pipeline freeze if some stages are still consuming a buffer after some waiting-for-max-frames stages have dropped out.  Not a common situation to be in but i've found myself there once or twice.

lib/stages/dropAllFrames.cpp
rhaas80: maybe `max_frames` should move into `utils` after all. Not a big deal for now though.

PR #1427, opened 2026-02-27
Correct indexing error in F-Engine file reader

eschnett:


rhaas80:
This is for a file `lib/stages/FEngine.old.cpp` which is all new according to git. Is this the correct change?

rhaas80 (approved):

PR #1428, opened 2026-02-28
New option `create_single_file` for HDF5 writer

eschnett:
This is convenient because it reduces the number of output files. It also simplifies reading results.

When set:
- write all frames into a single file, appending to an extensible dataset
- keep the file open until Kotekan shuts down
- use SWMR mode (and flush the file) to ensure the file is always consistent
- run `h5clear -s` to make half-written files readable


lib/stages/hdf5FileWrite.cpp
```diff
@@ -104,6 +103,51 @@ class hdf5FileWrite : public kotekan::Stage {
...
+        if (ierr) {
+            if (errno != EEXIST && errno != EISDIR) {
+                const char* const msg = strerror(errno);
+                FATAL_ERROR("Could not create directory \"{:s}\":\n{:s}", base_dir.c_str(), msg);
+            }
+        }
+
+        // Create HDF5 file
+        FileAccessProps fapl;
+        // We need the latest version for SWMR
```
rhaas80: is there a actual version number one can use instead? In case 2 months from now it no longer is the latest but "latest or the one before that"? 

Hmm, seems like HDF5 does not consider this something to offer. Right now the only versions defined are:

```
H5F_LIBVER_V200 = 5 , H5F_LIBVER_LATEST = 5
```

(in https://support.hdfgroup.org/documentation/hdf5/latest/_h5_fpublic_8h.html#a2d963b599894f684571fbd4d5e8a96a2a7e8e06822de9c57c31c7dd404520974d) so maybe `H5F_LIBVER_V200` is sufficient?

lib/stages/hdf5FileWrite.cpp
```diff
@@ -112,114 +156,172 @@ class hdf5FileWrite : public kotekan::Stage {
...
+            const auto extents = frame_desc->get_extents();
+            std::vector<std::size_t> initial_shape(extents.begin(), extents.end());
+            std::vector<std::size_t> maximum_shape(extents.begin(), extents.end());
+            initial_shape.at(0) = 0;
+            maximum_shape.at(0) = DataSpace::UNLIMITED;
+            const DataSpace space(initial_shape, maximum_shape);
+            {
+                // [[maybe_unused]] std::ptrdiff_t npoints = frame_desc->get_size();
+                // assert(std::ptrdiff_t(space.getElementCount()) == npoints);
+            }
```
rhaas80: remove

rhaas80 (approved):

lib/stages/hdf5FileWrite.cpp
eschnett: We're using the library version 1.10 at the moment on cx67, so I'll require that.

lib/stages/hdf5FileWrite.cpp
eschnett: Done.

PR #1433, opened 2026-03-04
changes supporting intel compilers

jbmertens:
Mostly build system changes to support intel's compilers.


cmake/Toolchain.cmake
```diff
@@ -237,7 +280,7 @@ check_cxx_compiler_flag(-fcx-limited-range HAVE_CX_LIMITED_RANGE)
 if(HAVE_CX_LIMITED_RANGE
    AND NOT ${IWYU}
    AND CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fcx-limited-range")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fcx-limited-range -ffast-math")
```
rhaas80: if this makes it slower... ?

lib/stages/EigenN2Iter.cpp
```diff
@@ -171,7 +171,7 @@ void EigenN2Iter::main_thread() {
         for (uint32_t i = 0; i < _num_eigenvectors; i++) {
             str_evals = fmt::format("{} {}", str_evals, evals[i]);
         }
-        DEBUG("Found eigenvalues: {:s}, with RMS residuals: {:e}, in {:.3f} s. Took {:d}/{:d} "
+        INFO("Found eigenvalues: {:s}, with RMS residuals: {:e}, in {:.3f} s. Took {:d}/{:d} "
```
rhaas80: should this be output during regular runs? It would be one output every 10s or so (or however often a the eigenvalues are being computed).

lib/stages/N2Subset.cpp
```diff
@@ -72,6 +72,9 @@ N2Subset::N2Subset(Config& config, const std::string& unique_name,
                     in_desc->get_num_ev(), out_desc->get_num_ev());
     }
 
+    _in_num_elements = in_desc->get_num_elements();
+    _out_num_elements = out_desc->get_num_elements();
+
```
rhaas80, line 77: this is more than code to support a different compiler, is it?

tests/test_n2_time_downsample.py
```diff
@@ -513,10 +514,10 @@ def test_contents(n2_data):
                 )
             )
         """
-        assert np.all(frame.vis == model_vis)
-        assert np.all(frame.evec == model_evec)
-        assert np.all(frame.eval == model_eval)
-        assert frame.erms == 1.0
+        assert np.allclose(frame.vis, model_vis, rtol=1e-6, atol=0)
```
geoffryan: I'm guessing changing from == to allclose is because of adding -ffast-math?  That's a pretty steep cost, being able to test equality is pretty valuable.

cmake/Toolchain.cmake
eschnett: We cannot not use `-ffast-math`. It enables optimizations that are too far away from the standard.

For example, fast-math assumes that there are no nans or infinities. The compiler then replaces calls to `isnan` or `isinf` by `false`. This makes even cursory consistency checks impossible.

jbmertens: Can definitely revert this, although that check may not be as useful in production. But I was seeing an odd behavior where adding this flag was making eigenvalue calculations even slower.

I added this to see if gcc could be made as fast as icx. The eigenvalue calculation compiled with clang was measured as ~2x faster than clang, and ~100x faster than gcc. I'm unsure why this is so discrepant.

lib/stages/EigenN2Iter.cpp
jbmertens: This should be reverted.

lib/stages/N2Subset.cpp
jbmertens: Yes, sneaking this fix in.

jbmertens: (A few of these seemingly-unrelated changes are just fixed compiler warnings.)

cmake/Features/FeatureDPDK.cmake
```diff
@@ -20,6 +20,15 @@ macro(_kotekan_detect_dpdk OUT_FOUND)
 
     pkg_check_modules(DPDK libdpdk>=19.11)
     if(DPDK_FOUND)
+        # Distro DPDK pkg-config hardcodes -march=corei7 as a portability baseline,
```
andrerenard, line 23: That's a strange choice actually now that I think of it, corei7 is really old now.  Is there harm in using the more optimal march settings from the main compiler?

tests/boost/test_N2Subset.cpp
```diff
@@ -148,8 +148,8 @@ BOOST_AUTO_TEST_CASE(test_fulluppertri_to_autocorrelations) {
     // Check vis values: should be diagonal elements (i, i)
     for (uint32_t i = 0; i < num_elements; ++i) {
         N2::cfloat expected{float(i), float(i)};
-        BOOST_CHECK_EQUAL(out_fv.vis[i].real(), expected.real());
-        BOOST_CHECK_EQUAL(out_fv.vis[i].imag(), expected.imag());
+        BOOST_CHECK_SMALL(out_fv.vis[i].real() - expected.real(), 1e-5f);
```
andrerenard: Maybe make `1e-5f` a constant/define?

cmake/Features/FeatureDPDK.cmake
jbmertens: Should be identical behavior.

It looks like gcc just uses the last one specified,
https://stackoverflow.com/questions/67097458/what-would-happen-if-i-specify-march-twice-in-cxxflags
while intel (and maybe clang) throw a warning or error.

andrerenard (approved):
I think this all looks good to me.   Although seeing that DPDK is built for `corei7`, I'm wondering now if we should manually build and install DPDK for the production nodes.  Less convenient than using the system packages, but has the potential to be more efficient.

jbmertens:
> I think this all looks good to me. Although seeing that DPDK is built for `corei7`, I'm wondering now if we should manually build and install DPDK for the production nodes. Less convenient than using the system packages, but has the potential to be more efficient.

Agreed - should be done on the sysops side though.
https://github.com/chord-observatory/chord-sysops/issues/6

tests/boost/test_N2Subset.cpp
jbmertens: Tolerances have been standardized some now.

PR #1437, opened 2026-03-10
Add DUT1 updatable config to chime_keep_gpu_warm.yaml

ketiltrout:
The top of this file points out that this is the same config as the `no_output` config except without the DPDK packet capture.

But that's not quite true anymore: a few differences other than the one I fix here have evolved over time.  I wonder if we should do a more thorough review of the differences at some point.


rikvl:
Closes https://github.com/chime-experiment/RTPipeline/issues/138 maybe?

andrerenard (approved):

ketiltrout:
> Closes https://github.com/chime-experiment/RTPipeline/issues/138 maybe?

Well, let's see...

PR #1438, opened 2026-03-11
SK kernel/wrapper unit tests

geoffryan:
Adds a CPU implementation of the SK kernel in `n2k`, unit tests for `lib/cuda/cudaRFISKbar.cpp` and `lib/cuda/cudaRFISKtilde.cpp` (which pass!), along with some helper stages.


lib/testing/n2k_sk_globals.hpp
```diff
@@ -0,0 +1,171 @@
+#ifndef KOTEKAN_TESTING_N2K_SK_GLOBALS
+#define KOTEKAN_TESTING_N2K_SK_GLOBALS
+
+/*
+ * Copied directly from external/n2k/src_lib/sk_globals.cu
+ * and external/n2k/include/n2k/internals/sk_globals.hpp
+ *
+ * These must be updated (re-copied) if the corresponding values
+ * are updated in N2K.
```
rhaas80, line 9: I guess there's no way to use an `#include` to avoid this copy&paste?

lib/testing/n2k_sk_globals.hpp
```diff
@@ -0,0 +1,171 @@
...
+    1.3902232458190005, 1.3271144885868689, 1.2660717040934295, 1.2072289354745493,
+    1.1504730149337898, 1.0955172211479256, 1.0419791330717345, 0.98945098875391424,
+    0.93755502085384212, 0.88598069388445311, 0.83450441662847263, 0.78299462265858932,
+    0.73140610926187644, 0.67976752202025059, 0.62816527585279858, 0.57672635752475354,
+    0.52560158203161289, 0.47495007561770608, 0.42492502403072885, 0.37565993578312867,
+    0.32725350941353021, 0.27974882722545447, 0.23309648937654681, 0.18707216705210078,
+    0.14104322735678684, 0.09303757829465914, 0.030149958054777281};
+} // namespace n2k_globals
+
+#endif
```
rhaas80, line 171: I assume these are just randomly chosen numbers?

lib/testing/testRFIS012Gen.cpp
```diff
@@ -0,0 +1,337 @@
...
+                                    "For s0 = {:d}, s1 = {:d}, min_s2 = {:d} > max_s2 = {:d}", s0,
+                                    s1, min_s2, max_s2);
+
+                            s2 = min_s2 + rng() % (max_s2 - min_s2 + 1);
+                        } else {
+                            // s0 is 0, so the others must be too.
+                            s1 = 0;
+                            s2 = 0;
+                        }
+                    }
```
rhaas80, line 313: I'd add an "else" to catch a user having a typo (or a developer adding a type and forgetting to add the required case to all if/elseif blocks).

rhaas80 (changes requested):
looks good. I'd add an "else" go avoid the not checking for incorrect user input case.

lib/testing/n2k_sk_globals.hpp
geoffryan: Nothing simple that doesn't bring in a lot of cruft.  Advantage of this being a separate file in `lib/testing` is that it is completely independent from `n2k`, can be run without cuda, etc.  I prefer the simplicity and the logical separation over the (potential) pain of having to copy this file again in the future if these are ever updated.

lib/testing/n2k_sk_globals.hpp
geoffryan: Nope, this is a table computed from the expected spectral kurtosis distribution.  Also built in to N2K.

lib/testing/testRFIS012Gen.cpp
geoffryan: Good catch.

geoffryan: Fixed in [258397a](https://github.com/kotekan/kotekan/pull/1438/commits/258397a7b208c1b7b7705a20ab0b0f91915db294)

rhaas80 (approved):

PR #1440, opened 2026-03-11
Send RFI metrics to external receiver

ljgray:
Reimplementation of functionality from the old `rfiBadInputFinder` and `rfiBroadcast` stages and the `hsaRfiBadInput` kernel. These items were removed in 42d3af4ef2311d48efc381e494dd147a1eda8a3f.

Full information about the old stages/kernel and the rfi receiver are in [this](https://bao.chimenet.ca/doc/documents/2390) document (for those with access). There are some subtle changes to the actual metrics being produced, but I don't expect that these will really affect the eventual feed flagging.

# Note
~I've temporarily recovered the old stages to review for comparison. However, `rfiBadInputFinder`, `rfiBroadcast_old`, `hsaRfiBadInput`, and `rfiRecord` are all deprecated and won't actually be merged.~

# Details
This stage sends the following metrics as UDP packets. Each packet contains:
- Header
- Frequency indices (on this node)
- Fraction of flagged samples per frequency
- Average SK value per frequency (from the high-cadence feed-averaged SK)
- Number of samples exceeding N sigma median-absolute-deviations (computed independently across frequencies) for each frequency and element

The stage actually sends a separate packet for each frequency in order to limit packet size.

The number of frames accumulated per packet is configurable. In the live CHIME system it's set to 10


ljgray:
This is still a draft because it isn't yet tested, but I think that all the expected work is done so additional eyes would be useful.

config/chime_upgrade_test_rfi_broadcast.j2
```diff
@@ -0,0 +1,2777 @@
...
+
+num_components: 2 # Re/Im
+num_freq_host_buf: 1
+{% set num_freq_gpu_buf = 16 %}
+num_freq_gpu_buf: {{ num_freq_gpu_buf }}
+num_ev_to_calculate: 4            # from n2 test case
+samples_per_data_set: 16384
+sample_period_ns: 2560
+network_frame_arrival_period: 16384 * sample_period_ns / 1e9
+baseband_buffer_depth: 3*272 # 3*272 = ~33 seconds after accounting for active frames
```
rhaas80: reducing this will use less (than 500GB) memory and also speed up startup time

lib/stages/rfiBadInputFinder.cpp
```diff
@@ -0,0 +1,244 @@
...
+    _frames_per_packet = json_request["frames_per_packet"].get<int>();
+    config.update_value(unique_name, "bi_frames_per_packet", _frames_per_packet);
+    // Send reply indicating success
+    conn.send_empty_reply(HTTP_RESPONSE::OK);
+    // Unlock mutex
+    rest_callback_mutex.unlock();
+}
+
+float rfiBadInputFinder::median(float array[], uint32_t num) {
+    float new_array[num];
```
rhaas80: variable size arrays are not strictly allowed in C++ (one is supposed to use `std::vector`. 

https://en.wikipedia.org/wiki/Variable-length_array#C.2B.2B

lib/stages/rfiBadInputFinder.cpp
```diff
@@ -0,0 +1,244 @@
...
+    for (uint32_t x = 0; x < num; x++) {
+        new_array[x] = array[x];
+    }
+    // sort array (super simply)
```
rhaas80: https://en.cppreference.com/cpp/algorithm/sort#Example

but I'm no CS major so really I don't mind :-)

lib/stages/rfiBadInputFinder.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+void rfiBadInputFinder::main_thread() {
+
+    // Intialize frame variables
+    N2::frameID frame_id(rfi_buf);
+    uint32_t frame_counter = 0;
+
+    // Initialize arrays
+    float rfi_data[_num_local_freq * _num_elements];
+    stream_t StreamIDs[total_links];
```
rhaas80: comment about variable length arrays in C++

lib/stages/rfiBadInputFinder.cpp
```diff
@@ -0,0 +1,244 @@
...
+                                   .num_total_freq = _num_total_freq,
+                                   .num_local_freq = _num_local_freq,
+                                   .frames_per_packet = _frames_per_packet,
+                                   .seq_num = 0,
+                                   .streamID = 0};
+
+    // Intialize empty packet
+    uint32_t packet_length =
+        sizeof(rfi_header) + _num_local_freq * sizeof(uint32_t) + sizeof(faulty_counter);
+    char* packet_buffer = (char*)malloc(packet_length);
```
rhaas80: this is all resurrected code I guess?

lib/stages/rfiBadInputFinder.cpp
```diff
@@ -0,0 +1,244 @@
...
+        // Adjust frame counter
+        frame_counter++;
+
+        // After 10 frames
+        rest_callback_mutex.lock();
+        if (frame_counter == _frames_per_packet) {
+
+            // Get current frequency bin
+            uint32_t current_freq_bin = tel.to_freq_id(StreamIDs[0]);
+            // TODO JSW: Handle num_local_freq > 1
```
rhaas80: that *is* the case for CHIME upgrade (and production CHIME as well). Is the comment still relevant?

lib/stages/rfiBadInputFinder.cpp
```diff
@@ -0,0 +1,244 @@
...
+                   _num_local_freq * sizeof(uint32_t));
+            // Add Data to packet
+            memcpy(packet_buffer + sizeof(rfi_header) + _num_local_freq * sizeof(uint32_t),
+                   faulty_counter, sizeof(faulty_counter));
+            // Send Packet
+            bytes_sent = sendto(socket_fd, packet_buffer, packet_length, 0,
+                                (struct sockaddr*)&saddr_remote, sizeof(sockaddr_in));
+            // Check if packet sent properly
+            if (bytes_sent != packet_length)
+                ERROR("SOMETHING WENT WRONG IN UDP TRANSMIT");
```
rhaas80: (not your code I know), but I would at least output `errno` or `strerror(errno)` in this case

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,330 @@
...
+    RFIPayload payload(num_local_freq, num_elements);
+
+    // Store information of the shape of the skbar array
+    const size_t _arr_size = num_local_freq * num_elements;
+    const size_t _arr_mid = num_elements / 2;
+    const bool _arr_size_is_even = num_elements % 2 == 0;
+    // Need an array to store the sum of skbar
+    // over time samples, since we don't explicitly assume
+    // that there is only one time sample per frame
+    float skbar_sum[num_local_freq * num_elements];
```
rhaas80: variable length array

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+        // Release the frame since we've copied out to skbar_sum
+        sk_bar_buf->mark_frame_empty(unique_name, skbar_frame_id);
+
+        // Iterate over frequencies since we need to compute the statistics
+        // for each frequency independently
+        for (size_t f = 0; f < _arr_size; f += num_elements) {
+            // Compute the median of skbar_sum by sorting in place
+            memcpy(_median_tmp, skbar_sum + f, _sizeof_mediantmp);
+            std::sort(_median_tmp, _median_tmp + num_elements);
```
rhaas80: :-)

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+            for (const auto& packet : packets) {
+                // Send and log if something went wrong
+                ssize_t bytes_sent = sendto(socket_handler, packet.data(), packet.size(), 0,
+                                            (struct sockaddr*)&dest_addr, sizeof(sockaddr_in));
+
+                if (bytes_sent != (ssize_t)packet.size()) {
+                    DEBUG("Error sending UDP packet. Expected to send {:d} bytes, actually sent "
+                          "{:d}.",
+                          packet.size(), bytes_sent);
```
rhaas80: I'd include `strerror(errno)`.

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,330 @@
...
+                } else {
+                    DEBUG("Sent UDP packet with {:d} bytes", bytes_sent);
+                }
+            }
+
+            // Reset payload members used in accumulation to zero
+            std::fill(payload.frac_flagged.begin(), payload.frac_flagged.end(), 0.0f);
+            std::fill(payload.sktilde_avg.begin(), payload.sktilde_avg.end(), 0.0f);
+            std::fill(payload.bad_feed_counts.begin(), payload.bad_feed_counts.end(), 0u);
```
rhaas80, line 348: `std::assign` may (but I do not know) be faster in this case.

ljgray:
This seems to be working and the metrics it produces seem to be fairly reasonable, so marking as ready for review

rhaas80:
Fails a CI test.

config/chime_upgrade_n2.j2
```diff
@@ -701,6 +709,45 @@ run_rfi_SKbar:
...
+    kotekan_stage: rfiBroadcast
+    rfi_sktilde_buf: host_rfi_SKtilde_buffer
+    rfi_skbar_buf: host_rfi_SKbar_buffer
+    rfi_mask_buf: host_rfi_RFImask_buffer
+    num_local_freq: num_freq_gpu_buf
+    num_global_freq: 1024
+    num_sigma_deviations: 3
+    frames_per_packet: 10
+    destination_port: "3000"
+    destination_ip: "10.5.60.1"
```
rhaas80, line 807: is this the life system IP or a dummy for testing?

ljgray:
> Fails a CI test.

Ah - just a lint failure on one of the old stages (which won't get merged in).

I can remove all the `donotmerge` commits is it's easier for review - just figured they could act as a reference

rhaas80:
nah, I can review. Just need to do it now so that I do not forget again.

config/chime_upgrade_n2.j2
```diff
@@ -518,6 +518,14 @@ host_rfi_S012tilde_ringbuffer:
...
     metadata_pool: main_pool
 
+host_rfi_SKtilde_buffer:
+    kotekan_buffer: standard
+    num_frames: gpu_buffer_depth
+    frame_size: sizeof_float32 * 3 * num_freq_gpu_buf * rfi_num_times
+    type: float32
+    dimnames: ["Trfi", "F", "S"]
+    shape: [rfi_num_times, num_freq_gpu_buf, 3]
+    metadata_pool: main_pool
```
rhaas80, line 542: note: while nice ot have the "type", "dimnames" and "shape" attributes are no used by kotekan (yet! There is the future plan of having the buffers know their shape instead of burying it deep in C++ code).

config/chime_upgrade_n2.j2
```diff
@@ -701,6 +709,45 @@ run_rfi_SKbar:
...
+          ring_buffer_size: gpu_buffer_depth * output_size
+          out_buf: host_rfi_SKtilde_buffer
+
+rfi_broadcast:
+    kotekan_stage: rfiBroadcast
+    rfi_sktilde_buf: host_rfi_SKtilde_buffer
+    rfi_skbar_buf: host_rfi_SKbar_buffer
+    rfi_mask_buf: host_rfi_RFImask_buffer
+    num_local_freq: num_freq_gpu_buf
+    num_global_freq: 1024
```
rhaas80: might be best to have a variable at global level or a Jinja variable for this 1024 (similar to say `num_elements` or whatnot) most so that one can distinguish between different 1024 that may show up.

config/chime_upgrade_n2.j2
```diff
@@ -701,6 +709,45 @@ run_rfi_SKbar:
...
+    num_sigma_deviations: 3
```
rhaas80, line 799: I'd add a comment at least for now stating that (if) this matches the life CHIME system's value. I assume this is a number that was found to empirically work without any deeper theory behind it.

config/chime_upgrade_test_rfi_broadcast.j2
```diff
@@ -0,0 +1,2777 @@
...
+#
+# Subset of chime_upgrade_n2.j2
+#
+# Pipeline for running the n2k correlator, accumulate onto a visibility matrix
+# and compute eigen-decomposition.
+#
+# Currently only uses one NUMA domain, but it shouldn't be hard to extend to
+# the second NUMA domain.
+#
+# Author: Andre Renard, Roland Haas, Erik Schnetter
```
rhaas80: you probably should add yourself to share in the ~~blame~~ praise

config/chime_upgrade_test_rfi_broadcast.j2
```diff
@@ -0,0 +1,2777 @@
...
+baseband_buffer_depth: 3*272 # 3*272 = ~33 seconds after accounting for active frames
+# see where / if / how to use host_buffer_depth and gpu_buffer_depth
+# must be at least size of gpu_buffer_depth
+host_buffer_depth: 12
+# this must be 4 since cudaTranspose2048_chime.cpp hard-codes the outermost
+# size of the voltage ring-buffer as 4
+gpu_buffer_depth: 4
+num_gpus: 2
+# 14,15,16 for blaze in EigenN2Iter
+# 17,18,19 for N2Accumulate
```
rhaas80: are these still needed?

config/chime_upgrade_test_rfi_broadcast.j2
```diff
@@ -0,0 +1,2777 @@
...
+updatable_config:
+  enable_vis_out:
+    kotekan_update_endpoint: "json"
+    enabled: false
+  enable_failed_out:
+    kotekan_update_endpoint: "json"
+    enabled: false
+  enable_sk_out:
+    kotekan_update_endpoint: "json"
+    enabled: false
```
rhaas80: these 3 are no not needed, are they?

config/chime_upgrade_test_rfi_broadcast.j2
```diff
@@ -0,0 +1,2777 @@
...
+  use_dataset_broker: False
+
+main_pool:
+  kotekan_metadata_pool: chordMetadata
+  num_metadata_objects: 30 * host_buffer_depth + 5 * baseband_buffer_depth
+
+# Baseband metadata pool
+baseband_pool:
+  kotekan_metadata_pool: BasebandMetadata
+  num_metadata_objects: 30 * host_buffer_depth * 32
```
rhaas80: not needed

config/chime_upgrade_test_rfi_broadcast.j2
```diff
@@ -0,0 +1,2777 @@
...
+    kotekan_stage: rfiBroadcast
+    rfi_sktilde_buf: host_rfi_SKtilde_buffer
+    rfi_skbar_buf: host_rfi_SKbar_buffer
+    rfi_mask_buf: host_rfi_RFImask_buffer
+    num_local_freq: num_freq_gpu_buf
+    num_global_freq: 1024
+    num_sigma_deviations: 3
+    frames_per_packet: 10
+    destination_port: "3000"
+    destination_ip: "10.5.60.1"
```
rhaas80: if this is to e kept I would strip out most of the stages and only keep the needed ones.

lib/stages/rfiBadInputFinder.cpp
```diff
@@ -0,0 +1,244 @@
+#include "rfiBadInputFinder.hpp"
```
rhaas80: is this file compiled? I do not see it in the changes to `CMakeLists.txt`

lib/stages/rfiBadInputFinder.cpp
```diff
@@ -0,0 +1,244 @@
...
+    // Register stage as consumer
+    rfi_buf->register_consumer(unique_name);
+
+    // Intialize internal config
+    _num_local_freq = config.get<uint32_t>(unique_name, "num_local_freq");
+    _num_total_freq = config.get_default<uint32_t>(unique_name, "num_total_freq", 1024);
+    _num_elements = config.get<uint32_t>(unique_name, "num_elements");
+    _samples_per_data_set = config.get<uint32_t>(unique_name, "samples_per_data_set");
+    // Rfi paramters
+    _sk_step = config.get_default<uint32_t>(unique_name, "sk_step", 256);
```
rhaas80: I believe there's a "new" variable that has taken over what `sk_step` did. Might be best to remove `sk_step` (or at least remove its default) so that there are no surprises if / when someone changes the new variable and this stage does not react to the change

lib/stages/rfiBadInputFinder.cpp
```diff
@@ -0,0 +1,244 @@
...
+    _rfi_combined = config.get_default<bool>(unique_name, "rfi_combined", true);
+    _frames_per_packet = config.get_default<uint32_t>(unique_name, "bi_frames_per_packet", 10);
```
rhaas80: `10` is what live CHIME uses?

lib/stages/rfiBadInputFinder.cpp
```diff
@@ -0,0 +1,244 @@
...
+    // Stage-specific paramters
+    total_links = config.get_default<uint32_t>(unique_name, "total_links", 1);
+    dest_port = config.get<uint32_t>(unique_name, "destination_port");
+    dest_server_ip = config.get<std::string>(unique_name, "destination_ip");
+    dest_protocol = config.get_default<std::string>(unique_name, "destination_protocol", "UDP");
+
+    // Set stats variables
+    stats_sigma = 3;
```
rhaas80: a `3` showed up in the input file before, this is that `3`, yes?

lib/stages/rfiBadInputFinder.cpp
```diff
@@ -0,0 +1,244 @@
...
+    _samples_per_data_set = config.get<uint32_t>(unique_name, "samples_per_data_set");
+    // Rfi paramters
+    _sk_step = config.get_default<uint32_t>(unique_name, "sk_step", 256);
+    _rfi_combined = config.get_default<bool>(unique_name, "rfi_combined", true);
+    _frames_per_packet = config.get_default<uint32_t>(unique_name, "bi_frames_per_packet", 10);
+    // Stage-specific paramters
+    total_links = config.get_default<uint32_t>(unique_name, "total_links", 1);
+    dest_port = config.get<uint32_t>(unique_name, "destination_port");
+    dest_server_ip = config.get<std::string>(unique_name, "destination_ip");
+    dest_protocol = config.get_default<std::string>(unique_name, "destination_protocol", "UDP");
```
rhaas80: `dest_protocol` is ignored, is it?

lib/stages/rfiBadInputFinder.hpp
```diff
@@ -0,0 +1,123 @@
...
+    /// The address for UDP stream to be sent to
+    std::string dest_server_ip;
+    /// The streaming protocol, only UDP is supported
+    std::string dest_protocol;
+    /// Internal socket error holder
+    int socket_fd;
+    /// Rest server callback mutex
+    std::mutex rest_callback_mutex;
+    /// String to hold endpoint
+    std::string endpoint;
```
rhaas80: I'd remove the unused ones

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+using kotekan::Config;
+using kotekan::Stage;
+
+REGISTER_KOTEKAN_STAGE(rfiBroadcast);
+
+#define BITS_PER_BYTE 8
+#define RFI_BUF_S 3
+
+namespace {
+/// Helper to compute array median. Modifies `arr` in-place.
```
rhaas80, line 32: why does a `median` function modify its input array? Does remove the median from the values?

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+float median(float* arr, int n) {
+    auto mid = n / 2;
+
+    if (mid == 0) {
+        return 0.0f;
+    }
+
+    std::nth_element(arr, arr + mid, arr + n);
```
rhaas80: ah, that's why the modification

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+    // Standard config params defining the expected buffer shapes
+    num_elements = config.get<size_t>(unique_name, "num_elements");
+    num_local_freq = config.get<size_t>(unique_name, "num_local_freq");
+    num_global_freq = config.get<size_t>(unique_name, "num_global_freq");
+    samples_per_data_set = config.get<size_t>(unique_name, "samples_per_data_set");
+    rfi_downsampling_factor = config.get<size_t>(unique_name, "rfi_downsampling_factor");
+    rfi_second_downsampling_factor =
+        config.get<size_t>(unique_name, "rfi_second_downsampling_factor");
+    // Stage-specific params
+    ndev_sigma = config.get<uint16_t>(unique_name, "num_sigma_deviations");
```
rhaas80: I'd try and use the same C++ variable name as the name of the yaml element, so that it is easier for future readers to follow the flow information

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+    // Packet/network config params
+    frames_per_packet = config.get_default<size_t>(unique_name, "frames_per_packet", 1);
+    dest_port = config.get<size_t>(unique_name, "destination_port");
+    dest_ip = config.get<std::string>(unique_name, "destination_ip");
+
+    // Compute the derived number of time samples for sktilde and skbar, respectively
+    _downsampled_samples_per_data_set = samples_per_data_set / rfi_downsampling_factor;
+    _second_downsampled_samples_per_data_set =
+        _downsampled_samples_per_data_set / rfi_second_downsampling_factor;
```
rhaas80: I'd add asserts that these divisions are without remainder (which I assume they must be for things to work as expected)

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+    // Check that input buffers have the expected frame size
+    // per the provided config parameters
+    // rfi_mask is type uint1x8, so uint8 frame size is 8x smaller than
+    // the number of samples
+    size_t _rfi_mask_expected_frame_size = num_local_freq * samples_per_data_set / BITS_PER_BYTE;
```
rhaas80, line 108: let's assume samples_per_data_set is actually multiple of 8 :-)

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+    header.num_elements = static_cast<uint32_t>(num_elements);
+    header.samples_per_data_set = static_cast<uint32_t>(samples_per_data_set);
+    header.num_total_freq = static_cast<uint32_t>(num_global_freq);
+    header.num_local_freq = static_cast<uint32_t>(num_local_freq);
+    header.frames_per_packet = static_cast<uint32_t>(frames_per_packet);
+    // These get set by the first frame per packet
+    header.seq_num = 0;
+
+    // Create the socket
+    int socket_handler = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
```
rhaas80, line 177: protocol type stage option is ignored

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+    if (socket_handler == -1) {
+        FATAL_ERROR("Could not create UDP socket for output stream.");
```
rhaas80: output `errno` and / or `strerror(errno)` ? Though I suspect errors to be very rare here.

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+    };
+    // Otherwise, connection is successful
+    DEBUG("UDP Connection establish: {:d} {:s}", dest_port, dest_ip);
+    // Directly track frames since `buffer_depth` for the incoming
+    // buffers could be less than `frames_per_packet`
+    uint16_t _packet_num_frame_counter = 0;
```
rhaas80: are 16 bit enough? `frames_per_packet` is a `uint32_t` so someone could set it to a (unrealistically large) value of say 100000

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+            break;
+
+        std::shared_ptr<chordMetadata> meta = get_chord_metadata(rfi_mask_buf, rfi_mask_frame_id);
+
+        // Set additional header metadata, just once per packet
+        if (rfi_mask_frame_id % frames_per_packet == 0) {
+            header.seq_num = meta->get_fpga_seq_num();
+
+            for (size_t f = 0; f < num_local_freq; ++f) {
+                payload.freq_ids[f] = (uint32_t)meta->get_coarse_freq()[f];
```
rhaas80: I'd use `.at(f)` (and maybe ensure that here's only one call to `get_coarse_freq` per frame since it allocates a `std::vector`) to make sure there's a nice error message if `num_local_freq` is set to a larger value than `coarse_freq.size()`

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+            }
+        };
+
+        // Sum the rfi mask for each freq. and release, keeping in mind
+        // that the rfi_mask has type uint1x8 and split time axis
+        size_t _rfi_t_hi = meta->dim[meta->dims - 1];
```
rhaas80: these fields are expected to go away in favor of the corresponding ones in the ndarray frame description (accessible via `buffer->get_ndarray_frame_desc()` or so [no frame index needed]).

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+        size_t _rfi_step = num_local_freq * _rfi_t_hi;
+
+        // Iterate outer time samples
+        // TODO: at this cadence, the RFI mask also contains information about samples
+        // lost at the network layer, so it isn't a perfect representation of the true
+        // flagging rate. Once that metric exists (produced elsewhere), we'll just
+        // change this step to grab the relevant buffer and pull the `n_rfi` value
+        // from it directly
```
rhaas80: ok

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+            (uint8_t*)sk_tilde_buf->wait_for_full_frame(unique_name, sktilde_frame_id);
+        if (sk_tilde_frame == nullptr)
+            break;
+
+        // Sum SK tilde over time for each frequency (second dimension)
+        for (size_t i = 0; i < num_local_freq; ++i) {
+            // Sum over time samples, which is the first dimension
+            for (size_t j = RFI_BUF_S * i; j < sk_tilde_buf->frame_size;
+                 j += RFI_BUF_S * num_local_freq) {
+                payload.sktilde_avg[i] += (float)sk_tilde_frame[j];
```
rhaas80: `i` and `j` are mostly fine, though `f` and `s` (???) maybe closer to what is used otherwise for frequency index (`s` is unique to this stage though I think)

rhaas80: you may want to have a local temp variable to accumulate into, just in case the compiler cannot guarantee that `payload.sktilde_avg` and `payload.sktilde_avg` cannot point to the same memory location (in which case it cannot eg keep the accumualted value in a register and has to push it to memory [well L1 cache] after each addition)

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+            (uint8_t*)sk_bar_buf->wait_for_full_frame(unique_name, skbar_frame_id);
+        if (sk_bar_frame == nullptr)
+            break;
+
+        // Accumulate in time. Outer index is time samples
+        for (size_t i = 0; i < sk_bar_buf->frame_size; i += RFI_BUF_S * _arr_size) {
+            // Index each frequency and element
+            for (size_t j = 0; j < _arr_size; ++j) {
+                // Skip forward by 3 to grab just the SK value
+                skbar_sum[j] += sk_bar_frame[i + RFI_BUF_S * j];
```
rhaas80: same comment about temp accumulator variable as above. no comment about `i`, `j` since they are just indexes into the raw array without much meaning

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+            }
+        }
+
+        // Release the frame since we've copied out to skbar_sum
+        sk_bar_buf->mark_frame_empty(unique_name, skbar_frame_id);
+
+        // Iterate over frequencies since we need to compute the statistics
+        // for each frequency independently
+        for (size_t f = 0; f < _arr_size; f += num_elements) {
```
rhaas80, line 272: not really counting frequencies in `f` here

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+            // Compute the median of skbar_sum. `median` modifies in-place
+            // so copy to a temp array
+            std::copy_n(skbar_sum + f, num_elements, _median_tmp);
+            float med = median(_median_tmp, num_elements);
```
rhaas80: awkward, but I guess save allocating and freeing temp array in `median`.

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+            // Compute the MAD, re-using `_median_tmp`
```
rhaas80: what's `MAD`?

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+            for (size_t i = 0; i < num_elements; ++i) {
+                // subtract the median in-place since this will be used twice
+                skbar_sum[f + i] = std::abs(skbar_sum[f + i] - med);
+                // copy into the temp array to compute the second median
+                _median_tmp[i] = skbar_sum[f + i];
+            }
+            float mad_thresh = ndev_sigma * 1.4826 * median(_median_tmp, num_elements);
```
rhaas80: 1.4826 is a magic number here (what is it? It's not recognizable like say 3.1415927)

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+        // Reset skbar_sum, since it's used in every iteration
+        std::fill_n(skbar_sum, _arr_size, 0.0f);
+
+        // Increment and wrap to number of frames per packet
+        _packet_num_frame_counter = (_packet_num_frame_counter + 1) % frames_per_packet;
+
+        // Send a packet every `frames_per_packet` frames
+        if (_packet_num_frame_counter == 0) {
+            // Reset the counter
+            _packet_num_frame_counter = 0;
```
rhaas80: this sets `_packet_num_frame_counter` to `0`, but the code has just checked that is (already) `0`. Should this be a different variable that is being reset / checked?

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+            // Normalize by the number of time samples in the packet.
+            // Cast the first operand to a float to get rid of a warning about
+            // loss of precision, even though we require that these values are
```
rhaas80: loss of precision is likely due to differently sized integer types, not int vs. float (the division being without remainder)

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,322 @@
...
+            // evenly divisible
+            float divisor = static_cast<float>(frames_per_packet) * samples_per_data_set;
+            float second_divisor = divisor / rfi_downsampling_factor;
+
+            // Normalize the average SK by the number of time samples, accounting
+            // for the fact that the input was already downsampled. Also, convert
+            // the number of flagged samples to a fraction.
+            for (size_t i = 0; i < num_local_freq; ++i) {
```
rhaas80: `f`

lib/stages/rfiBroadcast.hpp
```diff
@@ -0,0 +1,123 @@
...
+ *    - Number of time samples exceeding `ndev_sigma` standard deviations
+ *      from the median-subtracted per-feed SK `sk_bar_buf`.
+ *
+ * All of these inputs are accumulated for `frames_per_packet` frames and packed
+ * into a UDP packet with format:
+ *    - header: struct `RFIHeader`
+ *    - freq_ids: uint32 [num_local_freq]
+ *    - num_flagged: float [num_local_freq]
+ *    - sktilde_avg: float [num_local_freq]
+ *    - bad_feed_count: uint8 [num_local_freq * num_elements]
```
rhaas80, line 40: really `unit8`? `num_elements` is 2048, `num_local_freq` is 16 which larger than what a byte can hold

lib/stages/rfiBroadcast_old.cpp
```diff
@@ -0,0 +1,290 @@
+#include "rfiBroadcast.hpp"
```
rhaas80: Is this `_old` to be reviewed?

lib/stages/rfiRecord.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+    // Initialize variables
+    N2::frameID frame_id(rfi_buf);
+    int64_t fpga_seq_num;
+
+    auto& tel = Telescope::instance();
+
+    int fd = -1;
+    // File name
+    char file_name[100];
```
rhaas80: *probably* ok, but why so stingy and not use `PATH_MAX`?  https://pubs.opengroup.org/onlinepubs/009695399/basedefs/limits.h.html#tag_13_24_03_02

lib/stages/rfiRecord.cpp
```diff
@@ -0,0 +1,161 @@
...
+        if (write_flag) {
+
+            fpga_seq_num = meta->get_fpga_seq_num();
+
+            // Only write a new file every _frames_per_file frames
+            if (fpga_seq_num % (_samples_per_data_set * _frames_per_file) == 0) {
+                // Close the old file if it's open
+                if (fd != -1) {
+                    INFO("Closing last file");
+                    close(fd);
```
rhaas80: `close()` can return an error if data is being flushed

lib/stages/rfiRecord.cpp
```diff
@@ -0,0 +1,161 @@
...
+                struct tm timeinfo;
+
+                if (gmtime_r(&gps_time.tv_sec, &timeinfo) == nullptr) {
+                    ERROR("Cannot gerate time info");
+                }
+
+                strftime(data_time, sizeof(data_time), "%Y%m%dT%H%M%S", &timeinfo);
+                // TODO: should the file name include all frequencies?
+                snprintf(file_name, sizeof(file_name), "%s/%sN%09ld_%d.rfi", write_dir.c_str(),
+                         data_time, gps_time.tv_nsec, meta->get_coarse_freq()[0]);
```
rhaas80: should check for errors (buffer too small)

lib/stages/rfiRecord.cpp
```diff
@@ -0,0 +1,161 @@
...
+                fd = open(file_name, O_WRONLY | O_APPEND | O_CREAT, 0666);
+                if (fd < 0) {
+                    ERROR("Cannot open file {:s}, error {:d} ({:s})", file_name, errno,
+                          strerror(errno));
+                }
+
+                // Write seq number into the start of the file
+                ssize_t bytes_writen = write(fd, &fpga_seq_num, sizeof(int64_t));
+                if (bytes_writen != sizeof(int64_t)) {
+                    ERROR("Failed to write seq_num to disk");
```
rhaas80: output `errno` and `sterror(errno)`

lib/stages/rfiRecord.cpp
```diff
@@ -0,0 +1,161 @@
...
+                }
+            }
+
+            // Write the current frame into the file
+            ssize_t bytes_writen = write(fd, frame, rfi_buf->frame_size);
+            if (bytes_writen != static_cast<ssize_t>(rfi_buf->frame_size)) {
+                ERROR("Failed to write buffer to disk");
```
rhaas80: `errno`

lib/stages/rfiRecord.cpp
```diff
@@ -0,0 +1,161 @@
...
+            }
+        } else {
+            // Reset the file tracker and close any file that might be open
+            if (fd != -1) {
+                INFO("Closing open file...");
+                close(fd);
```
rhaas80: `close()` can return an error

lib/utils/rfi_functions.hpp
```diff
@@ -0,0 +1,105 @@
...
+ */
+struct RFIPayload {
+    /// List of local frequencies. Has length `num_local_freq`.
+    std::vector<uint32_t> freq_ids;
+    /// Fraction of flagged samples per freq. Has length `num_local_freq`.
+    std::vector<float> frac_flagged;
+    /// Average SK per frequency. Has length `num_local_freq`.
+    std::vector<float> sktilde_avg;
+    /// Count of bad feeds per freq and element. Has length
+    // `num_local_freq` * `num_elements`
```
rhaas80: missing `/`, at least of the doygen comment is expected to contain the 2nd line

lib/utils/rfi_functions.hpp
```diff
@@ -0,0 +1,105 @@
...
+    std::vector<uint8_t> bad_feed_counts;
```
rhaas80, line 54: how are the `feeds` it is counting different from the `elements` it is being indexed by?

rhaas80 (changes requested):
I think mostly the long j2 input files should be reduced (but maybe those are donotpush)

lib/stages/rfiBroadcast.hpp
ljgray: just a misnomer - this is a per-feed, per-freq count of the number of sample which exceed a flagging threshold (3-sigma MAD in this case). So this value can only increment by 1 per frame, so unless we are accumulating 2^16 frames this shouldn't be able to overflow

lib/stages/rfiBroadcast.cpp
ljgray: Open to suggestions on alternatives, but I figured that it's probably faster to do the single allocation + copy-per-loop instead of a new allocation each time

lib/stages/rfiBroadcast.cpp
ljgray: median-absolute-deviations - I'll add a note expanding on this

lib/stages/rfiBroadcast.cpp
ljgray: Good catch - I think this is just a residual line that I forgot to remove

lib/stages/rfiBroadcast.cpp
ljgray: Yeah these should be the same type

config/chime_upgrade_n2.j2
ljgray: This is the mock receiver IP - can make a note that it'll have to be changed eventually

lib/stages/rfiBroadcast.cpp
rhaas80: I have no alternative. Let's leave it like this. It works after all.

config/chime_upgrade_n2.j2
ljgray: Is there a preference, then, to leave these in the config with the assumption that they'll be used eventually, or to remove them entirely?

rhaas80: oh I'd leave them in. Sure they could get out of sync with what the code is actually doing, but they are also a handy reminder about "wouldn't this be nice" functionality :-)

lib/stages/rfiBroadcast.cpp
ljgray: Yeah - this is just a clangd warning (Result of integer division used in a floating point context; possible loss of precision).

lib/stages/rfiBroadcast_old.cpp
ljgray: Nope - removed

ljgray:
- Removed all `donotmerge` commits and resolved corresponding review comments
- Addressed all remaining review comments
- Added some fixes to `config/chime_upgrade_n2.j2` according to updates in #1500

lib/stages/rfiBroadcast.cpp
```diff
@@ -0,0 +1,355 @@
...
+
+    // Compute the derived number of time samples for sktilde and skbar, respectively
+    if (samples_per_data_set % rfi_downsampling_factor != 0) {
+        FATAL_ERROR("`rfi_downsampling_factor` does not evenly divide the number of time samples: "
+                    "{:d} / {:d}",
+                    rfi_downsampling_factor, samples_per_data_set);
+    } else {
+        _downsampled_samples_per_data_set = samples_per_data_set / rfi_downsampling_factor;
+    }
+    if (rfi_second_downsampling_factor % _downsampled_samples_per_data_set != 0) {
```
rhaas80: is this the correct ordering? I'd have expected 

`_downsampled_samples_per_data_set % rfi_second_downsampling_factor`

that is number-of-samples-after-first-downsampling divided-by second-downsampling-factor

rhaas80 (approved):

lib/stages/rfiBroadcast.cpp
ljgray: Fixed

PR #1441, opened 2026-03-12
Minor tweaks to `bufferBadInputs`

ljgray:
This was originally made under the assumption that we would no longer need to reorder the bad input list before passing it to the SK kernels. That no longer seems to be the case, so now this PR just makes a couple of minor tweaks without changing any functionality.

### DIFF
--- a/lib/stages/bufferBadInputs.cpp
+++ b/lib/stages/bufferBadInputs.cpp
@@ -1,5 +1,3 @@
-#define DONTUSECHIMEMETA 1
-
 #include "bufferBadInputs.hpp"
 
 #include "Config.hpp"         // for Config
@@ -8,15 +6,10 @@
 #include "chordMetadata.hpp"  // for get_chord_metadata, chordMetadata
 #include "configUpdater.hpp"  // for configUpdater
 #include "kotekanLogging.hpp" // for DEBUG, ERROR
-#include "visUtil.hpp"        // for parse_reorder_default
 
-#include "fmt.hpp" // for compile_string_to_view
-
-#include <algorithm>  // for copy, max
 #include <exception>  // for exception
 #include <functional> // for bind, function, _1
 #include <memory>     // for __shared_ptr_access, shared_ptr
-#include <tuple>      // for get
 
 using kotekan::bufferContainer;
 using kotekan::Config;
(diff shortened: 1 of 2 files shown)


ljgray:
Not yet tested so still a draft. I guess I can make a unit test for this, but it isn't exactly complicated...

ljgray:
@rhaas80 Given that a reorder is still required, this is ready to review/test.

lib/stages/bufferBadInputs.cpp
```diff
@@ -1,22 +1,17 @@
-#define DONTUSECHIMEMETA 1
-
 #include "bufferBadInputs.hpp"
 
 #include "Config.hpp"         // for Config
+#include "N2Util.hpp"         // for frameID
```
rhaas80: isn't there a `frameID` already in `visUtil.hpp`?

lib/stages/bufferBadInputs.cpp
```diff
@@ -47,49 +41,41 @@ bool bufferBadInputs::update_bad_inputs_callback(nlohmann::json& json) {
...
-    }
-
-    // Add current bad input mask
-    for (auto element : bad_inputs_correlator) {
-        if (element < (int)input_mask_len && element >= 0) {
-            host_mask[element] = 0;
+    // Add current bad input to the mask
+    for (int element : bad_inputs) {
+        if (element < (int)num_elements && element >= 0) {
+            host_mask[input_remap[element]] = 0;
```
rhaas80: hmm, good point. I wonder what we actually need in the end. Namely the data will go into beamformer order  (or cylinder order) though this here will make the host_mask in correlator order (at least if `input_remap` is what I think it is namely lists the adc_id for each channel id in channel id order).

lib/stages/bufferBadInputs.cpp
```diff
@@ -47,49 +41,41 @@ bool bufferBadInputs::update_bad_inputs_callback(nlohmann::json& json) {
...
     }
 
     // Create new metadata
-    out_buf->allocate_new_metadata_object(out_buffer_ID);
+    out_buf->allocate_new_metadata_object(frame_id);
 
     // Set no. of bad inputs in the metadata
-    get_chord_metadata(out_buf, out_buffer_ID)
-        ->set_rfi_num_bad_inputs(bad_inputs_correlator.size());
+    get_chord_metadata(out_buf, frame_id)->set_rfi_num_bad_inputs(bad_inputs.size());
```
rhaas80, line 65: looks good to me. I am unsure if anything will correctly propagate this number out of the bad feeds mask and into the voltage buffers or products computed from those

lib/stages/bufferBadInputs.hpp
```diff
@@ -55,19 +56,22 @@ class bufferBadInputs : public kotekan::Stage {
...
-    std::vector<int> bad_inputs_cylinder;
+    /// List of current bad inputs in received order, which
+    // is expected to be cylinder order
+    std::vector<int> bad_inputs;
 
-    /// List of current bad inputs in correlator order.
-    std::vector<int> bad_inputs_correlator;
+    /// List of current bad inputs in output order,
+    /// based on the input map
+    std::vector<int> bad_inputs_remap;
```
rhaas80: this variable is not used, is it?

lib/stages/bufferBadInputs.cpp
```diff
@@ -29,16 +24,15 @@ bufferBadInputs::bufferBadInputs(Config& config_, const std::string& unique_name
...
+    num_elements = config.get<size_t>(unique_name, "num_elements");
     auto input_reorder = parse_reorder_default(config, unique_name);
+
     input_remap = std::get<0>(input_reorder);
 
     out_buf = get_buffer("out_buf");
     out_buf->register_producer(unique_name);
 
-    out_buffer_ID = 0;
+    N2::frameID frame_id(out_buf);
```
rhaas80: this seems wrong. it declares a local variable `frame_id` instead of setting the class member `frame_id` (which should be of type `frameID` instead of `uint32_t`).

lib/stages/bufferBadInputs.hpp
```diff
@@ -55,19 +56,22 @@ class bufferBadInputs : public kotekan::Stage {
...
-    /// The mapping from correlator to cylinder element indexing.
+    // Mapping between orders
     std::vector<uint32_t> input_remap;
 
-    uint32_t out_buffer_ID = 0;
+    /// The size of the bad input mask.
+    size_t num_elements;
+
+    // Need to use the frame_id outside of the main thread
+    uint32_t frame_id;
```
rhaas80: should be `frameID` I think.

rhaas80 (changes requested):

lib/stages/bufferBadInputs.cpp
ljgray: Not as far as I can tell - seems to be in N2Util now

lib/stages/bufferBadInputs.cpp
ljgray: Yeah, I haven't checked if this actually gets used, but I assume the SK kernels get this information somehow

lib/stages/bufferBadInputs.cpp
ljgray: Ah, yeah, I guess this isn't the reorder that we need (or, at least, the one loaded from `parse_reorder_default`), which seems to be a big list of (adc_id, chan_id, correlator_input). Does the mapping to beamformer order live anywhere currently?

rhaas80: no, the only map that is listed in the input files is the one that has adc_id and channel_id columns with channel_id being contiguous (ie the index). So the mapping from channel_id (aka cylinder order) to beamformer order which is computable (since the relation between channel id and feed location is known, except of course for the feed that have been hijacked for non-sky data) this had always been computed. My feeling is that for the visibility data that is being written out, even though everybody should consult the input-reorder metadata stored along with it, I want to store it in cylinder order (and also write metadata that claims this order). For the same reason I suspect I will not want to actually change the `input_reorder` yaml file entry (since it is read by the baseband archiver script).

lib/stages/bufferBadInputs.cpp
rhaas80: might be copy. I find (git hash 55389fd27) in `visUtil.hpp`

```
 924 /**
 925  * @brief Class to hold buffer indices.
 926  **/
 927 class frameID : public modulo<int> {
 928 public:
 929     /**
 930      * @brief Create a frameID for a given buffer.
 931      *
 932      * @param buf   Buffer to use.
 933      **/
 934     frameID(const Buffer* buf) : modulo<int>(buf->num_frames) {}
 935 };
 ```
 
 Both seem to be the wrong place to put this to me :-) Should be in `buffer.hpp` :-)

File: lib/stages/bufferBadInputs.cpp [hunk 7]
```diff
@@ -27,18 +22,19 @@ REGISTER_KOTEKAN_STAGE(bufferBadInputs);
 
 bufferBadInputs::bufferBadInputs(Config& config_, const std::string& unique_name,
                                  bufferContainer& buffer_container) :
-    Stage(config_, unique_name, buffer_container, std::bind(&bufferBadInputs::main_thread, this)) {
-
-    uint32_t num_elements = config.get<uint32_t>(unique_name, "num_elements");
-    input_mask_len = sizeof(uint8_t) * num_elements;
+    Stage(config_, unique_name, buffer_container, std::bind(&bufferBadInputs::main_thread, this)),
+    // Required since frameID has no default constructor
+    frame_id(0) {
```
rhaas80: that initializes `frame_id` with a NULL pointer, will fail with a SEGFAULT.

rhaas80: *is* `frame_id` actually used outside of the `main_thread`? If not then it can just be defined in that function.

ljgray: It's used in `update_bad_inputs_callback`. I suppose I can rework the logic somehow...

ljgray: Just reverted to using a standard integer and manually wrapping.

rhaas80: :-) The alternative is something like `frame_id(get_buffer("out_buf"))` instead of the `frame_id(0)` 

But the manual wrapping will work.

rhaas80 (approved):

lib/stages/bufferBadInputs.cpp
ljgray: So I suppose this means we need an additional table in the config to map to beamformer order? How is that ordering passed to the reorder kernel?

rhaas80: One could add an option to the parseInputReorder stage. There is no explicit table that records the order in which either the channel_ids or the adc_ids would be to be in beamformer order. Naturally such an option would be very CHIME specific (but so is the whole xpose2048 kernel).

The reorder kernel gets a list that tells it for each element it finds in the incoming frame where to put it in the outgoing frame (so the inverse mapping of what is recorded in the yaml file actually).

ljgray: I'm almost tempted to just hard-code the reorder in here and make the stage CHIME-specific, since it's such a simple mapping and it's unlikely to change. Thoughts?

PR #1442, opened 2026-03-12
added gps week rollover offset to CHORDTelescope

geoffryan:
~Adds `gps_week_rollover_offset` parameter to CHORD Telescope.  Set it to the number of rollovers to add to the GPS time.~

~e.g. 1 => 1024 weeks of nanoseconds are added.~

Adds `auto_correct_gps_week_rollover` parameter to CHORD Telescope. If True, compares the GPS `frame0_nano` to GPS `start_ctime` and adjusts `frame0_nano` to be on the same GPS 1024 week rollover period.

Updated Earth Orientation Parameter scripts and documentation in `tools/earth_orientation`.


lib/utils/CHORDTelescope.hpp
```diff
@@ -398,13 +403,15 @@ struct GPSTimeParams {
      **/
     static GPSTimeParams from_config(const kotekan::Config& config, const std::string& path);
 
+    /// Retrieve the time0_ns from a remote server (fpga_master)
+    static bool get_gps_time0_ns_from_remote(const GPSTimeParams& gps, uint64_t& time0_ns);
```
andrerenard: Add a warning this is a blocking call.  If we try to use it in anything remotely time critical we'll get into trouble if there there are network/fpga_master issues.

lib/utils/CHORDTelescope.cpp
```diff
@@ -341,36 +342,59 @@ void GPSTimeParams::set_gps_time_params_from_config(GPSTimeParams& gps,
...
+        return false;
     }
 
-    gps.time0_ns = json_reply["frame0_nano"].get<uint64_t>();
-    INFO_NON_OO("GPS frame0 time set to {:d}", gps.time0_ns);
-    gps.gps_enabled = true;
+    // Number of nanoseconds to adjust for the 1024 week rollover.
+    uint64_t week_rollover_offset_ns = gps.gps_week_rollover_offset * 1024ul * 7ul * 86400ul * GIGA;
+
+    time0_ns = json_reply["frame0_nano"].get<uint64_t>() + week_rollover_offset_ns;
```
andrerenard: I feel like we should do a quick sanity check with system time to see if we are within at least the same ~decade?  Although cannot be too tight of a check, since frame0 could be maybe weeks ago if we've been running for a while.

andrerenard (dismissed):
Looks great, thanks!   Two suggestions inline.

lib/utils/CHORDTelescope.cpp
```diff
@@ -341,36 +342,59 @@ void GPSTimeParams::set_gps_time_params_from_config(GPSTimeParams& gps,
...
+        return;
+    }
+
+    // Otherwise, we successfully have the time!
+    gps.time0_ns = time0_ns;
+    gps.gps_enabled = true;
+    INFO_NON_OO("GPS frame0 time set to {:d}", gps.time0_ns);
+}
+
+bool GPSTimeParams::get_gps_time0_ns_from_remote(const GPSTimeParams& gps, uint64_t& time0_ns) {
```
jbmertens: 🙏

jbmertens (approved):

lib/utils/CHORDTelescope.cpp
jbmertens: And a check that not in the future.

lib/utils/CHORDTelescope.hpp
geoffryan: Done, also added a `timeout` parameter the caller can set.

lib/utils/CHORDTelescope.cpp
geoffryan: Updated the approach, `fpga_master` also returns its start time which is a good UNIX time.  We use that to dynamically determine the GPS epoch.

geoffryan:
Few updates:
- We (optionally) dynamically detect a GPS rollover problem by comparing GPS `frame0_nano` with GPS `start_ctime`, we then adjust `frame0_nano` to be on the same rollover period as `start_ctime`.
- Substantially updated the python scripts in `tools/earth_orientation` to generate & send EOP tables for Kotekan. They can generate a static EOP good for starting kotekan with, and also generate periodic updates which will ensure continuity as kotekan runs and the EOP change.

PR #1443, opened 2026-03-14
test xpose2048

rhaas80:
adds a test case for the xpose2048 transpose GPU stage

### DIFF
--- a/lib/testing/testXpose.cpp
+++ b/lib/testing/testXpose.cpp
@@ -1,0 +1,197 @@
+#include "testXpose.hpp"
+
+#include "Config.hpp" // for Config
+#include "Metadata.hpp"        // for GenericNDArray
+#include "StageFactory.hpp"    // for REGISTER_KOTEKAN_STAGE
+#include "buffer.hpp"          // for Buffer
+#include "bufferContainer.hpp" // for bufferContainer
+#include "chordMetadata.hpp"   // for get_chord_metadata, chordMetadata
+#include "visUtil.hpp"         // for frameID
+
+#include "json.hpp" // for basic_json, json, iter_impl
+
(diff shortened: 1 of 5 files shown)


lib/testing/testXpose.cpp
```diff
@@ -0,0 +1,197 @@
...
+        FATAL_ERROR("Unexpected frames sizes for out_xposed_buf, expected {:d} received {:d}",
+                    out_xposed_buf->frame_size, num_times * num_xposed_frequencies * num_polarizations * num_dishes * sizeof(uint8_t));
+    out_xposed_buf->allocate_ndarray_frame_desc(kotekan::int4x2_swapped_withoffset, "E", {num_times, num_xposed_frequencies, num_polarizations, num_dishes}, {"T", "F", "P", "D"});
+
+    scatter_indices_buf = get_buffer("scatter_indices_buf");
+    scatter_indices_buf->register_producer(unique_name);
+    if (scatter_indices_buf->frame_size != num_polarizations * num_dishes * sizeof(int32_t))
+        FATAL_ERROR("Unexpected frames sizes for scatter_indeces, expected {:d} received {:d}",
+                    scatter_indices_buf->frame_size, num_polarizations * num_dishes * sizeof(int32_t));
+    // TODO: this is not quite correct. Really if going to cylinder order
```
jbmertens, line 61: Flagging in case this should be fixed.

jbmertens (approved):

lib/testing/testXpose.cpp
rhaas80: well let's first see if the CHIME equivalent beamformer requires things to come in in beamformer order (which is Pol, Dish) anyway :-)

PR #1448, opened 2026-03-18
testDataGen: New parameter to set dish position metadata

eschnett:
These are old-style metadata that we want to phase out in favour of the telescope object, but we're still using them at the moment.


geoffryan (approved):
LGTM!

PR #1449, opened 2026-03-18
config changes (and merge CHORD and lint)

jbmertens:


andrerenard (approved):
Thanks for the fixes.

PR #1450, opened 2026-03-20
Rhaas/chime upgrade

rhaas80:
* corrections for passing ndarray metadata in buffer stages
* extra metadata in n2k_count from lostSamples
* thread control for blaze


ljgray (approved):
Seems fine to me besides the failing tests

rhaas80:
> Seems fine to me besides the failing tests

Interesting. I would not have expected to have touched anything that actually is tested. Well, on never stops learning new things  :-)

One other thing is maybe poke the kotekan-dev group about how to handle `buffer->set_frame_desk(nullptr)` ie accept this (and make sure that such a call errors out if the existing frame-desc is not NULL) or keep it like it is right now which means the callers (`bufferDelay`, `bufferMerge` etc.) have to check if their inbufs have a framedesk or not.

[...]

ah, tripped up (again) by a RelWithDebSymbols build setting `NDEBUG` that is disabling all `assert()` statements.

lib/stages/bufferQuiet.cpp
```diff
@@ -51,11 +58,24 @@ void bufferQuiet::main_thread() {
...
 
+    bool first_time = true;
     while (!stop_thread) {
 
-        timespec now;
-        clock_gettime(CLOCK_MONOTONIC, &now);
-        const timespec timeout = add_nsec(now, _quiet_time * 1e9);
+        // system_clock is the time used by wait_for_full_frame_timeout
+        const auto now = std::chrono::time_point_cast<std::chrono::nanoseconds>(
+            std::chrono::system_clock::now());
```
jbmertens, line 66: Steady clock, just in case?

lib/stages/EigenN2Iter.cpp
```diff
@@ -153,7 +156,10 @@ void EigenN2Iter::main_thread() {
     uint32_t num_elements = 0;
     bool initialized = false;
 
+    // these are kotekan wide (proces wide) settings
```
jbmertens: proces(s)?

lib/stages/EigenN2Iter.cpp
```diff
@@ -153,7 +156,10 @@ void EigenN2Iter::main_thread() {
...
     openblas_set_num_threads(1);
+    if (_num_blaze_workers > 0)
+        blaze::setNumThreads(_num_blaze_workers);
```
jbmertens, line 162: Does kotekan's thread affinity handle more than one worker correctly?

lib/stages/bufferQuiet.cpp
rhaas80: not really, no. `wait_for_full_frame_timeout` uses `system_clock` so unless that one is also updated there is not much point using a different clock here.  So, I'd rather not :-)

lib/stages/EigenN2Iter.cpp
rhaas80: yes to this one

lib/stages/bufferQuiet.cpp
jbmertens: Ah, `wait_for_full_frame_timeout` should probably use a steady clock then, but that is an issue for another PR.

lib/stages/EigenN2Iter.cpp
rhaas80: define "correctly". If I set `cpu_affinity = [n1, n2, n3]` and ask for 3 threads from blaze than those three threads inherit their parent's affinity mask ie are limit to cores `n1`, `n2`, `n3` but not to any single one of them. If I had asked for 4 threads from blaze they'd compete for the 3 available cores. At least for blaze's OpenMP parallelization scheme (since OpenMP behaves that way). Now how the pthread scheme or some others that we do not use behave, I have no clue :-)

lib/stages/bufferQuiet.cpp
rhaas80: exactly. that is someone else's problem

lib/stages/EigenN2Iter.cpp
jbmertens: Great, yeah this is the behavior I would expect. May be worth issuing a warning if the number of threads exceeds cores, but this is not essential.

PR #1453, opened 2026-03-25
cudaCopyFromRingbuffer: only unregister memory if actually registered

rhaas80:
matches CopyToRingBuffer, fails otherwise and leaves behind corrupted files

### DIFF
--- a/lib/cuda/cudaCopyFromRingbuffer.cpp
+++ b/lib/cuda/cudaCopyFromRingbuffer.cpp
@@ -79,7 +79,8 @@ cudaCopyFromRingbuffer::cudaCopyFromRingbuffer(Config& config, const std::string
 cudaCopyFromRingbuffer::~cudaCopyFromRingbuffer() {
     if (out_buffer && out_buffer->frame_size) {
         uint flags;
-        if (cudaErrorInvalidValue == cudaHostGetFlags(&flags, out_buffer->frames[instance_num])) {
+        // only unregister if it's already been registered
+        if (cudaSuccess == cudaHostGetFlags(&flags, out_buffer->frames[instance_num])) {
             CHECK_CUDA_ERROR(cudaHostUnregister(out_buffer->frames[instance_num]));
         }
     }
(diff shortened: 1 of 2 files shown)


rhaas80:
@dstndstn care to review?

dstndstn:
ahhh, sorry, sure, will review :)

dstndstn:
Could you also modify `cudaOutputData.cpp`?  It has this same stanza (that's probably where I copied this from!!)

Actually, maybe this bit of (frankly non-obvious) code should get pulled out into a couple of utility functions to register/unregister?  It appears to be called from ~5 places (cuda{Input/Output}Data and cudaCopy{From,To,NTo}Ringbuffer).

(Oh, actually, cudaHostRegister gets called from more places.... could I ask you to check for Register/Unregister pairs?)

rhaas80:
sure

rhaas80:
`cudaOutputData` was the only one not yet fixed. If one really wants to collect his logic (which seems fairly straightforward to me, similar to checking if a pointer is actually non-`NULL` before `free`ing it (admittedly `free` will happily ignore a `NULL` pointer passed to it). 

If consolidating this should be part of a `gpuBuffer` or so class. Though there seem to be other issues with the gpu buffers right now (such as no effective checks for typos or cross checks that only one ringbuffer claims control of a given gpu memory block). So maybe that is a larger project.

For the record, these files (not in `external`, which is on its own): use `cudaHostUnregister`:

```
$ git grep -l cudaHostUnregister | grep -v external
lib/cuda/cudaCopyFromRingbuffer.cpp
lib/cuda/cudaCopyNToRingbuffer.cpp
lib/cuda/cudaCopyToRingbuffer.cpp
lib/cuda/cudaInputData.cpp
lib/cuda/cudaOutputData.cpp
```

The kernels for the Julia generated code also contain `cudaHostRegister`, but never unregister their data (which is fine, since the program will terminate anyway when they are no longer needed).

rhaas80:
please re-review

dstndstn (approved):

PR #1454, opened 2026-03-31
Correct FRB beamformer

eschnett:
There was a bug in the CHORD FRB beamforming kernel.


jbmertens:
In autogenerated code we trust

jbmertens (approved):

config/fengine/include/frb_chime.j2
```diff
@@ -13,8 +13,8 @@ frb_num_beams_Q: 2 * num_dishes_x
 frb1_swap_xy: true
 frb1_input_scale: 1.0 / 1024
 
-frb2_num_beams_x: 4
-frb2_num_beams_y: 256
+frb2_num_beams_x: 8
+frb2_num_beams_y: 512
```
rhaas80, line 17: one per polarization I guess?

rhaas80 (commented):
so what was the actual bug? The Julia code makes multiple changes each which looks innocuous to me?

PR #1458, opened 2026-04-01
Fix n2counts rfi indexing bug

ljgray:
Still a draft because I'm still working on getting the python test working, but the fixes to the actual stage should be done and can probably be reviewed and tested using #1459

### DIFF
--- a/lib/stages/lostSamplesToN2Counts.cpp
+++ b/lib/stages/lostSamplesToN2Counts.cpp
@@ -154,9 +154,11 @@ void lostSamplesToN2Counts::main_thread() {
         // Get the RFI mask low num samples, which is the last dimension
         std::shared_ptr<chordMetadata> rfi_meta =
             get_chord_metadata(rfi_mask_buf, rfi_mask_buf_frame_id);
-        // rfi_mask fine time samples (last dimension)
+        // rfi_mask fine time samples (last dimension). Note that this
+        // is in *bytes*, not bits
         size_t _rfi_t_lo = rfi_meta->dim[rfi_meta->dims - 1];
-        // Stride between coarse time samples
+        size_t _rfi_t_lo_bits = BITS_PER_BYTE * _rfi_t_lo;
+        // Stride between coarse time samples, also in bytes
         size_t _rfi_f_stride = _rfi_t_lo;
         size_t _rfi_t_hi_stride = num_n2k_freq * _rfi_f_stride;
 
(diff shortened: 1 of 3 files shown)


lib/stages/lostSamplesToN2Counts.cpp
```diff
@@ -140,8 +140,8 @@ void lostSamplesToN2Counts::main_thread() {
     while (!stop_thread) {
         // Just need one complete frame with all frequencies here
         DEBUG("Waiting on empty n2k_counts frame.");
-        int32_t* n2k_count_frame =
-            (int32_t*)n2k_counts_buf->wait_for_empty_frame(unique_name, n2k_counts_buf_frame_id);
+        uint8_t* n2k_count_frame =
```
geoffryan: counts is an `int32_t`, this was correct before.  Reverting this should make the Python test pass.

lib/stages/lostSamplesToN2Counts.cpp
```diff
@@ -222,13 +223,11 @@ void lostSamplesToN2Counts::main_thread() {
...
-                        // base offset + coarse_time * stride + fine_time
-                        size_t tsr =
-                            ridx_f + (ts / _rfi_t_lo) * _rfi_t_hi_stride + (ts % _rfi_t_lo);
-                        // Convert bit index in a uint1 buffer to byte index in a uint8 buffer
-                        tsr /= BITS_PER_BYTE;
-                        // Assume that this is a uint1 stream viewed as uint8, meaning
-                        // that the MSB is the first sample in a given byte
+                        // base offset (in bytes) + coarse_time_sample * stride (in bytes)
+                        // + fine_time_sample (in bytes)
+                        size_t tsr = ridx_f + (ts / _rfi_t_lo_bits) * _rfi_t_hi_stride
```
geoffryan, line 228: Looks correct to me!

lib/stages/lostSamplesToN2Counts.cpp
```diff
@@ -222,13 +223,11 @@ void lostSamplesToN2Counts::main_thread() {
...
+                                     + (ts % _rfi_t_lo_bits) / BITS_PER_BYTE;
+                        // Iterate through individual bits in the rfi mask
                         uint8_t shiftval = (BITS_PER_BYTE - 1) - ts % BITS_PER_BYTE;
```
geoffryan: This is wrong.  The rfimask layout is logically like a `uint1`.  Time counts upwards with the bit index.  can just be the simple thing:
`shiftval = ts % BITS_PER_BYTE`

This will break the python test as is but passes the yaml test against a random rfi mask!

lib/stages/lostSamplesToN2Counts.cpp
geoffryan: For safety, I'd suggest `lost_samples_frame[ts] ^ 1u` --> `(lost_samples_frame[ts] & 1u) ^ 1u`  just in case other bits in lost_samples get set accidentally.

tests/test_lostsamples_to_n2counts.py
geoffryan, line 1: You'll want to change this to read the rfi mask in ascending bit order (instead of the current decreasing)

geoffryan (changes requested):
One change to pass the pytest, two to pass the yaml test (with random rfi).  Corresponding change in the pytest will be needed as well.

lib/stages/lostSamplesToN2Counts.cpp
```diff
@@ -140,8 +140,8 @@ void lostSamplesToN2Counts::main_thread() {
     while (!stop_thread) {
         // Just need one complete frame with all frequencies here
         DEBUG("Waiting on empty n2k_counts frame.");
-        int32_t* n2k_count_frame =
-            (int32_t*)n2k_counts_buf->wait_for_empty_frame(unique_name, n2k_counts_buf_frame_id);
+        uint32_t* n2k_count_frame =
```
geoffryan: Unlikely we'll count high enough for this to matter, but counts is int32_t not uint32_t

geoffryan (changes requested):

lib/stages/lostSamplesToN2Counts.cpp
ljgray: Fixed. What's the reason for using signed here? I'd assume that `counts` should always be positive

geoffryan: It certainly should be! I'm not entirely sure of the reason, that's the data type in the cuda kernel and we're just preserving that.

geoffryan (approved):

geoffryan:
LGTM!

PR #1459, opened 2026-04-01
Lost Samples consistency test

geoffryan:
Adds a CI test comparing the `cudaPL1bitCorrelator` and `lostSamplesToN2Counts` stages and ensuring they produce identical output.

~~Currently the test uses a trivial RFI mask.  This should be changed to a random mask once #1458 is completed.~~
Test uses a random RFI mask.

Note: ~~currently in draft~~ due to GPU issue. Test (with trivial mask) passes using CPU pipeline.


ljgray (approved):

geoffryan:
With merge of #1458 the harder test with a random RFI mask passes!

This test (still) currently runs on CPU, comparing `lostSamplesToN2Counts` against the CPU `PLMaskExpander` -> `PL1bitCorrelator` stages. These stages are also verified in CI against their GPU equivalents.  Leaving as a task for later fixing the test config to get the GPU comparison working as well.

PR #1462, opened 2026-04-13
ksgpu build updates

geoffryan:
Tweaks to the `ksgpu` CMake file enabling compilation on CTC and suppressing a warning.

### DIFF
--- a/external/ksgpu/CMakeLists.txt
+++ b/external/ksgpu/CMakeLists.txt
@@ -23,14 +23,12 @@ add_custom_command(
 # Define source files and headers
 file(GLOB HFILES "include/ksgpu/*.hpp" "include/*.hpp")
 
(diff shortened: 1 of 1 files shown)


rhaas80:
it also works to git revert 1c0c57e5988ab73a53f7c54f3c912b0d6a785e2a (since that compiles the .cpp files in ksgpu using nvcc (which is what upstream does). This may actually be cleaner than passing CUDA include files to the host compiler directly (just in case nvcc does some extra magic).

jbmertens:
LGTM, but tests aren't running on the force push, so just re-run by hand before merging.
(Or maybe the force-push has done the right thing, I'm not familiar with how that interacts with tests.)

jbmertens (approved):

rhaas80 (approved):

PR #1463, opened 2026-04-13
Read and combine gain files for multiple frequencies

ljgray:
Iterates over all local freqs and gathers them into a single gain buffer. For the tracking beamformer, supports different gain files per beam, but produces a single buffer with a `beam_id` axis (this is consistent with the production CHIME approach).

Stores the gains in an internal array and emits a new kotekan buffer in every pass of the main thread, which can then be copied to the GPU at the same cadence as other data.

# TODO
- [ ] Split into two stages - loader (emits single buffer only on update) and processor (copies to internal array, upchannelizes, and multiplies in a bad feed mask and pointing phases)


ljgray:
This could probably do with a unit test, but we should be able to test it as part of the upgrade pipeline before I get to that

lib/stages/ReadGain.cpp
```diff
@@ -48,41 +45,49 @@ REGISTER_KOTEKAN_STAGE(ReadGain);
...
+    // Initiqlize the vectors holding freq ID and freq MHz
+    freq_idx.resize(_num_local_freq, FREQ_ID_NOT_SET);
+    // NB: I don't think that this is actually used anywhere
+    freq_MHz.resize(_num_local_freq, -1);
 
     // Gain for FRB
     gain_frb_buf = get_buffer("gain_frb_buf");
-    gain_frb_buf_id = 0;
     gain_frb_buf->register_producer(unique_name);
+    gain_frb_buf_id = 0;
```
rhaas80: doesn't this reset the `gain_frb_buf_id(get_buffer("gain_frb_buf"))` from the constructor?

lib/stages/ReadGain.cpp
```diff
@@ -48,41 +45,49 @@ REGISTER_KOTEKAN_STAGE(ReadGain);
...
     update_gains_frb = false;
 
     // Gain for Tracking Beamformer
     gain_tracking_buf = get_buffer("gain_tracking_buf");
-    gain_tracking_buf_id = 0;
     gain_tracking_buf->register_producer(unique_name);
+    gain_tracking_buf_id = 0;
```
rhaas80: same as for `gain_frb_buf_id`

lib/stages/processGains.cpp
```diff
@@ -48,41 +45,49 @@ REGISTER_KOTEKAN_STAGE(ReadGain);
...
     update_gains_tracking = false;
 
-    // Create a space to store gains for all the tracking beams
-    // Initial values will be set by the call back functions below on startup
-    tracking_beam_gains = (float*)malloc(gain_tracking_buf->frame_size);
-    if (tracking_beam_gains == nullptr) {
-        throw std::runtime_error("Could not allocate memory in ReadGain for tracking gains");
+    // Check that buffers are as-expected
+    auto nsamples = _num_elements * _num_local_freq * 2;
+    if (gain_frb_buf->frame_size != nsamples) {
```
rhaas80: `frame_size` is in bytes, are the gain values byte sized?

lib/stages/processGains.cpp
```diff
@@ -48,41 +45,49 @@ REGISTER_KOTEKAN_STAGE(ReadGain);
...
+        FATAL_ERROR("gain_frb_buf does not have the expected size. Expected {:d}, got {:d}",
+                    nsamples, gain_frb_buf->frame_size);
+    }
+
+    if (gain_tracking_buf->frame_size != nsamples * _num_beams) {
```
rhaas80: same comment about bytes as above

lib/stages/ReadGain.cpp
```diff
@@ -155,129 +157,173 @@ void ReadGain::read_gain_frb() {
...
+
+    // Each frequency is provided in a different file
+    for (size_t i = 0; i < freq_idx.size(); ++i) {
+        auto fid = freq_idx[i];
+        bool ffailed = false; // set if this frequency can't be read
+        auto foffset = i * fstride;
+
+        // Construct the gain file name for this freq
+        snprintf(filename, sizeof(filename), "%s/quick_gains_%04d_reordered.bin",
+                 _gain_dir_frb.c_str(), fid);
```
rhaas80: should strictly speaking check that the return value (non-NUL bytes that are required to store the full output) is less than the size of the `filename` array.

lib/stages/ReadGain.cpp
```diff
@@ -155,129 +157,173 @@ void ReadGain::read_gain_frb() {
...
+        ptr_myfile = fopen(filename, "rb");
+
+        if (ptr_myfile == nullptr) {
+            WARN("GPU Cannot open gain file {:s}", filename);
+            any_failed = true;
+            ffailed = true;
+        } else if (_num_elements
+                   // The actual read happens here
+                   != fread(out_frame_frb + foffset, sizeof(float), 2 * _num_elements,
+                            ptr_myfile)) {
```
rhaas80: this will fail. You are asking for `2*_num_elements` to be read but check against `_num_elements`

lib/stages/ReadGain.cpp
```diff
@@ -155,129 +157,173 @@ void ReadGain::read_gain_frb() {
...
-        if (_num_elements != fread(out_frame_frb, sizeof(float) * 2, _num_elements, ptr_myfile)) {
-            WARN("Gain file ({:s}) wasn't long enough! Something went wrong, using default "
-                 "gains",
-                 filename);
-            gains_last_update_success_metric.labels({"frb"}).set(0);
-            for (uint i = 0; i < _num_elements; i++) {
-                out_frame_frb[i * 2] = default_gains[0] * scaling;
-                out_frame_frb[i * 2 + 1] = default_gains[1] * scaling;
+
+        fclose(ptr_myfile);
```
rhaas80: this will call `fclose(nullptr)` if the `fopen()` failed, won't it?

> The behaviour of fclose() is undefined if the stream parameter is an illegal pointer,

lib/stages/ReadGain.cpp
```diff
@@ -155,129 +157,173 @@ void ReadGain::read_gain_frb() {
...
-                         _num_elements, ptr_myfile)) {
-                WARN("Gain file ({:s}) wasn't long enough! Something went wrong, using default "
-                     "gains",
-                     filename);
-                gains_last_update_success_metric.labels({beam_label}).set(0);
-                for (uint i = 0; i < _num_elements; i++) {
-                    tracking_beam_gains[(beam_id * _num_elements + i) * 2] = default_gains[0];
-                    tracking_beam_gains[(beam_id * _num_elements + i) * 2 + 1] = default_gains[1];
+
+            fclose(ptr_myfile);
```
rhaas80: pretty much the same comments as for the other reader routine :-)

lib/stages/processGains.cpp
```diff
@@ -155,129 +157,173 @@ void ReadGain::read_gain_frb() {
...
-    for (int beam_id = 0; beam_id < _num_beams; ++beam_id)
-        DEBUG("Gain_tracking_buf[{:d}]: {:.2f} {:.2f} ", beam_id,
-              out_frame_tracking[beam_id * _num_elements * 2],
-              out_frame_tracking[beam_id * _num_elements * 2 + 1]);
-#endif
+        if (any_failed) {
+            gains_last_update_success_metric.labels({beam_label}).set(0);
+        } else {
+            gains_last_update_success_metric.labels({beam_label}).set(1);
+        }
```
rhaas80, line 383: no distinction between tracking and frb gains?

lib/stages/processGains.cpp
```diff
@@ -155,129 +157,173 @@ void ReadGain::read_gain_frb() {
...
+    gain_tracking_buf_id++;
 }
 
 void ReadGain::main_thread() {
-
-    uint8_t* frame =
-        metadata_buf->wait_for_full_frame(unique_name, metadata_buffer_precondition_id);
+    // don't need to track metadata buffer ID here - they just get used
+    // once at startup
+    uint8_t* frame = metadata_buf->wait_for_full_frame(unique_name, 0);
```
rhaas80, line 391: if metadatabuf if a voltage (network) buffer then it will not have all frequencies listed since dpdk creates frames with 1 frame per frequency initially

lib/stages/ReadGain.cpp
ljgray: Ah - I think it's just a relic that I forgot to remove

lib/stages/processGains.cpp
ljgray: Apparently not - this is a relic from before, so I'm hesitant to change it, but easy to update if that's what we want to do

lib/stages/processGains.cpp
ljgray: Oh, interesting. In the old config (where the old version of this stage was used), the network buffer seems to have included all local freqs. This can just be fixed by a config change, though

rhaas80: Now I need to check to see if I misremembered. Lets see: 

https://github.com/kotekan/kotekan/blob/eede49e02c537927287cb2db02bd1ac597be8078/lib/dpdk/iceBoardShuffle.hpp#L467

So indeed a change compared to CHIME's hotfix branch in that only the single frequency in the frame is recorded in the `coarse_freq` metadata (assuming this was what  I complained about).

Lets see what we really need once things are put together in an actual pipeline.

ljgray: Looking at the upgrade config, there's a `num_freq_host_buf` which is used in the network buffer frame shape, and is set to 1. So most likely this is just a general change from the old system

ljgray:
I've made changes to store the gains in an internal buffer and copy to a kotekan buffer every loop, which lets us copy them to the GPU at the same cadence as any other data.

lib/stages/processGains.cpp
```diff
@@ -0,0 +1,416 @@
...
+    // Set frame descriptions for each produced
+    gain_frb_buf->allocate_ndarray_frame_desc<kotekan::GetType_t<kotekan::int32>, 3>(
+        "gain_frb", {static_cast<long>(_num_local_freq), static_cast<long>(_num_elements), (long)2},
+        {"F", "E", "ReIm"});
+
+    gain_tracking_buf->allocate_ndarray_frame_desc<kotekan::GetType_t<kotekan::int32>, 4>(
+        "gain_tracking",
+        {static_cast<long>(_num_local_freq), static_cast<long>(_num_beams),
+         static_cast<long>(_num_elements), (long)2},
+        {"F", "B", "E", "ReIm"});
```
ljgray: Not certain that this is the correct representation for the `beam` and `element` axes (should elements always be split into `D` and `P`?)

config/chime_upgrade_test_gains.j2
```diff
@@ -0,0 +1,738 @@
+##########################################
+#
+# chime_upgrade_n2
+#
+# Pipeline for running the n2k correlator, accumulate onto a visibility matrix
+# and compute eigen-decomposition.
```
rhaas80, line 6: likely incorrect comment

lib/stages/processGains.cpp
```diff
@@ -0,0 +1,456 @@
...
+// Tracking Beamformer
+// curl localhost:12048/updatable_config/tracking_gain -X POST -H 'Content-Type: application/json' -d
+// '{"gain_dir":["path0","path1","path2","path3","path4","path5","path6","path7","path8","path9"]}'
+//
+// clang-format on
+
+processGains::processGains(Config& config, const std::string& unique_name,
+                           bufferContainer& buffer_container) :
+    Stage(config, unique_name, buffer_container, std::bind(&processGains::main_thread, this)),
+    gain_frb_buf_id(get_buffer("gain_frb_buf")),
```
rhaas80: not really an "id" anymore, but the actual buffer (at least to me "id" would be the string value of the gain_frb_buf option).

lib/stages/processGains.cpp
```diff
@@ -0,0 +1,456 @@
...
+    gain_frb_buf->register_producer(unique_name);
+    update_frb_flag.store(false);
+
+    // Gain for Tracking Beamformer
+    gain_tracking_buf = get_buffer("gain_tracking_buf");
+    gain_tracking_buf->register_producer(unique_name);
+    update_tracking_flag.store(false);
+
+    // Check that buffers are as-expected
+    auto frb_nsamples = _num_elements * _num_local_freq * _upchan_factor * 2 * sizeof(float);
```
rhaas80: `_samples` -> `_size` (each sample being a `float`)?

lib/stages/processGains.cpp
```diff
@@ -0,0 +1,456 @@
...
+    auto tracking_nsamples = _num_elements * _num_local_freq * _num_beams * 2 * sizeof(float);
+    if (gain_tracking_buf->frame_size != tracking_nsamples) {
+        FATAL_ERROR("gain_tracking_buf does not have the expected size. Expected {:d}, got {:d}",
+                    tracking_nsamples, gain_tracking_buf->frame_size);
+    }
+
+    // Set frame descriptions for each produced
+    gain_frb_buf->allocate_ndarray_frame_desc<kotekan::GetType_t<kotekan::float32>, 4>(
+        "gain_frb",
+        {static_cast<long>(_num_local_freq), static_cast<long>(_upchan_factor),
```
rhaas80: I think they are actually `ptrdiff_t`:

```
    /**                                                                                                      * @brief Allocates a new frame description object holding a D dimensional
     *        array of type T                                                                                * @param[in] value_type the kotekan type enumerator of the values stored
     * @param[in] rank dimensionality of the data array                                                      * @param[in] extents Array extentds in the D dimensions
     */
    void allocate_ndarray_frame_desc(kotekan::DataType value_type, kotekan::Symbol quantity_name,                                            const std::vector<std::ptrdiff_t>& extents,
                                     const std::vector<kotekan::Symbol>& dimnames) {
```

which is the same (signed 64bit) type as a `ptrdiff_t` admittedly.

lib/stages/processGains.cpp
```diff
@@ -0,0 +1,456 @@
...
+    if (gain_frb_buf->frame_size != frb_nsamples) {
+        FATAL_ERROR("gain_frb_buf does not have the expected size. Expected {:d}, got {:d}",
+                    frb_nsamples, gain_frb_buf->frame_size);
+    }
+
+    auto tracking_nsamples = _num_elements * _num_local_freq * _num_beams * 2 * sizeof(float);
+    if (gain_tracking_buf->frame_size != tracking_nsamples) {
+        FATAL_ERROR("gain_tracking_buf does not have the expected size. Expected {:d}, got {:d}",
+                    tracking_nsamples, gain_tracking_buf->frame_size);
+    }
```
rhaas80, line 82: interesting, why one set of gains per beam? The gains computed from the visibility matrix would not know anything about beams, do they? So something else comes into play here as well? Eg these are already the coefficients used to form each beam given voltage data?

lib/stages/processGains.cpp
```diff
@@ -0,0 +1,456 @@
...
+    bool any_failed = false;                             // set if any freq can't be read
+
+    // Each frequency is provided in a different file
+    for (size_t i = 0; i < freq_idx.size(); ++i) {
+        auto fid = freq_idx[i];
+        bool ffailed = false; // set if this frequency can't be read
+        auto foffset = i * fstride;
+
+        // Construct the gain file name for this freq
+        auto name = snprintf(filename, sizeof(filename), "%s/quick_gains_%04d_reordered.bin",
```
rhaas80: `namelen`? `filenamelen_needed`?

lib/stages/processGains.cpp
```diff
@@ -0,0 +1,456 @@
...
+            WARN("Filename is too long - need {:d} bits but buffer only has {:d}", name,
+                 sizeof(filename));
+            ptr_myfile = nullptr;
+        } else {
+            INFO("[FRB] Loading gains from {:s}", filename);
+            ptr_myfile = fopen(filename, "rb");
+        }
+
+        if (ptr_myfile == nullptr) {
+            WARN("Failed to open gain file {:s}", filename);
```
rhaas80, line 233: will trigger (and warn) if the file name is too long as well (since `ptr_myfile` is `NULL`)

lib/stages/processGains.cpp
```diff
@@ -0,0 +1,456 @@
...
+/// endpoint for each beam.
+void processGains::update_gain_tracking() {
+    if (!update_tracking_flag.load()) {
+        return;
+    }
+
+    // If we should update, wait for a moment in case multiple beam updates
+    // are received in a short period of time (so that we can try to handle
+    // them in one pass)
+    std::this_thread::sleep_for(std::chrono::seconds(2));
```
rhaas80: still a race, in case we do not finish (and reset the flag) before the next one comes in. 2s has been historically sufficient wait time I guess?

lib/stages/processGains.cpp
```diff
@@ -0,0 +1,456 @@
...
+    };
+
+    // Assign the frequency indices and MHz values
+    for (size_t f = 0; f < freq_idx.size(); ++f) {
+        freq_id_t fid = meta_freq_idx[f];
+        freq_idx[f] = fid;
+    }
+
+    metadata_buf->mark_frame_empty(unique_name, 0);
+    metadata_buf->unregister_consumer(unique_name);
```
rhaas80, line 408: (assuming this is not old code) is it guaranteed that unregistering myself will mark all frames that may have already arrived as empty?

lib/stages/processGains.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+    /// Destructor.
+    ~processGains();
+
+    void main_thread() override;
+
+    /// Endpoint for providing new directory path for FRB gain updates
+    bool update_gains_frb_callback(nlohmann::json& json);
+    /// Endpoint for providing new directory path for <span class="x x-first x-last">tracking
+    /// beamformer</span> gain updates
```
rhaas80: why HTML code?

rhaas80 (approved):
minor naming comments

config/chime_upgrade_test_gains.j2
ljgray: Config is `donotmerge` so won't matter

lib/stages/processGains.cpp
ljgray: Apparently the idea is that some external user could provide whatever gains they want to each individual beam (and that's how it was implemented in the past so I lean towards not changing it...)

lib/stages/processGains.cpp
ljgray: It's pretty arbitrary, and 2s is just a vague guess. More of a symptom of iffy design (independent callbacks for each beam but all beams loaded in the same place).

lib/stages/processGains.hpp
ljgray: It's whatever was written before - might be some doxygen thing? Not entirely sure...

lib/stages/processGains.cpp
ljgray: I could be wrong but in this case I would expect that only one frame from `metadata_buf` should ever arrive/be consumed here?

lib/stages/processGains.hpp
ljgray: Oh nvm, not doxygen. Should just be removed

PR #1464, opened 2026-04-14
Add benchmark config files

eschnett:
Also add two stages to efficiently invent voltages for CHORD and CHIME.


geoffryan (approved):
The stages could use some doc comments so others can use them, but otherwise LGTM!

PR #1465, opened 2026-04-16
Bad input buffer reorder

ljgray:
Adds a helper function in `visUtil` with the mapping from cylinder to beamformer order. `visUtil` seems like a reasonable place for this, but I'm fine with suggestions to put it elsewhere.

### DIFF
--- a/lib/stages/bufferBadInputs.cpp
+++ b/lib/stages/bufferBadInputs.cpp
@@ -6,6 +6,7 @@
 #include "chordMetadata.hpp"  // for get_chord_metadata, chordMetadata
 #include "configUpdater.hpp"  // for configUpdater
 #include "kotekanLogging.hpp" // for DEBUG, ERROR
+#include "visUtil.hpp"        // for get_cylinder_to_beamformer_reorder_table
 
 #include <exception>  // for exception
 #include <functional> // for bind, function, _1
@@ -39,6 +40,8 @@ bool bufferBadInputs::update_bad_inputs_callback(nlohmann::json& json) {
     uint8_t* host_mask = (uint8_t*)out_buf->wait_for_empty_frame(unique_name, frame_id);
     // Reset the mask (1 == good)
     std::memset(host_mask, 1U, num_elements);
+    // Get the reorder mapping
+    constexpr const auto reorder = get_cylinder_to_beamformer_reorder_table();
 
     bool all_good = true;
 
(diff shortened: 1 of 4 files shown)


rhaas80:
> Adds a helper function in `visUtil` with the mapping from cylinder to beamformer order. `visUtil` seems like a reasonable place for this, but I'm fine with suggestions to put it elsewhere.

good place. visUtil (like this reorder) is CHIME specific. N2Util (if that exists) would be its replacement (same as visAccumulate -> N2Accumulate)

the other option would be in CHIMETelescope, but that one may be used for other chime-like (if those existed) instruments I guess

lib/stages/bufferBadInputs.cpp
```diff
@@ -39,6 +40,8 @@ bool bufferBadInputs::update_bad_inputs_callback(nlohmann::json& json) {
     uint8_t* host_mask = (uint8_t*)out_buf->wait_for_empty_frame(unique_name, frame_id);
     // Reset the mask (1 == good)
     std::memset(host_mask, 1U, num_elements);
+    // Get the reorder mapping
+    constexpr auto reorder = cylinder_to_beamformer_order();
```
rhaas80: may get_cylinder_to_beamfrmer_reorder_table(). yes, it's a mouthfull :-)

rhaas80: `constexpr const` (not all `constexpr` are `const`)

lib/utils/visUtil.hpp
```diff
@@ -463,6 +463,20 @@ void map_vis_triangle(const std::vector<uint32_t>& inputmap, size_t block, size_
 std::tuple<std::vector<uint32_t>, std::vector<input_ctype>>
 parse_reorder_default(kotekan::Config& config, const std::string base_path);
 
+/**
+ * @brief Fixed mapping from CHIE cylinder ordering to beamformer ordering.
```
rhaas80: typo CHIE -> CHIME

lib/utils/visUtil.hpp
```diff
@@ -463,6 +463,20 @@ void map_vis_triangle(const std::vector<uint32_t>& inputmap, size_t block, size_
...
+ * @return Vector of indices mapping cylinder order to beamformer oreder.
```
rhaas80: not a `std::vector` but a `std::array`

rhaas80 (approved):

lib/stages/bufferBadInputs.cpp
ljgray: Of course not, that would be too easy

PR #1466, opened 2026-04-22
Implement CHIME's 8-bit quantizer

eschnett:
This implements the actual kernel (probably – @rhaas80 to check details). Compared to the CHORD 8-bit quantizer, the main difference is that the GPU loops over a 2d array (both time and frequency), not just time.

It does not yet implement the Kotekan stage.

### DIFF
--- a/lib/cuda/cudaQuantize8.cpp
+++ b/lib/cuda/cudaQuantize8.cpp
@@ -1,3 +1,6 @@
+// 8-bit FRB beam quantizer for CHIME, using the float16 FRB beamformer and CHIME's
+// sending-to-Bonsai network format
+
 #include <cuda_fp16.h>              // for __half
 #include <cuda_runtime_api.h>       // for cudaGetLastError
 #include <algorithm>                // for max
@@ -23,6 +26,7 @@
 #include "bufferContainer.hpp"      // for bufferContainer
 #include "driver_types.h"           // for cudaEvent_t, CUevent_st, CUstream_st, cudaStream_t
 #include "fmt.hpp"                  // for format
+#include "metadata.hpp"             // for metadataObject
 
 using kotekan::bufferContainer;
 using kotekan::Config;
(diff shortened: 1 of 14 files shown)


eschnett:
All quantizers are implemented and have unit tests which they pass.

lib/cuda/cudaQuantize8.cpp
```diff
@@ -66,8 +69,8 @@ cudaQuantize8::cudaQuantize8(Config& config, const std::string& unique_name,
     _num_beams(config.get<std::int64_t>(unique_name, "num_beams")),
     _num_frequencies(config.get<std::int64_t>(unique_name, "num_frequencies")),
     _num_times(config.get<std::int64_t>(unique_name, "num_times")),
-    _num_chunks(
-        kotekan::div_noremainder(1LL * _num_beams * _num_frequencies * _num_times, CHUNK_SIZE)),
+    _num_chunks(1LL * _num_beams * _num_frequencies
+                * kotekan::div_noremainder(_num_times, CHUNK_SIZE)),
```
rhaas80: the PR comment says that the kotekan stage is not implemented so I am ingnoring all strangeness here (eg having a chunksize only in time)

lib/cuda/testQuantizeKernel8Chime.cpp
```diff
@@ -0,0 +1,421 @@
...
+            for (int freq_outer = 0; freq_outer < out_nfreqs_outer; ++freq_outer)
+                for (int time_outer = 0; time_outer < out_ntimes_outer; ++time_outer)
+                    for (int beam_packet = 0; beam_packet < out_nbeams_packet; ++beam_packet)
+                        for (int freq_packet = 0; freq_packet < out_nfreqs_packet; ++freq_packet)
+                            for (int freq_chunk = 0; freq_chunk < out_nfreqs_chunk; ++freq_chunk)
+                                for (int time_chunk = 0; time_chunk < out_ntimes_chunk;
+                                     ++time_chunk)
+                                    beamsat(beam_outer, freq_outer, time_outer, beam_packet,
+                                            freq_packet, freq_chunk, time_chunk) = 0;
+        if (std::find(beams.begin(), beams.end(), 1) != beams.end())
```
rhaas80, line 152: does not (edited) catch the issue where the 8-nested loops above set an element to 0 multiple times.  A better test would be to fill with `0` the increment by `1` each loop, ,then check that all outputs are `1` afterwards

rhaas80 (commented):
it's not obviously wrong, though the indexing is complicated enough that I will try and check against my own (independent) implementation in the testing routines

lib/cuda/cudaQuantize8.cpp
eschnett: The kotekan stage _is_ implemented. Probably buggy, if you are right. I only ran self-tests of the kernels, not with the stages.

The respective stages are called `cudaQuantize4`, `cudaQuantize8`, and `cudaQuantize8Chime`.

rhaas80:
> it's not obviously wrong, though the indexing is complicated enough that I will try and check against my own (independent) implementation in the testing routines

I looked over the indexing (in `testQuantizekernel8Chime.cpp` so far only, admittedly) and it matches what I had coded up myself, which lends some confidence about not having too many typos (due to independent implementations). The test code is, up to obvious differences  due to float16 vs float32 (and differences in the amount of information output when a comparison fails), the same for `testQuantizekernel8.cpp` so that one is likely also fine by the same argument.

rhaas80 (approved):
leftover issues in kotekan stage to be addressed by @rhaas80 while testing

PR #1468, opened 2026-04-24
Update list of extra N2 freqs to write

ljgray:
The upgrade node has a different set of frequencies now due to a frequency remap. So, we should write out these new frequencies on the production system as well

### DIFF
--- a/config/chime_science_run_recv.yaml
+++ b/config/chime_science_run_recv.yaml
@@ -124,7 +124,8 @@ vis_buffer:
 # Subset of good frequencies to output whole for monitoring
 mfreq: &mfreq [107, 344, 619, 938]
 # Subset of frequencies to save to a separate acquisition for comparing to upgrade test node
-extra_mfreq: &extra_mfreq [269, 277, 285, 293, 301, 477, 485, 493, 541, 717, 749, 901, 957, 981, 989, 1005]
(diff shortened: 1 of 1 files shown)


config/chime_science_run_recv.yaml
```diff
@@ -124,7 +124,8 @@ vis_buffer:
 # Subset of good frequencies to output whole for monitoring
 mfreq: &mfreq [107, 344, 619, 938]
 # Subset of frequencies to save to a separate acquisition for comparing to upgrade test node
-extra_mfreq: &extra_mfreq [269, 277, 285, 293, 301, 477, 485, 493, 541, 717, 749, 901, 957, 981, 989, 1005]
+# extra_mfreq: &extra_mfreq [269, 277, 285, 293, 301, 477, 485, 493, 541, 717, 749, 901, 957, 981, 989, 1005]
+extra_mfreq: &extra_mfreq [666, 698, 698, 730, 738, 826, 898, 906, 978, 58, 66, 818, 962, 234, 746, 802, 954]
```
tristpinsm: confirmed this list corresponds to the frequency ids present in `upgrade-gpu1:/home/rhaas/shared/tristpinsm/20260424T231220Z_chime_corr`.

However, `698` is listed twice.

Should we also add a comment to explain that the commented out line was used from 4 March 2026 until the remap ?

tristpinsm (changes requested):
There's a duplicate entry.

Also, I suggest we add a comment to explain that the commented out line was used between March 4 2026 and the date of the remap.

ketiltrout (approved):

PR #1469, opened 2026-04-28
Various corrections to FRB beam reformer

eschnett:
- Correct `fpga_seq_num` in output
- Correct weight matrix

### DIFF
--- a/lib/stages/calcFRB2Weights.cpp
+++ b/lib/stages/calcFRB2Weights.cpp
@@ -216,32 +216,11 @@ public:
 
         // Set W2
         {
-            // Kendrick's FRB beamforming note, equation 7:
-            //   theta = M (nhat ⋅ sigma) / lambda
-            // where nhat is the unit vector in the direction of the sky location
-            // sigma is the dish displacement in meters East-West.
-            //
-            // We'll assume that the dishes are pointed along the meridian, so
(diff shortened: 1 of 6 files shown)


lib/stages/calcFRB2Weights.cpp
```diff
@@ -253,14 +232,23 @@ class calcFRB2Weights : public kotekan::Stage {
...
+            // TODO: Take these from the telescope object
+            const float sigmax_x = 6.3;
+            const float sigmax_y = 0;
+            const float sigmax_z = 0;
+
+            const float sigmay_x = 0;
+            const float sigmay_y = 8.5;
+            const float sigmay_z = 0;
+
 #pragma omp parallel
```
geoffryan, line 252: Should there be some logic here for setting the number of threads that get thrown at this (like a config parameter)? Or are we just trusting the cpu_affinity in the config?

geoffryan (approved):
One question about setting the parallelization logic, otherwise LGTM!

lib/stages/calcFRB2Weights.cpp
eschnett: Actually, this parallelization is probably not working at all. This will need to be fixed in the future. I don't see any speedup, and the kernel is busy for a few minutes at startup while all consumers are waiting.

PR #1470, opened 2026-04-28
Update Julia-generated GPU kernels

eschnett:
- Choose an explicit compute capability version
- Choose an explicit PTX version
- Rename output directory (don't call it "A40")
- Remove Smallfinder kernels (not used any more)


julia/bin/bb.sh
```diff
@@ -8,20 +8,17 @@ set -euxo pipefail
 scriptdir=$(dirname "$0")
 cd "$scriptdir/.."
 
-card='A40'
```
jbmertens, line 11: Might we want files for other cards eventually? Or will this also work with L40S, etc, given the other changes I see below?

julia/kernels/upchan.jl
```diff
@@ -1080,47 +1142,47 @@ function upchan!(emitter)
...
-                    #TODO         emitter,
-                    #TODO         :E2,
-                    #TODO         [:E2, Symbol(:E_ringbuf_m, m), Symbol(:W_m, m)],
-                    #TODO         (E2, E, W) -> :(muladd($(isodd(m) ? :(-$W) : :(+$W)), $E, $E2)),
-                    #TODO     )
-                    #TODO end
+                    # # m = M-1
+                    # split!(emitter, [Symbol(:W_m, m) for m in 0:(M - 1)], :Wpfb, Register(:mtap, 1, M))
+                    # apply!(emitter, :E2, [:E, Symbol(:W_m, M - 1)], (E, W) -> :($(isodd(M - 1) ? :(-$W) : :(+$W)) * $E))
+                    # # m ∈ 0:M-2
```
jbmertens, line 1148: Commented comments

lib/cuda/CMakeLists.txt
```diff
@@ -81,13 +80,6 @@ if(NOT ${IWYU})
                 generated/cudaFRBBeamformer_pathfinder_U16.cpp
                 generated/cudaFRBBeamformer_pathfinder_U32.cpp
                 generated/cudaFRBBeamformer_pathfinder_U64.cpp
-                generated/cudaFRBBeamformer_smallfinder_U1.cpp
```
jbmertens, line 84: No more smallfinder :(

jbmertens (approved):
Looks good, trusting the julia without reading it.

julia/bin/bb.sh
eschnett: No. Each kernel is designed for one instrument which runs on one hardware. The mapping to "cards" now happens further downstream, in the instrument-specific setup.

julia/kernels/upchan.jl
eschnett: Yes, yes...

lib/cuda/CMakeLists.txt
eschnett: We never used it.

PR #1472, opened 2026-05-06
Cleaning EigenN2Iter stage

jbmertens:


lib/stages/EigenN2Iter.cpp
```diff
@@ -203,86 +202,79 @@ void EigenN2Iter::main_thread() {
...
 
-                // "hack" consistent with behavior below, negative RMS indicates failure to converge
-                failed_frame.erms = -std::numeric_limits<float>::max();
-                failed_frame.emethod = N2EigenMethod::none; // No method used since we failed
+        N2FrameView dest_frame(dest_buf, dest_frame_id);
+        // Copy over data, but skip all ev members which we'll set explicitly below
+        dest_frame.copy_data(input_frame, {N2Field::eval, N2Field::evec, N2Field::erms});
+
+        if (failed) {
+            // HACK: negative RMS indicates a failed/non-converged calculation.
```
rhaas80, line 226: hmm

rhaas80: isthis the same hack already used say in chime? if not, maybe ask Andre what chime is using

lib/stages/EigenN2Iter.cpp
```diff
@@ -203,86 +202,79 @@ void EigenN2Iter::main_thread() {
...
+                dest_frame.eval[i] = evals[indr];
 
-            for (uint32_t j = 0; j < num_elements; j++) {
-                output_frame.evec[i * num_elements + j] = evecs(j, indr);
+                for (uint32_t j = 0; j < num_elements; j++) {
+                    dest_frame.evec[i * num_elements + j] = evecs(j, indr);
+                }
             }
+            // HACK: return the convergence state in the RMS field (negative == not
+            // converged)
```
rhaas80, line 270: hmm

rhaas80 (changes requested):
there should be no hacks.

lib/stages/EigenN2Iter.cpp
jbmertens: This code was coped over in the interest of reproducing/preserving what CHIME currently does. If I remember correctly, production uses the 2021.11_hotfix branch along with this stage,
https://github.com/kotekan/kotekan/blob/2021.11_hotfix/lib/stages/EigenVisIter.cpp#L186

This is a "hack" in the sense that negative RMS is used to indicate the solver failed, since there was no dedicated flag, so more of an obfuscated convention. I'm not sure if/how this is used downstream.

I added N2EigenMethod::failed_iterative, which I hope is a more clear way of indicating any eigeninfo should not be considered valid. I suppose we could add another flag to differentiate if the solver failed with an error, just failed to converge, or succeeded.

rhaas80 (approved):
let the record show that I am told that the hack existed before :-)

PR #1474, opened 2026-05-07
Tracking FPGA configs

jbmertens:
Add config tracker support for the FPGA.

Differentiate between local and upstream configs to avoid naming conflicts.


lib/core/kotekanMode.cpp
```diff
@@ -154,29 +111,40 @@ void kotekanMode::initalize_stages() {
...
+                                                 get_git_commit_hash(), get_cmake_build_options());
     } catch (const std::exception& e) {
-        ERROR_NON_OO("Failed to insert local config into ConfigTracker: {:s}", e.what());
+        FATAL_ERROR_NON_OO("Failed to set local config in ConfigTracker: {:s}", e.what());
+    }
+
+    // Optionally fetch the upstream FPGA controller's config + timing. If the
+    // block is present, both endpoints must succeed at startup (otherwise
+    // ConfigTracker FATALs).
+    if (config.exists("/", "fpga_controller")) {
```
andrerenard: I'm not sure if it makes sense to do this here with a new config block.  The telescope object already has the option for FPGA parameters.  So this is a bit of duplication, and also only applied to kotekan configs which interact with an F-engine - which while the most common case, is not the only way to run kotekan. 

Could we consider moving this logic into the telescope object to live alongside the frame0 time fetch logic?

andrerenard (changes requested):
Could we also update the configurations that use this new logic?  e.g. the CHORD Pathfinder configs?  It would be worth just running this in something like a production configuration to make sure it's working to first order.

lib/core/kotekanMode.cpp
jbmertens: Hm. Having this in the telescope introduces inter-dependency between the configTracker and Telescope objects. The telescope and configTracker might both independently want the fpga information (or not), so I'd prefer to keep the top level config block and change the telescope object refer to that. (@geoffryan, any opinion?)

I think the config tracker itself is the right place for storing information, at least, since there is some information at that endpoint beyond the frame0 time,
`{"start_ctime": 1778471155.3895762, "frame0_time": [2006, 9, 25, 3, 47, 16, 2.88], "frame0_ctime": 1159156036.0000029, "frame0_nano": 1159156036000002880}`

geoffryan: The Telescope is first one in and the last one to shut the door, right?  It's guaranteed to exist once the configTracker comes up.  If we wanted the Telescope to be the sole entity responsible for talking to `fpga_master`, we could just move the `fetchAndRegisterFpgaTracking` logic to the Telescope, it'll just save the json blob and configTracker can request it when it wants.

geoffryan: ~That wouldn't be too big a change, and would remove the duplication.~  This would be a *little* annoying because the specific F-engine comms logic is in the Telescope sub-classes. So we'd need to add a virtual function `json& Telescope::get_fpga_config()` and then implement it for ICE, CHIME (?), and CHORD Telescopes

geoffryan: That being said, this particular duplication in the config doesn't bug me.  It's difficult to refer to the insides of other config blocks I think, so to have the configTracker and Telescope share we'd just need `fpga_host` etc be at the root level config.

jbmertens: The telescope and configTracker (and other things like the factories) are all independent, I think, so initialization order doesn't (and shouldn't) matter much. Kotekan could be used for things that aren't telescopes 🤷

I don't have a strong preference, although it does seem "correct" to keep the telescope and configTracker independent.

lib/dpdk/crs16BoardCaptureWorker.hpp
```diff
@@ -72,6 +72,14 @@ class crs16BoardCaptureWorker : public dpdkRXhandler {
     const kotekan::FrameInfo* active_f0 = nullptr;
     const kotekan::FrameInfo* active_f1 = nullptr;
 
+    /// Sample-based FPGA seq monotonicity check. At each sample the seq must
```
jbmertens, line 75: @andrerenard - looking for feedback about this too.

config/chime_science_run_gpu_no_output.yaml
geoffryan, line 1: I don't know if this config should be touched?  It cannot run with this version of kotekan anyways (uses hsa), so no need to keep it up to date. Better to keep it as a record of what CHIME did.

lib/core/kotekanMode.cpp
```diff
@@ -154,29 +111,63 @@ void kotekanMode::initalize_stages() {
...
-        // loopback only if no external IPv4 is available (e.g. local tests).
-        std::string host = restServer::instance().bind_address();
-        if (host == "0.0.0.0" || host == "localhost") {
-            std::string resolved = first_non_loopback_ipv4();
-            host = resolved.empty() ? std::string("127.0.0.1") : resolved;
+    // ConfigTracker setup. The optional /config_tracker block (an object)
+    // configures the tracker; defaults apply when absent.
+    const std::string ct_path = "/config_tracker";
+    bool ct_enabled = true;
+    if (config.exists("/", "config_tracker")) {
```
andrerenard, line 105: Can we make `ct_enabled` iff `config.exists("/", "config_tracker")`?

jbmertens: The default is off now, if this block doesn't exist,

lib/stages/bufferSend.cpp
```diff
@@ -64,7 +64,12 @@ bufferSend::bufferSend(Config& config, const std::string& unique_name,
 
     socket_fd = -1;
 
-    use_config_tracker = config.get_default<bool>(unique_name, "use_config_tracker", true);
+    // The stage-local use_config_tracker overrides the global default; if it
+    // isn't set, fall back to /config_tracker/enabled (default true).
+    const bool ct_default = config.exists("/", "config_tracker")
+                                ? config.get_default<bool>("/config_tracker", "enabled", true)
+                                : true;
```
andrerenard: This feels like it should be a function of the config tracker so we don't have to hardcode this config structure it possibly many places.

lib/stages/bufferRecv.cpp
```diff
@@ -52,12 +52,18 @@ REGISTER_KOTEKAN_STAGE(bufferRecv);
...
     dropped_frame_counter(
         Metrics::instance().add_counter("kotekan_buffer_recv_dropped_frame_total", unique_name)),
     transfer_time_seconds(Metrics::instance().add_gauge("kotekan_buffer_recv_transfer_time_seconds",
                                                         unique_name, {"source"})) {
 
+    // The stage-local use_config_tracker overrides the global default; if it
+    // isn't set, fall back to /config_tracker/enabled (default true).
+    const bool ct_default = config.exists("/", "config_tracker")
+                                ? config.get_default<bool>("/config_tracker", "enabled", true)
+                                : true;
```
andrerenard: Same here.

lib/utils/CHIMETelescope.cpp
```diff
@@ -27,12 +27,25 @@ CHIMETelescope::CHIMETelescope(const kotekan::Config& config, const std::string&
...
     bool require_gps = config.get_default<bool>(path, "require_gps", true);
     _query_gps = config.get_default<bool>(path, "query_gps", false);
-    _gps_host = config.get_default<std::string>(path, "gps_host", "10.1.13.1");
-    _gps_port = config.get_default<uint32_t>(path, "gps_port", 54321);
-    _gps_endpoint = config.get_default<std::string>(path, "gps_endpoint", "/get-frame0-time");
+    std::string default_gps_endpoint = "/get-frame0-time";
+    if (config.exists(path, "gps_host_info")) {
+        const std::string ref = config.get<std::string>(path, "gps_host_info");
+        _gps_host = config.get<std::string>(ref, "host");
+        _gps_port = config.get<uint32_t>(ref, "port");
```
andrerenard: This could have a default port, since fpga_master (almost) always uses that same port.

lib/utils/CHIMETelescope.cpp
```diff
@@ -27,12 +27,25 @@ CHIMETelescope::CHIMETelescope(const kotekan::Config& config, const std::string&
...
+    // gps_endpoint). Required when query_gps is true.
     bool require_gps = config.get_default<bool>(path, "require_gps", true);
     _query_gps = config.get_default<bool>(path, "query_gps", false);
-    _gps_host = config.get_default<std::string>(path, "gps_host", "10.1.13.1");
-    _gps_port = config.get_default<uint32_t>(path, "gps_port", 54321);
-    _gps_endpoint = config.get_default<std::string>(path, "gps_endpoint", "/get-frame0-time");
+    std::string default_gps_endpoint = "/get-frame0-time";
+    if (config.exists(path, "gps_host_info")) {
+        const std::string ref = config.get<std::string>(path, "gps_host_info");
+        _gps_host = config.get<std::string>(ref, "host");
```
andrerenard: You could keep default IP for CHIME.

lib/utils/CHORDTelescope.cpp
```diff
@@ -314,9 +314,22 @@ GPSTimeParams GPSTimeParams::from_config(const kotekan::Config& config, const st
...
-    gps.gps_port = config.get_default<uint32_t>(path, "gps_port", 54321);
-    gps.gps_endpoint = config.get_default<std::string>(path, "gps_endpoint", "/get-frame0-time");
+    // gps_host_info names a sibling block (typically /fpga_controller)
+    // holding ``host``, ``port``, and optionally ``timing_endpoint`` (the
+    // default for gps_endpoint). Required when query_gps is true.
+    std::string default_gps_endpoint = "/get-frame0-time";
+    if (config.exists(path, "gps_host_info")) {
+        const std::string ref = config.get<std::string>(path, "gps_host_info");
+        gps.gps_host = config.get<std::string>(ref, "host");
+        gps.gps_port = config.get<uint32_t>(ref, "port");
```
andrerenard: Port can have a default.

lib/utils/ICETelescope.cpp
```diff
@@ -28,9 +28,22 @@ ICETelescope::ICETelescope(const kotekan::Config& config, const std::string& pat
...
-    _gps_port = config.get_default<uint32_t>(path, "gps_host", 54321);
-    _gps_endpoint = config.get_default<std::string>(path, "gps_endpoint", "/get-frame0-time");
+    // gps_host_info names a sibling block (typically /fpga_controller)
+    // holding ``host``, ``port``, and optionally ``timing_endpoint`` (the
+    // default for gps_endpoint). Required when query_gps is true.
+    std::string default_gps_endpoint = "/get-frame0-time";
+    if (config.exists(path, "gps_host_info")) {
+        const std::string ref = config.get<std::string>(path, "gps_host_info");
+        _gps_host = config.get<std::string>(ref, "host");
+        _gps_port = config.get<uint32_t>(ref, "port");
```
andrerenard: Same.

python/kotekan/runner.py
```diff
@@ -104,8 +105,49 @@ def __init__(
...
+        path from ``run()``.
+        """
+        # The gdb path uses subprocess.run(), which returns a CompletedProcess
+        # with no poll()/terminate(); there is nothing to clean up there.
+        if p is None or not hasattr(p, "poll"):
+            return
+        if p.poll() is None:
+            p.terminate()
+            try:
+                p.wait(timeout=10)
```
andrerenard, line 135: Is this enough time?

andrerenard (changes requested):
Minor stuff.

andrerenard (approved):
Approved once minor comments addressed.

PR #1477, opened 2026-05-08
n2accum test fix

jbmertens:


geoffryan (approved):
Can't believe I got pwned by claude

PR #1478, opened 2026-05-08
Update config  & code to run on site

jbmertens:


config/chord_pathfinder.j2
```diff
@@ -602,6 +733,37 @@ run_send_rfi_RFImask:
...
+          ring_buffer_size: buffer_depth * input_size
+          in_buf: host_pl_mask_buffer_in
+          signal_buf: host_pl_mask_ringbuffer
+          gpu_mem_output: pl_mask_buffer
+    gpu_1:
+        kotekan_stage: cudaProcess
+        gpu_id: 1
+        cpu_affinity: [25, 26, 27]
+        in_buffers:
+            host_pl_mask_buffer_in: host_pl_mask_buffer_1
```
jbmertens, line 793: Not familiar with the details of gpu naming on multiple gpus, could these use _1's?

jbmertens (commented):
I can't actually approve this PR since I made it. But @geoffryan you can approve it! 🚀

config/chord_pathfinder.j2
geoffryan: Which fields?  The name resolution is a little confusing, but I *believe* this is the right thing to do.  The "_1"s need to be on the RHS of the keys in `in_buffers` and `out_buffers`.  The LHS's are set by the values further down in `commands` list (here `in_buf` and `signal_buf`)

geoffryan:
This shouldn't be merged yet, it doesn't run.

geoffryan:
The test modular configs in `config/tests/chord_pathfinder_*.j2` run, and the main `chord_pathfinder.j2` config has been updated to match.  We can try running it on site, and this can probably be merged.

geoffryan (approved):
LGTM

PR #1483, opened 2026-05-11
Optimize N2Accumulate:

eschnett:
- move some expensive metadata queries outside of loops
- move `if` statement outside of loop kernel, duplicating code (but what wouldn't we do for efficiency?)
- simplify the OpenMP parallelization scheme


rhaas80:
is it actually faster? And was the cause for slowdown identified? Ie if the `if` statement could stay in the loop, that would be helpful to keep the code readable. Moving the `std::vector` allocating calls in `chordMetadata` out of the loop is much less disruptive to reading flow.

lib/stages/N2Accumulate.cpp
```diff
@@ -1145,11 +1176,11 @@ bool N2Accumulate::output_and_reset(frameID& in_frame_id, frameID& in_rfiframema
 #ifdef WITH_OMP
     [[maybe_unused]] double prof_out_end_time = omp_get_wtime();
 
-    DEBUG("Outputting {:d} frames took {:f} ms\n    setup: {:f} ms\n    work:  {:f} ms\n    free:  "
-          "{:f} ms\n    fill:  {:f} ms",
-          _num_freq_per_n2k_frame, 1000 * (prof_out_end_time - prof_out_start_time),
-          1000 * prof_out_setup_time, 1000 * prof_out_work_time, 1000 * prof_out_free_time,
-          1000 * (prof_out_end_time - prof_out_fill_time));
+    INFO("Outputting {:d} frames took {:f} ms\n    setup: {:f} ms\n    work:  {:f} ms\n    free:  "
```
rhaas80: does this really have to be output for "regular" runs (be `INFO` level)

geoffryan:
I'd be curious to see the timing outputs of `configs/test/thrash_n2accum_chime.yaml`, if this is substantially faster happy to include it but most of this stage is memory-bound, I suspect those if statements it the output loop are not too bad.

Happy to change the default on the output batch size but I don't know if removing it entirely is the right call, in pathfinder we'll have to wait for 384 buffer frames to be available at once.

rhaas80:
I note: fails tests.

lib/stages/N2Accumulate.cpp
eschnett: Done.

eschnett:
@geoffryan By this time I moved the loop kernels around twice, first adding, then removing code duplication. Please check whether this is still correct! I might have overlooked a difference between the two branches.

eschnett:
@geoffryan There is a failing test. Can you have a brief look at the test and see whether its parameter settings are outdated or whether there is a bug that I introduced? I do not know the test; if you do, it might be faster for you to have a look.

geoffryan:
@eschnett The failing test is the N2Accumulate test. The parameters should not have to change, since the update should produce identical output.  The failure is on a timeout, kotekan is not completing the test run in 60s (it should take two to three seconds at most).  

The variance-calculating lambdas look correct.  The output is checked in the pytest, so if you can get the test passing it should be correct!

I'd appreciate if you could run the `thrash_n2accum_chime` config as well to make sure there aren't any performance regressions.  On a debug build with 2 workers on the chord branch I get:
```
INFO: /N2_accumulate: Adding input frame pair took 73.921204 ms + 0.055729 ms idle
...
INFO: /N2_accumulate: Outputting 16 frames took 78.717312 ms
    setup: 0.035081 ms
    work:  64.422127 ms
    free:  0.007096 ms
    fill:  14.241168 ms
```
and, more importantly, on a release build (still with 2 workers):
```
INFO: /N2_accumulate: Adding input frame pair took 24.378178 ms + 0.042449 ms idle
...
INFO: /N2_accumulate: Outputting 16 frames took 86.153193 ms
    setup: 0.011159 ms
    work:  66.022617 ms
    free:  0.006646 ms
    fill:  20.107485 ms
```

geoffryan:
@eschnett  #1487 Adds an option so the profiling timings can be INFO, so its easier to run the thrash test on a release build.

tests/test_n2_accumulate.py
```diff
@@ -6,7 +6,7 @@
 import kotekan.telescope as tel
 
 prod_config = {
-    "buffer_depth": 5,
```
rhaas80, line 9: if the CUDA code expects at least 12, does it check (and abort if not found)?

eschnett: There is no CUDA code, this is just C++. No, it does not check, it just waits until the buffers are available. Hence the timeout in the previous test runs.

eschnett:
The doc build failure is unrelated.

tests/test_n2_accumulate.py
geoffryan: This test has been running fine with a buffer depth of 5, what changed to require the increase?

tests/test_n2_accumulate.py
```diff
@@ -6,7 +6,7 @@
...
+    "buffer_depth": 12,
```
geoffryan: You should not have to change the total buffer depth for this test (and i'd rather it stay small so the test covers buffer wrap-arounds).

The issue is that the output buffer in the `accum_data` fixture below only gets initialized with a `buffer_depth` of frames.  You'll need to modify the `DumpN2Buffer` class in `python/kotekan/runner.py` to accept a number of frames (or a multiplier), and pass in the number of frequencies when the buffer gets created.

geoffryan (changes requested):
- Python test needs to be modified
- Needs timing output of `thrash_n2accum_chime` run

tests/test_n2_accumulate.py
eschnett: Done.

eschnett:
Benchmark results for CHIME running on upgrade-gpu1.

`num_workers = 4`, all other respective parameters at their default.

OLD:
```
INFO: /run_N2_accumulate: Outputting 16 frames took 224.158003 ms
    setup: 0.028534 ms
    work:  212.644498 ms
    free:  0.090382 ms
    fill:  11.388219 ms
INFO: /run_N2_accumulate: Adding input frame pair took 253.077662 ms + 74.616246 ms idle
INFO: /run_N2_accumulate: Adding input frame pair took 27.069552 ms + 0.104152 ms idle
INFO: /run_N2_accumulate: Adding input frame pair took 26.923150 ms + 0.080996 ms idle
INFO: /run_N2_accumulate: Adding input frame pair took 27.054733 ms + 0.072831 ms idle
```

NEW (this PR):
```
INFO: /run_N2_accumulate: Outputting 16 frames took 79.746518 ms
    setup: 0.021822 ms
    work:  68.612916 ms
    free:  0.009532 ms
    fill:  11.091718 ms
INFO: /run_N2_accumulate: Adding input frame pair took 110.929828 ms + 97.077983 ms idle
INFO: /run_N2_accumulate: Adding input frame pair took 25.421311 ms + 0.058276 ms idle
INFO: /run_N2_accumulate: Adding input frame pair took 25.498711 ms + 0.069207 ms idle
INFO: /run_N2_accumulate: Adding input frame pair took 25.534654 ms + 0.060510 ms idle
```

geoffryan (approved):
LGTM!

geoffryan:
timing looks great!  Thanks!

PR #1484, opened 2026-05-12
More metadata in N2 files

geoffryan:
Adds more metadata fields to the N2 files:
- frame0 time (as an attribute)
- instrument time at time_center
- all EOP data at bin center (where it is relevant)


lib/stages/hdf5N2Write.cpp
```diff
@@ -716,6 +735,18 @@ N2FileData::AddFrameStatus N2FileData::add_frame(const N2FrameView& fv, size_t t
     auto ns_close = [](int64_t a, int64_t b, int64_t tol_ns = 2) {
         return std::llabs(a - b) <= tol_ns;
     };
+    // Accept timing differences up to 2 ns (e.g. fuzz on EOP table updates)
```
jbmertens, line 739: May be helpful to put this in `timeUtil.hpp` or somewhere else. But not necessary.

lib/stages/hdf5N2Write.cpp
```diff
@@ -757,13 +788,38 @@ N2FileData::AddFrameStatus N2FileData::add_frame(const N2FrameView& fv, size_t t
...
+        add_failure(fmt::format("bin_t_inst_ns[t={}] mismatch: stored {} != incoming {} (ns)",
+                                t_index, bin_t_inst_ns[t_index], fv.bin_eop.t_inst_ns));
+    if (bin_ut1_ns[t_index] > 0 && !ns_close(bin_ut1_ns[t_index], fv.bin_eop.t_ut1_ns))
+        add_failure(fmt::format("bin_ut1_ns[t={}] mismatch: stored {} != incoming {} (ns)", t_index,
+                                bin_ut1_ns[t_index], fv.bin_eop.t_ut1_ns));
+    if (bin_delta_ut1_inst[t_index] > 0
+        && !sec_close(bin_delta_ut1_inst[t_index], fv.bin_eop.delta_UT1_inst))
+        add_failure(fmt::format("bin_delta_ut1_inst[t={}] mismatch: stored {} != incoming {} (deg)",
+                                t_index, bin_delta_ut1_inst[t_index], fv.bin_eop.delta_UT1_inst));
+    if (bin_era_deg[t_index] > 0 && !deg_close(bin_era_deg[t_index], fv.bin_eop.ERA_deg))
```
jbmertens: Technically `bin_era_deg` could actually be zero, right?
And maybe `bin_delta_ut1_inst`?

lib/stages/hdf5N2Write.hpp
```diff
@@ -121,8 +121,14 @@ class N2FileData {
     // t-dependent metadata
     std::vector<uint64_t> fpga_start_tick;           // (t)
     std::vector<uint64_t> frame_length_fpga_ticks;   // (t)
-    std::vector<int64_t> time_center_ut1;            // (t)
-    std::vector<int64_t> bin_ut1;                    // (t)
+    std::vector<int64_t> time_center_t_inst_ns;      // (t)
+    std::vector<int64_t> time_center_ut1_ns;         // (t)
+    std::vector<int64_t> bin_t_inst_ns;              // (t)
+    std::vector<int64_t> bin_ut1_ns;                 // (t)
```
jbmertens, line 127: uint?

jbmertens (approved):
Looks good! Couple small nitpicks that don't actually need to be fixed.

lib/stages/hdf5N2Write.cpp
geoffryan: Oh, that's a copy/paste mistake.  Good catch.

lib/stages/hdf5N2Write.cpp
geoffryan: These should all be bit-identical in operation.  I'd rather not let `timeUtil.hpp` have an idea of "slop".

geoffryan: But just for safety having the checks here with some wiggle room makes sense.

lib/stages/hdf5N2Write.hpp
geoffryan: They're `int64_t` in the EOP definition.

geoffryan: You can in principle get intermediate negative values when manipulating these (e.g. when converting) so its easier if everything is sign-ed the whole way.

PR #1486, opened 2026-05-13
Correct SASS code generation

eschnett:
- Correct the Julia code that outputs SASS. Previously the kernel was cached and thus there was no SASS output.
- Regenerate the kernels, including the SASS code.


jbmertens (approved):
Image

PR #1487, opened 2026-05-13
N2Accum optionally prints profiling to INFO

geoffryan:
Just an option to emit N2Accumulate's profiling timing as INFO instead of DEBUG. Defaults to false, useful for testing performance on release builds.


jbmertens (approved):
You mean you don't want to look at all the other debug output? 😜

PR #1489, opened 2026-05-13
Reduce lcore startup delay in the CHORD Pathfinder config.

andrerenard:
This fixes an issue that got added in the CHIME support changes.   CHIME has many more NICs so the startup time needed before starting the lcores is much longer.  However in the CHORD case, we only have 2 NICs so the startup is faster.  

Having said that, I think I understand better why some of the synchronization delays are needed - and I have some ideas to reduce them with better startup handling.   I will implement those changes in a separate PR.


geoffryan (approved):
LGTM

jbmertens (approved):

PR #1490, opened 2026-05-13
Rhaas/updates

rhaas80:
* 519e9e1b - test_parseReorderDefault: test mapping to beamformer order
* 4076fb21 - parseReorderDefault: simplify logic, correct use of reorder table
* a96bdf37 - get_cylinder_to_beamformer_reorder_table: be more verbose describing it
* 2e554780 - parseReorderDefault: add a bit of type-safety
* 14698197 - parseReorderDefault: generalize input and output ordering
* d1c5536d - calcFRB2Weights: remove dependency on CHORDTelescope
* eb5494ca - Config: work around unsigned(-1) == unsigned_max issue
* bb5b11d0 - UpchannelizationSchedule: remove assumption that Telescope is CHORD
* 90510b28 - lint.sh: exempt L0_L1 struct documentation file
* 2209ec2e - FrbPacket: add from_buffer method
* 7f67daea - frbbuffer.py: add fpga0_ns field
* a34b3d47 - frb_functions: add kludge code to make pseudo-C++ compile, add missing fields
* eba37a92 - frb_functions: import packet structure declaration from CHIMEFRB project
* 2ba4193e - tx_utils: use sscanf when parsing IPv4 address strings

before I accumulate ever more private changes, here's a pull request for (hopefully) the more straightforward changes.

### DIFF
--- a/lib/core/Config.hpp
+++ b/lib/core/Config.hpp
@@ -31,10 +31,13 @@ namespace kotekan {
 
 // Define a "large" type for every arithmetic type
 template<typename Type>
-using LargeType = typename std::conditional<
-    std::is_integral<Type>::value,
-    typename std::conditional<std::is_unsigned<Type>::value, unsigned long long, long long>::type,
-    double>::type;
+using LargeType =
+    typename std::conditional<std::is_integral<Type>::value,
+                              // "long double" will have at least 64bits for the significant plus an
+                              // additional sign bit, enough to hold both int64_t and uint64_t
+                              // do not use "unsigned long long" to avoid unsigned(-1) ==
+                              // unsigned_max issues
+                              long double, double>::type;
 
 // Convert types, with checking for overflows
 template<typename Ret, typename Type>
--- a/lib/stages/UpchannelizationSchedule.cpp
+++ b/lib/stages/UpchannelizationSchedule.cpp
@@ -58,7 +58,7 @@ std::map<int, std::set<int>> UpchannelizationSchedule::make_upchan_channels_to_f
 }
 
 void UpchannelizationSchedule::output_statistics() const {
-    const auto& chord_telescope = Telescope::instance().cast<CHORDTelescope>();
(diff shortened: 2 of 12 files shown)


rhaas80:
@geoffryan this is the code I mentioned about doing transformations between different dish orderings: https://github.com/kotekan/kotekan/commit/14698197e9abcc738f3e05c8fdc55adc455e5105

lib/core/Config.hpp
```diff
@@ -31,10 +31,13 @@ namespace kotekan {
...
-    std::is_integral<Type>::value,
-    typename std::conditional<std::is_unsigned<Type>::value, unsigned long long, long long>::type,
-    double>::type;
+using LargeType =
+    typename std::conditional<std::is_integral<Type>::value,
+                              // "long double" will have at least 64bits for the significant plus an
+                              // additional sign bit, enough to hold both int64_t and uint64_t
+                              // do not use "unsigned long long" to avoid unsigned(-1) ==
+                              // unsigned_max issues
+                              long double, double>::type;
```
rhaas80, line 40: note: in 1392dcfe15217f6708f759a6ab572a2feb0dfd67 a type `__int128` was introduced, which would allow handling `int64_t` and `uint64_t` as well (but not `unsigned __int128` of course).

geoffryan: I wouldn't worry about config supporting `__int128`?  It's only used for a couple EOP calculations internally and not passed between functions.

rhaas80: admittedly I also never expected an `int64_t` for the number of dishes :-P

This is more along the lines to see if using an `__int128` is more palatable to people than a `long double` (for integers, note though that `long double` is exact for the `int64_t` range).

geoffryan: oh i see!  Could go either way.  I don't love that its a compiler extension and not in the standard we're using, but it *is* very common.

ljgray (approved):

PR #1495, opened 2026-05-14
Upchannelizers: Use "better" PFB window function

eschnett:
After discussions with Harrison Siegel, I think that the zero for the PFB window function should be located between sample points. This makes the length of the Window function the same as the length of the time span which is analyzed.


geoffryan (approved):
This looks fine, but I'm not entirely sure of the consequences of this change. We should run some offline (independent from kotekan) tests of upchannelization window functions to see what choice actually performs the best.

PR #1497, opened 2026-05-15
Kv/airspy crosscorr fix

kvand:
With Claude's help, I *finally* got around to pulling the little airspy producer forward to modern kotekan.

This is the first of 2 PRs, getting basic airspy things working, both auto-and cross-corr. The second PR will add the html / javascript viewer, all cleaned up for airspy ops; should in principle work for ARO ops as well, though that may take some downstream debugging.

These largely leave the core kotekan untouched, though definitely worth double-checking!

### DIFF
--- a/config/airspy_autocorr.yaml
+++ b/config/airspy_autocorr.yaml
@@ -27,13 +27,14 @@ one: 1
 two: 2
 sizeof_float: 4
 
-meta_pool:
-  kotekan_metadata_pool: chordMetadata
-  num_metadata_objects: 5 * buffer_depth
+# No stage in the airspy pipeline reads, writes, or passes frame metadata,
+# so the buffers below declare `metadata_pool: none` and there is no
+# metadata-pool block at all (rather than the heavyweight, CHORD-specific
+# chordMetadata pool the forward-port inherited).
 
 input_buf:
   kotekan_buffer: standard
-  metadata_pool: meta_pool
+  metadata_pool: none
   num_frames: buffer_depth
   frame_size: samples_per_data_set * bytes_per_sample
 
@@ -56,7 +57,7 @@ spectrum_output:
 
   post_fengine_buf:
     kotekan_buffer: standard
-    metadata_pool: meta_pool
+    metadata_pool: none
     num_frames: buffer_depth
     frame_size: samples_per_data_set * sizeof_float # * 2 for complex / 2 for half-size
 
@@ -69,7 +70,7 @@ spectrum_output:
 
   post_corr_buf:
     kotekan_buffer: standard
-    metadata_pool: meta_pool
+    metadata_pool: none
     num_frames: buffer_depth
     frame_size: ( spectrum_length + one ) * sizeof_float
 
--- a/config/airspy_autocorr.yaml
+++ b/config/airspy_autocorr.yaml
@@ -28,7 +28,7 @@ two: 2
 sizeof_float: 4
 
 meta_pool:
-  kotekan_metadata_pool: chimeMetadata
+  kotekan_metadata_pool: chordMetadata
   num_metadata_objects: 5 * buffer_depth
 
 input_buf:
@@ -46,6 +46,8 @@ airspy_input:
   gain_mix: 10 #0-15
   gain_if:  10 #0-15
   biast_power: false
+  autostart: true
+#  serial: 0x03A464C834658167  # uncomment & edit to pick a specific dongle
 
 spectrum_output:
   integration_length: 64
@@ -60,6 +62,8 @@ spectrum_output:
 
   fengine:
     kotekan_stage: fftwEngine
+    # airspyInput now produces real (RAW) int16 samples; use the real-input FFT.
+    input_type: real
     in_buf: input_buf
     out_buf: post_fengine_buf
 
--- a/lib/core/buffer.hpp
+++ b/lib/core/buffer.hpp
@@ -31,8 +31,6 @@
 
 #ifdef MAC_OSX
 #include "osxBindCPU.hpp"
-
-#include <immintrin.h>
 #endif
 
 /// The system page size, this might become more dynamic someday
--- a/lib/stages/AirspyAlign.cpp
+++ b/lib/stages/AirspyAlign.cpp
@@ -4,6 +4,7 @@
 #include "StageFactory.hpp"    // for REGISTER_KOTEKAN_STAGE
 #include "buffer.hpp"          // for Buffer
 #include "bufferContainer.hpp" // for bufferContainer
+#include "fftwPlannerLock.hpp" // for fftw_planner_mutex
 #include "kotekanLogging.hpp"  // for DEBUG
 
 #include <algorithm>          // for min
@@ -73,15 +74,21 @@ void AirspyAlign::start_callback(kotekan::connectionInstance& conn, bool send_co
 
     float* samplesA = (float*)fftwf_malloc(sizeof(float) * fft_len);
     fftwf_complex* spectrumA = (fftwf_complex*)fftwf_malloc(sizeof(fftwf_complex) * n_bins);
-    fftwf_plan fft_planA = fftwf_plan_dft_r2c_1d(fft_len, samplesA, spectrumA, FFTW_ESTIMATE);
-
     float* samplesB = (float*)fftwf_malloc(sizeof(float) * fft_len);
     fftwf_complex* spectrumB = (fftwf_complex*)fftwf_malloc(sizeof(fftwf_complex) * n_bins);
-    fftwf_plan fft_planB = fftwf_plan_dft_r2c_1d(fft_len, samplesB, spectrumB, FFTW_ESTIMATE);
-
     float* samplesC = (float*)fftwf_malloc(sizeof(float) * fft_len);
     fftwf_complex* spectrumC = (fftwf_complex*)fftwf_malloc(sizeof(fftwf_complex) * n_bins);
-    fftwf_plan fft_planC = fftwf_plan_dft_c2r_1d(fft_len, spectrumC, samplesC, FFTW_ESTIMATE);
+
+    // FFTW's planner is not thread-safe; this callback runs on a REST
+    // handler thread and can race other stages' plan create/destroy.
+    // Serialize plan creation (fftwf_malloc above is thread-safe).
+    fftwf_plan fft_planA, fft_planB, fft_planC;
+    {
+        std::lock_guard<std::mutex> planner_lock(fftw_planner_mutex());
+        fft_planA = fftwf_plan_dft_r2c_1d(fft_len, samplesA, spectrumA, FFTW_ESTIMATE);
+        fft_planB = fftwf_plan_dft_r2c_1d(fft_len, samplesB, spectrumB, FFTW_ESTIMATE);
+        fft_planC = fftwf_plan_dft_c2r_1d(fft_len, spectrumC, samplesC, FFTW_ESTIMATE);
+    }
 
     for (uint32_t i = 0; i < _lag_window; i++) {
         samplesA[i] = localA[i];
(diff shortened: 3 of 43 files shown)


kvand:
From what I can tell, the failing CI tests are pre-existing conditions, nothing to do with anything changed here.

jbmertens:
> From what I can tell, the failing CI tests are pre-existing conditions, nothing to do with anything changed here.

Yeah, the failing test is a non-deterministic one that is flaky. Re-running, it passes.

The linter should be run though. It doesn't lint consistently on different black/clang-format versions, despite repeated attempts to fix that, so may need to be linted from inside the docker image.

lib/core/buffer.hpp
```diff
@@ -31,8 +31,6 @@
 
 #ifdef MAC_OSX
 #include "osxBindCPU.hpp"
-
-#include <immintrin.h>
```
jbmertens, line 35: @andrerenard is this an include you want to keep?  (See also stage.hpp)

config/airspy_autocorr.yaml
```diff
@@ -28,7 +28,7 @@ two: 2
 sizeof_float: 4
 
 meta_pool:
-  kotekan_metadata_pool: chimeMetadata
+  kotekan_metadata_pool: chordMetadata
```
jbmertens: Is metadata: none a possible type?

lib/stages/CMakeLists.txt
```diff
@@ -122,6 +124,10 @@ if(LIBAIRSPY_FOUND)
     target_include_directories(kotekan_stages SYSTEM PRIVATE ${LIBAIRSPY_INCLUDE_DIR})
     target_include_directories(kotekan_stages SYSTEM PRIVATE ${FFTW_INCLUDES})
     target_link_libraries(kotekan_stages PRIVATE ${LIBAIRSPY_LIBRARIES})
+    if(FFTW_FOUND)
```
jbmertens, line 127: There is also a `-DUSE_AIRSPY` flag we may want to wrap the airspy files in.
And throw an error/warning if airspy is required, but fftw isn't found, or also `-DUSE_FFTW=OFF`.

lib/stages/fftwEngine.cpp
```diff
@@ -18,57 +20,96 @@ REGISTER_KOTEKAN_STAGE(fftwEngine);
...
+    Stage(config, unique_name, buffer_container, std::bind(&fftwEngine::main_thread, this)),
+    real_samples(nullptr), complex_samples(nullptr), spectrum(nullptr), fft_plan(nullptr) {
 
     in_buf = get_buffer("in_buf");
     in_buf->register_consumer(unique_name);
     out_buf = get_buffer("out_buf");
     out_buf->register_producer(unique_name);
 
-    spectrum_length = config.get_default<int>(unique_name, "spectrum_length", 1024);
+    _spectrum_length = config.get_default<int>(unique_name, "spectrum_length", 1024);
```
jbmertens: Should there be a default here?

lib/stages/AirspyAlign.cpp
```diff
@@ -0,0 +1,214 @@
...
+                                      std::bind(&AirspyAlign::get_correlation_callback, this, _1));
+
+    frame_inA = 0;
+    frame_inB = 0;
+    uint32_t copied = 0;
+    const uint32_t samples_per_frame = buf_inA->frame_size / sizeof(short);
+
+    // Assumes the two input buffers are aligned frame-for-frame.
+    while (!stop_thread) {
+        short* inA = (short*)buf_inA->wait_for_full_frame(unique_name, frame_inA);
```
jbmertens, line 204: In general, we may want to set/check frame descriptions for the buffers produced/consumed by this stage and others below, to validate types.

lib/stages/AirspyAlign.cpp
```diff
@@ -0,0 +1,214 @@
...
+        float norm = (float)_lag_window - i;
+        float val = fabs(samplesC[i] / norm);
+        if (val > maxval) {
+            lag = i;
+            maxval = val;
+        }
+        meanval += val;
+        corr_pos.push_back(val);
+
+        val = fabs(samplesC[fft_len - i] / norm);
```
jbmertens: `samplesC[fft_len - i - 1]` ?
(And similar below)

lib/stages/airspyInput.cpp
```diff
@@ -24,64 +32,274 @@ airspyInput::airspyInput(Config& config, const std::string& unique_name,
...
+                }
+                adcstat_cv.notify_one();
+                INFO("Airspy ADC mean: {:f}, RMS: {:f}, rail fraction {:f}", adcmean, adcrms,
+                     adcrailfrac);
+            }
+
+            // RAW samples are unsigned 12-bit centred on 2048: subtract DC bias.
+            // Also flip the sign on every other sample to shift the spectrum into
+            // the first Nyquist zone (multiply by (-1)^idx).
+            for (uint32_t i = 0; i < n_samples; i += 2) {
```
jbmertens, line 316: Consider validating for even frame sizes

lib/stages/airspyInput.cpp
```diff
@@ -24,64 +32,274 @@ airspyInput::airspyInput(Config& config, const std::string& unique_name,
...
+        in = (void*)((char*)in + copy_length);
         frame_loc = (frame_loc + copy_length) % buf->frame_size;
 
         if (frame_loc == 0) {
             DEBUG("Airspy Buffer {:d} Full", frame_id);
+
+            short* fr = (short*)frame_ptr;
+            const uint32_t n_samples = buf->frame_size / BYTES_PER_SAMPLE;
+
+            // Read dump_adcstat without locking (single bool, set under lock by REST thread).
```
jbmertens: Why read without locking?

lib/stages/fftwEngine.cpp
```diff
@@ -18,57 +20,96 @@ REGISTER_KOTEKAN_STAGE(fftwEngine);
...
+            fftwf_plan_dft_1d(_spectrum_length, complex_samples, spectrum, -1, FFTW_ESTIMATE);
+    }
 }
 
 fftwEngine::~fftwEngine() {
-    fftwf_free(samples);
-    fftwf_free(spectrum);
-    fftwf_free(fft_plan);
+    if (fft_plan)
+        fftwf_destroy_plan(fft_plan);
```
jbmertens, line 71: FFTW plan creation/destruction is not thread safe. Stages creation is serial, so that is fine, but destruction here may run into issues if there are multiple instances of this stage (or other fftw usage).

lib/stages/fftwEngine.cpp
```diff
@@ -18,57 +20,96 @@ REGISTER_KOTEKAN_STAGE(fftwEngine);
...
+        } else {
+            // Complex IQ: each input sample is an int16 pair, so step by 2*_spectrum_length ints.
+            for (int j = 0; j < samples_per_input_frame / 2; j += _spectrum_length) {
+                DEBUG("Running complex FFT, {:d}", in_local[2 * j]);
+                for (int i = 0; i < _spectrum_length; i++) {
+                    complex_samples[i][0] = in_local[2 * (i + j)];
+                    complex_samples[i][1] = in_local[2 * (i + j) + 1];
+                }
+                fftwf_execute(fft_plan);
+                // Shift DC into the centre.
```
jbmertens, line 119: Worth noting this convention in a doxystring.

tests/test_sample_autocorr.py
```diff
@@ -0,0 +1,90 @@
+# === Start Python 2/3 compatibility
+from __future__ import absolute_import, division, print_function, unicode_literals
+from future.builtins import *  # noqa  pylint: disable=W0401, W0614
+from future.builtins.disabled import *  # noqa  pylint: disable=W0401, W0614
+
+# === End Python 2/3 compatibility
+
+"""SampleAutocorr is a REST-triggered one-shot stage; its compute kernel is the
```
jbmertens: "on-demand" instead of "one-shot"?

jbmertens (changes requested):
Some comments - mostly for config-writer QOL.

lib/core/buffer.hpp
kvand: This was an old osx include, I don't think it's used in modern versions -- though I'm thinking I might try to get a macos build working again once this is in, I do build a lot on my laptop, and it's a pain that nothing works there anymore.

lib/stages/CMakeLists.txt
kvand: I think strictly they're meant to be independent and toggle-able features: airspy gives that little dongle, and there's no intrinsic reason you'd need fftw with it; similarly, you might run another SDR and want the FFTW for that.

lib/stages/fftwEngine.cpp
kvand: I'm ok with no default, though I guess we probably should add one. Will set to 128 since often times people don't need much resolution.

lib/stages/fftwEngine.cpp
kvand: Right -- process-wide threadlock for fftw ops added. As you say, not necessary as used so far, but good to future-proof.

lib/stages/fftwEngine.cpp
kvand: Noted.

tests/test_sample_autocorr.py
kvand: Yup!

lib/stages/AirspyAlign.cpp
kvand: Corrected -- it should actually be a modulo to make sure everything always stays in bounds and the indices end up right.

lib/stages/airspyInput.cpp
kvand: Good catch, validation check added.

lib/stages/airspyInput.cpp
kvand: Ah, that *can* actually produce a rade condition -- harmless, but good to fix. Updated in latest commit.

lib/stages/AirspyAlign.cpp
kvand: Is that possible with 'none' type metadata? It seems like the frame descriptor stuff lives in e.g. chordMetadata?

jbmertens: Hm, good point, the `GenericNDArray` frame descriptions doesn't as play well with buffers without a metadata type.

Just using `buf->allocate_ndarray_frame_desc` should be sufficient for setting the frame description, that function also has its own checking mechanism. (A better name for it might have been `allocate_or_validate_ndarray_frame_desc` or something, but a consumer can/"should" also check that the frame description has appropriate values.)

config/airspy_autocorr.yaml
kvand: Yes, updated to use that!

lib/stages/AirspyAlign.cpp
kvand: Alright, I think this is added in, with two caveats:
1. post_corr_buf is consumer-side-only (networkPowerStream sets/records; producers don't tag it), because rawFileWrite has a hard guard against NDArray-tagged buffers (it can't round-trip the layout metadata), and the corr unit tests use rawFileWrite to capture output. Tagging buf_out in simpleAutocorr/SimpleCrosscorr would break those tests. Consumer-side still documents the contract in code, just no producer↔consumer cross-check on that one buffer.
2. The power_corr descriptor describes the heterogeneous [float32 spectra + uint32 count] × elems layout as a flat float32 array — byte_size and the spectrum dtype validate exactly, only the trailing count word's reported dtype is cosmetic. A MixedNDArray subtype could model it properly, I can try to get that into a future PR follow-up.

kvand:
Looks like the CI tests are all passing now, I think comments are all addressed -- @jbmertens, want to check back through when you've got a few minutes? Thanks! (Zero urgency here.)

lib/stages/airspyInput.hpp
```diff
@@ -33,117 +37,119 @@
...
+    int _biast_power;
+    /// Optional specific device serial-number; 0 means open any.
+    long _airspy_sn;
+    /// Optional file path to read raw samples from instead of a device.
+    std::string _airspy_fn;
+    /// Whether to begin streaming immediately on startup.
+    bool _autostart;
+
+    // ADC statistics. The REST adcstat handler requests a dump and waits on
+    // @c adcstat_cv until the producer fills @c adc{rms,mean,railfrac} on the
```
jbmertens, line 140: I see some doxygen-style comments in here, should this be a comment doxygen will process? (`///`)

jbmertens (approved):
LGTM! Just a comment on a comment.

PR #1500, opened 2026-05-20
N2 metadata update

geoffryan:
- adds `abs_time_index` field to N2 files
- refactored `ERAL` to `ERAL_deg` in metadata, N2 files, and tests.

### DIFF
--- a/lib/stages/hdf5N2Write.cpp
+++ b/lib/stages/hdf5N2Write.cpp
@@ -561,6 +561,8 @@ std::unique_ptr<HighFive::File> N2FileData::_open_or_create_file(const std::stri
                               HighFive::create_datatype<uint64_t>(), props_empty);
         _check_create_dataset(*file, "/frame_length_fpga_ticks", {num_file_t_}, {"time"},
                               HighFive::create_datatype<uint64_t>(), props_empty);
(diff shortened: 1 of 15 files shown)


lib/stages/hdf5N2Write.cpp
```diff
@@ -698,6 +700,7 @@ N2FileData::N2FileData(FileMode file_mode_, uint64_t num_file_t_, const N2FrameV
     // Additional metadata
     fpga_start_tick.assign(num_file_t, 0);
     frame_length_fpga_ticks.assign(num_file_t, 0);
+    bin_abs_index.assign(num_file_t, std::numeric_limits<uint64_t>::max());
```
jbmertens, line 703: Most other fields default to 0. Why nonzero here?

jbmertens (approved):
LGTM

lib/stages/hdf5N2Write.cpp
geoffryan: For most of the fields 0 can also serve as a sentinel value to see if the field has been written to yet.  Would like to be able to make that check, but bins _could_ have 0 index (especially in testing).  Far more unlikely to get to a bin that saturates a uint64_t ;)

PR #1501, opened 2026-05-21
Automatic PRs to lint, test failure fixes

jbmertens:
So you can merge a PR instead of linting by hand.
Also lint yamls, bump some versions.

Unrelated fixes that prevent spurious CI test failures:
 - Set n2 acccum test seed and relax tolerances slightly
 - REST server port binding is blocking (prevents race failure with config tracker)


config/tests/chime_frb_verification.yaml
geoffryan, line 1: Why this deletion?

tools/j2lint.py
```diff
@@ -0,0 +1,78 @@
...
+    # Track multi-line Jinja `{# ... #}` comment blocks so their contents
+    # are exempt from the indent rule (free-form prose, not yaml content).
+    in_comment_block = False
+    for lineno, line in enumerate(text.splitlines(), start=1):
+        if in_comment_block:
+            if "#}" in line:
+                in_comment_block = False
+            continue
+
+        stripped = line.lstrip(" ")
```
geoffryan: should this be a `lstrip()` to get all whitespace, not just spaces?

config/tests/chime_verify_upchan.yaml
```diff
@@ -161,8 +161,6 @@ compressed_lost_samples_buffer_cpy:
     frame_size: samples_per_data_set * num_data_sets / (factor_upchan * downsample_time) * sizeof_int
     metadata_pool: main_pool
 
-cpu_affinity: [2,3,8,9]
```
geoffryan, line 164: deleted on purpose?

config/frb_test.yaml
```diff
@@ -82,8 +82,6 @@ cpu_output_buffers:
     cpu_output_buffer_0:
         kotekan_buffer: standard
 
-cpu_affinity: [2,3,8,9]
```
geoffryan, line 85: see above

geoffryan (changes requested):
The PR looks mostly fine, comments below, but "requesting changes" because the auto-formatted yaml's have some problems, so the parameters here should likely be tweaked.

config/tests/chime_frb_verification.yaml
jbmertens: Duplicate yaml block (the same thing exists earlier in this file). These can't be auto-fixed, really... they'll need to be fixed by hand, but at least the lint job will fail.

jbmertens:
> The PR looks mostly fine, comments below, but "requesting changes" because the auto-formatted yaml's have some problems, so the parameters here should likely be tweaked.

Hopefully better / less intrusive now... see #1504

config/tests/testCHORDTelescope.yaml
```diff
@@ -17,51 +17,55 @@ sec_in_day: 86400
...
-  origin_itrs_lat_deg: 49.3207514444444444
-  grid_x_axis: [1.0, 0.0, 0.0]
-  grid_y_axis: [0.0, 1.0, 0.0]
-  dish_elev_axis: [1.0, 0.0, 0.0]
-  dish_vert_axis: [0.0, 0.0, 1.0]
-  dish_coelev_deg: 0.0
+    name: CHORDTelescope
+    require_gps: false
+    origin_itrs_lon_deg: -119.62081125
+    origin_itrs_lat_deg: 49.3207514444444470
```
geoffryan, line 42: Why did this value get altered?

config/tests/point_source.yaml
```diff
@@ -19,102 +19,100 @@ fft_length: 16384
...
-  require_gps: false
-  eop_updatable_config: /earth_rotation_data
-  num_dishes_x: 8
-  num_dishes_y: 12
+    name: CHORDTelescope
+    origin_itrs_lat_deg: 49.32075144444
+    origin_itrs_lon_deg: -119.62081125
+    grid_x_axis: [0.999974342398359362, -0.000037539331442772, -0.007163318767675494]
+    grid_y_axis: [0.000065403387739210, 0.999992433220348809, 0.003889630373557614]
+    dish_elev_axis: [0.99999999838132392, -0.000056897733584327 0.0]
```
geoffryan, line 27: first component of dish_elev_axis was changed.

lib/core/restServer.cpp
```diff
@@ -147,6 +147,67 @@ void restServer::start(const std::string& bind_address, u_short port) {
...
+        socklen_t len = sizeof(sin);
+        if (getsockname(sock, (struct sockaddr*)&sin, &len) == -1) {
+            ERROR_NON_OO("restServer: Failed getting socket name ({:s}:{:d})", _bind_address,
+                         _port);
+            exit(1);
+        }
+        _port = ntohs(sin.sin_port);
+    }
+    // This INFO line is parsed by the python runner to get the RESTserver port. Don't edit.
+    INFO_NON_OO("restServer: started server on address:port {:s}:{:d}", _bind_address, _port);
```
geoffryan: Should this still be in the `http_server_thread` instead of out here?

tests/test_n2_accumulate.py
```diff
@@ -1262,9 +1268,10 @@ def test_weight(accum_data, expected_accum, accum_list, setup, accum_setup):
         if accum_setup["variance_mode"] == "CHIMEv1":
             # The bias subtraction can introduce a LOT of truncation error on random
             # data, so need to set the tolerances wide (in lieu of replicating the truncation
-            # error in this test)
-            rtol = 1.0e-2
-            atol = 1.0e-2
+            # error in this test). Increased to 2e-2 because tighter bounds frequently
+            # fail on the CI runner for various seeds.
+            rtol = 2.0e-2
```
geoffryan, line 1273: If the seed is fixed could we roll these tolerances back to `1.0e-2`?

geoffryan (approved):
Quibbles discussed.  LGTM

lib/core/restServer.cpp
jbmertens: Still thinking about this one. I'll ask for feedback from Andre too.

tools/lint.sh
```diff
@@ -44,6 +44,9 @@ usage() {
                               -DCMAKE_EXPORT_COMPILE_COMMANDS=ON first
                               (include-what-you-use.org)
         - cmakelint           lint CMakeList files
+        - yamlfix_kotekan.py  ruamel.yaml round-trip: fixes indent/--- in place,
```
andrerenard, line 47: We might want to either remove or update the kotekan command line `yamllint` linting, since that's not going to produce the same output anymore.

lib/core/restServer.cpp
```diff
@@ -147,6 +147,67 @@ void restServer::start(const std::string& bind_address, u_short port) {
...
+    event_config* ev_config = event_config_new();
+    if (!ev_config) {
+        ERROR_NON_OO("Failed to create config for libevent");
+        exit(1);
+    }
+    if (event_config_avoid_method(ev_config, "select")) {
+        ERROR_NON_OO("Failed to exclude select from the libevent options");
+        exit(1);
+    }
+    event_base = event_base_new_with_config(ev_config);
```
andrerenard, line 170: Double check safe to free.

lib/core/restServer.cpp
andrerenard: Is there a case where `pytest` gets this line before the event loop is started?  Should this maybe be triggered after the event loop is running, and not just right before we start it?

andrerenard (changes requested):

tools/lint.sh
jbmertens: It is fairly permissive, so I don't think will conflict with ruamel. I'll remove the line-length spec, but I assume it's useful for making sure yaml is valid otherwise, so beyond that will keep it.

andrerenard (approved):
LGTM

PR #1505, opened 2026-05-22
Fix for blaze race condition when multiple stages use blaze

jbmertens:
Applies a small patch to blaze files. Multiple eigensolver stages should work now.


rhaas80 (approved):
neat

andrerenard:
Nice fix!

PR #1510, opened 2026-05-26
New telescope CHARTS

eschnett:
CHARTS https://astro.uc.cl/charlas/canadian-chilean-array-for-radio-transients-studies-charts-tomas-cassanelli/ is a new radio telescope in Chile. They are interested in using our X-Engine design. This PR add kernels and an example pipeline for CHARTS.


julia/output/bb_charts.out
jbmertens, line 1: Should this file be included?

eschnett: I think so; this is the log file for the kernel generation, and we have these log files for all kernels.

jbmertens (approved):
LGTM

PR #1512, opened 2026-05-26
Kv/airspy followup fixes

kvand:
airspy follow-up fixes: simpleAutocorr float32, ADC-stats doxygen, R820T dither config

### DIFF
--- a/lib/stages/airspyInput.cpp
+++ b/lib/stages/airspyInput.cpp
@@ -220,17 +220,18 @@ void airspyInput::main_thread() {
         airspy_opened = true;
     }
 
+    // init_device() already FATAL_ERROR'd with the specific reason on any
+    // failure; just exit main_thread quietly here.
     a_device = init_device();
-    if (a_device == nullptr) {
-        FATAL_ERROR("Error in airspyInput. Cannot find device.");
+    if (a_device == nullptr)
         return;
-    }
 
     if (_autostart) {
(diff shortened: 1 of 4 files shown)


lib/stages/airspyInput.cpp
```diff
@@ -438,6 +436,32 @@ struct airspy_device* airspyInput::init_device() {
...
+    // write only updates the firmware's register shadow; airspy_start_rx later
+    // bulk-writes that shadow before locking the PLL, so the lock happens with
+    // the chosen dither setting baked in. Disabling dither eliminates the
+    // per-restart inter-unit LO drift at fractional tunes (coherent mode) at
+    // the cost of deterministic fractional-N spurs.
+    {
+        uint8_t v;
+        result = airspy_r820t_read(dev, 0x12, &v);
+        if (result != AIRSPY_SUCCESS) {
+            ERROR("airspy_r820t_read(0x12) failed: {:s} ({:d})",
```
jbmertens: Should this be FATAL_ERROR?

jbmertens (approved):
Minor comment, otherwise LGTM.

lib/stages/airspyInput.cpp
kvand: Oh, good catch! There are actually a couple of those, so a slightly larger commit will show up in a moment correcting.

PR #1514, opened 2026-05-28
Kv/airspy viewer refactor

kvand:
Rewrite of the web browser / javascript viewer interface for airspy auto- and cross-correlation visualization / interaction. Should be most of the way generalized to being usable on ARO and others.

36676e95  config/airspy_crosscorr: pin A by serial, tweak gains/bandwidth for testing
b88f47d9  viewer: lag-align "Nudge A/B +1" buttons; drop "Restart & Align"
3a1b8a1c  viewer: live LO retune from the browser
4bdc116a  airspy: drop /restart REST endpoint + autostart config
a6c830a9  python/scripts/js_viewer: livebeam_server bridge + browser viewer
a9a191c8  stages: add SpawnProcess for launching helper child processes
dabab8ab  core/restServer: CORS allowlist for the browser viewer


kvand:
Ok, last of the (immediate) airspy PRs -- this one is in no way urgent, and should be ~orthogonal to all the actual CHORD work going on, but it finally gets a nice UI in place for simple setups like lab airspy use and extension to places like ARO.

jbmertens:
Worth discussion with @andrerenard, who has also been thinking about including an http server in kotekan. This PR is another way to go about that, which could also work: a python layer that plays this role between kotekan and user.

kvand:
Good point re: http server -- though note the little python script is *also* acting as a websocket server. That's a little more complex and probably not something we're going to bake into kotekan in the near-term?

For the viewer specifically, I wanted to keep this as a distinct process that I can hive off onto an independent host, consuming kotekan's networkpowerstream and serving http & websocket to as many clients might want it, without overloading kotekan. I wanted to avoid general loading of the kotekan server -- the example configs used here keep everything local and that SpawnProcess stage to automate the spin-up, but that's not intended to be the only mode of use.

jbmertens (commented):
I've submitted a PR (#1528) which goes ahead and addresses some generalization for the viewer, other random details including items from claude.

LGTM otherwise, still interested to get an opinion from @andrerenard.

jbmertens (approved):
Good by me to merge, though may still be worth running the server idea by @andrerenard .

PR #1516, opened 2026-05-28
CHIME HFB beams

eschnett:
Implement HFB beams for CHIME:
- add additional upchannelization setup for U=128
- add additional FRB-like HFB stage
- simulated F-Engine data are processed to completion


config/fengine/chime.j2
```diff
@@ -10,6 +10,10 @@
 log_level: debug
 poison_buffers: true
 
+run_buffer_status:
+    kotekan_stage: bufferStatus
+    time_delay: 10.0e+6         # in μs
```
rhaas80: this seems extraneous to HFB beams being formed

config/fengine/include/frb_chime.j2
```diff
@@ -42,39 +41,39 @@ host_frb1_U16_phase_buffer:
...
     num_frames: buffer_depth
-    ring_buffer_size: buffer_depth * sizeof_float16 * frb_num_beams_P * frb_num_beams_Q * upchan_all_max_num_output_channels * frb_num_output_times
+    ring_buffer_size: buffer_depth * sizeof_float16 * frb_num_beams_P * frb_num_beams_Q * upchan_U16_max_num_output_channels * frb_num_output_times
     metadata_pool: main_pool
 
 host_frb2_beams_buffer:
     kotekan_buffer: standard
     num_frames: buffer_depth
-    frame_size: sizeof_float16 * frb2_num_beams * (upchan_all_max_output_channel - upchan_all_min_output_channel) * frb_num_output_times
+    frame_size: sizeof_float16 * frb2_num_beams * upchan_U16_max_num_output_channels * frb_num_output_times
```
rhaas80, line 63: seems wrong, replaces `max -min` by just `max` (admittedly `min` may well be 0).

config/fengine/include/frb_chime.j2
```diff
@@ -42,39 +41,39 @@ host_frb1_U16_phase_buffer:
...
     metadata_pool: main_pool
 
 host_frb3_beams_buffer:
     kotekan_buffer: standard
     num_frames: buffer_depth
-    frame_size: sizeof_uint8 * frb2_num_beams * (upchan_all_max_output_channel - upchan_all_min_output_channel) * frb_num_output_times
+    frame_size: sizeof_uint8 * frb2_num_beams * upchan_U16_max_num_output_channels * frb_num_output_times
```
rhaas80, line 69: max-min

config/fengine/include/frb_chime.j2
```diff
@@ -42,39 +41,39 @@ host_frb1_U16_phase_buffer:
...
     metadata_pool: main_pool
 
 host_frb3_beams_offsetscale_buffer:
     kotekan_buffer: standard
     num_frames: buffer_depth
     # the output is quantized in chunks of `frb_chunk_size`
-    frame_size: sizeof_float16 * 2 * frb2_num_beams * (upchan_all_max_output_channel - upchan_all_min_output_channel) * frb_num_output_times / frb_chunk_size
+    frame_size: sizeof_float16 * 2 * frb2_num_beams * upchan_U16_max_num_output_channels * frb_num_output_times / frb_chunk_size
```
rhaas80, line 76: max-min

config/fengine/include/frb_chime.j2
```diff
@@ -117,9 +116,10 @@ run_frb1_U16:
 
 run_calc_frb2_weights:
     kotekan_stage: calcFRB2Weights
-    frb2_num_frequencies: upchan_all_max_output_channel - upchan_all_min_output_channel
+    frb2_num_frequencies: upchan_U16_max_num_output_channels
     frb2_beam_positions: host_frb2_beam_positions_buffer
     frb2_weights: host_frb2_weights_buffer
+    num_threads: 32
```
rhaas80, line 122: not directly related to HFB beamforming

config/fengine/include/frb_chime.j2
```diff
@@ -117,9 +116,10 @@ run_frb1_U16:
 
 run_calc_frb2_weights:
     kotekan_stage: calcFRB2Weights
-    frb2_num_frequencies: upchan_all_max_output_channel - upchan_all_min_output_channel
+    frb2_num_frequencies: upchan_U16_max_num_output_channels
```
rhaas80, line 119: these are changed from `upchan_all...` to `upchan_U16` to not conflict with changes made for HFB ?

lib/stages/calcFRB2Weights.cpp
```diff
@@ -1,34 +1,43 @@
...
 
 class calcFRB2Weights : public kotekan::Stage {
     // Telescope setup
     const int num_dishes = config.get<int>(unique_name, "num_dishes");
 
+    // Upchannelization setup
+    const std::string upchannelization_schedule_name =
+        config.get_default<std::string>(unique_name, "upchannelization_schedule_name", "");
+
     // FRB1 beamformer setup
```
rhaas80, line 41: not necessarily FRB1 anymore. Could be HFB1, yes?

lib/stages/calcFRB2Weights.cpp
```diff
@@ -313,8 +325,13 @@ class calcFRB2Weights : public kotekan::Stage {
                             }
                         }
                     }
+                    ++nfreqs_done;
+                    std::printf("\rcalcFRB2Weights: freqs: %d/%d...", int(nfreqs_done),
```
rhaas80, line 329: leftover debug code?

julia/kernels/chimefrb.jl
```diff
@@ -43,7 +43,7 @@ const F̄ = F_per_U[U] * U
 
 const Tbar = T ÷ U
 # We use U = 128 for HFB beams; downsample to the max
-const Tds = U < 128 ? idiv(Tds_U1, U) : Tbar
```
rhaas80, line 46: why  the `/ 4` ?

rhaas80 (changes requested):
contains changes unrelated to HFB. The HFB changes themselves look ok.

config/fengine/include/frb_chime.j2
eschnett: Yes, the min is zero for CHIME, both for FRB and HFB beams. Things were set up like this in many other, but not in all places. This cleans things up.

config/fengine/include/frb_chime.j2
eschnett: No, the upchannelization setup for CHIME is generally much simpler than for CHORD because there is only a single upchannelization factor, so none of the `min` values are necessary.

lib/stages/calcFRB2Weights.cpp
eschnett: Yes, this is the change. We are using the "FRB" algorithm, but the beams are now called "HFB" beams, and they have a different upchannelization schedule (U=128 instead of U=16).

lib/stages/calcFRB2Weights.cpp
eschnett: I want to keep it. Calculating the weights takes a long time (minutes), and without progress output it's difficult to judge whether the system has stalled.

julia/kernels/chimefrb.jl
eschnett: `Tbar` is the total length of the ring buffer. To prevent stalling we want to fill only 1/4 at a time. (That's the ratio I chose for other kernels as well.)

config/fengine/chime.j2
eschnett: Removed.

config/fengine/include/frb_chime.j2
rhaas80: in that case, rather than `_max` use `num_`?

config/fengine/include/frb_chime.j2
rhaas80: ok

julia/kernels/chimefrb.jl
rhaas80: ok, so not a critical value that describes a specific shape. Sure. A comment "To prevent stalling we want to fill only 1/4 at a time." may be helpful.

rhaas80 (commented):
down-grading my objections to comments

rhaas80 (commented):
and once more to make github happy

rhaas80 (approved):
...

PR #1518, opened 2026-05-28
Update n2k setup for recent changes

eschnett:
- The CPU parts of n2k changed; update the config files
- We now need `constu8` (UInt8) inputs; support this in testDataGen


lib/testing/inventVoltage.cpp
```diff
@@ -43,9 +45,17 @@ class inventVoltage : public kotekan::Stage {
     const std::string input_kind = config.get<std::string>(unique_name, "input_kind");
     const bool reuse_frames = config.get_default<bool>(unique_name, "reuse_frames", false);
 
+    const std::vector<std::string> startup_buffer_names =
```
jbmertens, line 53: Add a bit of documentation (in the header file?) describing intent of this.

lib/testing/inventVoltage.cpp
```diff
@@ -85,6 +118,37 @@ class inventVoltage : public kotekan::Stage {
...
+                using std::fmax, std::fmin, std::sqrt;
+                time_min = fmin(time_min, time);
+                time_max = fmax(time_max, time);
+                time_count += 1;
+                time_sum += time;
+                time_sum2 += time * time;
+                if (frame_index % out_frame_every == 0) {
+                    double time_avg = time_sum / time_count;
+                    double time_sdv = sqrt(fmax(0.0, time_sum2 / time_count - time_avg * time_avg));
+                    INFO("Timing statistics:\n"
```
jbmertens, line 139: INFO or DEBUG?

jbmertens (approved):
Couple quick items - otherwise LGTM! Preemptively approving.

lib/testing/inventVoltage.cpp
eschnett: I'm using INFO because I'm using this stage for benchmarking to generate data very quickly. While benchmarking I can't use DEBUG output (there would be too much) so I'm using INFO output. This output shows how quickly Kotekan is generating input data, which (once the pipeline filled up) is the same speed with which Kotekan is able to process data.

lib/testing/inventVoltage.cpp
eschnett: Will do. Quick answer here: Kotekan takes a long time to start up because e.g. beamforming matrices need to be calculated. We want to delay starting the benchmark timings until this startup has finished. The "startup buffers" are those that we wait for before starting to push data through the pipeline as fast as we can.

eschnett: Done.

PR #1520, opened 2026-05-28
bufferStatus: Output ring buffer statistics as well

eschnett:
- bufferStatus: Output ring buffer statistics
- ring buffers: New function to output buffer status
- regular buffers: Add more detail to buffer status output

"More detail" means there are now four states for each frame: empty, filling, full, and emptying. Before there were only two, full and empty.


lib/core/buffer.cpp
```diff
@@ -401,19 +403,52 @@ std::string Buffer::get_dot_node_label() {
...
+            producer_snapshots.push_back({p.second.last_frame_acquired, p.second.is_done});
+        consumer_snapshots.reserve(consumers.size());
+        for (const auto& c : consumers)
+            consumer_snapshots.push_back({c.second.last_frame_acquired, c.second.is_done});
+    }
+    // Unicode symbols (matching RingBuffer::print_buffer_status)
+    const std::string emptying_symbol = "▼";
+    const std::string full_symbol = "█";
+    const std::string filling_symbol = "▲";
+    const std::string empty_symbol = "·";
```
rhaas80: I'd like to lobby to use plain ASCII and not utf-8 here (since utf-8 gets mangled in my somewhat deep nesting of screen, ssh, screen, ssh, ssh accessing eg the upgrade node from my laptop via my worstation and the chime network gatekeeper node.

lib/core/buffer.cpp
```diff
@@ -401,19 +403,52 @@ std::string Buffer::get_dot_node_label() {
...
+                    break;
+                }
+            }
+            buf << (filling ? filling_symbol : empty_symbol);
+        }
     }
-    status_string[num_frames] = '\0';
-    INFO("Buffer {:s}, status: {:s}", buffer_name, std::string(status_string));
+    buf << "]";
+    INFO_NON_OO("Buffer {:40s}, status: {:s}", buffer_name, buf.str());
```
rhaas80, line 454: why remove the `OO`?

rhaas80 (commented):
I would try and avoid unicode

lib/core/buffer.cpp
eschnett: The `OO` output prefixes all output with a badly formatted buffer name. This line then adds a nicely formatted buffer again. Combined, this uses about the first 60 character of every line, every line wraps, and the output is not nicely aligned vertically. We could also use `INFO_F` instead (that skips the `fmt` formatting as well, we'd do it ourselves), or output to `std::cout`. I think` INFO_NON_OO` is the highest-level output that makes sense here.

lib/core/buffer.cpp
eschnett: So sad. But noted.

eschnett: ASCII it is.

rhaas80:
fails 11 tests

eschnett:
All green now! It turns out that `"v"` is a string, not a character, and those are apparently different types in C++! Fixed now.

rhaas80 (approved):

PR #1522, opened 2026-06-02
Update rfi broadcast SK bar statistic

ljgray:
Just sends the raw SK data integrated to some number of frames instead of median-absolute-deviations counts.

### DIFF
--- a/lib/stages/rfiBroadcast.cpp
+++ b/lib/stages/rfiBroadcast.cpp
@@ -28,33 +28,6 @@ REGISTER_KOTEKAN_STAGE(rfiBroadcast);
 
 constexpr int BITS_PER_BYTE = 8;
 constexpr int RFI_BUF_S = 3;
-// Conversion between median absolute deviations and sigma for
-// a normal distribution
-constexpr float MAD_TO_SIGMA = 1.4826;
-
-namespace {
-/// Helper to compute array median. Modifies `arr` in-place.
-float median(std::vector<float>& arr) {
-    auto n = arr.size();
-    auto mid = n / 2;
-
-    if (mid == 0) {
-        return 0.0f;
-    }
-
-    std::nth_element(arr.begin(), arr.begin() + mid, arr.end());
-    auto upper = arr[n / 2];
-
-    // array length is odd - return the central value
-    if (n % 2 != 0)
-        return upper;
-
-    // array length is even
-    std::nth_element(arr.begin(), arr.begin() + mid - 1, arr.end());
-
-    return (upper + arr[n / 2 - 1]) / 2.0;
-}
-}
 
 rfiBroadcast::rfiBroadcast(Config& config, const std::string& unique_name,
                            bufferContainer& buffer_container) :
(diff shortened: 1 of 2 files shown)


lib/stages/rfiBroadcast.cpp
```diff
@@ -234,79 +198,53 @@ void rfiBroadcast::main_thread() {
 
         // Wait for the sk_tilde frame
         DEBUG("Waiting on empty sk_tilde frame.");
-        uint8_t* sk_tilde_frame =
-            (uint8_t*)sk_tilde_buf->wait_for_full_frame(unique_name, sktilde_frame_id);
+        float* sk_tilde_frame = reinterpret_cast<float*>(
+            sk_tilde_buf->wait_for_full_frame(unique_name, sktilde_frame_id));
```
rhaas80, line 202: so it's a `float` and not a `uint8`? 

If there's a frame desc then you can also try and do:

```
assert(kotekan::GetDataType_v<std::remove_reference_t<decltype(*sk_tilde_frame)>> == sk_tilde_buf->get_ndarray_frame_desc()->get_value_datatype());
```

which checks that the type of your pointer matches the declared type of the frame.

lib/utils/rfi_functions.hpp
```diff
@@ -37,26 +35,28 @@ struct __attribute__((packed)) RFIHeader {
...
+using SkType = float;
+
 /*
  * @struct RFIPayload
  * @brief A structure defining the payload included in each rfiBroadcast packet.
  */
 struct RFIPayload {
     /// List of local frequencies. Has length `num_local_freq`.
-    std::vector<uint32_t> freq_ids;
+    std::vector<FreqIDType> freq_ids;
```
rhaas80: there's a `freq_id_t` defined in `Telescope.hpp`  maybe use that one instead?

lib/stages/rfiBroadcast.cpp
```diff
@@ -31,33 +28,6 @@ REGISTER_KOTEKAN_STAGE(rfiBroadcast);
 
 constexpr int BITS_PER_BYTE = 8;
 constexpr int RFI_BUF_S = 3;
-// Conversion between median absolute deviations and sigma for
-// a normal distribution
-constexpr float MAD_TO_SIGMA = 1.4826;
-
-namespace {
-/// Helper to compute array median. Modifies `arr` in-place.
-float median(std::vector<float>& arr) {
```
rhaas80, line 40: no more median?

lib/utils/rfi_functions.hpp
rhaas80: though I guess since you are sending them across network as binary data you need to know that indeed it's a `uint32_t` since the receiving side expects that explicit type.

rhaas80 (approved):

lib/stages/rfiBroadcast.cpp
ljgray: Logic moved to the external receiver

PR #1523, opened 2026-06-02
fix(python): Add pyproject.toml for build requirements.

tristpinsm:
Recent `pip` uses build isolation and was failing in the current `setup.py` scheme because it was missing the `future` and `versioneer` dependencies.

Transitioning the rest of the build to `pyproject.toml` might be a useful follow-up.


jbmertens (approved):
LGTM.

PR #1529, opened 2026-06-05
Add ndarray buffer type

jbmertens:
Converges ndarray frame_descs with other buffer types: buffers own descriptions, stages validate.

  - adds ndarray buffer type
  - buffer factory allocates based on ndarray value_type and extents
  - Buffer::require_frame_desc/ensure_frame_desc for validation
  - Many stages migrated from standard to ndarray buffers, configs updated appropriately
  - Does not address ringbuffer types.
  - Does not address some producers (dpdk, cudacopy stages)

### DIFF
--- a/docs/sphinx/dev/dev_buffers.rst
+++ b/docs/sphinx/dev/dev_buffers.rst
@@ -69,7 +69,7 @@ Type-specific parameters are required unless marked *optional*:
        - ``quantity_name`` -- label for the stored quantity
        - ``dimnames`` -- list of axis labels, one per extent
 
-       Omitted labels are supplied by the producing stage (see below).
+       Omitted labels may be supplied by a producing stage (see below).
        ``frame_size`` is derived from the descriptor.
      - ``GenericNDArray``, built from the config block and attached at
        startup
@@ -84,8 +84,9 @@ Type-specific parameters are required unless marked *optional*:
        startup
    * - ``standard``
      - - ``frame_size`` -- frame size in bytes, set explicitly
-     - none declared; stages may attach one at runtime via
-       ``ensure_frame_desc`` / ``require_frame_desc``
+     - none declared; a stage may attach one at runtime via
+       ``ensure_frame_desc`` (``require_frame_desc`` would be fatal, as a
+       ``standard`` buffer has no declared descriptor)
    * - ``vis``
      - - ``num_elements`` -- number of correlator inputs
        - ``num_ev`` -- number of eigenvectors
@@ -128,8 +129,14 @@ therefore declare only the structure, leaving labels for the producing
 stage to supply; when both producer and consumer expect a label it must agree.
 
 The model is: **buffers carry their own description; stages validate
-against it** (and optionally complete any labels the config omitted to minimize
-repetition).
+against it.** The buffer factory guarantees the baseline at startup -- a
+declared descriptor's byte size matches ``frame_size`` and its structure
+matches the config -- so a stage never re-checks the whole shape for that
+reason. Beyond that baseline, validation lives in the stage: there is no
+shared field-validation helper, because the checks stages need (e.g. dimension
+type, buffer size divisibility, etc) are too varied to fold into one. A stage
+should read a descriptor and check any specific properties it depends on,
+completing any labels the config omitted to minimize repetition.
 
 Rationale
 ---------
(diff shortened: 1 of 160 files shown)


eschnett (approved):

lib/metadata/NDArray.hpp
```diff
@@ -77,67 +77,115 @@ make_arrays_from_pairs(std::initializer_list<std::pair<T, U>> values) {
...
+                                                    const Symbol quantity_name,
+                                                    const std::vector<std::ptrdiff_t>& extents,
+                                                    const std::vector<Symbol>& dimnames);
+
+    /// create an NDArray frame descriptor from a config block. Reads the
+    /// structural `value_type` and `extents` (required) and the semantic labels
+    /// `quantity_name` and `dimnames` (optional -- left unset when omitted, for
+    /// the producing stage to fill in). Extent entries may be arithmetic
+    /// expressions referencing other (scoped) config values. The data pointer is
+    /// null; the result describes shape only (e.g. for Buffer::ensure_frame_desc).
```
andrerenard: "The data pointer is null"?

docs/sphinx/dev/dev_buffers.rst
```diff
@@ -4,5 +4,201 @@
...
+  interpreting them, so a buffer's description cannot always come from the
+  stages that touch it.
+- Validation is the primary safety mechanism: producer, consumer, and
+  config descriptions should be cross-checked at startup, so mismatches
+  appear at launch (with both descriptions printed) rather than downstream.
+- Shapes stay visible in the config, next to the pipeline wiring, for
+  review and debugging without consulting stage source.
+- Descriptor requirements depend on stage requirements and buffer types:
+  ``standard`` buffers carry none, ``ndarray`` buffers carry
+  partial descriptors, and ``N2`` buffers carry the full descriptor.
```
andrerenard, line 154: This seems more like an AI "thinking" block than something useful?

andrerenard (approved):
It's great to be moving in this direction!  Thanks for the PR.   I do think this is worth merging, but I think this is really a first step.   The main thing I worry about is duplication of logic in both the buffer _and_ stages.   Currently this forms a strong check, which is nice to have, however that appears to require that both the buffer and stage agree on exact names, sizes, etc. based on config derived parameters.   It feels it will be easier to work with if we say that the buffer is the "source of truth".  Instead of the stage checking it has the exact same state (requiring a copy of the config parameters for the stage), it could use the frame_desc to determine _if_ it can process the incoming data and what parameters to use.   

It might be worth having some kind of formalism for saying what a stage can process. e.g.
n_dims = 3
allowed dim sizes: [>0, 1<128, 2]
type: int32 or float32

And then use the dim sizes as the internal loop variables instead of separately getting those from the config.   

This does however, mean that things like output dimensions as they related to input dimensions now need to be checked in the stages.  

Also I'm not sure how important the dimension names as a check value, compared with the dimension sizes and type...

docs/sphinx/dev/dev_buffers.rst
jbmertens: I left this in intentionally, sort of a set of reminders as to "why this direction?". For anyone thinking about further changes down the line.

rhaas80:
renaming an existing and widely used function to a new name is a terrible idea.

jbmertens:
Many stages used `allocate_ndarray_frame_desc`, but per discussions over the past year, my understanding is the buffer factory rather than stages was agreed as where frame descs should be allocated. The wide use is just because this change is overdue?

rhaas80:
I was away on a conference last week so any discussion last week would be lost on me. I cannot say I recall intensive discussion about this though. If it was local to UoT / PI then that is not really helpful in that respect. 

I would not even mind if there was a new function, what I *do* mind is the loss of the existing function and requirement to change / update all development branches. 

While CHORD still has a longer time horizon to productive use, CHIME is somewhat closer and we would ideally want to actually use the code in production end of July. In fact the FRB fixes that Erik just proposed were the last major blocker (assuming there is nothing wrong in the BeamReformer).

Since Kotekan serves as  basis for multiple telescopes it would be nice if wide ranging changes were discussed in (one of) the common telecons and things were not changed in such a manner as to break common code used by all telescopes. 

Basically: if it ain't broken, don't fix it. And as far as I am concerned: "uglyness is not a reason to fix code".

Use of `allocate_ndarray_frame_desc` to declare the array shape is indeed somewhat old (though given that I introduced it, it cannot be much more than 1 year old), while (ab)use of it by consumers to check that the array has the shape they expect is somewhat newer. 

Anyhow, it would be nice is these changes happened less frequent during summer. Though with the FRB stuff now fixed, it seems feasible to consider forking off a CHIMEproduction branch of chord.

jbmertens:
I do have a similar gripe about changing core functionality - it causes supply chain shocks. I have no problem with reverting https://github.com/kotekan/kotekan/pull/1529/changes/0769c66a544367c4ac82c017a6dcebbe4f77974a or other parts of the PR if it helps.

But I would like to have support for the ndarray buffer type support before working on flagging and calibration in CHORD. This PR also "brings forward" a lot of unrelated stages, which can be undone...

rhaas80:
I am all in favor of having the buffer carry the array shape and that being set via the input files is ok. The consumers (and producers) will likely still need to check that whatever is set in the input files is actually what they expect. So the information still gets scattered around the same way.

eschnett:
It won't be as scattered. Many stages currently require parameters (e.g. `num_local_freq`) that specify the array shape. Instead of using that parameter and checking whether the array has that shape, they would determine `num_local_freq` from the frame description. The information is then not scattered, and stages are easier to configure.

Of course, if a stage has a hard-coded requirement (e.g. "number of time samples must be a multiple of 256"), then they still need to check that.

rhaas80:
I think you need to check more than just the sizes. As a producer, since I am no longer the one setting the description, I need to check that the dimensionality is correct, the extents are ones that I can handle, but also that the quantity name is the expected one, and that all the dimension names are the correct ones. There should then be functionality that makes this easy so that stage authors actually do so.

Looking at `GenericNDArray::reconcile` it seems to fill in partially missing information which seems odd. No stage should set only partial information (eg no empty quantity name is allowed).

eschnett:
Yes, I don't like the `reconcile` either. I plan to open a PR that removes this functionality, if that works with our code base. I think it's mostly tests that use this. Test are written to be simple and benefit from defaults. But a production pipeline shouldn't depend on defaults.

I am currently going through the F-Engine simulator pipelines for all the telescopes and fixing them. I already found a bug or two.

PR #1530, opened 2026-06-10
CHARTS: Update test input files

eschnett:


config/fengine/include/frb_charts.j2
```diff
@@ -62,19 +62,19 @@ host_frb2_beams_buffer:
...
-    frame_size: frb2_num_beams * (upchan_all_max_output_channel - upchan_all_min_output_channel) * frb_num_output_times / 2
-    metadata_pool: main_pool
-
-host_frb3_beams_offsetscale_buffer:
-    kotekan_buffer: standard
-    num_frames: buffer_depth
-    # the output is quantized in chunks of `frb_chunk_size`
-    frame_size: sizeof_float16 * 2 * frb2_num_beams * (upchan_all_max_output_channel - upchan_all_min_output_channel) * frb_num_output_times / frb_chunk_size
-    metadata_pool: main_pool
+#TODO host_frb3_beams_buffer:
```
jbmertens, line 65: Do you want to keep these TODOs?

config/fengine/include/output_charts.j2
```diff
@@ -239,11 +239,7 @@ write_data:
         in_buf: host_n2_vis_buffer
         file_name: n2_vis
         create_single_file: false # not implemented
-        max_frames: 1
-    # write_n2_vis:
-    #     kotekan_stage: hdf5N2Write
-    #     num_file_t: 4
-    #     in_buf: host_n2_vis_buffer
+        max_frames: 1 * num_frequencies
```
jbmertens, line 242: 1* ?

jbmertens (approved):
LGTM, just a couple small comments, preemptively approving.

config/fengine/include/frb_charts.j2
eschnett: Yes, i want to keep them at the moment. I want the CHARTS team to decide whether to quantize to 4 bits (enable) or keep as float64 (remove).

config/fengine/include/output_charts.j2
eschnett: The current setup runs n2accumulate in a mode where all data are collected into 1 sample spanning about 1 second. this could be changed, and then there would be multiple samples. n2accumulate also splits its output into one-buffer-per-frequency, so that total number of frames is the number of samples times the number of frequencies.

PR #1533, opened 2026-06-11
Correct x/y vs. M/N directions in FRB beamformers

eschnett:
CHARTS, CHORD, and Hirax arrange dishes and beams differently from CHIME in the FRB beamformer. Clean this up, add comments, and correct errors. (Probably.)

### DIFF
--- a/config/fengine/include/hfb_chime.j2
+++ b/config/fengine/include/hfb_chime.j2
@@ -43,12 +43,12 @@ host_hfb2_weights_buffer:
 host_hfb1_beams_buffer:
     kotekan_buffer: standard
     num_frames: buffer_depth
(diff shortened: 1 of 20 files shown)


config/fengine/include/hfb_chime.j2
```diff
@@ -141,8 +141,6 @@ run_hfb23:
           frb1_max_num_frequencies: upchan_hfb_upchan_U128_max_num_output_channels
           frb1_max_num_times: buffer_depth * hfb_num_output_times
           frb2_num_frequencies: upchan_hfb_upchan_U128_max_num_output_channels
```
jbmertens, line 143: Pre-existing issue, but noting that the following is set twice:

jbmertens (approved):
LGTM

config/fengine/include/hfb_chime.j2
eschnett: Thanks! Will fix in another PR.

PR #1535, opened 2026-06-12
Telescope metadata cleanup

eschnett:
Clean up writing telescope metadata:
- Omit superfluous `num_elements`
- Do not write `input_order`; these metadata describe the telescope, not just one dataset. (The dataset is still unambiguously described by the dataset metadata.)
- Write dish array as 2D HDF5 array, not as 1D array of 1D arrays

Read and check telescope metadata:
- When reading a file, check metadata for consistency, and abort if inconsistent (wrong number of dishes etc.)

### DIFF
--- a/lib/stages/hdf5FileRead.cpp
+++ b/lib/stages/hdf5FileRead.cpp
@@ -1,36 +1,37 @@
-#include <Config.hpp>                             // for Config
-#include <DataType.hpp>                           // for string_to_type, DataType
-#include <Stage.hpp>                              // for Stage
-#include <StageFactory.hpp>                       // for REGISTER_KOTEKAN_STAGE
-#include <Symbol.hpp>                             // for Symbol
-#include <buffer.hpp>                             // for Buffer
-#include <bufferContainer.hpp>                    // for bufferContainer
-#include <chordMetadata.hpp>                      // for chordMetadata, metadata_is_chord, get_c...
-#include <hdf5Files.hpp>                          // for chord_metadata_version
-#include <highfive/H5Attribute.hpp>               // for Attribute, Attribute::read
-#include <highfive/H5DataSet.hpp>                 // for DataSet, AnnotateTraits::getAttribute
-#include <highfive/H5DataSpace.hpp>               // for DataSpace, DataSpace::getDimensions
-#include <highfive/H5Exception.hpp>               // for FileException
-#include <highfive/H5File.hpp>                    // for File, File::File, NodeTraits::getDataSet
-#include <highfive/bits/H5Slice_traits_misc.hpp>  // for SliceTraits::read_raw
-#include <kotekanLogging.hpp>                     // for DEBUG, FATAL_ERROR, ERROR
-#include <metadata.hpp>                           // for metadataObject
-#include <prometheusMetrics.hpp>                  // for Metrics, Gauge
-#include <unistd.h>                               // for gethostname, sleep
-#include <visUtil.hpp>                            // for current_time
-#include <algorithm>                              // for copy
-#include <array>                                  // for array
-#include <cassert>                                // for assert
-#include <cstddef>                                // for ptrdiff_t
-#include <cstdint>                                // for int64_t, uint8_t
-#include <functional>                             // for function
-#include <iomanip>                                // for operator<<, setfill, setw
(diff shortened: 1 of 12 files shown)


lib/stages/hdf5FileRead.cpp
```diff
@@ -60,6 +64,103 @@ class hdf5FileRead : public kotekan::Stage {
...
+        check_bool_attr("gps_time_enabled", telescope.gps_time_enabled());
+        check_int_attr("num_polarizations", num_polarizations);
+        check_int_attr("num_dishes", num_dishes);
+        check_float_attr("itrs_lat_deg", telescope.get_itrs_lat_deg());
+        check_float_attr("itrs_lon_deg", telescope.get_itrs_lon_deg());
+        // node.getAttribute("grid_orientation", telescope.get_grid_orientation());
+        check_uint_attr("grid_size_x", telescope.get_grid_size_x());
+        check_uint_attr("grid_size_y", telescope.get_grid_size_y());
+        check_float_attr("feed_separation_x_m", telescope.get_feed_separation_x_m());
+        check_float_attr("feed_separation_y_m", telescope.get_feed_separation_y_m());
```
jbmertens, line 131: Consider writing/validating element ordering as well.

jbmertens (approved):
Just one thought - otherwise LGTM.

rhaas80:
is it to late to suggest using "channel id" instead of "station id"? (I keep using the latter but it is apparently the former that is the correct term).

lib/stages/hdf5FileRead.cpp
eschnett: These attributes describe the telescope, not the dataset. Anything dataset specific shouldn't be here. We could add an attribute "these data are for CHIME", or "these data are for CHORD", but the way voltage is laid out in memory doesn't fit here – the file may not even contain voltages.

I know that @geoffryan added a "dataset order" to the telescope object, but I think this information really shouldn't be in the telescope. What if we reorder the data in memory? Do we then need a second telescope object?

eschnett:
"Channels" are for frequencies, "stations" are the antenna elements. Those are different. I think the telescope object mostly deals with stations, whereas the F-engine creates the channels.

lib/stages/hdf5FileRead.cpp
jbmertens: Ah, fair, iirc he did leave a to-do item that element ordering should really be in the frame_desc or metadata.

geoffryan: "ElementOrder" is not stored on the telescope. It (should) be stored on the dataset. It is passed to the telescope object for relevant functions on a dataset, exactly to support multiple in-memory orders.

geoffryan: (i swear this is the last time, poof I'm gone!)

rhaas80:
> "Channels" are for frequencies, "stations" are the antenna elements. Those are different. I think the telescope object mostly deals with stations, whereas the F-engine creates the channels

Maybe? What do the actual radio-astronomers say? Eg the VDIF file format https://vlbi.org/wp-content/uploads/2019/03/VDIF_specification_Release_1.1.1.pdf has a "station ID" field:

> The 16-bit ‘Station ID’ field will accommodate the standard globally assigned 2-character
ASCII ID. Or, if the number of stations is very large (SKA, for example), the Station ID
may be numeric; in this case, the 16-bit field will be interpreted as an unsigned 16-bit
integer. The two cases can be distinguished by the value of the first 8-bit character of the
field; if this character has ASCII value <4810 (representing the zero character ‘0’), a
numeric ID is assumed.

There;s also "channels" defined but they seem a bit murky in that it says that

> For simplicity, and in accordance with historical VLBI practice, the VDIF specification for
multi-channel Data Arrays supports only 2n channels with 2k bits/sample; maximum #channels is
231=2,147,483,648 and maximum bits/sample is 25=32. Extension of these formats to include an
arbitrary number of bits/sample is not contemplated. In such cases, users are strongly
encouraged to use single-channel Data Threads, which do not impose this constraint.

so seem to be mostly a legacy feature. 

For purposes of kotekan the item we label is a single polarization of a single dish (to avoid the ambiguity of whether an "antenna" is only for one polarization or both, plus there's non-antenna data streams, eg the noise source).

PR #1536, opened 2026-06-12
Separate `load` and `process` gain stages

ljgray:
Removes the existing `processGains` stage in favour of two separate stages.

# loadFeedGains
Load gain files for a specific beamformer type. Expects one updatable config endpoint per "beam", and produces separate output buffers for each beam (complex [nfreq, nelement]). Buffers are only produced/updated when the corresponding gain file is updated. Gains are sent exactly as stored.

# processFeedGains
Collects `N` buffers for `N` beams for a beamformer type, upchannelizes, concatenates over a new 0th axis (called `beam`), applies a feed mask, and stores gains in a permanent buffer. `kotekan` buffers are emitted continuously by copying the current contents of the permanent buffer on every pass. The feed mask is also stored in a permanent buffer, which is updated whenever a new feed mask buffer is received (this can be changed if the feed mask ends up being sent continuously).

# bad inputs
Updates `chime_upgrade_n2` to get the bad input mask from `bufferBadInputs` and adds the corresponding updatable config.

Has been tested with mock gain files and produces the expected upchannelized buffers (for FRB, HFB, and tracking gains simultaneously).

A basic config exists in `chime_upgrade_test_gains.j2`. We'll want to remove that config file eventually, but there aren't any `FRB`/`tracking` beamformer chime_upgrade configs in `chord` yet.

Unit test not done yet but I'll get to it eventually.


config/chime_upgrade_test_gains.j2
```diff
@@ -710,52 +711,132 @@ run_recv_rfi_RFImask:
...
+{%- endfor %}
+
+gain_frb_upchan_buffer:
   kotekan_buffer: standard
   num_frames: 1
   frame_size: num_elements * num_freq_gpu_buf * frb_upchan_factor * 2 * sizeof_float32
   type: float32
-  dimnames: ["F", "U", "E", "ReIm"]
-  shape: [num_freq_gpu_buf, frb_upchan_factor, num_elements, 2]
+  dimnames: ["B", "F", "U", "E", "ReIm"]
```
rhaas80: what does "B" stand for? If "beam" then "R" is used in the FRB stuff

config/chime_upgrade_test_gains.j2
```diff
@@ -710,52 +711,132 @@ run_recv_rfi_RFImask:
...
+    - gain_frb_buffer
+  in_mask_buf: host_bf_mask_buffer
+  out_buf: gain_frb_upchan_buffer
+
+process_HFB_gains:
+  kotekan_stage: processFeedGains
+  num_local_freq: num_freq_gpu_buf
+  upchan_factor: hfb_upchan_factor
+  gain_buffers:
+    - gain_frb_buffer
```
rhaas80: `gain_hfb_buffer`?

lib/stages/loadFeedGains.cpp
```diff
@@ -0,0 +1,289 @@
...
+                any_failed = true;
+                ffailed = true;
+            };
+        }
+
+        // If the read failed at any point, just use default gains, but still
+        // return true
+        if (ffailed) {
+            float* dst = frame + foffset;
+            for (uint j = 0; j < 2 * num_elements; j += 2) {
```
rhaas80, line 203: shouldn't be `2. * num_elements` but `2 * num_elements` (int instead of double), though it won't make a difference in the end.

lib/stages/processFeedGains.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+            // check if this buffer has an available frame, using a short
+            // timeout to avoid blocking
+            timespec timeout = double_to_ts(current_time());
+            int status = buf->wait_for_full_frame_timeout(unique_name, frame_id, timeout);
+
+            if (status == 0) {
+                DEBUG("Received gain update for beam {:d}", beam_id);
+                // frame available
+                float* frame = (float*)buf->wait_for_full_frame(unique_name, frame_id);
```
rhaas80: One must not wait for a buffer more than once. Instead you need to get the frame (that you already own)  via `buffer->frames.at(frame_id)`. While only `wait_for_full_frame` carries the explicit warning, both functions are the same (except for a call to `wait` vs `wait_until`) so the warning should also apply here.

lib/stages/processFeedGains.cpp
```diff
@@ -0,0 +1,218 @@
...
+        }
+        // Check for mask updates and copy to the mask buffer
+        timespec timeout = double_to_ts(current_time());
+        int status =
+            in_mask_buf->wait_for_full_frame_timeout(unique_name, in_mask_frame_id, timeout);
+        if (status == 0) {
+            DEBUG("Received bad input mask update");
+            // frame available, so update
+            uint8_t* in_mask_frame =
+                (uint8_t*)in_mask_buf->wait_for_full_frame(unique_name, in_mask_frame_id);
```
rhaas80: same as above

lib/stages/processFeedGains.cpp
rhaas80: fails to handle the `status == -1` case which is used to signal thread exit (link `NULL` for the blocking calls).

rhaas80 (changes requested):

rhaas80 (approved):
given that all comments are marked as resolved

PR #1540, opened 2026-06-18
Update GPU kernels

eschnett:
- Add new U=128 upchannelizer for Pathfinder
- Use proper C++ mutex to ensure PTX is built only once


julia/kernels/upchan.jl
```diff
@@ -104,16 +104,7 @@ elseif U == 4
...
-    const W = 16
-    const B = 2
-elseif U == 32
-    const W = 16
-    const B = 2
-elseif U == 64
-    const W = 16
-    const B = 2
-elseif U == 128
+elseif U >= 16
```
jbmertens: Does this remove a check for supported U values?

jbmertens (approved):
One small comment, otherwise LGTM.

julia/kernels/upchan.jl
eschnett: D'oh.

PR #1544, opened 2026-06-19
Add new metadata field dimscalings

eschnett:
A "dimension scaling" specifies by how much a particular array dimension is rescaled. This generalized time downsampling to all other quantities. For example, a dish index `D` with a `dimscaling` of 8 means that this index really specifies the dish `8*D`, not `D`. Such rescalings are used in many places in Kotekan – mostly for time downsampling, but not only.

A more extreme example is the packet loss mask `pl_mask`. We currently describe it by the axes `{"T2hi64", "F4", "P", "D8", "T2lo64"}`, describing the rescaling implicitly by e.g. `T2` (`T div 2`) and `T2lo64` (`(T div 2) mod 64`). This now becomes

old: `dimnames = {T2hi64, F4, P, D8, T2lo64}`
new: `dimnames = {T, F, P, D, T}` with `dimscalings = {128, 4, 1, 8, 16}`.

This PR introduces the new metadata fields, add respective `dimscalings` arguments to its constructor, and updates all places where NDArrays are created.

This PR does not yet change the dimnames from e.g. `F4` to `F`, although it could – this will be a later cleanup.


lib/metadata/chordMetadata.hpp
```diff
@@ -90,6 +90,7 @@ class chordMetadata : public metadataObject {
     int dims;
     int dim[CHORD_META_MAX_DIM];
     char dim_name[CHORD_META_MAX_DIM][CHORD_META_MAX_DIMNAME]; // "F", "T", "D", etc
+    int64_t dim_scaling[CHORD_META_MAX_DIM];
```
jbmertens, line 93: If we have this in the frame description, do we also need it in metadata? For network sends?

jbmertens (commented):
Ah, merge conflicts - will review more after those are resolved.

lib/metadata/chordMetadata.hpp
eschnett: It's not strictly necessary. In the future we want to remove all these entries from the metadata. At the moment we probably can't because we don't yet have frame descriptors for GPU arrays, so the individual GPU kernels create their own, but just locally, and compare to the metadata to check correctness.

config/ci-tests/gpu_batch/test_n2k_writeHDF5.yaml
```diff
@@ -89,6 +89,7 @@ gen_data_voltage:
     name: "E"
     array_shape: [2048, 16, 2, 64]
     dim_name: ["T", "F", "P", "D"]
+    dim_scaling: [1, 1, 1, 1]
```
jbmertens, line 92: Hm. It's probably safe to leave this a default that doesn't need to be specified? Although I do appreciate the explicitness.

lib/cuda/cudaPLMaskAccumulator.cpp
```diff
@@ -122,11 +123,13 @@ cudaPLMaskAccumulator::cudaPLMaskAccumulator(kotekan::Config& config,
...
                                           div_noremainder(num_frequencies, 4), num_polarizations,
                                           div_noremainder(num_dishes, 8), 64 / 8},
-            std::array<std::string, 5>{"T2hi64", "F4", "P", "D8", "T2lo64"}, *this),
+            std::array<std::string, 5>{"T2hi64", "F4", "P", "D8", "T2lo64"}, {128, 4, 1, 8, 16},
+            *this),
     pl_counts(pl_counts_name, "pl_counts",
               std::array<std::ptrdiff_t, 4>{num_subintegrations, num_frequencies, num_polarizations,
                                             num_dishes},
-              std::array<std::string, 4>{"Tc", "F", "P", "D"}, *this)
+              std::array<std::string, 4>{"Tc", "F", "P", "D"}, {sub_integration_ntime, 1, 1, 8},
```
jbmertens: Upscaling factor of 8 for the dishes, is this correct?

lib/testing/gpuSimulateN2kCorr.cpp
```diff
@@ -258,13 +258,14 @@ void gpuSimulateN2kCorr::main_thread() {
         meta_out->type = kotekan::int32;
         meta_out->dims = 6;
         assert(meta_out->dims <= CHORD_META_MAX_DIM);
-        meta_out->set_array_dimension(0, nt_outer, "Tc");
-        meta_out->set_array_dimension(1, _num_local_freq, "F");
+        meta_out->set_array_dimension(
+            0, nt_outer, "Tc", meta_in->get_time_downsampling_fpga() * _sub_integration_ntime);
```
jbmertens, line 262: Does this work when time_downsampling_fpga != 1?

jbmertens (approved):
Couple questions - otherwise LGTM.

I also see dim_scaling/dim_scalings/dimscalings. Could use "dim_scalings" everywhere?

lib/testing/gpuSimulateN2kCorr.cpp
eschnett: In principle it does. It is, however, different from the frame description above, which assume that `time_downsampling_fpga` is 1, so this should be cleaned up.

lib/cuda/cudaPLMaskAccumulator.cpp
eschnett: It is not. (Was not – corrected.)

PR #1546, opened 2026-06-19
Include extra SK capture mode

ljgray:
I figure that we should actually merge this.

- enables second-stage flagging by default
- fixes the `EigneN2Iter` diagonal band error
- includes the SK dump config setup put together by @rhaas80

### DIFF
--- a/config/chime_upgrade_n2.j2
+++ b/config/chime_upgrade_n2.j2
@@ -101,7 +101,7 @@ updatable_config:
   # Updatable Config for second stage RFI excision.
   rfi_second_stage_enabled:
       kotekan_update_endpoint: json
-      enabled: false
+      enabled: true
       valid_at_time_ns: 0
   
   rfi_second_stage_thresholds:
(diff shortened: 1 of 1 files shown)


config/chime_upgrade_n2.j2
```diff
@@ -1030,7 +1030,7 @@ eigencalc:
     krylov: 2
     subspace: 1
     # Masking config from chime science run
-    bands_filled: [ [0, 30], [251, 262] ]
+    diagonal_bands_filled: [ [0, 30], [251, 262] ]
```
rhaas80, line 1033: yes! that one really, really should be included

config/chime_upgrade_n2.j2
```diff
@@ -126,7 +126,7 @@ dataset_manager:
 
 main_pool:
   kotekan_metadata_pool: chordMetadata
-  num_metadata_objects: 30 * host_buffer_depth + 5 * baseband_buffer_depth
+  num_metadata_objects: 30 * host_buffer_depth + 5 * baseband_buffer_depth + sk_buffer_depth
```
rhaas80: depends a bit on the value of `sk_buffer_depth`. Making this very large can actually (in combination with the baseband buffer depth) lead to out of memory errors.

config/chime_upgrade_n2.j2
```diff
@@ -1155,13 +1164,23 @@ switch_sk_out:
       - enabled: host_rfi_SKtilde_buffer
     out_buf: host_rfi_SKtilde_buffer_post
     updatable_config: /updatable_config/enable_sk_out
+delay_sk_out:
```
rhaas80: I am not quite sure if this actually does what I hoped it would do. 

Maybe best to least out the delaying of output related stages? That also reduces the memory usage.

ljgray:
Removed output buffering.

I guess there's still a lingering issue that this process will always allocate the SK buffers and copy into them, even if we aren't writing anything. This is fine for the test system, but not really ideal in production (where we only really expect to be writing SK data once in a blue moon)

rhaas80 (approved):

PR #1551, opened 2026-06-20
Kv/macos

kvand:
A smattering of updates to get kotekan compiling & running locally on my M3 macbook.

### DIFF
--- a/lib/core/errors.h
+++ b/lib/core/errors.h
@@ -58,6 +58,7 @@ extern int __enable_syslog;
 
 // to store error messages before exiting with error code
 extern char __err_msg[1024];
+#define MAX_LOG_MSG_LEN 1024
 extern const int __max_log_msg_len;
 
 void internal_logging_f(int log, const char* format, ...);
--- a/lib/core/errors.h
+++ b/lib/core/errors.h
@@ -57,9 +57,8 @@ extern int _global_log_level;
 extern int __enable_syslog;
 
 // to store error messages before exiting with error code
-extern char __err_msg[1024];
 #define MAX_LOG_MSG_LEN 1024
-extern const int __max_log_msg_len;
+extern char __err_msg[MAX_LOG_MSG_LEN];
 
 void internal_logging_f(int log, const char* format, ...);
 void exit_kotekan(enum ReturnCode code);
--- a/lib/stages/HFBTruncate.cpp
+++ b/lib/stages/HFBTruncate.cpp
@@ -15,9 +15,8 @@
 #include <cstdint>     // for int32_t, uint32_t
 #include <cstring>     // for memset, size_t
 #include <functional>  // for bind, function
-#if defined(__x86_64__) || defined(__i386__)
+#ifdef __AVX2__
 #include <immintrin.h> // for __m256, _mm256_div_ps, _mm256_loadu_ps, _mm256_set1_ps
-#include <mm_malloc.h> // for _mm_free, _mm_malloc
 #endif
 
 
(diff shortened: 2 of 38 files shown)


jbmertens (approved):
Changes look good to me, just needs linting merged (#1552) and conflicts resolved.

lib/core/errors.h
```diff
@@ -58,6 +58,7 @@ extern int __enable_syslog;
 
 // to store error messages before exiting with error code
 extern char __err_msg[1024];
+#define MAX_LOG_MSG_LEN 1024
```
andrerenard, line 60: If we are going to use a define instead, we should use it in all locations that `__max_log_msg_len` was used - and remove that variable.  Also while we are at it, use this define in the string above.

lib/stages/hdf5N2Write.cpp
```diff
@@ -1221,7 +1221,7 @@ hdf5N2Write::hdf5N2Write(kotekan::Config& config, const std::string& unique_name
 
 hdf5N2Write::~hdf5N2Write() {}
 
-std::uint64_t hdf5N2Write::_get_abs_file_idx(const N2FrameView& fv) const {
+size_t hdf5N2Write::_get_abs_file_idx(const N2FrameView& fv) const {
```
andrerenard, line 1224: Should this be fixed by adding an include for `unit64_t` instead?

lib/stages/HFBTruncate.cpp
```diff
@@ -52,11 +54,12 @@ HFBTruncate::HFBTruncate(Config& config, const std::string& unique_name,
 void HFBTruncate::main_thread() {
 
     frameID frame_id(in_buf), output_frame_id(out_buf);
-    const float err_init = 0.5 * err_sq_lim;
-
     float err, tr_hfb;
+#if defined(__x86_64__) || defined(__i386__)
```
andrerenard: Wouldn't it be better to use an AVX define?  e.g. `__AVX2__`?

lib/stages/HFBTruncate.cpp
```diff
@@ -88,15 +90,18 @@ void HFBTruncate::main_thread() {
         // Copy frame into output buffer
         auto output_frame = HFBFrameView::copy_frame(in_buf, frame_id, out_buf, output_frame_id);
 
-        // truncate absorber data and weights (8 at a time)
-        for (i_vec = 0; i_vec < int32_t(data_size) - 7; i_vec += 8) {
+        // truncate absorber data and weights (8 at a time on x86)
+        i_vec = 0;
+#if defined(__x86_64__) || defined(__i386__)
```
andrerenard: Same here?

lib/stages/rfiAVXVDIF.cpp
```diff
@@ -12,7 +12,9 @@
 #include "util.h"               // for e_time
 #endif
 
+#if defined(__x86_64__) || defined(__i386__)
```
andrerenard, line 15: Looks like this is in a lot of locations.

andrerenard (approved):
Minor comments.  Also needs to be updated with `develop` and linted before we can merge.

lib/core/errors.h
kvand: Good point -- updated throughout!

lib/stages/HFBTruncate.cpp
kvand: Updated to __AVX2__ guards throughout.

lib/stages/hdf5N2Write.cpp
kvand: On macos, size_t and uint64_t are apparently formally different types: `unsigned long` and `unsigned long long` -- they compile to 64b, but the compiler doesn't guarantee that, and the rest of this file uses size_t... We could swap the entire file to uint64_t, but it's a much more invasive edit.

(They're both `unsigned long` on linux, so the compiler is happe there.)

andrerenard: If it's consistently `size_t` for that class, then let's leave it that way as `size_t`

PR #1554, opened 2026-06-22
cudaPLMaskUpchannelizer: Add packet loss mask upchannelizer

eschnett:


eschnett:
Oh no! Erik submitted a PR without any autogenerated PTX code?

jbmertens:
I like the idea, but this can significantly increase CI time with a 53s test. I'd like to keep GPU CI tests to a few seconds, _maybe_ a couple tens, since the GPU tests can only run one-at-a-time across all PRs/commits.

If this is already a minimal exhaustive test, I might move this (along with possibly other tests) to nightly/weekly, once the chord branch becomes the default branch.

eschnett:
"I like the idea": This PR implements a new feature we need (upchannelizing the packet loss mask), so this isn't optional. I assume you like the idea of exhaustive testing? I have reduced the test coverage, and things should now run much faster. Also, there are currently no merge conflicts.

jbmertens:
Yes, referring specifically to the test there. It looks like it runs 10x faster now, great! Need to read this a bit more carefully, but will review soon.

jbmertens (approved):
LGTM. I have not thought through the actual upchannelization logic.
(I think there are notes on this I don't have access to?)

rhaas80:
Shouldn't it just copy? If a source time-sample is lost then all upsampled frequencies are lost. So far the upchannelization is not looking at any neighbouring frequencies or frames, is it? We do seem to have a proliferation of packet-loss type buffers now. They also seem to converge to CHIME's `lost_samples` buffer (ie one bool per time unit, no distinction between frequencies or feeds). :-P

eschnett:
There is a comment in the source code. There is also a CPU version that is very simple and "obviously correct"; that's one of the reasons the self-tests are so slow.

The upchannelizer looks at `M*U` input time samples to produce `1` output time sample. The mask propagates in this way -- any masked input sample means the output sample is masked.

rhaas80:
Right, not a "copy" but an "and" of the input mask. Still the logic should be very simple. I may be getting old, though even the obviously correct CPU code doe snot look "simple" to me :-)

eschnett:
The complicated bit is that this uses bit-wise addressing, so the time index needs to be split into a memory address and a bit address, with bit shifts and `&` and `|` and `~` to get and set bits.

PR #1555, opened 2026-06-22
various fixes to quantizer and frb beams codes

rhaas80:
* corrects number of beams per FRB network packet in quantizers
* corrects documentation in L0_L1_packet file
* removes "custom" names from parseReorderDefault and uses the CHIMEfoo names introduced recently
* Python fixes
* float16 / float64 support in testDataGenFloat
* new float support in givenDataGen (at least support without having to provide integer values that have the correct bit-patterns when interpreted as a float...)

### DIFF
--- a/lib/stages/givenDataGen.cpp
+++ b/lib/stages/givenDataGen.cpp
@@ -1,16 +1,11 @@
 #include "givenDataGen.hpp"
 
 #include "DataType.hpp"        // for type_total_bytes, string_to_type, DataType
-#include "DataType.hpp"        // for type_total_bytes, string_to_type, DataType
 #include "NDArray.hpp"         // for GenericNDArray, Config
 #include "StageFactory.hpp"    // for REGISTER_KOTEKAN_STAGE
-#include "StageFactory.hpp"    // for REGISTER_KOTEKAN_STAGE
 #include "bufferContainer.hpp" // for bufferContainer
-#include "bufferContainer.hpp" // for bufferContainer
-#include "chordMetadata.hpp"   // for chordMetadata, get_chord_metadata, CHORD_META_MAX_DIM
 #include "chordMetadata.hpp"   // for chordMetadata, get_chord_metadata, CHORD_META_MAX_DIM
 
-#include "fmt.hpp" // for format
 #include "fmt.hpp" // for format
 
 #include <assert.h>   // for assert
@@ -34,7 +29,8 @@ REGISTER_KOTEKAN_STAGE(givenDataGen);
 givenDataGen::givenDataGen(Config& config, const std::string& unique_name,
                            bufferContainer& buffer_container) :
     Stage(config, unique_name, buffer_container, std::bind(&givenDataGen::main_thread, this)),
-    _out_buf(get_buffer("out_buf")), _values(config.get<std::vector<int>>(unique_name, "values")),
+    _out_buf(get_buffer("out_buf")),
+    _values(config.get<std::vector<long double>>(unique_name, "values")),
     _name(config.get<kotekan::Symbol>(unique_name, "name")),
     _datatype(kotekan::string_to_type(config.get<std::string>(unique_name, "datatype"))),
     _array_shape(config.get<std::vector<ptrdiff_t>>(unique_name, "array_shape")),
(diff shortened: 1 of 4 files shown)


lib/stages/givenDataGen.cpp
```diff
@@ -76,12 +76,45 @@ void givenDataGen::main_thread() {
             break;
 
         const size_t type_size = kotekan::type_total_bytes(_datatype);
+        const bool is_float = kotekan::type_to_string(_datatype).find("float") != std::string::npos;
+        const bool is_int = kotekan::type_to_string(_datatype).find("int") != std::string::npos;
+        const bool is_complex = kotekan::type_to_string(_datatype)[0] == 'c';
+        assert(is_float + is_int == 1);
```
jbmertens: Consider throwing a FATAL error here as well for release builds.

jbmertens: (And other asserts below. Maybe worth having some logic that replaces the FATAL throws with aborts for release builds, or per some other compile flag.)

lib/testing/testDataGenFloat.cpp
```diff
@@ -46,55 +47,152 @@ testDataGenFloat::testDataGenFloat(Config& config, const std::string& unique_nam
...
+            sz *= s;
+        if (sz != buf->frame_size)
+            // clang-format off
+            throw std::invalid_argument("testDataGen: product of 'array_shape' config setting must equal the buffer frame size");
+        // clang-format on
+    }
+    _dim_name = config.get_default<std::vector<std::string>>(unique_name, "dim_name",
+                                                             std::vector<std::string>({"D"}));
+    if (_array_shape.size() != _dim_name.size()) {
+        throw std::invalid_argument("testDataGen: 'array_shape' and 'dim_name' config "
```
jbmertens, line 77: Consider FATAL_ERROR here.

lib/testing/testDataGenFloat.cpp
```diff
@@ -46,55 +47,152 @@ testDataGenFloat::testDataGenFloat(Config& config, const std::string& unique_nam
     _samples_per_data_set = config.get<uint32_t>(unique_name, "samples_per_data_set");
     _first_frame_index = config.get_default<uint32_t>(unique_name, "first_frame_index", 0);
     _gen_all_const_data = config.get_default<bool>(unique_name, "gen_all_const_data", false);
+
+    _name = config.get_default<std::string>(unique_name, "name", "E");
+
+    _value_type = kotekan::string_to_type(config.get_default<std::string>(unique_name, "value_type", "float32"));
+    assert(_value_type == kotekan::float16 ||_value_type == kotekan::float32 ||  _value_type == kotekan::float64);
```
jbmertens: Consider adding a user-facing error here. (FATAL_ERROR or so.)

lib/testing/testDataGenFloat.cpp
```diff
@@ -46,55 +47,152 @@ testDataGenFloat::testDataGenFloat(Config& config, const std::string& unique_nam
...
+
+        chordmeta->set_frame_counter(frame_id_abs);
+
+        // this needs the decoded type
+        // could be moved into constructor, but need the bit of code above
+        /* new style array description */
+        const std::vector<ptrdiff_t> extents(_array_shape.begin(), _array_shape.end());
+        const std::vector<kotekan::Symbol> dimnames(_dim_name.begin(), _dim_name.end());
+
+        buf->allocate_ndarray_frame_desc(chordmeta->type, _name, extents, dimnames);
```
jbmertens: Consider using ensure_frame_desc.

jbmertens (approved):
Good enough as-is, but a few small comments.

lib/testing/testDataGenFloat.cpp
rhaas80: there's a new variant? This is output, if anyone else outputs to the same buffer, it is a bug.

lib/testing/testDataGenFloat.cpp
rhaas80: given that this is my usual comment I can hardly refuse :-)

lib/stages/givenDataGen.cpp
rhaas80: not sure if that makes much of a difference. No such check should happen in a hot loop.

rhaas80: > Consider throwing a FATAL error here as well for release builds.

this stage should probably *not* be used in release runs. It is slow. For constant data a dedicated stage that creates that data may be more useful (and yes, now i will be asked to move it to `testing`, yes?).

jbmertens: We do use some fake/manual data generation in production CHORD currently (e.g. for the bad feed mask, which has no broker yet). So it is, arguably, useful not just for testing!

If we generate a single frame and just keep re-reading it, speed isn't as relevant.

lib/testing/testDataGenFloat.cpp
jbmertens: Ah, per #1529. The bufferFactory may now allocate the frameDesc, so ensure is meant for "allocate if it doesn't exist, then set/validate".

lib/stages/givenDataGen.cpp
rhaas80: Let me fix these then.

rhaas80: Done.

lib/testing/testDataGenFloat.cpp
rhaas80: it's a `throw` in `testDataGen`.

lib/testing/testDataGenFloat.cpp
rhaas80: Updated, under protest.

PR #1556, opened 2026-06-23
Send frame descs in buffer send/recv

jbmertens:
Optionally send and validate frame_descs in buffer send/recv.
Feature addition, should not affect current functionality.


docs/sphinx/dev/dev_buffers.rst
```diff
@@ -202,3 +202,17 @@ Avoid:
...
+Transmitting descriptors between nodes
+--------------------------------------
+
+``bufferSend`` and ``bufferRecv`` can carry a buffer's frame descriptor over the
+network so the receiver validates its config-declared descriptor against the
+sender's. Enable it with ``use_frame_desc: true`` on **both** stages -- a matched
+flag, like ``use_config_tracker`` -- and the wire format is unchanged when it is
+off (independent of the config tracker). The descriptor is serialized once per
+connection; on receive it is reconciled with ``ensure_frame_desc``, attaching it
+to an undeclared (``standard``) buffer or validating a declared one (fatal on
```
rhaas80: I thought these got renamed to "ndarray" buffers?

lib/metadata/FrameDesc.cpp
```diff
@@ -0,0 +1,33 @@
...
+        throw std::runtime_error("FrameDesc::deserialize: input too small for the wire tag");
+
+    uint32_t tag;
+    std::memcpy(&tag, bytes, sizeof(tag));
+    const char* payload = bytes + sizeof(tag);
+    const size_t payload_size = size - sizeof(tag);
+
+    switch (static_cast<WireType>(tag)) {
+        case WireType::generic_ndarray:
+            return GenericNDArray::deserialize_payload(payload, payload_size);
```
rhaas80: why "payload" in the name of the serialization function? it knows nothing of the existence of a `tag` before itself in memory.

lib/metadata/FrameDesc.hpp
```diff
@@ -3,12 +3,91 @@
...
+inline size_t str_size(const std::string& s) {
+    return sizeof(uint32_t) + s.size();
+}
+
+/// Bounds-checked sequential reader. Every accessor throws std::runtime_error
+/// rather than reading past the end, so it is safe on untrusted network input.
+class Reader {
+public:
+    Reader(const char* data, size_t size) : p(data), end(data + size) {}
+    uint16_t get_u16() {
```
rhaas80: I'd expect this to be all templated on types instead.

lib/metadata/FrameDesc.hpp
```diff
@@ -49,6 +128,40 @@ class FrameDesc {
     T* as() {
         return dynamic_cast<T*>(this);
     }
+
+    /// @name Wire serialization
```
rhaas80: I have to ask: why re-invent this wheel? We already have a couple of serialization / deserialization routines and formats. Wouldn't one of them do? This is not at all performance critical since it happens only once, right?

lib/metadata/N2FrameDesc.cpp
```diff
@@ -74,6 +74,76 @@ Symbol N2FrameDesc::get_quantity_name() const {
...
+    size_t n = 4 * sizeof(uint32_t) + wire::str_size(N2Layout_to_string(n2_layout));
+    if (layout_requires_product_list(n2_layout))
+        n += product_list.size() * 2 * sizeof(uint16_t);
+    return n;
+}
+
+void N2FrameDesc::serialize_payload(char* out) const {
+    out = wire::put_u32(out, num_elements);
+    out = wire::put_u32(out, num_ev);
+    out = wire::put_u32(out, num_products);
```
rhaas80: might need to check that these members are of sufficiently small type (just in case there are more `uint64_t` for say `num_elements` lurking in the code...)

lib/metadata/N2FrameDesc.cpp
```diff
@@ -74,6 +74,76 @@ Symbol N2FrameDesc::get_quantity_name() const {
...
+    // Encode the layout by name (matching the NDArray value_type encoding) so the
+    // wire form survives any reordering of the N2Layout enum.
+    out = wire::put_str(out, N2Layout_to_string(n2_layout));
+    // Subset layouts cannot be regenerated from num_elements alone, so send the
+    // explicit product list; other layouts the receiver regenerates (count 0).
+    if (layout_requires_product_list(n2_layout)) {
+        out = wire::put_u32(out, static_cast<uint32_t>(product_list.size()));
+        for (const auto& prod : product_list) {
+            out = wire::put_u16(out, prod.input_a);
+            out = wire::put_u16(out, prod.input_b);
```
rhaas80: seems strange to have `uint32_t` for elements but then `uint16` for the product indices (they are the same thing, aren't they)?

lib/metadata/N2FrameDesc.cpp
```diff
@@ -74,6 +74,76 @@ Symbol N2FrameDesc::get_quantity_name() const {
     return Symbol("N2");
 }
 
+FrameDesc::WireType N2FrameDesc::wire_type() const {
+    return FrameDesc::WireType::n2;
+}
+
+size_t N2FrameDesc::serialized_payload_size() const {
+    // num_elements, num_ev, num_products, product_list count (u32 each) + layout name
```
rhaas80: product list count is redundant for layouts that don't need it (since the layout type is also transmitted)

lib/metadata/N2FrameDesc.cpp
```diff
@@ -74,6 +74,76 @@ Symbol N2FrameDesc::get_quantity_name() const {
...
+    // dropped here.
+    const size_t expected_num_products = get_num_prod(num_elements, n2_layout, product_list);
+    if (num_products != expected_num_products)
+        throw std::runtime_error(fmt::format(
+            fmt("N2FrameDesc::deserialize: num_products {:d} inconsistent with layout (expected "
+                "{:d})"),
+            num_products, expected_num_products));
+
+    // The N2FrameDesc constructor validates layout/product_list consistency and
+    // throws on an unknown layout, so malformed input is rejected here.
```
rhaas80: well presumably you never get here b/c `get_num_prod` also aborts on unknown layout?

lib/metadata/NDArray.cpp
```diff
@@ -47,6 +94,85 @@ void GenericNDArray::output_framedesc(std::ostream& os) const {
...
+size_t GenericNDArray::serialized_payload_size() const {
+    size_t n = wire::str_size(type_to_string(get_value_datatype()));
+    n += sizeof(uint32_t);             // rank
+    n += get_rank() * sizeof(int64_t); // extents
+    n += wire::str_size(symbol_str(get_quantity_name()));
+    n += sizeof(uint32_t);             // dimnames count
+    for (const auto& dimname : get_dimnames())
+        n += wire::str_size(symbol_str(dimname));
+    n += sizeof(uint32_t);             // dimscalings count
+    n += get_rank() * sizeof(int64_t); // dimscalings
```
rhaas80: not correct. This *should* be true, but later you loop over `dimscalings` which could (would be a bug...) have size different from `extents`. Same for `dimname` I guess.

lib/metadata/NDArray.cpp
```diff
@@ -31,6 +31,53 @@ std::string format_vector(const std::vector<T>& vec) {
     return buf.str();
 }
 
+// An unset (empty) Symbol has no string; serialize it as a zero-length string.
+std::string symbol_str(const Symbol& s) {
+    return s.valid() ? s.get_string() : std::string();
```
rhaas80: seems strange to allow serialization of a Symbol explicitly marked as invalid. But sure.

(if valid was required, then `Symbol::get_string()` could be used, since it aborts for invalid Symbols)

rhaas80: also (I don't think this should really matter) means one cannot distinguish between invalid Symbols and Symbols of the empty string (which I'd admittedly also consider invalid)

lib/metadata/NDArray.cpp
```diff
@@ -47,6 +94,85 @@ void GenericNDArray::output_framedesc(std::ostream& os) const {
...
+        out = wire::put_i64(out, static_cast<int64_t>(scaling));
+}
+
+std::shared_ptr<const FrameDesc> GenericNDArray::deserialize_payload(const char* bytes,
+                                                                     size_t size) {
+    wire::Reader r(bytes, size);
+    const std::string type_name = r.get_str();
+    const DataType value_datatype = string_to_type(type_name);
+
+    // Bound rank and the dimname count before reserving so a corrupt length cannot
```
rhaas80: also dimscalings?

lib/stages/bufferRecv.cpp
```diff
@@ -485,11 +488,72 @@ void connInstance::internal_read_callback() {
...
 
-                    state = connState::metadata;
+                    state = use_frame_desc ? connState::frame_desc_size : connState::metadata;
                     bytes_read = 0;
                 }
 
             } break;
+            case connState::frame_desc_size: {
+                // Independent of the config tracker; only read when use_frame_desc
+                // is set on both ends. The size rides every frame (0 = no
```
rhaas80: not sure how I would know whether `use_frame_desc` is set on the *other* end. But presumably things will eventually fail in that case :-)

rhaas80: "rides"?

lib/stages/bufferRecv.cpp
```diff
@@ -485,11 +488,72 @@ void connInstance::internal_read_callback() {
...
+                    // Validate on read -- before the frame/drop decision -- so a
+                    // dropped frame still validates the descriptor. Malformed bytes
+                    // close the connection; a real config-vs-received mismatch is
+                    // fatal in ensure_frame_desc, the intended contract check.
+                    std::shared_ptr<const kotekan::FrameDesc> recv_desc;
+                    try {
+                        recv_desc = kotekan::FrameDesc::deserialize(frame_desc_space.data(),
+                                                                    frame_desc_size);
+                    } catch (const std::exception& e) {
+                        ERROR("Failed to deserialize frame descriptor from {:s}: {:s}", client_ip,
```
rhaas80: generally a bad idea. since it will make it impossible to actually backtrace into the cause of the failure using gdb.

lib/stages/bufferRecv.hpp
```diff
@@ -142,6 +142,9 @@ class bufferRecv : public kotekan::Stage {
     /// Whether to use the config tracker
     bool use_config_tracker;
 
+    /// Expect a serialized frame descriptor on the wire (matched with the sender)
```
rhaas80: "has to match with the sender"?

lib/stages/bufferSend.cpp
```diff
@@ -142,6 +145,42 @@ void bufferSend::main_thread() {
             config_tracker_combined_hash = ConfigTracker::instance().getTrackerHash();
             DEBUG2("Sent header: {:d}", n_sent);
 
+            // Send the frame descriptor once per connection (independent of the
+            // config tracker). The 4-byte size rides every frame for uniform
+            // framing; the serialized descriptor follows only on the first frame.
+            if (use_frame_desc) {
+                auto desc = desc_sent ? nullptr : buf->get_frame_desc();
```
rhaas80: this is all just to skip the while loop? why not `if (use_frame_dec && !desc_sent)`?

lib/stages/bufferSend.hpp
```diff
@@ -138,6 +142,9 @@ class bufferSend : public kotekan::Stage {
     /// Set to true if this is the first transmission
     std::atomic<bool> first_transmission_sent;
 
+    /// Set once the frame descriptor has been sent on the current connection
+    std::atomic<bool> desc_sent;
```
rhaas80, line 148: I am surprised this needs to be `atomic`. Where do the multiple threads using it exist?

rhaas80 (changes requested):

lib/metadata/FrameDesc.hpp
jbmertens: Indeed - json serialization is used now (same as metadata).

lib/metadata/FrameDesc.hpp
jbmertens: Using JSON serialization instead.

lib/metadata/FrameDesc.cpp
jbmertens: Using JSON serialization instead.

lib/metadata/N2FrameDesc.cpp
jbmertens: Should be ok with the json serialization, which range checks when parsing.

lib/stages/bufferRecv.cpp
jbmertens: It should be a FATAL_ERROR, yes, if that is what you mean.

rhaas80: the `catch` is what prevents being able to backtrace

jbmertens: I'm not too sure what gdb should be catching here, or what behavior would help. The backtrace would be lost to stack unwinding with or without the catch, right? Validation failures should now FATAL_ERROR (SIGTERM) where they fail so gdb can stop there, but I see that gdb can catch throws anyways.

If FATAL_ERROR should be a hard abort for debug builds, I'm fine with that, but I think that is another PR.

docs/sphinx/dev/dev_buffers.rst
jbmertens: Reworded some. `ndarray` was added, but `standard` was not removed and can still be used.

lib/metadata/N2FrameDesc.cpp
jbmertens: This comes from prod_ctype, just a vestigial convention not cleaned up here. Not as relevant with the json serialization.

lib/metadata/NDArray.cpp
jbmertens: Moved to Symbol.*pp. The optional ndarray fields make this possible, which is... awkward, I agree. For the scope of this PR I don't have a better fix. An empty string constructs an invalid Symbol (there is no valid empty-string symbol),

lib/stages/bufferRecv.cpp
jbmertens: The behavior is similar to the configTracker, yeah it should (now) cause a FATAL_ERROR.

"rides on the wire" is a claude-ism 🤷 I'm not sure where it comes up with some of its slang. Most of these edits were done by Claude, in case that isn't obvious 😅

lib/stages/bufferRecv.hpp
jbmertens: Reworded, the receiver's flag must match the sender.

lib/stages/bufferSend.cpp
jbmertens: Done.

lib/stages/bufferSend.hpp
jbmertens: The `main_thread` launches a `connect_thread`, both access desc_sent (same pattern as first_transmission_sent). These could both be made thread-local I suppose.

lib/metadata/N2FrameDesc.cpp
jbmertens: Right. Switches should now cover all cases, FATAL if something isn't caught.

jbmertens: Should be removed, and cleaned up / treated similarly in other switches.

lib/stages/bufferSend.hpp
rhaas80: I'd treat member variables the same as global variables, ie have as few as possible (for stages that means many things can be just local variables in `main_thread`).

lib/stages/bufferRecv.cpp
rhaas80: without the catch the backtrace is in fact not lost and gdb lets me go back to the exact location of the error. a `FATAL_ERROR` of course brings me only to the location of the `FATAL_ERROR` so that one now better have sufficient information present that I never have to ask "now what did *actually* cause this :- )).

rhaas80 (approved):
Doesn't really affect anything I plan to use, so: sure...

lib/stages/bufferRecv.cpp
jbmertens: I guess this applies to the try/catch in Stage::start as well?

Google tells me there is a "fork and dump" approach that gives us both a gentle shutdown and a dump,
https://mocelik.com/c/generating-and-debugging-core-dumps/
...could just make that the behavior for fatal errors.

PR #1557, opened 2026-06-23
chimefrb.jl: Correct twiddle factors

eschnett:


rhaas80:
I'll give it at try. Thank you!

rhaas80 (approved):
Works fine!

image

PR #1559, opened 2026-06-25
julia: correct poison check in upchan template

rhaas80:


eschnett (approved):
This change is correct, but the kernels also need to be regenerated.

rhaas80:
oh right. I think I ran Julia, let me get the update.

PR #1560, opened 2026-06-25
Clean up F-Engine input files

eschnett:


config/fengine/include/output_chime.j2
```diff
@@ -329,11 +329,6 @@ write_data:
         in_buf: host_n2_vis_buffer
         file_name: n2_vis
         create_single_file: false # not implemented
-        max_frames: 1 * num_frequencies
-    # write_n2_vis:
```
jbmertens, line 333: Should this be removed instead of just commented?

jbmertens (approved):
LGTM, small comment below.

config/fengine/include/output_chime.j2
eschnett: This PR removes commented-out code, and removes the `max_frames` option.

jbmertens: Ack, indeed!

PR #1561, opened 2026-06-26
Fix chime_upgrade_n2 config

ljgray:
- Use `extents` instead of 'shape' parameter
- Remove duplicate definitions of `host_rfi_SKtilde_buffer` and `host_rfi_SKtilde_ringbuffer`

### DIFF
--- a/config/chime_upgrade_n2.j2
+++ b/config/chime_upgrade_n2.j2
@@ -151,7 +151,7 @@ host_voltage_buffers:
   num_frames: baseband_buffer_depth
   value_type: int4x2_swapped_withoffset
   extents: [1, samples_per_data_set * num_elements * num_freq_host_buf / 2048, 2048]
-  dimnames: [F, T, E]
+  dimnames: ["F", "T", "E"]
   metadata_pool: main_pool
   use_hugepages: true
   numa_0:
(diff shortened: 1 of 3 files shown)


config/chime_upgrade_n2.j2
```diff
@@ -1127,30 +1127,15 @@ write_failed_vis:
     in_buf: failed_vis_buffer_post
 
 # System to optionally dump sk statistics data
-host_rfi_SKtilde_buffer:
```
rhaas80, line 1130: are these no longer needed?

ljgray: Ah - it's already defined on line 548, so this was a duplicate

rhaas80: I see.

rhaas80 (approved):

lib/stages/lostSamplesToN2Counts.cpp
```diff
@@ -130,6 +114,10 @@ lostSamplesToN2Counts::lostSamplesToN2Counts(Config& config, const std::string&
                     n2k_counts_buf->frame_size, _expected_counts_frame_size);
 
     // Set the frame description
+    rfi_mask_buf->require_frame_desc(kotekan::GenericNDArray::describe(
+        kotekan::uint1x8, "RFImask",
+        {static_cast<long>(samples_per_data_set) / 1024, static_cast<long>(num_n2k_freq), 1024 / 8},
```
rhaas80: `ptrdiff_t` according to the definition in https://github.com/kotekan/kotekan/blob/88420414ac4e400f5cd3df8606b42be27b489c51/lib/metadata/NDArray.cpp#L83

for is `ptrdiff_t` is `long` though (since we don't care about portability).

rhaas80 (approved):

PR #1563, opened 2026-06-26
Add optional config usage tracking

jbmertens:
## Summary

Adds an opt-in feature to track which config leaf items are actually read while a pipeline runs, and by which stage, plus the ability to **fail the run** when a config carries parameters no stage consumes. The `config/ci-tests` configs are cleaned of dead parameters and set to enforce this, so `run_tests.sh` rejects stale config going forward.

Set the top-level `log_config_usage`; at shutdown a summary lists the accessed items (with the requesting base paths, typically stage `unique_name`s) and the items never accessed. Off by default; no overhead unless enabled.

| value | behavior |
|---|---|
| `false` (default) | disabled |
| `true` / `info` | summary at INFO |
| `warn` / `error` | summary at that severity when any item went unused |
| `fatal_error` | **non-zero exit** when any item went unused (for CI) |

## Design

- **Single choke point.** All typed reads (`get<T>`, `get_default`, expression-variable lookups via `eval`) funnel through `Config::get_value(base_path, name)`, refactored to a single exit where accesses are recorded. Existing scope-walking resolution is unchanged (covered by the existing `test_config` cases).
- **Leaf, not branch**, attributed to the requesting `base_path`. A value resolved by walking up the tree is stored at its resolved leaf path.
- **Whole-block reads.** When `get_value` returns an *object* (e.g. `gpuProcess` reading its `in_buffers`/`out_buffers` map), every descendant leaf is recorded — the caller consumes the whole block.
- **Exclusions.** Framework dispatch markers (`kotekan_buffer`/`kotekan_stage`/`kotekan_metadata_pool`/`kotekan_update_endpoint`) and configUpdater-managed (`kotekan_update_endpoint`) blocks are excluded, since the factories and config updater consume those via a `get_full_config_json()` tree walk rather than `get()`.
- **Levels** parsed by `Config::parse_usage_report_level` (bool or string, case-insensitive; invalid value fails startup). The fatal path uses `exit_kotekan(FATAL_ERROR)` directly — `FATAL_ERROR_NON_OO` throws, and the summary runs from `~kotekanMode`, a `noexcept` destructor.

A `std::mutex` is intentionally not a `Config` member (would break the by-value `make_config` test helpers); the tracking map is guarded by a file-scope mutex.

## CI-test cleanup + enforcement

Removed parameters no stage reads in each pipeline — boilerplate `sizeof_*` constants, and genuinely-dead params verified against source: `retry_time` on `bufferSend`, `counts_max`/`counts_min` on `testN2kGen`, `input_order` on `hdf5FileWrite` writers, etc. (per-config: e.g. `frame_arrival_period` is dead in CPU configs but kept in GPU configs where `gpuProcess` reads it). Each config now sets `log_config_usage: fatal_error`.

## Testing

- `test_config` boost cases for access tracking and `parse_usage_report_level`.
- Ran all 21 `config/ci-tests/{cpu,gpu}_batch/*.yaml` on the real binary (2× A40): every config reports **N of N leaf items accessed** and exits 0. A negative test (injecting a bogus param) exits non-zero, and the level matrix behaves as specified (`info`/`warn`/`error`→exit 0, `fatal_error`→exit 1).

## Caveat

The summary reflects *runtime* accesses through `get()`. "Unused" means "not read in this execution" — a parameter read only on an untaken branch shows as unused. (This is the basis for the per-config cleanup above.)

### DIFF
--- a/lib/core/Config.cpp
+++ b/lib/core/Config.cpp
@@ -3,6 +3,7 @@
 #include "fmt.hpp"  // for compile_string_to_view, format, fmt
 #include "json.hpp" // for json, basic_json, iter_impl, operator>>
 
+#include <cctype>    // for tolower
 #include <cstddef>   // for size_t
 #include <cstdint>   // for int32_t, uint8_t
 #include <fstream>   // for basic_ifstream, basic_istream, ifstream
@@ -124,10 +125,10 @@ json Config::get_value(const std::string& base_path, const std::string& name) co
 
         json::json_pointer value_pointer(value_path);
         const json& value = _json.at(value_pointer);
-        // Note the access of a leaf value (not a branch/object) when tracking
-        // is enabled, attributed to the requesting base path.
-        if (_track_access && !value.is_object())
-            record_access(value_path, base_path);
+        // Note the access when tracking is enabled, attributed to the
+        // requesting base path. A whole-object fetch records all its leaves.
+        if (_usage_report_level != UsageReportLevel::off)
+            record_access(value_path, value, base_path);
         return value;
     }
     throw std::runtime_error(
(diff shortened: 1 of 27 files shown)


lib/core/Config.cpp
```diff
@@ -158,6 +173,150 @@ void Config::dump_config() const {
...
+    std::string accessed;
+    std::string not_accessed;
+    std::size_t n_accessed = 0;
+
+    {
+        std::lock_guard<std::mutex> lock(access_lock);
+        for (const std::string& leaf : leaves) {
+            auto it = _accessed.find(leaf);
+            if (it == _accessed.end()) {
+                not_accessed += fmt::format(fmt("\n    {:s}"), leaf);
```
eschnett: Repeated string concatenation is `O(n^2)`. I like to use `std::ostringstream` for this. But I think that code is not performance critical.

eschnett (approved):
I like the idea, this will help us a lot.

lib/core/Config.cpp
jbmertens: Switched `accessed`/`not_accessed`/`base_paths` to `std::ostringstream` in 5de98fdf.

PR #1564, opened 2026-06-27
Fix frame hand-off race in applyGains and visCompression

jbmertens:
Fixes the intermittent `test_applygains` CI failure, root-caused to a frame hand-off race in `applyGains` (and the same bug in `baselineCompression`/`visCompression`) when run with `num_threads > 1`.

The worker threads share wrapping `frame_id` counters and a single buffer producer/consumer registration, so when a counter wraps onto a slot another thread is still using, the buffer hands the same frame to two threads — duplicating an output frame (occasionally uninitialized). Both stages run `num_threads: 3` in production (`chime_science_run_recv.yaml`).

Fix: each worker takes exclusive ownership of its input and output slot before processing; the gain/stack computation still runs concurrently on distinct frames. `test_applygains` and `test_compression` now run these stages with `num_threads: 16` as a regression guard (the bug surfaced in ~84% of runs at that thread count, 0% after).

### DIFF
--- a/lib/stages/applyGains.cpp
+++ b/lib/stages/applyGains.cpp
@@ -34,7 +34,7 @@
 #include "prometheusMetrics.hpp"     // for Metrics, Counter, Gauge
 #include "restClient.hpp"            // for restClient
 #include "visBuffer.hpp"             // for VisFrameView, VisField
-#include "visUtil.hpp"               // for modulo, cfloat, double_to_ts, ts_to_double, frameID
+#include "visUtil.hpp"               // for cfloat, double_to_ts, ts_to_double
 #include "fmt.hpp"                   // for compile_string_to_view, format, fmt, format_string
 #include "gsl-lite.hpp"              // for span
 
@@ -57,9 +57,9 @@ REGISTER_KOTEKAN_STAGE(applyGains);
 applyGains::applyGains(Config& config, const std::string& unique_name,
                        bufferContainer& buffer_container) :
     Stage(config, unique_name, buffer_container, std::bind(&applyGains::main_thread, this)),
-    in_buf(get_buffer("in_buf")), out_buf(get_buffer("out_buf")), frame_id_in(in_buf),
-    frame_id_out(out_buf), update_age_metric(Metrics::instance().add_gauge(
-                               "kotekan_applygains_update_age_seconds", unique_name)),
+    in_buf(get_buffer("in_buf")), out_buf(get_buffer("out_buf")),
+    update_age_metric(Metrics::instance().add_gauge("kotekan_applygains_update_age_seconds",
+                                                    unique_name)),
     late_update_counter(
         Metrics::instance().add_counter("kotekan_applygains_late_update_count", unique_name)),
     late_frames_counter(
@@ -92,6 +92,11 @@ applyGains::applyGains(Config& config, const std::string& unique_name,
     num_threads = config.get_default<uint32_t>(unique_name, "num_threads", 1);
     if (num_threads == 0)
         throw std::invalid_argument("applyGains: num_threads has to be at least 1.");
+
+    // One "in use" flag per buffer slot, so two apply threads never process the
+    // same frame concurrently when the cursor wraps onto a slot still in use.
+    in_frame_busy.assign(in_buf->num_frames, false);
+    out_frame_busy.assign(out_buf->num_frames, false);
 
     // FIFO for gains and weights updates
     gains_fifo.resize(num_kept_updates);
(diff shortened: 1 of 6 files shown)


lib/stages/applyGains.cpp
```diff
@@ -263,14 +299,24 @@ void applyGains::apply_thread() {
         // Report number of frames received late and skip the frame entirely
         if (late) {
             late_frames_counter.inc();
-            std::lock_guard<std::mutex> lock_frame_ids(m_frame_ids);
             in_buf->mark_frame_empty(unique_name, input_frame_id);
-            input_frame_id = frame_id_in++;
+            release_frame(in_frame_busy, input_frame_id);
             continue;
```
rhaas80, line 320: this changes behaviour. The old old code increased `frame_id_in`, the new one dies not seem to do so (or only once one enters `claim_input_frame()`. Is that accounted for?

lib/stages/applyGains.hpp
```diff
@@ -154,6 +155,24 @@ class applyGains : public kotekan::Stage {
...
     std::mutex m_frame_ids;
 
+    /// Per-slot "in use" flags so two apply threads never process the same
+    /// buffer frame at once: the shared frame_id can wrap onto a slot another
+    /// thread is still holding, and the buffer would otherwise hand the same
+    /// frame to both. Guarded by m_frame_ids; frame_cv wakes a thread waiting
+    /// for a slot to be released.
+    std::condition_variable frame_cv;
+    std::vector<bool> in_frame_busy;
+    std::vector<bool> out_frame_busy;
```
rhaas80, line 162: I note that having more threads than frames will just have the extra threads stall. Since this essentially just implements a ring-buffer (since you never jump ahead) you could implement this with just a "next_frame" and "frames_available" pair (or similar set of two counters). But this will work as well I guess.

lib/stages/visCompression.cpp
```diff
@@ -285,12 +324,9 @@ void baselineCompression::compress_thread(uint32_t thread_id) {
...
 
-        // Get the current values of the shared frame IDs and increment them.
-        {
-            std::lock_guard<std::mutex> lock_frame_ids(m_frame_ids);
-            output_frame_id = frame_id_out++;
-            input_frame_id = frame_id_in++;
-        }
+        // Release the slots so other threads may reuse them.
+        release_frame(out_frame_busy, output_frame_id);
+        release_frame(in_frame_busy, input_frame_id);
```
rhaas80: same comments as above

tests/test_compression.py
```diff
@@ -176,16 +179,24 @@ def smallchime_data(tmpdir_factory):
...
 def test_metadata(diagonal_data):
 
-    freq_ids = np.array([frame.metadata.freq_id for frame in diagonal_data])
-    fpga_seqs = np.array([frame.metadata.fpga_seq for frame in diagonal_data])
-    dset_ids = np.array([frame.metadata.dataset_id for frame in diagonal_data])
-    nprod = np.array([frame.metadata.num_prod for frame in diagonal_data])
+    total_frames = diag_global_params["total_frames"]
+    freq_ids = diag_global_params["freq_ids"]
+
+    # baselineCompression runs multi-threaded (see diag_stage_params), so the
```
rhaas80: I think this runs on the N2 receiver and that one (at least for CHIME) won't be updated.

rhaas80 (approved):
sure

lib/stages/applyGains.cpp
jbmertens: Hm. Yes, output order might be different. Will aim to make it consistent...

lib/stages/applyGains.hpp
jbmertens: Yeah, before this change, adding more threads than frames made the tests reliably fail.

I'm not sure two scalars is sufficient, since frames aren't guaranteed to be released in order. Maybe there's a version that works where the threads are also ordered. But then a single thread stalling could bottleneck things. 🤷 Will think about it.

lib/stages/applyGains.cpp
jbmertens: Huh, so even before this PR there was no guarantee frames remain time-ordered, which is an assumption I've seen a number of stages make. Including the vis writer, so this is another bug (also likely unrealized in practice). So I guess I will also change the order intentionally...

jbmertens:
Pushed a modified version which addresses review comments. The changes _should_ affect behavior, but in a way that acts to preserve frame ordering (at the risk of a stall in one thread blocking other threads).

The code before this PR can alter frame ordering, which can cause downstream frames to be silently dropped by the vis writer stage. That is, CHIME's writer stage might see frames with a larger fpga_seq_num (later time) _before_ frames with a smaller fpga_seq_num (earlier time), and drop them. I think that is more likely to happen in production than the race condition, so may occasionally occur in practice.

rhaas80 (approved):

PR #1565, opened 2026-06-29
chimefrb: Correct twiddle factors

eschnett:
This time the twiddle factors were correct, but a desctructuring assignment in Julia was wrong.


julia/kernels/chimefrb.jl
```diff
@@ -1810,8 +1797,8 @@ if CUDA.functional()
     end
     fix_ptx_kernel()
 
-    # # Self-test (compares the kernel against a CPU reference beamformer)
-    # main(; run_selftest=true)
+    # Self-test (compares the kernel against a CPU reference beamformer)
+    main(; run_selftest=true)
```
rhaas80, line 1801: yes!

julia/kernels/chimefrb.jl
```diff
@@ -435,14 +435,6 @@ function make_chimefrb_kernel()
...
                     Γ2_dish2_1_dish5_0.re,
                     Γ2_dish2_1_dish5_0.im,
-                    Γ2_dish2_1_dish5_0.re,
-                    Γ2_dish2_0_dish5_0.im,
-                    Γ2_dish2_0_dish5_0.re,
-                    Γ2_dish2_1_dish5_0.im,
-                    Γ2_dish2_1_dish5_0.re,
-                    Γ2_dish2_0_dish5_1.im,
-                    Γ2_dish2_0_dish5_1.re,
-                    Γ2_dish2_1_dish5_1.im,
```
rhaas80, line 445: huh? Just deletes the elements? Sure if this fixes the issue. Why did this compile in the first place? There's not checks that would realize that the number of values available is larger than the expected one?

rhaas80 (approved):
Looks good to me. Passes my tests so far.

image

("48" is the correct value of the ratio, and this is a histogram showing differences Python / CUDA)

julia/kernels/chimefrb.jl
eschnett: Oh, no... You see, Julia is a convenient language. If the RHS has too many elements then the surplus element are silently ignored. What else could be the intended meaning of the code!

rhaas80: :scream:

PR #1567, opened 2026-06-30
minor updates and fixes

rhaas80:
* 5107f3560 - CHIMETelescope: correct error reporting code getting frequency map
* 0905a6429 - L0_L1_packet: correct number of beams per stream
* 6a28fb2c0 - parseReorderDefault: use "common" order names rather legacy ones
* e0573924e - frbbuffer: corect calling class method when constructing from file

the `parseReorderDefault` will cause failures if an explicit order had been used using the "old" names (eg beamformer instead of CHIMEBeamformer)


ljgray (approved):

jbmertens:
The failing test is unrelated - due to the IERS website being down. So, merging.

PR #1568, opened 2026-06-30
fixes to FRB quantizers for CHIME

rhaas80:
This pull request contains fixes (handling metadata, some incorrect size of arrays) for the quantizers for CHIME, both for the achromatic beamformer (in `cudaQuantize8`) and for the chromatic (legacy CHIME look-alike) beamformer (testQuantizeKernel8Chime).

### DIFF
--- a/lib/cuda/testQuantizeKernel8.cpp
+++ b/lib/cuda/testQuantizeKernel8.cpp
@@ -1,30 +1,31 @@
 // Test 8-bit FRB beam quantizer for CHIME, using the float16 FRB beamformer and CHIME's
 // sending-to-Bonsai network format
 
-#include <algorithm>                // for fill, fill_n
-#include <cassert>                  // for assert
-#include <cfloat>                   // for FLT_MAX
-#include <cmath>                    // for fabs, isfinite, isnan, fmax, fmin
-#include <cstdint>                  // for uint8_t
-#include <cstdlib>                  // for abort
-#include <string>                   // for allocator, string
-#include <vector>                   // for vector
-#include <functional>               // for function
+#include "Config.hpp"              // for Config
+#include "DataType.hpp"            // for float16_t
+#include "Stage.hpp"               // for Stage
+#include "StageFactory.hpp"        // for REGISTER_KOTEKAN_STAGE
+#include "bufferContainer.hpp"     // for bufferContainer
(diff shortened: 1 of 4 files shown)


lib/cuda/testQuantizeKernel8.cpp
```diff
@@ -291,7 +289,8 @@ class testQuantizeKernel8 : public kotekan::Stage {
                                                 beam, expected_offset, offset, scale_may_collapse);
 
                                 // Check absolute error of scale
```
eschnett: That's checking the relative error, not the absolute error.

lib/cuda/testQuantizeKernel8Chime.cpp
```diff
@@ -287,7 +285,8 @@ class testQuantizeKernel8Chime : public kotekan::Stage {
                                 }
 
                                 // Check absolute error of scale
```
eschnett: Ditto.

eschnett:
I do not see any array size changes for the legacy CHIME look-alike beamformer, only different tolerance settings.

rhaas80:
> I do not see any array size changes for the legacy CHIME look-alike beamformer, only different tolerance settings.

It is quite possible that I did not change `R8` to `R4` for legacy CHIME yet (have never gotten far enough to actually check it). Given it's intended use I am not even sure if a GPU based one is required or if one could just use the CPU based one in `frbPostProcess` and `frbNetworkSend`.

lib/cuda/testQuantizeKernel8.cpp
rhaas80: I'll let the author know :-)

```
934499609e (Erik Schnetter 2026-05-06 15:43:33 -0400 293)                                 // Check absolute error of scale
```

eschnett: This is your change: You are changing the RHS from `epsilon` to `epsilon * fabs(expected_scale)`, i.e. from absolute to relative error.

rhaas80: let me see. Oh I see, you are referring to code *after* the line to which the GitHub comment is attached to. GitHub had me confused.

Right, yes, that does change the check from absolute to relative error. Will update.

eschnett: I am referring to the comment which is now wrong and needs to be updated...

eschnett (approved):

PR #1570, opened 2026-07-02
Various corrections (mostly switch CHIME to CHIMETelescope)

eschnett:
- Remove not-needed `input_order` arguments for some kernels (we are always using the fiducial dish order)
- Do not set `input_order` in examples
- Correct some dimscaling arguments
- F-Engine simulator: Switch CHIME to be a CHIMETelescope, not a CHORDtelescope any more
- Use correct CHIME frequency channel numbers (0 ... 1023)
- Set telescope frequencies etc. (previously just used default values)
- Downsample before calculating CHIME HFB beams
- Update CHIME FRB quantizer dimscalings in j2 files


config/fengine/include/fengine_charts.j2
```diff
@@ -54,6 +53,11 @@ telescope:
 {%   endfor %}
 {% endfor %}
 
+    sampling_rate_MHz: adc_frequency / 1.0e+6
+    fft_length: num_samples_per_frame
+    min_science_freq_MHz: 1000 * adc_frequency / num_samples_per_frame / 1.0e+6
```
rhaas80, line 60: why the 1000?

config/fengine/include/fengine_charts.j2
```diff
@@ -54,6 +53,11 @@ telescope:
...
+    max_science_freq_MHz: 1671 * adc_frequency / num_samples_per_frame / 1.0e+6
```
rhaas80, line 61: I assume 1000 and 1671 are instrument characteristics?

config/fengine/include/fengine_chime.j2
```diff
@@ -76,7 +80,8 @@ num_frequencies: 16       # 2048 frequencies / 64 gpus * 50% coverage
 #     There are 2.56 channels per MHz, with a channel width of 390625 Hz.
 # (These particular channels were chosen with a fixed step size between these channel boundaries.)
 frequency_channels: [
-    1025, 1089, 1153, 1217, 1281, 1345, 1409, 1473, 1537, 1601, 1665, 1729, 1793, 1857, 1921, 1985,
+    # 1025, 1089, 1153, 1217, 1281, 1345, 1409, 1473, 1537, 1601, 1665, 1729, 1793, 1857, 1921, 1985,
+    63, 127, 191, 255, 319, 383, 447, 511, 575, 639, 703, 767, 831, 895, 959, 1023
 ]
```
rhaas80, line 83: not really for this pull request, but in general: this number will not be constant across notes. Does the current code handle the case where there are frequency channels that are never encountered? Depending on where it is used this may be changing at runtime (if one takes the flagged frequencies into account).

config/fengine/include/fengine_chime.j2
```diff
@@ -52,6 +51,11 @@ telescope:
 {%   endfor %}
 {% endfor %}
 
+    sampling_rate_MHz: adc_frequency / 1.0e+6
+    fft_length: num_samples_per_frame
+    min_science_freq_MHz: 1025 * adc_frequency / num_samples_per_frame / 1.0e+6
+    max_science_freq_MHz: 2048 * adc_frequency / num_samples_per_frame / 1.0e+6
```
rhaas80: I note that `adc_frequency` and `num_samples_per_frame` are still not what the actual hardware does, they are prentend ones that give the correct result one takes a ratio. This will be confusing to anyone familiar with the instrument.

config/fengine/include/hfb_chime.j2
```diff
@@ -189,10 +218,11 @@ run_hfb23:
           frb2_num_frequencies: upchan_hfb_upchan_U128_max_num_output_channels
           frb2_num_beams: hfb2_num_beams
           frb2_num_times: hfb_num_output_times
-          in_signal: host_hfb1_beams_ringbuffer
-          frb1_beams_name: hfb1_beams
+          in_signal: host_hfb1_accumulated_beams_ringbuffer
+          frb1_beams_name: hfb1_accumulated_beams
           frb2_weights_name: hfb2_weights
           frb2_beams_name: hfb2_beams
+          frb_downsampling_factor: hfb_downsampling_factor * hfb_second_downsampling_factor
```
rhaas80, line 225: incovenient name, not indicating which stage in the FRB pipeline this is done.

julia/kernels/frb_template.cxx
```diff
@@ -555,7 +549,8 @@ cudaEvent_t cuda{{{kernel_name}}}::execute(cudaPipelineState& /*pipestate*/, con
         // First obtain from the Telescope the grid indices of every dish in the input data.
         // The input order is a Beamformer type (P slow, D fast), so the first num_dishes
         // entries will contain the grid indices for all connected dishes.
-        std::vector<grid_idx_2d_t> grid_indices = tel.get_main_array_grid_indices(cuda_number_of_dishes, input_order);
+        std::vector<grid_idx_2d_t> grid_indices =
+            tel.get_main_array_grid_indices(cuda_number_of_dishes, tel.fiducial_element_order());
```
rhaas80, line 553: could call it `tel.beamformer_element_order()`. There may be other fiducial ones.

lib/cuda/cudaHFB1Accumulate.cpp
```diff
@@ -0,0 +1,215 @@
...
+ *   @gpu_mem_type         @c float16
+ *   @gpu_mem_dim_name     [@c Ttilde][@c Fbar][@c beamQ][@c beamP]
+ * @gpu_mem Output time-averaged beams
+ *   @gpu_mem_buffer       @c ring
+ *   @gpu_mem_quantity     @c I
+ *   @gpu_mem_type         @c float16
+ *   @gpu_mem_dim_name     [@c Ttilde][@c Fbar][@c beamQ][@c beamP]
+ * @conf  buffer_depth                 Int.  The number of GPU frames used for pipelining.
+ * @conf  hfb_second_downsampling_factor                    Int.  Number of input time samples to
+ * average per output.
```
rhaas80: needs a comment explaining that the only value supported is that which averages down to a single sample for the whole input frame

lib/cuda/cudaHFB1Accumulate.cpp
```diff
@@ -0,0 +1,215 @@
...
+    // collapses them to a single averaged sample, so its Ttilde scaling grows by
+    // `hfb_second_downsampling_factor`.
+    hfb1_beams(hfb1_beams_name, "I",
+               std::array<std::ptrdiff_t, 4>{buffer_depth * hfb_second_downsampling_factor
+                                                 * num_output_times,
+                                             num_frequencies, frb1_num_beams_Q, frb1_num_beams_P},
+               std::array<std::string, 4>{"Ttilde", "Fbar", "beamQ", "beamP"},
+               {hfb_downsampling_factor, 1, 1, 1}, *this),
+    hfb1_accumulated_beams(
+        hfb1_accumulated_beams_name, "I",
```
rhaas80, line 118: different quantity (not identical to "I" or a subset of the values of "I")

lib/cuda/cudaHFB1Accumulate.cpp
```diff
@@ -0,0 +1,215 @@
...
+    if (pos_in + hfb_second_downsampling_factor > in_nd.get_extent(0))
+        FATAL_ERROR("hfb1_beams read window would wrap the ring buffer: pos_in ({:d}) + "
+                    "hfb_second_downsampling_factor ({:d}) > ring size ({:d}). The kernel is not "
+                    "ring-aware; size the ring as a multiple of hfb_second_downsampling_factor.",
+                    pos_in, hfb_second_downsampling_factor, in_nd.get_extent(0));
+
+    float16_t* const in_memory = in_nd.data() + pos_in * in_nd.get_stride(0);
+    float16_t* const out_memory = out_nd.data() + pos_out * out_nd.get_stride(0);
+
+    launch_accumulate_hfb1(reinterpret_cast<__half*>(out_memory),
```
rhaas80: if this is to ensure that `__half` is actually `float16_t` then it won't do it. For that an `is_same` would seem to be required.

lib/cuda/cudaHFB1Accumulate.cu
```diff
@@ -0,0 +1,86 @@
...
+        *reinterpret_cast<int4*>(p) = reinterpret_cast<const int4&>(v);
+    }
+
+} // namespace
+
+__global__ void accumulate_hfb1(__half* __restrict__ const accumulator,
+                                const __half* __restrict__ const input, const int num_frequencies,
+                                const int num_times) {
+  // Array indexing:
+  //     input[F][PQ][T]
```
rhaas80: middle endian order? In CUDA code which has C order `input[T][F][PQ]` seems better since that's the actual memory layout if the calculation of `input_offset` is to be believed

lib/cuda/cudaHFB1Accumulate.cu
```diff
@@ -0,0 +1,86 @@
...
+  //     accumulator[F][PQ]
+  const int frequency = blockIdx.x;
+  const int pq = 8 * threadIdx.x;
+
+  half2x4 accum;
+  accum.x = accum.y = accum.z = accum.w = __float2half2_rn(0.0f);
+  for (int time = 0; time < num_times; ++time) {
```
rhaas80, line 47: is there a danger of running out of range if `num_times` is very large?

lib/cuda/cudaHFB1Accumulate.cu
```diff
@@ -0,0 +1,86 @@
...
+                            const int num_frequencies, const int num_times,
+                            const cudaStream_t stream) {
+    assert(accumulator);
+    assert(input);
+    assert(num_frequencies > 0);
+    assert(num_times > 0);
+
+    // Each thread owns 8 contiguous halves; one block handles one frequency's P*Q halves.
+    const dim3 nthreads(P * Q / 8, 1, 1); // 512 threads = 4096 halves / 8
+    const dim3 nblocks(num_frequencies, 1, 1);
```
rhaas80, line 82: I'd actually move (or repeat) the assumption checks on the sizes) here. So that the same source code files contains the code with the assumption and the assertion on the assumption.

lib/stages/hdf5FileRead.cpp
```diff
@@ -155,7 +155,7 @@ class hdf5FileRead : public kotekan::Stage {
             std::vector<std::array<double, 3>> feed_positions_m(dims.at(0));
             attr.read_raw(reinterpret_cast<double*>(feed_positions_m.data()));
             const auto& expected_feed_positions_m =
-                telescope.get_feed_positions_m(num_dishes, ElementOrder::CHORDBeamformer);
+                telescope.get_feed_positions_m(num_dishes, telescope.fiducial_element_order());
             if (feed_positions_m != expected_feed_positions_m)
```
rhaas80, line 159: This will (still) not let me save data to a HDF5 file in different order (eg the voltages before the xpose stage) and re-read them.

lib/stages/parseReorderDefault.cpp
```diff
@@ -51,15 +51,16 @@ parseReorderDefault::parseReorderDefault(Config& config, const std::string& uniq
...
+            {"D", "P", "D"}, {_num_dishes / _num_chime_cylinders, 1, 1}));
     } else if (_input_order == ElementOrder::CHIMEBeamformer) {
         _out_buf->require_frame_desc(kotekan::GenericNDArray::describe(
             kotekan::int32, _name, {_num_polarizations, _num_dishes}, {"P", "D"}, {1, 1}));
     } else {
         FATAL_ERROR("Unexpected input_order {:s}", _input_order);
     }
 #else
-    // this is what xpose2048 expects
+    // xpose2048 expects its input in correlator order {"E"} and will produce output in beamformer
```
rhaas80, line 62: not true. Output depends on the scattering map

rhaas80 (approved):
It's no worse than the current code. Still fails to save the non-xpose voltages to HDF5 and read them back in (since those are not in the fiducial order). Also would fail if I was to store visibility matrix data in the order expected by the following stages (which is CHIMECylinder not CHIMEBeamformer)

config/fengine/include/rfi_chime.j2
```diff
@@ -246,3 +246,20 @@ run_recv_rfi_RFImask:
           output_size: num_frequencies * num_times / 8
           ring_buffer_size: buffer_depth * output_size
           out_buf: host_rfi_RFImask_buffer
+
+# TODO: copy this stanza to the other telescopes; remove them from output_*.j2
```
jbmertens: Flagging in case this TODO should be done.

lib/cuda/CMakeLists.txt
```diff
@@ -28,6 +28,8 @@ add_library(
     cudaCopyToRingbuffer.cpp
     cudaCopyNToRingbuffer.cpp
     cudaFRBBeamReformer.cpp
+    cudaHFB1Accumulate.cpp
```
jbmertens, line 31: Why the 1 in HFB1, or FRB1? A name that indicates intent would be nice.

lib/cuda/cudaHFB1Accumulate.cpp
jbmertens: Or if this isn't tunable, it shouldn't be exposed as a config parameter.

lib/cuda/cudaHFB1Accumulate.cpp
```diff
@@ -0,0 +1,215 @@
...
+cudaHFB1Accumulate::cudaHFB1Accumulate(kotekan::Config& config, const std::string& unique_name,
+                                       kotekan::bufferContainer& host_buffers,
+                                       cudaDeviceInterface& device, const int instance_num) :
+    cudaCommand(config, unique_name, host_buffers, device, instance_num, no_cuda_command_state,
+                "cudaHFB1Accumulate"),
+    // Parameters
+    buffer_depth(config.get<int>(unique_name, "buffer_depth")),
+    hfb_second_downsampling_factor(config.get<int>(unique_name, "hfb_second_downsampling_factor")),
+    num_frequencies(config.get<int>(unique_name, "num_frequencies")),
+    frb1_num_beams_P(config.get<int>(unique_name, "frb1_num_beams_P")),
```
jbmertens, line 101: If `P = 2*256` and `Q = 2*4`, why are these read from config? Will the kernel be generalized later?

config/fengine/include/fengine_charts.j2
eschnett: they use the second nyquist zone, same as chime. to make this work with the CHORDTelescope object you need to add the minimum channel number. chime/chord uses powers of 2; they don't, they have a very different mapping.

config/fengine/include/fengine_charts.j2
eschnett: yepp

config/fengine/include/fengine_chime.j2
eschnett: i think this is mostly used when inventing voltages. otherwise all the frequency channel assignments come from the voltage buffer.

config/fengine/include/fengine_chime.j2
eschnett: yes, that should be corrected.

config/fengine/include/hfb_chime.j2
eschnett: which name is inconvenient?

julia/kernels/frb_template.cxx
eschnett: no, there is only ever one "fiducial" order. that is what "fiducial" means.

lib/cuda/cudaHFB1Accumulate.cpp
eschnett: no, it's the same quantity. "I" is an intensity beam. the other weird names ("I2" etc.) should be changed to be "I" too. we currently encode which stage produced that intensity, but that's not a good idea. ideally there would be a julia script that sees the name "I" and then knows how to plot it.

lib/cuda/cudaHFB1Accumulate.cpp
eschnett: that cast should go away.

lib/stages/N2Accumulate.cpp
```diff
@@ -96,12 +96,11 @@ N2Accumulate::N2Accumulate(Config& config, const std::string& unique_name,
     _num_elements(_num_polarizations * _num_dishes),
     _num_workers(config.get_default<int>(unique_name, "num_workers", 1)),
     _do_fringestop(config.get_default<bool>(unique_name, "do_fringestop", false)),
-    _input_order(config.get<ElementOrder>(unique_name, "input_order")),
     _variance_mode(config.get<N2VarianceMode>(unique_name, "variance_mode")),
     _debug_accum_mode(config.get_default<bool>(unique_name, "debug_accum_mode", false)),
     _profile_info(config.get_default<bool>(unique_name, "profile_info", false)),
     _tel(Telescope::instance()),
-    _feed_positions_m(_tel.get_feed_positions_m(_num_elements, _input_order)),
+    _feed_positions_m(_tel.get_feed_positions_m(_num_elements, _tel.fiducial_element_order())),
```
jbmertens, line 81: the chord_pathfinder.j2 (and any other configs using CHORDEarly, just config/tests/chord_pathfinder_core.j2 it looks like?) will need to be updated to accommodate this.

jbmertens (dismissed):

lib/cuda/cudaHFB1Accumulate.cu
eschnett: in principle yes. another option would be to accumulate using 32-bit floats. this whole stage is memory bandwidth limited so float would come at zero cost.

lib/stages/hdf5FileRead.cpp
eschnett: the voltages before reordering have index `[E]` ("element"), after reordering they have `[P][D]`. it's not possible to describe the chime input via simply "dishes" and "polarizations" since these are mixed.

i think it should work to save the voltages – that's what the F-Engine simulator does, and that's what HDF5 reads.

lib/stages/parseReorderDefault.cpp
eschnett: yes. that scattering map defines how to go from correlator to beamformer order.

lib/cuda/CMakeLists.txt
eschnett: that stage post-processes the output of the `HFB1` stage (which is the `FRB1` kernel). we can change the name to something else, the accumulator itself is quite generic.

jbmertens (commented):
(Marking review as comment - config changes can be made later.)

lib/cuda/cudaHFB1Accumulate.cpp
eschnett: these are read from the config as safety measure. if the config has wrong values then the stage will abort instead of silently doing the wrong thing. (it could be made tunable, but i didn't because we wouldn't use that feature.) maybe we will make it tunable if other instruments also want such an accumulator, maybe Hirax will?

jbmertens: I hope CHORD will have an HFB pipeline at some point, maybe HIRAX, although that is far future.

lib/stages/N2Accumulate.cpp
eschnett: if we want to continue to use the CHORDEarly memory layout, which is a `[D][P]` layout, then i absolutely insist that we describe this properly in the metadata as `[D][P]`, and not as `[P][D]`.

we put a lot of effort into getting the metadata correct. we cannot now being to ignore the metadata, creating wrong metadata because this is more convenient, and rely one a setting defined in a parameter file to describe the content of any data files.

in other words: we already have a way to describe the "CHORDEarly" layout, we do not need to invent a new name for this.

jbmertens: Ah, I'm not suggesting a new name or code changes, but rather updating the chord_pathfinder.j2 so it no longer uses CHORDEarly at all. Which can be a separate PR.

config/fengine/include/hfb_chime.j2
rhaas80: `frb_downsampling_factor` (in that most other names give an indication about which stages is looking at it)

julia/kernels/frb_template.cxx
rhaas80: ok, I had to look up the word... https://www.merriam-webster.com/dictionary/fiducial claims "taken as standard of reference" which I take is the meaning used here, but also "founded on faith or trust" if which there can be multiples of. 

How about "[canonical](https://en.wiktionary.org/wiki/canonical_form#English)" since it is only a convention and not that I trust one order any more than any other?

config/fengine/include/fengine_charts.j2
eschnett: I added a comment explaining this.

config/fengine/include/hfb_chime.j2
eschnett: This parameter name is defined by the stage which runs, which is called `CHIMEFRB`. We can rename the stage, but not in this cleanup PR.

julia/kernels/frb_template.cxx
eschnett: This function is defined here https://github.com/kotekan/kotekan/blob/0b2a66aa8c3416344cb34fac848f8761e208aabe/lib/utils/Telescope.hpp#L420. Changing its name would be a separate PR.

lib/cuda/cudaHFB1Accumulate.cpp
eschnett: This doesn't process a "whole input frame", it processes a ring buffer. At each iteration it bites off `hfb_second_downsampling_factor` samples and produces `1` output sample. That's an internal decision that's not visible to the outside, only the downsampling factor is.

The downsampling factor is tunable.

lib/cuda/cudaHFB1Accumulate.cpp
eschnett: Removed.

lib/cuda/cudaHFB1Accumulate.cu
eschnett: Comment fixed.

lib/cuda/cudaHFB1Accumulate.cu
eschnett: At the moment `num_times` is 20 for testing, which accumulates to 800 ms. I don't know what future usage will be. If people want to accumulate to 10 s, then a different algorithm is necessary. Maybe the whole FRB2 stage will need to be changed to 32-bit floats.

config/fengine/include/rfi_chime.j2
eschnett: Good idea. Done.

lib/cuda/CMakeLists.txt
eschnett: The original FRB names: In CHORD, we split the CHIME FRB stage into 3 bits, named FRB1, FRB2, and FRB3. They run in this order. FRB1 is the FFT intensity beamformer, FRB2 is the "beam re-former", and FRB3 is the "beam quantizer".

We can rename the stages. I wouldn't want to do it in this PR, though.

config/fengine/include/fengine_chime.j2
eschnett: Fixed.

lib/cuda/cudaHFB1Accumulate.cu
eschnett: The functions in the file `cudaHFB1Accumulate.cu` are only called from `cudaHFB1Accumulate.cpp`. The only reason they are in different files is that one contains CUDA, the other contains C++ code.

To check these assumptions, one would pass the run-time values as arguments. Otherwise the CUDA kernel does not have access to the run-time values. The CUDA kernel is a stand-alone function, not a stage, and only sees the arguments that are passed in.

Otherwise you say "don't expose parameters if there is only one legal value". Here you suggest "add parameters even if there is only one legal value"...

lib/stages/N2Accumulate.cpp
eschnett: That's a good idea.

PR #1571, opened 2026-07-02
Add n_valid, n_pl, n_rfi, and n_rfi_only prometheus gauges to `N2Accumulate`

ljgray:


lib/stages/N2Accumulate.cpp
```diff
@@ -1003,6 +989,13 @@ bool N2Accumulate::output_and_reset(frameID& in_frame_id, frameID& in_rfiframema
         }
         N2FrameView out_vis(out_buf, out_frame_id + f);
 
+        // Update some prometheus metrics here, since metadata is already available
```
ljgray, line 1003: Open to arguments about putting this elsewhere (in `main_thread`?)

jbmertens: Just at output is fine with me, so this seems good. I think we'll usually be interested in something like fractions of frames lost rather than an absolute number at arbitrary times. It could go in the main_thread as a fraction? But this is good as-is too.

jbmertens (approved):

ljgray:
@jbmertens If I'm interpreting things correctly, we would just have to divide by `ticks_in_accum` to convert to a fraction, correct? If so, I'm perfectly happy to just produce these as fractions instead.

There's also the `skipped_frame_counter` metric that Geoff added in the header but didn't get a chance to implement - I think I'll update this PR to include that metric as well.

lib/stages/N2Accumulate.cpp
```diff
@@ -1,51 +1,29 @@
...
-#include "buffer.hpp"            // for Buffer
 #include "bufferContainer.hpp"   // for bufferContainer
-#include "bufferContainer.hpp"   // for bufferContainer
-#include "chordMetadata.hpp"     // for chordMetadata, get_chord_metadata
 #include "chordMetadata.hpp"     // for chordMetadata, get_chord_metadata
 #include "dataset.hpp"           // for dset_id_t
-#include "dataset.hpp"           // for dset_id_t
 #include "div.hpp"               // for div_ceil, num_triangle_blocks
-#include "div.hpp"               // for div_ceil, num_triangle_blocks
-#include "kotekanLogging.hpp"    // for FATAL_ERROR, DEBUG, FATAL_ERROR_NON_OO, INFO
```
jbmertens, line 36: Noticing a lot of headers are removed here - is this intentional?

ljgray: This was just automatic from clang-format - looking back at `develop`, these were all duplicates for whatever reason

jbmertens:
Sounds good! Adding `skipped_frame_counter` is great too. I'm happy with either the fraction or absolute number - they are ~equivalent, maybe the fraction is a bit easier to understand from someone working in (only) grafana?

lib/stages/N2Accumulate.cpp
```diff
@@ -102,8 +80,14 @@ N2Accumulate::N2Accumulate(Config& config, const std::string& unique_name,
     _profile_info(config.get_default<bool>(unique_name, "profile_info", false)),
     _tel(Telescope::instance()),
     _feed_positions_m(_tel.get_feed_positions_m(_num_elements, _input_order)),
-    skipped_frame_counter(Metrics::instance().add_counter(
-        "kotekan_N2accumulate_skipped_frame_total", unique_name, {"freq_id", "reason"})) {
```
rhaas80, line 106: since the commit removes an existing metric: I assume no one is using this anyway yet?

rhaas80 (approved):

lib/stages/N2Accumulate.cpp
ljgray: This was included in `N2Accumulate.hpp` with a `TODO`, but was never actually used. It's essentially replaced with `kotekan_N2accumulate_skipped_frames`, which is initialized in `main_thread` and provides a count of skipped input frames for each output frame, instead of just an absolute count

PR #1573, opened 2026-07-03
misc updates including first set of FRB input files

rhaas80:


config/chime_upgrade_frb.j2
```diff
@@ -0,0 +1,3917 @@
+##########################################
+#
+# chime_upgrade_n2
```
ljgray, line 3:
```suggestion
# chime_upgrade_frb
```

config/chime_upgrade_frb.j2
```diff
@@ -0,0 +1,3917 @@
...
+#
+# Pipeline for running the n2k correlator, accumulate onto a visibility matrix
```
ljgray, line 5: Should also mention FRB beamforming

ljgray (approved):
I'll trust that the beamforming config works, but everything else looks good.

I assume that this is intended to be an incremental update to `chime_upgrade_n2`, instead of an independent config. If that's the case, should we think about combining the chime_upgrade_<...> configs into a single config at some point soon?

rhaas80:
> I'll trust that the beamforming config works, but everything else looks good.
> 
> I assume that this is intended to be an incremental update to `chime_upgrade_n2`, instead of an independent config. If that's the case, should we think about combining the chime_upgrade_<...> configs into a single config at some point soon?

Uh, yes. The idea was keep he working configs untouched so as to not accidentally clobber them while hacking away at a new thing. I *think* I had done a `gvimdiff` between `n2` and `frb` configs to bring `frb` back in sync with `n2` (for the `extents` if for nothing else). 

Generally a single file would be nice. What I would try and avoid is many `include` since they tend to hide information. Similarly I would try and avoid defining variables at global scope as much as possible and instead try and have a `n2:`, `frb:` section instead, each of which can define its own helpers without running into danger of variables being visible elsewhere. Similarly the buffer declarations could be clustered this way. At least that's my hope and wishes. No idea if that is actually practical or not.

rhaas80:
> I'll trust that the beamforming config works, but everything else looks good.

The beamforming stuff may well fail if I forgot to shift everything back from `dataGenFewHot` to reading FEngine simulator files. But it at least shows all the bits and pieces required.

PR #1574, opened 2026-07-06
Include external gain interface in chime_upgrade_frb config

ljgray:
Mechanically, this works (requires a few cherry-picked commits from `rhaas/frbNetwork4` in order to use the `FewHot` test stage). How it affects the data will require additional tests.

There are two sets of "gains" discussed in the context of the FRB beamformer - first, those produced by `setUpchanGain`. These are just a single, real value for each upchannelized frequency, applied to the frequency after upchannelizing (assuming my reading of the kernels is correct). For CHIME's sake, I think we just leave these all to be 1.

Then, there are the FRB1 phases, which were previously produced by `setFRB1Phase`. As far as I can tell (again, from reading the kernels), these are effectively the complex per-element calibration gains (times some scaling factor?). `setFRB1Phase` just produces some constant value (equal to the scaling factor), so I've updated `processFeedGains` to include a fixed, real scaling factor as well.

The one potential _TODO_ here involves the `upchan_min_channel` and `upchan_max_channel` parameters in `setFRB1Phase`, but I don't think these are really used here for CHIME in a way that matters (they're just set to the range [0, 16) for an upchan factor of 16).

This PR also only includes gains for the FRB beamformer stage - full config (including HFB and tracking beams as well) is still demonstrated in the `chime_upgrade_test_gains` config file (which I'll keep around for now).

### DIFF
--- a/config/chime_upgrade_frb.j2
+++ b/config/chime_upgrade_frb.j2
@@ -952,13 +952,8 @@ rfi_broadcast:
     rfi_sktilde_buf: host_rfi_SKtilde_buffer
     rfi_skbar_buf: host_rfi_SKbar_buffer
     rfi_mask_buf: host_rfi_RFImask_buffer
-    # 3-sigma deviations seems reasonable for identifying outliers,
-    # and empirically produces vaguely comparable results to the
-    # old system
-    num_sigma_deviations: 3
-    # Simply selected from the old system, and accounted for by
-    # the receiver. Can be increased or decreased based on the rate
-    # at which we want to be sending packets
+    # Same value as legacy system. Can be increased or decreased based
+    # on the rate at which we want to be sending packets
     frames_per_packet: 10
     destination_port: "3000"
     # This is the IP of `vm-steelpan-recv1`. Needs to be updated when
--- a/config/chime_upgrade_frb.j2
+++ b/config/chime_upgrade_frb.j2
@@ -89,6 +89,10 @@ earth_rotation_data:
     - {t_inst_ns: 0, delta_UT1_inst: 0.0, xp_as: 0.0, yp_as: 0.0}
 
 updatable_config:
+  frb_gain:
+    0:
+      kotekan_update_endpoint: "json"
+      gain_directory: /mnt/frb-archiver/daily_gain_solutions/Latest_FRB
   # bad input mask, assumed to be in cylinder order
   bad_inputs:
       kotekan_update_endpoint: json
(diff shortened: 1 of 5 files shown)


config/chime_upgrade_frb.j2
```diff
@@ -1504,17 +1513,27 @@ FRB:
...
         upchan_factor: factor_upchan_out
-        # TODO: replace the stage with one that actually takes the gains into account
-        # TODO: need to worry about over / underflow
-        frb1_input_scale: 1.0 / 32
-        frb1_swap_MN: true
-        upchan_min_channel: upchan_U16_min_channel
-        upchan_max_channel: upchan_U16_max_channel
-        upchan_max_num_channels: upchan_U16_max_num_channels
-        frb1_phase: host_frb1_U16_phase_buffer
+        scaling_factor: 1.0 / 32
```
rhaas80, line 1532: I guess overall amplitude does not matter to the FRB system?

lib/stages/processFeedGains.cpp
```diff
@@ -32,6 +33,8 @@ processFeedGains::processFeedGains(Config& config, const std::string& unique_nam
     num_local_freq = config.get<uint32_t>(unique_name, "num_local_freq");
     num_beams = config.get_default<uint32_t>(unique_name, "num_beams", 1);
     upchan_factor = config.get_default<uint32_t>(unique_name, "upchan_factor", 1);
+    num_components = config.get<uint32_t>(unique_name, "num_components");
```
rhaas80, line 36: these are the components of a complex number? The only reason for this to be not 2 is for test cases?

lib/stages/processFeedGains.cpp
```diff
@@ -52,9 +55,10 @@ processFeedGains::processFeedGains(Config& config, const std::string& unique_nam
                         gain_buffers.at(0)->frame_size);
         }
 
-        if (buf->frame_size != num_local_freq * num_elements * 2 * sizeof(float)) {
+        if (buf->frame_size != num_local_freq * num_elements * num_components * sizeof(float)) {
             FATAL_ERROR("Input buffer does not have the expected size. Expected {:d}, got {:d}",
-                        num_local_freq * num_elements * 2 * sizeof(float), buf->frame_size);
+                        num_local_freq * num_elements * num_components * sizeof(float),
+                        buf->frame_size);
         }
```
rhaas80, line 62: I wonder if the `require_FOO` functions check this as well (does not matter really for this pull request).

lib/stages/processFeedGains.cpp
```diff
@@ -86,48 +90,61 @@ processFeedGains::processFeedGains(Config& config, const std::string& unique_nam
...
+        {static_cast<ptrdiff_t>(num_beams), static_cast<ptrdiff_t>(num_local_freq * upchan_factor),
+         static_cast<ptrdiff_t>(num_elements), static_cast<ptrdiff_t>(num_components)},
+        {"R", "Fbar", "E", "C"}, {1, 1, 1, 1}));
+
+    freq_upchan_factor = std::vector<int>(num_local_freq * upchan_factor, upchan_factor);
+    freq_upchan_index = std::vector<int>(num_local_freq * upchan_factor);
+    coarse_freq = std::vector<int>(num_local_freq * upchan_factor, -1);
+
+    // set the actual frequency upchan indices. Assume increasing index
+    // TODO: verify that this is correct
```
rhaas80: not for CHIME apparently, both coarse and fine decrease with frequency (https://chime-experiment.slack.com/archives/CR2K31TNJ/p1783041822971589). Correct for the new upchannelizer though (so coarse decreases, fine increases).

rhaas80 (approved):

config/chime_upgrade_frb.j2
ljgray: Unsure - just using the same default as `setFRB1Phase`.

lib/stages/processFeedGains.cpp
ljgray: Yup - also just matching functionality from `setFRB1Phase`. Not sure when we would ever want to use any value other than 2, though

lib/stages/processFeedGains.cpp
ljgray: Ah, it probably does. I'll make another PR later on for metadata-related stuff

lib/stages/processFeedGains.cpp
ljgray: > coarse decreases, fine increases

Interesting. So for the beamformer (which seems to just want a single `Fbar` frequency axis instead of coarse and fine axes), we would end up with a frequency axis that is not monotonic?

rhaas80: kind of yes. If I am not mistaken than you will have:

```
int f_coarse = fbar / 16
```

being the coarse frequency bin that corresponds to the 16 upchannelized ones. And since `f_coarse` decreases with fequency and `fbar` increases with (fine) frequency you get a sawtooth pattern between `fbar` and the actual frequency in MHz.

At least I think so.

This is to-be-fixed of course.

PR #1579, opened 2026-07-10
Updates to run upgrade FRB pipeline indefinitely 

ljgray:
- Some cherry-picked commits for `testDataGenFewHot` stage, which should be merged in from elsewhere
- Removes `do_once` copy to GPU requirements for bad input mask and gain buffers
- Integrate real feed gain system


lib/stages/processFRBFeedGains.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+processFRBFeedGains::~processFRBFeedGains() {}
+
+void processFRBFeedGains::set_frame_desc(Buffer* buf) {
+    buf->require_frame_desc(kotekan::NDArray<kotekan::GetType_t<kotekan::float16>, 5>::describe(
+        "W",
+        {static_cast<ptrdiff_t>(num_local_freq * upchan_factor),
+         static_cast<ptrdiff_t>(num_polarizations), static_cast<ptrdiff_t>(num_dishes_N),
+         static_cast<ptrdiff_t>(num_dishes_M), static_cast<ptrdiff_t>(num_components)},
+        {"Fbar", "P", "dishN", "dishM", "C"}, {upchan_factor, 1, 1, 1, 1}));
```
jbmertens: Functionally the same, but - would a toggle suffice in lieu of another stage?

lib/stages/processFRBFeedGains.cpp
```diff
@@ -0,0 +1,53 @@
...
+using kotekan::Config;
+
+REGISTER_KOTEKAN_STAGE(processFRBFeedGains);
+
+processFRBFeedGains::processFRBFeedGains(Config& config, const std::string& unique_name,
+                                         bufferContainer& buffer_container) :
+    processFeedGains(config, unique_name, buffer_container) {
+    // get the additional config parameters needed for the frame desc
+    num_polarizations = config.get<uint32_t>(unique_name, "num_polarizations");
+    frb1_swap_MN = config.get_default<bool>(unique_name, "frb1_swap_MN", false);
```
jbmertens, line 21: Could frb1_swap_MN be determined automatically from the telescope type or element ordering?

lib/stages/bufferBadInputs.cpp
```diff
@@ -46,19 +48,26 @@ bufferBadInputs::bufferBadInputs(Config& config_, const std::string& unique_name
...
+    std::string badInputs = config.get<std::string>(unique_name, "updatable_config/bad_inputs");
+    configUpdater::instance().subscribe(
+        badInputs,
+        std::bind(&bufferBadInputs::update_bad_inputs_callback, this, std::placeholders::_1));
 }
 
 bufferBadInputs::~bufferBadInputs() {}
 
 bool bufferBadInputs::update_bad_inputs_callback(nlohmann::json& json) {
     DEBUG("update_bad_inputs_callback(): Update to bad inputs list.");
```
jbmertens: Maybe INFO here?

lib/stages/bufferBadInputs.cpp
```diff
@@ -70,31 +79,35 @@ bool bufferBadInputs::update_bad_inputs_callback(nlohmann::json& json) {
...
+
+        // Copy from the permanent buffer
+        {
+            std::lock_guard<std::mutex> lock(mtx);
+            std::copy_n(input_mask.begin(), num_elements, out_frame);
+        }
+
+        // Set metadata and release
+        out_buf->allocate_new_metadata_object(frame_id);
+        get_chord_metadata(out_buf, frame_id)->set_rfi_num_bad_inputs(num_bad_inputs);
```
jbmertens: `num_bad_inputs` may need to be set in the callback still

jbmertens (commented):
This will be great for CHORD too, thanks!

lib/stages/processFRBFeedGains.cpp
ljgray: My thinking was that this will also be needed for the tracking beamformer, which will have different indices. We could just have a toggle between "modes" (tracking or FRB) or something like that

lib/stages/processFRBFeedGains.cpp
eschnett: It could be determined from the telescope type.

eschnett:
Removing the `do_once` will require a new buffer with beam phases at every iteration. To make this work well, this `W` buffer will need to have its time cadence set properly, and we need to add the respective checks to ensure that the FRB beam calculation and the `W` buffers do not get out of sync. Currently the code probably expects a new `W` buffer every time it runs, and it will then consume a variable (!) number of time samples, depending on how many are currently available. That's unlikely to be useful. That is, we'll need to add some non-trivial logic to keep the `W` buffers and voltages in sync.

ljgray:
> Currently the code probably expects a new `W` buffer every time it runs, and it will then consume a variable (!) number of time samples, depending on how many are currently available.

Yeah, that's pretty much how it's implemented here. This shouldn't end up mattering too much for chime (for now) since these buffers only change once per day, but it probably isn't the correct long-term solution.

> That is, we'll need to add some non-trivial logic to keep the `W` buffers and voltages in sync.

There's an old HSA async copy kernel that would copy buffers to the GPU only as-needed (so every time the phases actually change(, I think that we would eventually just want to port that to cuda, unless there's some additional synchronization that we care about.

ljgray:
@jbmertens This should be good to go now, so I'd like to get it merged in somewhere. Are we still good to merge to `develop`? Given that the kernel "do_once" changes might affect Chord testing

jbmertens:
Thanks! Yeah, should be good to merge into develop, we'll want this for chord. I'll review now.

config/chime_upgrade_frb.j2
```diff
@@ -21,6 +21,9 @@ type: config
 # Note DEBUG and DEBUG2 require a build with (-DCMAKE_BUILD_TYPE=Debug)
 log_level: info
 
+{% set using_dpdk = False %}
```
jbmertens: Assuming these setting are just for testing now.

config/chime_upgrade_frb.j2
```diff
@@ -1793,19 +1821,21 @@ write_baseband:
...
-#  log_level: warn
-#  # Keep data in memory even if the receiver node is down
-#  # Baseband readout handles back pressure management
-#  drop_frames: false
-#  use_config_tracker: false # not used by legacy CHIME kotekan code
-#  baseband_send_0:
-#    kotekan_stage: bufferSend
-#    server_port: 11027
-#    buf: baseband_output_buffer_merged
+{#
```
jbmertens: Keep the commented code?

lib/stages/bufferBadInputs.cpp
```diff
@@ -46,19 +48,24 @@ bufferBadInputs::bufferBadInputs(Config& config_, const std::string& unique_name
...
-    // Get the next output frame
-    uint8_t* host_mask = (uint8_t*)out_buf->wait_for_empty_frame(unique_name, frame_id);
-    // Reset the mask (1 == good)
-    std::memset(host_mask, 1U, num_elements);
+    // hold lock for the entire update
+    std::lock_guard<std::mutex> lock(mtx);
 
-    bool all_good = true;
+    // Reset the mask (1 == good)
+    std::fill(input_mask.begin(), input_mask.end(), 1u);
```
jbmertens: This reset happens before the json input is actually read and validated. Is that intentional?

lib/stages/processFeedGains.cpp
```diff
@@ -23,8 +22,6 @@ using kotekan::configUpdater;
 using kotekan::Stage;
 using nlohmann::json;
 
-REGISTER_KOTEKAN_STAGE(processFeedGains);
```
jbmertens, line 26: This stage is still mentioned by some configs, chime_upgrade_test_gains.j2 .

lib/stages/bufferBadInputs.cpp
```diff
@@ -67,34 +74,45 @@ bool bufferBadInputs::update_bad_inputs_callback(nlohmann::json& json) {
...
+
+        // Copy from the permanent buffer
+        {
+            std::lock_guard<std::mutex> lock(mtx);
+            std::copy_n(input_mask.begin(), num_elements, out_frame);
+        }
+
+        // Set metadata and release
+        out_buf->allocate_new_metadata_object(frame_id);
+        get_chord_metadata(out_buf, frame_id)->set_rfi_num_bad_inputs(num_bad_inputs);
```
jbmertens: num_bad_inputs is read outside the lock here.

lib/stages/loadFeedGains.hpp
```diff
@@ -13,6 +13,7 @@
 #include "json.hpp" // for json
 
 #include <mutex>    // for mutex
+#include <mutex>    // for lock_guard, mutex
```
jbmertens, line 15: Duplicate of the line above.

lib/stages/processFeedGains.cpp
```diff
@@ -171,11 +132,17 @@ void processFeedGains::main_thread() {
 
             // check if this buffer has an available frame, using a short
             // timeout to avoid blocking
-            timespec timeout = double_to_ts(current_time());
+            timespec timeout;
+            if (set_coarse_freqs_once) {
+                // first frame - need to wait until we get something
+                timeout = double_to_ts(60 * 60 * 24);
```
jbmertens: Should this be `double_to_ts(current_time() + 60*60*24)`?

jbmertens (changes requested):
Mostly looks good - few comments below.

config/chime_upgrade_frb.j2
ljgray: Yes, although I'm actually going to swap the defaults (use_dpdk = True)

config/chime_upgrade_frb.j2
ljgray: Want to keep - changed to a configurable parameter

lib/stages/bufferBadInputs.cpp
ljgray: Nope - fixed

jbmertens:
Also apologies, the lint fix might cause some conflicts here 😬

ljgray:
Alright, I think that all comments are addressed. Let's see if we run into any lint issues...

jbmertens (approved):
LGTM!

PR #1581, opened 2026-07-16
Fixes for a few racy issues.

jbmertens:
Using up some claude tokens...

----------

Fixes four unrelated race/lifetime bugs, one per subsystem, each in its own commit.

- **cudaRechunk**: had a race, but the stage is now being removed altogether per discussion with Erik and Dustin.
- **buffer**: the frame waits dropped the lock before re-checking `shutdown_signal` and recording bookkeeping, and held a reference to the consumer's map entry across a wait that `unregister_consumer()` can erase. Waits now detect erasure via an epoch counter (common wake stays two array reads), and `/buffers` JSON is built from a snapshot instead of under the buffer lock.
- **restServer**: `handle_request` read the callback/alias maps with no lock. Dispatch now looks up under a shared lock and invokes a copy, and `remove_*_callback` drains an in-flight callback before returning, so destroying a callback's owner is safe.
- **datasetManager**: `_states` was read unlocked, and a failed broker request leaked its `_requested_states` claim — after one broker hiccup, `dataset_state()` retried into a permanent wait holding `_lock_dsets`, wedging all dataset lookups. Claims are now taken atomically and released on every exit path; a regression test covers this (hangs pre-fix).

Behavior notes for reviewers:
- `wait_for_full_frame` now returns NULL (documented) for an unregistered consumer instead of throwing — this also fixes a latent crash for stages that self-unregister and keep looping (e.g. `testDataCheck` with `leave_last` on a multi-consumer buffer).
- A REST callback stuck past the 5 s drain timeout is a **fatal shutdown** by design: the caller is a destructor about to free what the callback uses, and kotekan runs under a restarting daemon.
- `restServer::get_aliases()` is removed from the public header with no replacement (it handed out an unguarded reference).


lib/core/buffer.cpp
```diff
@@ -181,6 +181,7 @@ void GenericBuffer::send_shutdown_signal() {
 }
 
 void GenericBuffer::json_description(nlohmann::json& buf_json) {
+    buffer_lock lock(mutex);
```
andrerenard: These locks are potentially dangerous in another way, since they are shared with the critical path `wait_for*` and `mark_frame*` calls.  So a call to generate the json from say a rest call, now locks out a critical path call.   I'd prefer this data be returned potentially inconsistent than have this call slow down the pipeline in possibly critical parts. 

Having said that iterating a data structure like `consumers` which could change exterially is an unsafe pattern (previously this used a C style fixed array, so could only return bad data in a race condition).   Registering and unregistering consumers/producers is not in the critical path.  So those data structures could be protected by a different mutex to guard against this particular case.

lib/core/buffer.cpp
```diff
@@ -387,6 +388,7 @@ int Buffer::get_num_full_frames() {
 }
 
 void Buffer::json_description(nlohmann::json& buf_json) {
+    buffer_lock lock(mutex);
```
andrerenard: Here I think we can just remove the lock, and accept possibly inconsistent JSON data.

lib/core/buffer.cpp
andrerenard: Technically this call also takes the lock.  I think that's ok since it's out of the possibly slow JSON code, but if we wanted to be extra careful, we could have a non-locking version of this call.

andrerenard (dismissed):
Some comments on the buffer changes.  

We should make sure this gets tested with the CHORD and CHIME pipelines before it's merged, since there is some risk this will create issues I cannot foresee given how central some of this code is. 

It may also be worth having @dstndstn (@eschnett when he's back) take a second look at the `cudaRechunk` code to make sure this change reflects the intended behaviour.

jbmertens:
> Some comments on the buffer changes.
> 
> We should make sure this gets tested with the CHORD and CHIME pipelines before it's merged, since there is some risk this will create issues I cannot foresee given how central some of this code is.
> 
> It may also be worth having @dstndstn (@eschnett when he's back) take a second look at the `cudaRechunk` code to make sure this change reflects the intended behaviour.

These changes are live on chord pathfinder, so working there at least.

I'll pull out the "trivial" fixes into a separate PR. There are a number of additional changes coming in soon too.

lib/core/buffer.cpp
jbmertens: Yeah, it's likely never an issue in practice, but the change seems technically correct.

jbmertens:
Pulled out other changes into #1586 .

jbmertens:
@eschnett - just following up on this, the cudaRechunk changes could use another pair of eyes.

eschnett:
The `cudaRechunk` changes look correct.

The changes to `buffer.[ch]pp`: These are about a deadlock when consumer unregisters itself. Is this a real problem? I think each consumer runs on a single thread, so it couldn't possibly be waiting for a frame while trying to unregister itself. Or is the  consumer/stage handling more complicated?

I haven't understood the changes to `datasetManager` yet.

eschnett:
The `datasetManager` changes look fine. I don't understand the original problem, and I think maybe there wasn't a a run-time error that was ever observed and this is just a precaution, but the new code looks cleaner and correct. It also isn't performance critical, so being more cautious about locks and inserting into maps is fine.

eschnett (approved):

PR #1582, opened 2026-07-16
Patch up leaks valgrind found

jbmertens:


jbmertens:
Summary of the fixes (from running the `config/ci-tests/cpu_batch` configs under valgrind with clean shutdowns):

- **metadata (c7d3328):** the serialize() implementations that assign fields one by one (N2, Beam, HFB, Baseband, Vis) left struct padding uninitialized, and bufferSend sent it over the network while rawFileWrite wrote it to disk. VisMetadata's explicit trailing pad missed 4 bytes of padding between `freq_id` and `dataset_id`. Zero the output buffer first, as chordMetadata::serialize already does. VisMetadata members also get default initializers so fields a producing stage never sets are deterministic.

- **bufferRecv / restClient (e85faaf):** neither freed its libevent `event_config`; bufferRecv also never freed its event base, listener/timer events, or listener socket after `event_base_dispatch()` returns, and connInstance objects were only deleted on connection errors, so connections still open on exit leaked their frame and metadata staging memory. Open instances are now tracked and deleted when the event loop exits.

Result: all CI cpu_batch and gpu_batch configs run under valgrind with 0 errors and 0 bytes definitely/indirectly/possibly lost (before: `test_send_recv` lost ~600 KB across categories, `run_n2accum_chime` reported 178 uninitialized-write errors). The serialized formats are unchanged — padding bytes are now zero instead of garbage, and receivers' set_from_bytes never read them.

Verified with: run_tests.sh on cpu_batch (5/5) and gpu_batch (16/16, with USE_HDF5=ON); boost tests (27/27); serial and parallel pytests matching the develop baseline; tools/lint.sh.

andrerenard (approved):
LGTM.  What config did you run with valgrind?   I expect we might find other small issues like this in the larger pipelines.  Although only the ones with simulated data have any hope of running inside valgrind.  But Erik's full GPU tests should be possible to run through this.

jbmertens:
All the CI yaml configs, including GPU ones. So this isn't comprehensive.

PR #1583, opened 2026-07-16
processFeedGains: wait for first frames of mask and gains

rhaas80:


rhaas80:
Untested so far. but looks sane to me (unsurprisingly).

PR #1584, opened 2026-07-17
Improve DPDK startup for CHIME and NIC counter logging

andrerenard:
Changes the CHIME packet capture alignment process to avoid the random startup failures the result from the NIC startup delays.

Also adds a number of new NIC counter status to the metrics, and warnings about NIC drops related to mbuf starvation.

### DIFF
--- a/config/chime_upgrade_network.j2
+++ b/config/chime_upgrade_network.j2
@@ -34,6 +34,14 @@ sizeof_float: 4
 sizeof_int: 4
 sizeof_short: 2
 
+{% set num_polarizations = 2 %}
+{% set num_dishes = 1024 %}
+{% set num_elements = num_polarizations * num_dishes %}
+
+num_polarizations: {{ num_polarizations }}
+num_dishes: {{ num_dishes }}
+num_elements: {{ num_elements }}
+
 use_hugepages: true
 
 # Single source of truth for the upstream FPGA controller's REST interface.
@@ -111,7 +119,9 @@ dpdk:
   max_rx_pkt_len: 5000
   # Set to true when using a switch to replicate packets
   using_switch: true
-  alignment: samples_per_frame * 4
+  alignment: samples_per_frame
+  alignment_startup_delay: 20.0
+  lcore_start_delay: 1.0
   # List of PCIe devices we don't want to attach too
   # Required to exclude the ConnectX 6 Lx output NICs.
   # Excluded devices are listed by their PCIe address. 
--- a/config/chime_upgrade_network.j2
+++ b/config/chime_upgrade_network.j2
@@ -20,7 +20,6 @@ type: config
 # OFF, ERROR, WARN, INFO, DEBUG, DEBUG2 (case insensitive)
 # Note DEBUG and DEBUG2 require a build with (-DCMAKE_BUILD_TYPE=Debug)
 log_level: info
-num_elements: 2048
 num_freq_host_buf: 1
 num_freq_gpu_buf: 16
 samples_per_frame: 16384
--- a/lib/dpdk/dpdkCore.hpp
+++ b/lib/dpdk/dpdkCore.hpp
@@ -14,22 +14,22 @@
 // break with "templates must have C++ linkage".
 extern "C" {
 // IWYU pragma: begin_keep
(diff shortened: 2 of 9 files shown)


jbmertens:
Do you expect any impact on CHORD with this?

andrerenard:
> Do you expect any impact on CHORD with this?

There shouldn't be, CHORD uses a different mechanism (and code) for synchronization.  The only common change is the NIC statistics monitoring in `update_port_stats()`, which could be disabled if needed.

config/chime_upgrade_network.j2
```diff
@@ -34,6 +34,14 @@ sizeof_float: 4
 sizeof_int: 4
 sizeof_short: 2
 
+{% set num_polarizations = 2 %}
+{% set num_dishes = 1024 %}
+{% set num_elements = num_polarizations * num_dishes %}
+
+num_polarizations: {{ num_polarizations }}
+num_dishes: {{ num_dishes }}
+num_elements: {{ num_elements }}
```
jbmertens, line 42: num_elements is set twice in the config now.
(The yamllint doesn't check j2 files for this sort of conflict, unfortunately, although we could probably add that...)

lib/dpdk/dpdkCore.hpp
```diff
@@ -14,22 +14,22 @@
 // break with "templates must have C++ linkage".
 extern "C" {
 // IWYU pragma: begin_keep
-#include <rte_ethdev.h>         // for rte_eth_conf
-#include <rte_ring_core.h>      // for rte_ring
-// cinttypes needed by some CentOS systems.
```
jbmertens, line 19: Just checking this removal is intentional (since the comment flags it as "needed").

jbmertens (approved):
Few items from Claude,

1. rx_packets_total/rx_bytes_total regress to zero for iceBoardStandard and iceBoardVDIF. The increment moved out of the shared check_packet() into iceBoardShuffle::handle_packet() only. The other two handlers (used by aro_record, adc_record, hartrao_reduced_links, etc.) now report zero packets/bytes forever, so kotekan_dpdk_rx_packets_total/rx_bytes_total, the /port_data fields, and the periodic status log all read 0. Fix: add the increment (after check_order, matching shuffle) to both handlers. Note the shuffle move also changed the metric's meaning from "valid" to "accepted" packets (now excludes reordered/dupes) — intentional, just worth documenting.
2. chime_upgrade_frb.j2 sets alignment: samples_per_frame, which is undefined in that file (only samples_per_data_set exists; samples_per_frame lives only in chime_upgrade_network.j2). Config evaluation throws when run with using_dpdk: true. chime_upgrade_n2.j2 was correctly changed to samples_per_data_set; frb should match.
3. dpdkCore.cpp NIC-stat deltas use unsigned subtraction against the previous poll — a counter reset (e.g. the link-reset this same code logs) underflows to ~1.8e19 and emits a bogus WARN/gauge spike.
4. last_eth_stats starts zero-initialized and is never primed with an initial rte_eth_stats_get(), so the first poll can WARN with the full cumulative counter as if that many packets dropped "in the last ~1s" (conditional on whether the PMD zeros stats on rte_eth_dev_start).

(The lcore_start_delay: 1.0 float added to the configs is fine — Config::eval converts it to 1 cleanly.)

🤖 Generated with Claude Code (https://claude.com/claude-code)

jbmertens:
When running with these changes in chord (pathfinder) production, kotekan actually just crashes when the FPGA sent corrupt data. Which is nice in that we know the something is broken, but not nice in that kotekan just crashes without an indication why. There's a placeholder fix for this on the `chord` branch which discards obviously bad packets and WARNs, but we will likely want a better solution long-term.  (Even as simple as an error message saying as much before crashing.)

andrerenard:
> When running with these changes in chord (pathfinder) production, kotekan actually just crashes when the FPGA sent corrupt data. Which is nice in that we know the something is broken, but not nice in that kotekan just crashes without an indication why. There's a placeholder fix for this on the `chord` branch which discards obviously bad packets and WARNs, but we will likely want a better solution long-term. (Even as simple as an error message saying as much before crashing.)

I think that must be an unrelated issue, as this change only impacts the CHIME related packet handling pipeline.

PR #1585, opened 2026-07-18
Patch up leaks valgrind found

jbmertens:
Just ran CI tests through valgrind and let Claude fix up the results. So this is not exhaustive, and does not test production pipelines. And should not actually have any effects, other than cleaner shutdown.

------

* metadata: zero serialization buffers and VisMetadata members

The field-assigning serialize() implementations (N2, Beam, HFB, Baseband, Vis) left struct padding bytes uninitialized in the output, which bufferSend then sent over the network and rawFileWrite wrote to disk (flagged by valgrind). Zero the output buffer first, as chordMetadata::serialize already does; VisMetadata's explicit trailing pad did not cover the implicit padding between freq_id and dataset_id.

Also give VisMetadata default member initializers so fields a producing stage does not set are deterministic.

* bufferRecv/restClient: free libevent objects and open connections

Fix memory leaks found by valgrind: neither stage freed its libevent event_config (and bufferRecv never freed its event base, listener and timer events, or listener socket). bufferRecv also only deleted connInstance objects on connection close or errors, so instances open at shutdown leaked their frame and metadata staging memory; track them in an instance list and delete any still open after the event loop exits.

### DIFF
--- a/lib/metadata/BasebandMetadata.cpp
+++ b/lib/metadata/BasebandMetadata.cpp
@@ -1,6 +1,7 @@
 #include "BasebandMetadata.hpp"
 
 #include <assert.h>     // for assert
+#include <cstring>      // for memset
 #include <json.hpp>     // for json
 #include <memory>       // for allocator, shared_ptr, dynamic_pointer_cast, __shared_ptr_access
 #include <string>       // for basic_string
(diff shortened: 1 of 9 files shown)


andrerenard (approved):
LGTM

PR #1589, opened 2026-07-20
rfi: add option to disable first-stage RFI excision

jbmertens:
Add a rfi_first_stage_excision_enabled config option (default true) to cudaRFISKtilde. When false, the SK kernel skips the RFI mask computation. The SKtilde statistics are still computed, so second-stage excision and bad feed flagging is unaffected. This avoids using a "dummy buffer" to turn off RFI flagging.

(Note about motivation: this change is so the relevant GPU stages can run on chord pathfinder, without actually doing anything, as most feeds/dishes are not connected and so the SK statistic just flags everything.)

### DIFF
--- a/lib/stages/N2Accumulate.cpp
+++ b/lib/stages/N2Accumulate.cpp
@@ -533,7 +533,7 @@ void N2Accumulate::main_thread() {
                     continue;
                 }
 
-                // Third: accum RFI samples
+                // Third: accum RFI sampes
                 _n_rfi_samples_in_vis[f] += rficounts_t0[f] + rficounts_t1[f];
 
                 // Fourth: Normalization - accum the remaining good ticks.
--- a/lib/stages/N2Accumulate.cpp
+++ b/lib/stages/N2Accumulate.cpp
@@ -533,7 +533,7 @@ void N2Accumulate::main_thread() {
                     continue;
                 }
 
-                // Third: accum RFI sampes
+                // Third: accum RFI samples
                 _n_rfi_samples_in_vis[f] += rficounts_t0[f] + rficounts_t1[f];
 
                 // Fourth: Normalization - accum the remaining good ticks.
--- a/lib/stages/ProcessPacketMask.cpp
+++ b/lib/stages/ProcessPacketMask.cpp
@@ -37,8 +37,11 @@ STAGE_CONSTRUCTOR(ProcessPacketMask) {
     // The RFI mask here is just a hack to avoid running the actual RFI kernels
     // It just replays a mask of whatever is in the frames at startup
     // (This should be set to 0xff as the "zeroing" value.)
-    rfi_mask_buf = get_buffer("rfi_mask_buf");
-    rfi_mask_buf->register_producer(unique_name);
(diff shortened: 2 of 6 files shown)


lib/stages/N2Accumulate.cpp
```diff
@@ -533,7 +533,7 @@ void N2Accumulate::main_thread() {
                     continue;
                 }
 
-                // Third: accum RFI sampes
+                // Third: accum RFI samples
```
rhaas80: unrelated to the topic of the pull request

lib/stages/ProcessPacketMask.cpp
```diff
@@ -37,8 +37,11 @@ STAGE_CONSTRUCTOR(ProcessPacketMask) {
     // The RFI mask here is just a hack to avoid running the actual RFI kernels
     // It just replays a mask of whatever is in the frames at startup
     // (This should be set to 0xff as the "zeroing" value.)
-    rfi_mask_buf = get_buffer("rfi_mask_buf");
-    rfi_mask_buf->register_producer(unique_name);
+    // Optional: pipelines using the GPU-computed RFI mask can omit this buffer.
+    rfi_mask_buf =
+        config.exists(unique_name, "rfi_mask_buf") ? get_buffer("rfi_mask_buf") : nullptr;
```
rhaas80, line 43: did your check-for-unused options go in? If not "optionals are dangerous since they hide typos".

lib/stages/ProcessPacketMask.cpp
```diff
@@ -166,7 +171,8 @@ void ProcessPacketMask::main_thread() {
     frameID voltage_frame_id(voltage_buf);
     frameID combined_bitmap_frame_id(combined_receipt_bitmap_buf);
     frameID pl_mask_frame_id(pl_mask_buf);
-    frameID rfi_mask_frame_id(rfi_mask_buf);
+    // frameID needs a valid buffer even if rfi_mask_frame_id is never used.
+    frameID rfi_mask_frame_id(rfi_mask_buf != nullptr ? rfi_mask_buf : pl_mask_buf);
```
rhaas80: a terribly idea.

rhaas80 (changes requested):
contains unrelated changes, abuses `pl_mask_buf` as a dummy buffer from a frameID of a non-existent buffer

lib/stages/N2Accumulate.cpp
jbmertens: Reverted.

lib/stages/ProcessPacketMask.cpp
jbmertens: Er, one step back, one step forward.
Changing to std::optional.

lib/stages/ProcessPacketMask.cpp
jbmertens: Yes, the CI test config includes `log_config_usage: fatal_error`. Happy to change this behavior if something else would be better though.

rhaas80 (approved):

PR #1591, opened 2026-07-28
Update pipeline viewer infrastructure

jbmertens:
- Update dot visualizer output routines
 - Add option to make data in buffers accessible via rest endpoint. Changes may affect performance, but only if used...


lib/core/buffer.cpp
```diff
@@ -386,6 +430,60 @@ int Buffer::get_num_full_frames() {
...
 
+int Buffer::peek_newest_full_frame(std::vector<uint8_t>& data_out, size_t max_len,
+                                   std::shared_ptr<metadataObject>& metadata_out) {
+    // Size the destination before taking the lock: frame_size never changes,
+    // and doing the allocation (and its page faults) here keeps them out of
+    // the window where every producer and consumer on this buffer is blocked.
+    const size_t copy_len = std::min(max_len, frame_size);
+    data_out.resize(copy_len);
+
+    buffer_lock lock(mutex);
```
andrerenard: We shouldn't lock the entire buffer when doing the memcpy, we just need to make sure the frame being copied cannot be written into by a producer while the copy is taking place.  This would reduce the change of a peek blocking some critical code's buffer access.

andrerenard (changes requested):

andrerenard (approved):
This is looking good, I'm reasonably confident now all the new buffer logic works out.

PR #1592, opened 2026-07-30
restServer: quiesce request handling before pipeline teardown

kvand:
## Problem

`restServer::handle_request` looks a callback up under `callback_map_lock` but deliberately invokes a **copy** of it *after* releasing that lock (so callbacks like `/start` and `/stop` may re-register endpoints). On an external `SIGINT`/`SIGTERM` shutdown, `main()` runs `delete kotekan_mode` on the **main thread** while the libevent server thread may be part-way through a handler that captured a stage's `this`. The stage is freed under the running callback → **use-after-free** (observed as a `SIGSEGV` in a stage's `get_status` handler while a monitor polled it through a commanded shutdown).

The `/stop` and `/kill` handlers are *not* the trigger: they run on the single libevent thread, which serializes handlers, so tearing the mode down from one of them can't race another handler. Only the main-thread signal path runs concurrently with the server thread.

## Fix

A request-quiesce barrier in the core rest server:

- `handle_request` now holds a new `request_lock` **shared** for the whole handler (lookup + invoke).
- `stop_processing()` clears an `_accepting_requests` flag (further requests get `503`) and then takes `request_lock` **exclusively**, so it blocks until any in-flight handler has returned.
- The main-thread signal-shutdown path (`kotekan.cpp`) calls `stop_processing()` before `delete kotekan_mode`.
- The drain is skipped when `stop_processing()` is called from the server thread itself (a `/stop`-style handler): the single-threaded event loop guarantees no other handler is in flight, and draining there would self-deadlock on the caller's own shared hold.

`request_lock` is distinct from `callback_map_lock` (which only guards the map structure and is still released before a callback runs), so callbacks that re-register endpoints do not deadlock against the barrier. The server thread is left running (endpoints stay resolvable, just refused), so this composes with the framework's normal singleton teardown.

## Validation

Built CPU-only; `-DWERROR` clean. Tested against a minimal pipeline while polling endpoints through repeated shutdowns:

- With the copy→invoke window artificially widened: **16/20 shutdowns `SIGSEGV` without the fix, 0/20 with it.**
- At natural timing: **30/30 clean shutdowns** with and without the fix — no shutdown hang/deadlock, no regression.

https://claude.ai/code/session_019StS4mCetA7DUhRzUGA4ow

### DIFF
--- a/kotekan/kotekan.cpp
+++ b/kotekan/kotekan.cpp
@@ -707,6 +707,15 @@ int main(int argc, char** argv) {
 
         if (sig_value == SIGINT || sig_value == SIGTERM) {
             INFO_NON_OO("Got SIGINT or SIGTERM, shutting down kotekan...");
+            // Quiesce the REST server before destructing the stages: this
+            // teardown runs on the main thread while the server thread may
+            // still be handling a request that captured a stage's `this`
+            // (e.g. a monitor polling a stage's get_status endpoint), which
+            // would use-after-free once the stage is deleted below. Must
+            // happen before kotekan_state_lock is taken: the /start, /stop
+            // and /status handlers block on that lock, so draining while
+            // holding it would deadlock against any such handler in flight.
+            rest_server.stop_processing();
             std::lock_guard<std::mutex> lock(kotekan_state_lock);
             if (kotekan_mode != nullptr) {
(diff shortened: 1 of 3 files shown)


jbmertens:
Thanks. LGTM, although Claude found a potential deadlock condition, see PR #1593 for that and a couple other nits.

PR #1596, opened 2026-08-05
cuda: add DTV pilot-tone detection via the PilotProxy method

djgormley:
Succeeded by #1677.


jbmertens:
If not too much trouble, can the vendored code be moved to the `external` directory, and license file in there updated?

The `develop` branch would actually be best for this PR too - that branch should now be semi-stable and updated with changes for chord and chime upgrade. The `chord` branch currently contains a handful of developmental changes that need to be merged into develop anyways.

djgormley:
> If not too much trouble, can the vendored code be moved to the `external` directory, and license file in there updated?
> 
> The `develop` branch would actually be best for this PR too - that branch should now be semi-stable and updated with changes for chord and chime upgrade. The `chord` branch currently contains a handful of developmental changes that need to be merged into develop anyways.

The external and license shouldn't be a problem.  

Andre told me to use this branch so that I can get this thing running on the CHORD Pathfinder sooner than later.  But if you think I should put it under development that's fine with me too.  

I'll work on these when I have a minute.  Thanks.

jbmertens:
Ah, ok great. Ideally these changes end up in both chord and develop branches then. The `chord` branch is nominally based off of develop, so I would still suggest a PR into develop and then immediately rebasing the chord branch. (The chord branch isn't protected nor given special status since we switched to develop, it's just develop plus misc. unreviewed changs that are running on site.)

config/fengine/include/dtv_chord.j2
```diff
@@ -0,0 +1,112 @@
...
+# lib/cuda/pilotproxy/) on the voltage ring buffer. At its first detector
+# block the stage reads the chordMetadata coarse_freq identifiers and binds
+# any of this node's coarse channels that appear as `chord_channel_id` in
+# the PilotProxy runtime bundle; nodes carrying no DTV pilot channel run
+# DISABLED and emit all-zero products.
+#
+# The runtime bundle (pilot_profiles.json + weights.bin) is exported from
+# the pilot-proxy repository:
+#
+#   pilot-proxy export-runtime-weight-bundle \
```
jbmertens, line 16: This will need to be run to work in production, right? How about adding a feature flag that runs this as part of the build process? (Do the files take long to generate?)

config/tests/verify_pilotproxy_pipeline.yaml
```diff
@@ -0,0 +1,180 @@
...
+    dim_name: ["T", "F", "P", "D"]
+    dim_scaling: [1, 1, 1, 1]
+    # Coarse channels on this "node": 2408 (ATSC ch 14) and 2623 (ATSC ch 21)
+    # are pilot channels present in the ATSC 14-36 runtime bundle; 1600 and
+    # 4000 are non-pilot channels and must emit zeros.
+    manual_freq_ids: [2408, 1600, 2623, 4000]
+    meta_time_downsample_factor: 1
+
+dump_voltage:
+    kotekan_stage: rawFileWrite
```
jbmertens, line 118: Not important to change since this is just for testing, but: why not use the hdf5 file writer stage, which includes frame descriptions?

config/ci-tests/gpu_batch/test_pilotproxy_detector.yaml
```diff
@@ -0,0 +1,24 @@
...
+log_level: INFO
+
+cpu_affinity: [0]
+
+# Required by the default telescope setup (the stage itself reads no config).
+num_polarizations: 2
+num_dishes: 512
+
+test_pilotproxy_detector:
+    kotekan_stage: testPilotProxyDetector
```
jbmertens, line 29: If this is just for running test code, why a stage instead of boost test? I guess boost tests don't run on a gpu node, although we could modify ci for that.

lib/cuda/cudaPilotProxyPacker.hpp
```diff
@@ -0,0 +1,73 @@
+// PilotProxy detector-input packer kernel (A40 / sm_86).
```
jbmertens: Consider making this a doxygen comment block.

lib/cuda/cudaPilotProxyPacker.hpp
```diff
@@ -0,0 +1,73 @@
...
+// See cudaPilotProxyPacker.cu for the kernel and the CPU reference.
+
+#ifndef CUDA_PILOTPROXY_PACKER_HPP
+#define CUDA_PILOTPROXY_PACKER_HPP
+
+#include <cstddef>
+#include <cstdint>
+#include <cuda_runtime.h>
+
+// Pack one coarse channel's detector block from the voltage ring buffer.
```
jbmertens: Consider making this a doxygen comment block.

jbmertens (commented):
Mostly lgtm - just a few random thoughts. I have also not thought through how this integrates with the GPU code, hoping @andrerenard or someone more familiar with the GPU side of things will do that. I do see calls like cudaDeviceSynchronize(), so I'm not sure how well this plays with the other kernels.

lib/cuda/cudaPilotProxyPacker.hpp
djgormley: Done.  Block is doxygen now (@file/@brief/@param, formula in @code).

lib/cuda/cudaPilotProxyPacker.hpp
djgormley: Done.  Block is doxygen now (@file/@brief/@param, formula in @code).

djgormley:
> If not too much trouble, can the vendored code be moved to the `external` directory, and license file in there updated?
> 
> The `develop` branch would actually be best for this PR too - that branch should now be semi-stable and updated with changes for chord and chime upgrade. The `chord` branch currently contains a handful of developmental changes that need to be merged into develop anyways.

Done in the updated series. The core now lives under external/pilotproxy, built like ksgpu/n2k, with the upstream MIT license text in the directory, the repo LICENSE entry updated, and a COMMIT pin in the same format as the other external trees. The byte-identical-to-upstream check moved with it and now cross-checks the pin. The branch is rebased onto develop and the PR base is flipped. Happy to help with the chord rebase after this merges.

djgormley:
> Mostly lgtm - just a few random thoughts. I have also not thought through how this integrates with the GPU code, hoping @andrerenard or someone more familiar with the GPU side of things will do that. I do see calls like cudaDeviceSynchronize(), so I'm not sure how well this plays with the other kernels.

Every cudaDeviceSynchronize in the runtime path is an #ifndef NDEBUG error check inside the vendored core, so release builds compile them out. The rest are in the standalone test stage (the same pattern as testPLMaskUpchannelizer). The production stage never blocks the host: the packer and all copies run on the stage's assigned kotekan stream, and the vendored core's launches, which are legacy-default-stream by its documented execution policy, are bracketed with cudaEventRecord/cudaStreamWaitEvent on both sides. The caveat is that stream-0 launches serialize against other blocking streams at launch boundaries. That cost is bounded by the detector's duty cycle: one packer launch plus one fused kernel per bound channel per decision block, which works out to well under 1% on an A40. If profiling on a shared node ever shows contention, the fix is a stream-taking API variant in the upstream core. That is already noted as the proposed follow-up in the vendor README.

config/fengine/include/dtv_chord.j2
djgormley: Right, the stage reads the bundle at runtime. Generation is fast: the export synthesizes the integer weight rows for ATSC 14-36 from the packaged profiles and the validator re-checks the manifest, about a quarter second all-in. I added -DPILOTPROXY_EXPORT_BUNDLE=ON/AUTO, which exports and validates into <build>/pilotproxy_bundle as part of the build (AUTO only when the pilot-proxy CLI is on PATH). I left it OFF by default on purpose. The bundle is survey calibration data with epoch semantics, so production sites should pin a released bundle, but dev and test builds are now self-contained with one flag.

config/ci-tests/gpu_batch/test_pilotproxy_detector.yaml
djgormley: Exactly that. The sweep needs a CUDA device, and the boost tests build and run on the CPU-only CI jobs, while gpu_batch configs are what run_tests.sh executes on the GPU runner (the same pattern as test_pl_mask_upchannelizer and the verify_cuda_n2k harnesses). The sweep body is a self-contained function of invented inputs, so it can port directly if CI ever grows a GPU boost lane. I put that rationale in the yaml header.

config/tests/verify_pilotproxy_pipeline.yaml
djgormley: That one is deliberate, though I should have said so in the file (now I do). The offline verifier recomputes every product bit-for-bit from the exact dumped bytes plus the runtime bundle, the layout is pinned by the config, and the raw dump keeps the harness free of an HDF5 read dependency. hdf5FileWrite would be the right tool if anyone needs these dumps self-describing; for the parity check, the descriptors testDataGen attaches are dropped on purpose.

djgormley:
> Ah, ok great. Ideally these changes end up in both chord and develop branches then. The `chord` branch is nominally based off of develop, so I would still suggest a PR into develop and then immediately rebasing the chord branch. (The chord branch isn't protected nor given special status since we switched to develop, it's just develop plus misc. unreviewed changs that are running on site.)

Rebased onto develop and moved the PR base over. Along with the rebase: the vendored core now lives in external/pilotproxy (built like ksgpu/n2k, license and COMMIT pin included, and the byte-identical-to-upstream check moved with it), the packer header comments are doxygen now, there is a -DPILOTPROXY_EXPORT_BUNDLE=ON/AUTO flag that exports and validates the runtime bundle as part of the build (it takes about a quarter second), and the answers to the config questions are written into the yaml headers. I also tidied the tail commits while rewriting history anyway. The two revendor commits are squashed into one, and the geometry-mismatch commit got its full message back. Replies on the individual threads.

config/ci-tests/gpu_batch/test_pilotproxy_detector.yaml
jbmertens: Sounds good.

config/tests/verify_pilotproxy_pipeline.yaml
jbmertens: Perfect, thanks.

config/fengine/include/dtv_chord.j2
jbmertens: Great. For CHORD we can use this as part of the build deployment process. It may get re-generated when kotekan is re-built, but I assume that is no problem. ~~If the output is expected to change across versions, it may also be helpful to add that information to `lib/version/version.c.in`, as a record for reproducibility.~~ never mind, I see it is in there already!

jbmertens:
> Every cudaDeviceSynchronize in the runtime path is an #ifndef NDEBUG error check inside the vendored core, so release builds compile them out. The rest are in the standalone test stage (the same pattern as testPLMaskUpchannelizer). The production stage never blocks the host: the packer and all copies run on the stage's assigned kotekan stream, and the vendored core's launches, which are legacy-default-stream by its documented execution policy, are bracketed with cudaEventRecord/cudaStreamWaitEvent on both sides. The caveat is that stream-0 launches serialize against other blocking streams at launch boundaries. That cost is bounded by the detector's duty cycle: one packer launch plus one fused kernel per bound channel per decision block, which works out to well under 1% on an A40. If profiling on a shared node ever shows contention, the fix is a stream-taking API variant in the upstream core. That is already noted as the proposed follow-up in the vendor README.

Great, thanks for the detail. Happy to merge this in, I'd still like @andrerenard or @eschnett to confirm.

djgormley:
CI is passing now.  I'm going to stop working on it while Andre and Erik review.  Let me know if you guys need anything.  Thanks.

jbmertens:
Sorry, one more thought - it would be great to add the PILOTPROXY_EXPORT_BUNDLE build flag (and its included validation!) to CI somewhere. Maybe include the flag on the "Test gcc build + CUDA/OpenCL" job?
https://github.com/kotekan/kotekan/blob/develop/.github/workflows/main.yaml#L284

djgormley:
> Sorry, one more thought - it would be great to add the PILOTPROXY_EXPORT_BUNDLE build flag (and its included validation!) to CI somewhere. Maybe include the flag on the "Test gcc build + CUDA/OpenCL" job? https://github.com/kotekan/kotekan/blob/develop/.github/workflows/main.yaml#L284

Good call, added. There was one wrinkle worth naming, since it changes the shape of the fix.

`PILOTPROXY_EXPORT_BUNDLE` shells out to the `pilot-proxy` CLI at configure time, so putting the flag in `cmake_defs` by itself does not work on a runner that lacks the CLI: `ON` fails the configure outright, and `AUTO` skips silently, which would look like coverage without being any. The CLI has to be installed before cmake runs, and the reusable build workflow had no hook that early.

So the change is in two parts. First, `test_kotekan_build.yaml` gains an optional `pre_build` input, guarded by `if: inputs.pre_build` in the same form as the existing `post_build` hook. It is inert for every current caller, since a matrix entry that does not set it passes the empty string. Second, the CUDA/OpenCL job you pointed at uses that hook to install the CLI into its own venv and put the venv on `PATH`, and sets `-DPILOTPROXY_EXPORT_BUNDLE=ON`. The bundle target is an `ALL` target, so the export and the validator both run as part of that build rather than needing a separate invocation.

The CLI is pinned to the same upstream commit as the vendored detector core, read out of `external/pilotproxy/VENDOR.json` rather than repeated in the workflow. Thus the two cannot drift apart, and the bundle CI validates always carries the geometry the kernel in that same build was compiled for. The install itself is small, since the CLI's only hard dependency is numpy.

I ran the whole path locally against a pip install at that pin: the console script and the packaged profile JSONs land where the feature's prefix derivation expects them, `make` runs both commands, and the validator reports 0 errors on the exported bundle. The workflow files are yamlfix-clean and the diff is additive, 25 lines with no deletions.

Two judgment calls I would rather leave to you than make silently:

- This adds a network dependency (a pip install from GitHub) to a job in a matrix with `fail-fast: true`, so an outage there would cancel the rest of the GPU matrix. `AUTO` instead of `ON` would degrade to a skip rather than a failure, at the cost of no longer being a real gate.
- The export needs no GPU at all, since it is pure Python that writes the weight rows and re-derives the norms. A CPU job would exercise it just as well, and that matrix is `fail-fast: false`. I put it where you suggested, but say the word and I will move it.

jbmertens:
The CI image is a full ubuntu docker image (see the Dockerfile in tools/docker/24.04), so there is a shell available. There are (or were?) other commands invoked during the build, so I don't think that should be an issue, but let me know if it is. The relevant python package can be added to the docker image.

Since this is GPU code, I figure it makes sense to include it in a GPU build on hardware we own, but if a CPU build on a random github node works, we can save a bit on local compute.

djgormley:
Noticed a few things that need to be changed via other PRs before this can be integrate so closing for now.

PR #1599, opened 2026-08-08
Rhaas/upchan freq from voltage

rhaas80:
set list of frequencies in upchan schedule based on frequencies from dpdk, instead of hard-coding in input files. Input files are identical on all nodes while no two nodes handle the same frequencies.

### DIFF
--- a/lib/stages/UpchannelizationSchedule.cpp
+++ b/lib/stages/UpchannelizationSchedule.cpp
@@ -8,11 +8,6 @@
 #include <mutex>
 #include <sstream> // for basic_ostream, basic_ostringstream, operator<<, basic_ostream::...
 
-std::vector<int> UpchannelizationSchedule::make_frequency_channels() const {
-    const auto frequency_channels = config.get<std::vector<int>>(unique_name, "frequency_channels");
-    return frequency_channels;
-}
-
 std::map<int, int> UpchannelizationSchedule::make_frequency_channels_to_indices() const {
     std::map<int, int> channels_to_indices;
     for (std::size_t index = 0; index < frequency_channels.size(); ++index)
@@ -176,7 +171,7 @@ UpchannelizationSchedule::UpchannelizationSchedule(kotekan::Config& config,
     //
     unique_name(unique_name), config(config),
     //
-    frequency_channels(coarse_freq.size() ? coarse_freq : make_frequency_channels()),
+    frequency_channels(coarse_freq),
     frequency_channels_to_indices(make_frequency_channels_to_indices()),
     //
     upchan_factors(make_upchan_factors()),
--- a/lib/stages/UpchannelizationSchedule.cpp
+++ b/lib/stages/UpchannelizationSchedule.cpp
@@ -170,12 +170,13 @@ bool UpchannelizationSchedule::invariant() const {
 }
 
 UpchannelizationSchedule::UpchannelizationSchedule(kotekan::Config& config,
-                                                   const std::string& unique_name) :
+                                                   const std::string& unique_name,
+                                                   const std::vector<int>& coarse_freq) :
     kotekan::kotekanLogging(),
     //
     unique_name(unique_name), config(config),
     //
-    frequency_channels(make_frequency_channels()),
+    frequency_channels(coarse_freq.size() ? coarse_freq : make_frequency_channels()),
     frequency_channels_to_indices(make_frequency_channels_to_indices()),
     //
     upchan_factors(make_upchan_factors()),
@@ -188,14 +189,16 @@ UpchannelizationSchedule::UpchannelizationSchedule(kotekan::Config& config,
     assert(invariant());
 }
 
-const UpchannelizationSchedule& UpchannelizationSchedule::instance(kotekan::Config& config,
-                                                                   const std::string& unique_name) {
+const UpchannelizationSchedule&
+UpchannelizationSchedule::instance(kotekan::Config& config, const std::string& unique_name,
+                                   const std::vector<int>& coarse_freq) {
     static std::map<std::string, UpchannelizationSchedule> the_instances;
     static std::mutex the_mutex;
 
     std::lock_guard<std::mutex> lock(the_mutex);
     const UpchannelizationSchedule& the_instance =
-        the_instances.try_emplace(unique_name, config, unique_name).first->second;
+        the_instances.try_emplace(unique_name, config, unique_name, coarse_freq).first->second;
+    assert(coarse_freq.size() && the_instance.frequency_channels == coarse_freq);
     return the_instance;
 }
 
--- a/lib/stages/UpchannelizationSchedule.cpp
+++ b/lib/stages/UpchannelizationSchedule.cpp
@@ -198,7 +198,13 @@ UpchannelizationSchedule::instance(kotekan::Config& config, const std::string& u
     std::lock_guard<std::mutex> lock(the_mutex);
     const UpchannelizationSchedule& the_instance =
         the_instances.try_emplace(unique_name, config, unique_name, coarse_freq).first->second;
-    assert(coarse_freq.size() && the_instance.frequency_channels == coarse_freq);
+    if (!(the_instance.frequency_channels == coarse_freq)) {
+        const std::string coarse_freq_str = fmt::format("{:s}", fmt::join(coarse_freq, ","));
+        const std::string frequency_channels_str =
+            fmt::format("{:s}", fmt::join(the_instance.frequency_channels, ","));
+        FATAL_ERROR_NON_OO("Frequency channels [{:s}] and [{:s}] do not match", coarse_freq_str,
+                           frequency_channels_str);
+    }
     return the_instance;
 }
 
--- a/lib/stages/UpchannelizationSchedule.hpp
+++ b/lib/stages/UpchannelizationSchedule.hpp
@@ -51,7 +51,6 @@ class UpchannelizationSchedule : kotekan::kotekanLogging {
 
     ////////////////////////////////////////////////////////////////////////////////
 
-    std::vector<int> make_frequency_channels() const;
     std::map<int, int> make_frequency_channels_to_indices() const;
 
     std::vector<int> make_upchan_factors() const;
@@ -68,8 +67,8 @@ public:
     UpchannelizationSchedule& operator=(const UpchannelizationSchedule&) = delete;
     UpchannelizationSchedule& operator=(UpchannelizationSchedule&&) = delete;
 
-    UpchannelizationSchedule(kotekan::Config& config, const std::string& unique_name = "",
-                             const std::vector<int>& coarse_freq = std::vector<int>());
+    UpchannelizationSchedule(kotekan::Config& config, const std::string& unique_name,
+                             const std::vector<int>& coarse_freq);
 
     // Get the singleton instance
     static const UpchannelizationSchedule&
--- a/lib/stages/UpchannelizationSchedule.hpp
+++ b/lib/stages/UpchannelizationSchedule.hpp
@@ -68,11 +68,13 @@ public:
     UpchannelizationSchedule& operator=(const UpchannelizationSchedule&) = delete;
     UpchannelizationSchedule& operator=(UpchannelizationSchedule&&) = delete;
 
-    UpchannelizationSchedule(kotekan::Config& config, const std::string& unique_name = "");
+    UpchannelizationSchedule(kotekan::Config& config, const std::string& unique_name = "",
+                             const std::vector<int>& coarse_freq = std::vector<int>());
 
     // Get the singleton instance
-    static const UpchannelizationSchedule& instance(kotekan::Config& config,
-                                                    const std::string& unique_name = "");
+    static const UpchannelizationSchedule&
+    instance(kotekan::Config& config, const std::string& unique_name = "",
+             const std::vector<int>& coarse_freq = std::vector<int>());
 
     // The coarse frequency channels handled by this X-Engine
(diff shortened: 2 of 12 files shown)


rhaas80:
@eschnett: ping

lib/stages/UpchannelizationSchedule.cpp
```diff
@@ -166,12 +166,13 @@ bool UpchannelizationSchedule::invariant() const {
...
 UpchannelizationSchedule::UpchannelizationSchedule(kotekan::Config& config,
-                                                   const std::string& unique_name) :
+                                                   const std::string& unique_name,
+                                                   const std::vector<int>& coarse_freq) :
     kotekan::kotekanLogging(),
     //
     unique_name(unique_name), config(config),
     //
-    frequency_channels(make_frequency_channels()),
+    frequency_channels(coarse_freq.size() ? coarse_freq : make_frequency_channels()),
```
eschnett: I don't think this fallback is necessary. Since the function signature changes, all callers need to be updated, and can be changed to pass in the frequencies.

lib/stages/UpchannelizationSchedule.cpp
```diff
@@ -184,14 +185,16 @@ UpchannelizationSchedule::UpchannelizationSchedule(kotekan::Config& config,
...
+UpchannelizationSchedule::instance(kotekan::Config& config, const std::string& unique_name,
+                                   const std::vector<int>& coarse_freq) {
     static std::map<std::string, UpchannelizationSchedule> the_instances;
     static std::mutex the_mutex;
 
     std::lock_guard<std::mutex> lock(the_mutex);
     const UpchannelizationSchedule& the_instance =
-        the_instances.try_emplace(unique_name, config, unique_name).first->second;
+        the_instances.try_emplace(unique_name, config, unique_name, coarse_freq).first->second;
+    assert(coarse_freq.size() && the_instance.frequency_channels == coarse_freq);
```
eschnett: This is a public function, so there needs to be proper error checking. `assert` can be deactivated, an `if` and a fatal error would be better.

lib/stages/UpchannelizationSchedule.hpp
```diff
@@ -69,11 +69,13 @@ class UpchannelizationSchedule : kotekan::kotekanLogging {
     UpchannelizationSchedule& operator=(const UpchannelizationSchedule&) = delete;
     UpchannelizationSchedule& operator=(UpchannelizationSchedule&&) = delete;
 
-    UpchannelizationSchedule(kotekan::Config& config, const std::string& unique_name = "");
+    UpchannelizationSchedule(kotekan::Config& config, const std::string& unique_name = "",
+                             const std::vector<int>& coarse_freq = std::vector<int>());
```
eschnett: Is there any way in which the fallback definition is correct in CHIME? If not, remove it, and require the frequencies all the time.

lib/stages/UpchannelizationSchedule.hpp
```diff
@@ -69,11 +69,13 @@ class UpchannelizationSchedule : kotekan::kotekanLogging {
...
 
     // Get the singleton instance
-    static const UpchannelizationSchedule& instance(kotekan::Config& config,
-                                                    const std::string& unique_name = "");
+    static const UpchannelizationSchedule&
+    instance(kotekan::Config& config, const std::string& unique_name = "",
+             const std::vector<int>& coarse_freq = std::vector<int>());
```
eschnett, line 77: Ditto, the fallback should not be necessary.

eschnett:
This PR expects the caller to know the frequencies. The frequencies are only known after the first frame has arrived. This means that the upchan schedule is only known after the first frame has arrived.

Therefore it would be easier to move the "wait for the first voltage frame" logic into the upchan schedule itself. Add a parameter that specifies the voltage buffer name, wait for it, and extract the frequencies here.

lib/stages/UpchannelizationSchedule.cpp
rhaas80: fine with me. the fallback was only in there to let some (private maybe) tests that use upchan scheduler with the old arguments work. Note that the `hpp` file defines a default for the extra argument. Removing the default makes for easier code.

lib/stages/UpchannelizationSchedule.cpp
rhaas80: it's public yes, but not triggerable by user input. But sure, I can make this an `ERROR_NON_OO` instead.

lib/stages/UpchannelizationSchedule.hpp
rhaas80: This is not CHIME specific. The reason why the frequencies need to be passed in is that at all kotekan instances on all GPU nodes (at least in CHIME, though I would expect CHORD to be the same) are given the same config file, but each node handles only certain frequencies. So we cannot specify the set of frequencies for a single node in the input fiel.

Will remove the fallback.

lib/stages/UpchannelizationSchedule.hpp
rhaas80: 'k

lib/stages/UpchannelizationSchedule.hpp
rhaas80: actually.... I cannot remove it. C++ won't let me do:

```
    UpchannelizationSchedule(kotekan::Config& config, const std::string& unique_name = "", 
                                          const std::vector<int>& coarse_freq);
```

so I will also have to remove the default for `unique_name`.

eschnett: You can't go back to the state before your change???

rhaas80: not really, I need to pass in `coarse_freq` after all. I can move `coarse_freq` before `unique_name` however. 

While one is removing defaults though: should `unique_name` have a default (that one was not introduced by me).

rhaas80:
> Therefore it would be easier to move the "wait for the first voltage frame" logic into the upchan schedule itself. Add a parameter that specifies the voltage buffer name, wait for it, and extract the frequencies here.

the upchan schedule is not a stage so cannot register itself as a consumer so cannot itself get the data from the buffer (metadata), can it?

eschnett:
Maybe it should be stage?

The set of frequencies is essentially a global setting in Kotekan. It would be quite convenient if there was a good way to access it. Maybe the receiver (and the file reader) can set a global variable?

rhaas80:
> Maybe it should be stage?

That I cannot comment on. For CHIME it seems mostly overkill since it only contains a single number :-)

> The set of frequencies is essentially a global setting in Kotekan. It would be quite convenient if there was a good way to access it. Maybe the receiver (and the file reader) can set a global variable?

I think even the packet capture in dpdk has to wait until receives the first package from the FPGA nodes to find out what the contained frequencies are (though I may be wrong, @andrerenard would have to confirm this). A global variable would kind of circumvent the idea that all information in kotekan flows via the buffers. Though the frequencies could be attached to the Telescope object I guess, they are not valid all the time though (since dpdk has to receive a frame)? Having to grab them out of packet and then storing them (and having to register as  consumer for some random unrelated buffer just for that) is annoying though, yes.

Most likely a post-deployment issue for at least CHIME, so I have no specific preference.

eschnett:
Having to wait for a packet or a buffer to get the frequencies is annoying, yes, and it's also not trivial. Hence the idea to do this in one single place in the code, which then remembers the frequencies so that they are easy to get. Otherwise all the places that need the frequencies need to perform this dance.

If you don't like my suggestions above then you could create a new stage called `CaptureFrequencies` which waits for the first voltage buffer, extracts the frequencies, and stores them. This would use the same mechanism that currently all the upchan schedulers use, but would need to do it only in one place.

rhaas80:
Hmm, so this would be structure similar to this:

* UpchanSchedule stage
  * in its constructor gets a writer lock on a semaphore
  * in its main waits for a packet to arrive to extract frequencies from
  * releases lock
  * offers `get_frequency_channels` or so function that obtains a reader lock on the semaphore and returns the saved information (and then releases the lock of course)
* client stages
  * call `get_frequency_channels` in their `main_thread` but before the loop
  * will stall until `UpchanSchedule` has finished

This introduces a custom data path outside of kotekan's pipeline.

An alternative would be for upchanschedule itself to produce a buffer (similar to say the FRB2 weights) that is read only once by its consumers.

This does not introduce a new data path.

rhaas80:
> If you don't like my suggestions above then you could create a new stage called `CaptureFrequencies` which waits for the first voltage buffer, extracts the frequencies, and stores them. This would use the same mechanism that currently all the upchan schedulers use, but would need to do it only in one place.

Cannot really store the frequencies as a buffer since they are some sort of `std::map` that does not easily lend itself to be send around as a flat buffer.

rhaas80:
re-review, please? All that `UpchannelizationSchedule` really is, is a code fragment to parse a set of ranges into a C++ data structure, yes?

lib/stages/UpchannelizationSchedule.hpp
```diff
@@ -68,11 +67,13 @@ class UpchannelizationSchedule : kotekan::kotekanLogging {
...
 
-    UpchannelizationSchedule(kotekan::Config& config, const std::string& unique_name = "");
+    UpchannelizationSchedule(kotekan::Config& config, const std::string& unique_name,
+                             const std::vector<int>& coarse_freq);
 
     // Get the singleton instance
-    static const UpchannelizationSchedule& instance(kotekan::Config& config,
-                                                    const std::string& unique_name = "");
+    static const UpchannelizationSchedule&
+    instance(kotekan::Config& config, const std::string& unique_name = "",
```
andrerenard, line 75: Do we actually need this to be a singleton anymore?

andrerenard (changes requested):
DPDK also doesn't know about what frequencies it will get until it starts getting packets.  

Unless we build a larger system around kotekan (e.g. maybe in coco/choco) to determine what nodes are expected to get what frequencies and inject those in the the config with each run, we are always going to need a pattern like this I'm afraid.    Having a "metadata" read once buffer, still seems a bit cleaner to me than introducing another synchronization primitive into the pipeline logic. 

(If we were going to add another synchronization system, I'd propose having a blocking call in the telescope object that returns the frequency list - and then have DPDK setting that be the trigger which allows it to return.  However, we'd have to be careful with that, since it would be easy to dead-lock the startup by calling it in a constructor.  And it would add extra complexity for multiple NUMA domains, or multiple ports with separate processing pipes, etc.  So I still think waiting on a network generate frame is easier to understand and use)

Also it seems to me like with this pattern we no longer need `UpchannelizationSchedule` to be a singleton?  By adding the `coarse_freq` and the stage name, we've basically provided everything that's needed to make the schedule in every use case I think. 

Finally, this change will break existing production configs, which are not being updated here with the "metadata" buffer.  Ideally we'd update those here.

lib/stages/UpchannelizationSchedule.hpp
rhaas80: It's no longer a singleton (had to be removed b/c of the HFB beamformer also needing an upchannelization schedule).

rhaas80:
> Finally, this change will break existing production configs, which are not being updated here with the "metadata" buffer. Ideally we'd update those here.

production configs?

andrerenard:
> > Finally, this change will break existing production configs, which are not being updated here with the "metadata" buffer. Ideally we'd update those here.
> 
> production configs?

Maybe not truely production yet, but `chime_upgrade_frb.j2`, etc.  This change currently only updates the test config.

lib/stages/UpchannelizationSchedule.hpp
andrerenard: In a different branch/PR?   I still see all singleton related static variables, locks, etc.

andrerenard: I guess singleton is an odd term, since it's a map of singletons.   But do we actually need to keep that map, or can we use these as local objects?

rhaas80:
> > > Finally, this change will break existing production configs, which are not being updated here with the "metadata" buffer. Ideally we'd update those here.
> > 
> > 
> > production configs?
> 
> Maybe not truely production yet, but `chime_upgrade_frb.j2`, etc. This change currently only updates the test config.

I am running the frb captures with this change. The corresponding change to the frb input files is just not part of this pull request.

lib/stages/UpchannelizationSchedule.hpp
rhaas80: it's a multi-ton. One per config file yaml path. 

As far as I can tell this could just be a parse function (or object) that returns the `std::map<int,std::vector<int>>` that is the schedule (and is called with the same yaml object path by everybody who needs it). I don't think anything in here is expensive or large enough to warrant wanting to keep only one in-memory copy. Cannot be easily made into a kotekan buffer since a `std::map` is not a flat array.

I am however not he author nor do I know what may be used in CHORD's set where there may be many more frequencies and factors.

andrerenard: @eschnett Thoughts on this?  Can we remove the singleton/multi-ton behaviour after accounting for the other changes here?   The schedule is larger in CHORD, but it's not huge by any means.

eschnett: Yes, this can be a parse function. I'll change it.

andrerenard:
@eschnett can I ping you on this one?   I think there is only one remaining issue, and I'd like to merge this in for something related I'm working on.

lib/stages/UpchannelizationSchedule.hpp
eschnett: Ping received. Urgency increased to front-of-queue.

PR #1601, opened 2026-08-09
Speed up FRB2 weight calculation (30 s -> 30 ms)

eschnett:
Speed up the FRB2 weight calculation by converting it to CUDA. The speedup is higher than naively expected because we never SIMD-parallelized the compute-intensive trigonometric calculations on the CPU.

Add a CUDA version of calcFRB2Weights that computes the FRB2 beamforming weights W2 on the GPU. The CPU stage is unchanged and remains available; all production configs now use the GPU stage.

The kernel exploits that W2 is separable: one block per (freq, beamR) computes Up[P] and Uq[Q] in shared memory and writes the contiguous P x Q tile as their outer product. Ufunc is evaluated with the cosine recurrence cos((s+1)x) = 2 cos(x) cos(sx) - cos((s-1)x), needing a single cospif call per element instead of M+1. The host side chunks over frequencies to bound GPU memory (gpu_max_chunk_bytes, 2 GiB default) and pins the output frame for faster copies when possible.

On an A40 at full CHIME scale the kernel takes 9 ms (FRB, 2.1 GB table) and ~73 ms (HFB, 17 GB, 8 chunks); the previous CPU stage took minutes to hours on 32 cores. GPU results differ from the CPU at the ~1e-4 level (cospif with recurrence vs cosf(pi x) summed term by term), far below the float16 output resolution.

The new gpu_batch CI test test_calc_frb2_weights.yaml runs both stages on identical inputs and compares the results (tolerance 2e-3; observed max difference 4.9e-4, one float16 ulp).


lib/cuda/cudaCalcFRB2Weights.cpp
```diff
@@ -0,0 +1,277 @@
...
+            //
+            // TODO: When `frb1_swap_MN` is true, the N axis is the geographic x
+            // (east-west) direction, so `sigmaN_x` arguably should be
+            // `get_feed_separation_x_m()`. The value below matches the CPU stage
+            // calcFRB2Weights; review with the CHIME team before changing either.
+            const float sigmaM_x = frb1_swap_MN ? 0 : telescope.get_feed_separation_x_m();
+            const float sigmaM_y = frb1_swap_MN ? telescope.get_feed_separation_y_m() : 0;
+            const float sigmaM_z = 0;
+
+            const float sigmaN_x = frb1_swap_MN ? telescope.get_feed_separation_y_m() : 0;
```
rhaas80: has a bug that is fixed in https://github.com/kotekan/kotekan/pull/1597/commits/1fcf45b08bd8b0f1ab25261d09f7739549c876c6 namely `sigma{N,M}_x` must always be assigned the `_x` component of the separation vector(s), never the `_y` one.  Just now fixed in main.

rhaas80 (changes requested):
I expect the tests are testing against the CPU implementation?

Should be fine then. 30ms are way faster than needed, so simple is fine.

rhaas80 (approved):

PR #1606, opened 2026-08-11
Fix out-of-bounds frame access when a host buffer is shallower than the GPU pipeline

eschnett:
Fix out-of-bounds frame access when a host buffer is shallower than the GPU pipeline

The CUDA commands that page-lock host buffer frames indexed the frame array by command instance number:

    in_buffer->frames[instance_num]

with instance_num in [0, gpu_buffer_depth).  That only holds when the buffer happens to have exactly gpu_buffer_depth frames.  A buffer with fewer frames reads past the end of the vector; with a shallower buffer than the pipeline, this aborted during shutdown in a debug build (where _GLIBCXX_ASSERTIONS catches it) and silently corrupted the unregister call in a release build.

It was also asymmetric with registration: gpuProcess::init() registers every frame of every host buffer it uses, so frames beyond the pipeline depth were never unregistered.

Replace the per-instance indexing with register_host_buffer() and unregister_host_buffer() helpers on cudaCommand, which walk all frames of the buffer once, from instance 0 — matching what gpuProcess::init() does.

### DIFF
--- a/lib/cuda/cudaCommand.cpp
+++ b/lib/cuda/cudaCommand.cpp
@@ -98,6 +98,28 @@ cudaCommand::~cudaCommand() {
     DEBUG("post_events Freed: {:s}", unique_name.c_str());
 }
 
+void cudaCommand::register_host_buffer(Buffer* buf) {
+    if (instance_num != 0 || !buf->frame_size)
+        return;
+    for (int i = 0; i < buf->num_frames; i++) {
+        uint flags;
+        // only register the memory if it isn't already...
+        if (cudaErrorInvalidValue == cudaHostGetFlags(&flags, buf->frames[i]))
+            CHECK_CUDA_ERROR(cudaHostRegister(buf->frames[i], buf->frame_size, 0));
+    }
+}
+
+void cudaCommand::unregister_host_buffer(Buffer* buf) {
+    if (instance_num != 0 || !buf->frame_size)
+        return;
+    for (int i = 0; i < buf->num_frames; i++) {
+        uint flags;
+        // only unregister if it's actually been registered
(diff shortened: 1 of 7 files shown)


lib/cuda/cudaCommand.cpp
```diff
@@ -98,6 +98,28 @@ cudaCommand::~cudaCommand() {
     DEBUG("post_events Freed: {:s}", unique_name.c_str());
 }
 
+void cudaCommand::register_host_buffer(Buffer* buf) {
+    if (instance_num != 0 || !buf->frame_size)
+        return;
+    for (int i = 0; i < buf->num_frames; i++) {
+        uint flags;
+        // only register the memory if it isn't already...
+        if (cudaErrorInvalidValue == cudaHostGetFlags(&flags, buf->frames[i]))
```
jbmertens, line 107: Will this throw a dev-friendly message if the memory is already registered? If not, can we add one?

lib/cuda/cudaCommand.cpp
```diff
@@ -98,6 +98,28 @@ cudaCommand::~cudaCommand() {
...
+    }
+}
+
+void cudaCommand::unregister_host_buffer(Buffer* buf) {
+    if (instance_num != 0 || !buf->frame_size)
+        return;
+    for (int i = 0; i < buf->num_frames; i++) {
+        uint flags;
+        // only unregister if it's actually been registered
+        if (cudaSuccess == cudaHostGetFlags(&flags, buf->frames[i]))
```
jbmertens, line 118: Similar here, would be nice to have a dev-friendly message if cudaHostGetFlags fails.

jbmertens (approved):
LGTM, just a request for a helpful error message if not already there.

PR #1607, opened 2026-08-11
cudaDeviceInterface: report the CUDA error when loading a kernel module fails

eschnett:
`cuModuleLoadDataEx` failures printed a bare "Could not load module data from elf", with no error code and no indication of which kernel file was involved. Report `cuGetErrorString` and the file name instead, so that e.g. an architecture mismatch says "No kernel image is available for execution on the device" rather than nothing at all.


jbmertens (approved):
CI test failures are unrelated.

PR #1608, opened 2026-08-11
Update kernel generators to support SM 12.0

eschnett:
CHARTS uses modern RTX 5090 GPUs which don't run our old kernels any more. Add options to generate code for newer GPUs.


jbmertens (approved):
LGTM.

PR #1609, opened 2026-08-11
Add NDArrayRingbuffer functions to return the current position

eschnett:
Add NDArrayRingBuffer functions to return the current position even when the currently valid region is empty, i.e. when no data may be accessed. This is useful to align ringbuffers with different dimscalings.


jbmertens (approved):
LGTM.

PR #1610, opened 2026-08-11
bufferBadInputs: Don't hold a lock while parsing JSON

eschnett:
This change is not really necessary but the result is more elegant.


jbmertens (approved):
LGTM.

PR #1611, opened 2026-08-11
inventBFMask: Create bad feed masks continuously, not just once

eschnett:


lib/testing/inventBFMask.cpp
```diff
@@ -49,73 +50,91 @@ class inventBFMask : public kotekan::Stage {
         assert(buffer);
         buffer->register_producer(unique_name);
 
-        if ((mode != "all_good") && (mode != "all_bad") && (mode != "auto") && (mode != "manual")) {
+        if (!(mode == "all_good" || mode == "all_bad" || mode == "auto" || mode == "manual"))
```
jbmertens, line 53: !

lib/testing/inventBFMask.cpp
```diff
@@ -49,73 +50,91 @@ class inventBFMask : public kotekan::Stage {
...
+            } else if (mode == "manual") {
+
+                for (int polr = 0; polr < num_polarizations; ++polr) {
+                    for (int dish = 0; dish < num_dishes; ++dish) {
+                        const int idx = dish + num_dishes * polr;
+                        frame[idx] = 1;
+                    }
+                }
+                for (const auto feed : manual_bad_feeds) {
+                    assert(feed >= 0 && feed < std::ptrdiff_t(buffer->frame_size));
```
jbmertens, line 132: Maybe check if feeds are valid in the constructor rather than here?

lib/testing/inventBFMask.cpp
```diff
@@ -49,73 +50,91 @@ class inventBFMask : public kotekan::Stage {
...
+                    frame[feed] = 0;
+                }
+
+            } else {
+                std::abort();
```
jbmertens: Add an error message here.

jbmertens (approved):
Couple small comments.
CI tests failures are unrelated.

lib/testing/inventBFMask.cpp
eschnett: They can change over time.

lib/testing/inventBFMask.cpp
eschnett: Done.

lib/testing/inventBFMask.cpp
eschnett: Check in the constructor anyway.

PR #1612, opened 2026-08-12
test_timeUtil: derive the reference ERA from the UT1 

jbmertens:
Compute ERA4 from the integer nanoseconds instead, so the reference and the value under test refer to the same instant.
(Fixes incorrectly intermittently failing CI tests.)

### DIFF
--- a/tests/boost/timeUtil.py
+++ b/tests/boost/timeUtil.py
@@ -118,9 +118,15 @@ def get_era_nrot_fancy(t):
     return era_deg, n_rot
 
 
-def get_era_nrot_precise(t):
+def get_era_nrot_precise(ut1_ns):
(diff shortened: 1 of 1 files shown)


rhaas80 (approved):

eschnett:
This changes what the test is testing. Would it be better to loosen the tolarance?

jbmertens:
> This changes what the test is testing. Would it be better to loosen the tolarance?

It performs (or should be) the same rounding that is done in the c++ code. So we can actually tighten the tolerance with this change. Or yes, loosen the tolerance a bit, and measure the rounding error instead. 🥔🥔

I'll merge this in, but can change it if you prefer.

PR #1613, opened 2026-08-12
RFI calculation: Properly sync with the bad feed mask

eschnett:
The bad feed mask now changes over time, every `bf_mask_lifetime_in_samples` samples. Ask for a new bad feed mask when necessary.

Update the config files to allocate the bad feed mask correctly.


lib/cuda/cudaRFIS012tilde.cpp
```diff
@@ -96,16 +99,22 @@ cudaRFIS012tilde::cudaRFIS012tilde(kotekan::Config& config, const std::string& u
     num_frequencies(config.get<int>(unique_name, "num_frequencies")),
     num_polarizations(config.get<int>(unique_name, "num_polarizations")),
     num_dishes(config.get<int>(unique_name, "num_dishes")),
+    bf_mask_lifetime_in_samples(
```
jbmertens, line 103: This seems difficult to know in advance.

jbmertens (commented):
Just a note - in general, there will always be some lag between incoming voltage data and the bad feed flags. Not just because coco/choco can be slow, but because some heuristics can only be computed retrospectively, and that takes a bit of time. So even if we have perfect synchronization between nodes, there can be a delay between a feed becoming bad and when the mask changes; this is true even if kotekan starts pulling updates. So maybe it's best to just apply the mask asap, rather than waiting for synchronization.

lib/cuda/cudaRFIS012tilde.cpp
eschnett: It's the lifetime of the current frame. Whenever we create a frame we need to declare in the metadata for how long it is valid. Afterwards we need to create a new frame. (This is what CHIME is currently doing, I am only making their changes work with CHORD.) I'm choosing a bf mask lifetime of 1 second at the moment. We can reduce that if necessary.

lib/testing/gpuSimulateRFISK.cpp
```diff
@@ -101,6 +101,9 @@ using kotekan::Config;
  * @conf  rfi_total_downsampling_factor   Int. Number of fast times `t` in a `t_rfi`. Must divide
  *                              `samples_per_data_set`. Output SK buffers will contain
  *                              `samples_per_data_set` / `downsampling_factor` time samples.
+ * @conf  bf_mask_lifetime_in_samples     Int. Number of FPGA samples that one bad feed mask is
```
jbmertens, line 104: "should" = needs to be validated?

lib/utils/div.hpp
```diff
@@ -17,6 +17,17 @@ auto round_down(T x, U y) {
     return r;
 }
 
+// Round up `x` to the next higher multiple of `y`
```
jbmertens: Looks like this may be unsafe for some type combinations T and U.

jbmertens (approved):

jbmertens:
Couple small comments, but otherwise LGTM. We'll also need to think about how to record when masks have been applied where.

lib/testing/gpuSimulateRFISK.cpp
eschnett: Yes, this testing stage needs to check this at run time.

lib/utils/div.hpp
eschnett: Yes. Yes it is.

eschnett: Fixed.

rhaas80:
It's too late now, but this seems like a terrible commit to me. 

The subject line claims "RFI calculation: Properly sync with the bad feed mask" but then it, for example, makes changes to `chime_keep_gpu_warm.yaml` which is a production file and has nothing to do with the upgrade. The (long!) commit message continues to state that (and I assume it actually does):

```
    `hdf5N2Write` requires `input_order` and neither `write_n2_write.j2` nor
    `chord_pathfinder_recv.j2` supplies it. `763cb911c` made it required and added it
    to `n2_accumulate` in `chord_pathfinder.j2`, but not to any of the writers.
    Set `CHORDBeamformer`, once per config so all the writers inherit it.
```

and while these are true (and there's comments of mine in the upgrade files stating their sillyness) this has nothing to do with the bad feed mask. 

It introduces variables `input_order` and `output_order` for `bufferBadInputs` but that stage does not trasnform any array. It is given a list of channels ids and uses those to create a mask. The entries in the *mask* are in a specific order (which would be `output_order` but since there no "input" this is a bad word to use). The list of channels is always given as channel ids (this happens to be the index of a feed when feeds are sorted in cylinder order in CHIME but is fundamentally the *name* of the feed and not its location in an array). It adds special code for `input_order == output_order` to set up an identity ordering, however that already happens if one was to actually call the full transformation `input_index` -> station (channel) id -> `output_index` when `input_order` and `output_order` are the same. See eg https://github.com/kotekan/kotekan/blob/a52a01f46a4f8d3122226611c8020b397add76cd/lib/stages/N2Accumulate.cpp#L88

There is are so many changes in this commit that is very hard to understand. Multiple smaller commits that each only changes a single small set of functionality would have been much preferred (by me, had I paid attention, sorry).

Is the "read without consuming" functionality really required? If all we want to do is wait for new bad-feeds-mask to be available then we can have that one have a time dimension sufficiently downscaled then use the (known) dimscaling value to predict when a new value needs to be used and call `wait_for_full_frame` (or whatnot) only then. Less new functionality that needs to be mantained.

All too late to be really helpful, I know.

PR #1614, opened 2026-08-12
NDArrayRingBuffer: Allow reading data without claiming them

eschnett:
`wait_and_claim_readable` used to loop until the caller claimed a strictly positive number of elements. It now loops until the caller *reads* a strictly positive number of elements, which makes the case `claimed == 0 && read > 0` expressible: the data are read now and will be read again during the next iteration, without ever being consumed. This is what a slowly varying input (e.g. a bad feed mask) needs -- it is read on every iteration but advances only rarely.

When nothing is claimed we skip `RingBuffer::wait_and_claim_readable` (which requires a positive size) and start reading at the unchanged read head; `finish_read` likewise skips `RingBuffer::finish_read`. The data stay protected because the consumer's read tail does not move either.

This means the ringbuffer no longer guarantees that a consumer makes progress -- a consumer that always claims zero now spins instead of blocking. All callers were audited:

- The generated upchannelizer, FRB and CHIME-FRB kernels could return `claimed == 0` together with `read > 0` and relied on the old behaviour to wait for more data. They now return `read = 0` in that case. This happens for real configurations, e.g. the U128 upchannelizers (overlap 384 > granularity 256) and the U1/U2 FRB beamformers (downsampling 192 > granularity 48).
- All other callers already satisfied `claimed == 0 => read == 0`.

Add `NDArrayRingBuffer::check_read_progress`, and call it from every consuming stage's constructor. It verifies that the largest amount a stage will read at once is at least the smallest amount from which it can claim anything, so that a too small ringbuffer is diagnosed at startup instead of hanging at run time. Several stages sit exactly on this boundary today, e.g. cudaRFIS012bar (ringbuf/4 == 32 == rfi_second_downsampling_factor) and cudaBasebandBeamformer_chord (T_ringbuf/4 == granularity == 8192).

Also add `get_maybe_empty_read_claimed()`, since `get_read_claimed()` rejects an empty claim.

Generated kernels are updated in all three places they are kept: the Julia templates, `julia/output`, and `lib/cuda/generated`.


lib/cuda/NDArrayRingBuffer.hpp
```diff
@@ -335,28 +363,38 @@ class NDArrayRingBuffer : public kotekan::kotekanLogging {
...
+            // read tail does not move either.
+            read_begin = div_noremainder(read_head_bytes, granularity_in_bytes());
+        }
 
-        const std::ptrdiff_t read_begin = div_noremainder(claimed.value(), granularity_in_bytes());
         read_valid = extent_t(read_begin, read_begin + read_descriptor.read);
         read_claimed = extent_t(read_begin, read_begin + read_descriptor.claimed);
         assert(read_valid.size() > 0);
-        assert(read_claimed.size() > 0);
+        assert(read_claimed.size() >= 0);
```
jbmertens, line 385: Is this check actually checking anything now? Would the size ever be negative?

lib/utils/div.hpp
```diff
@@ -17,6 +17,17 @@ auto round_down(T x, U y) {
     return r;
 }
 
+// Round up `x` to the next higher multiple of `y`
+template<typename T, typename U>
```
jbmertens, line 21: This won't always work with mixed types. Also we don't usually need to worry about overflow, but for a function like this it could be important. Some explicit casting / behavior notes could help.

This also applies to the functions below.
I don't actually see this function used in this PR?

lib/cuda/NDArrayRingBuffer.hpp
jbmertens, line 327: Not sure I understand what's going on here. I'll need to read through the buffer class to understand.

jbmertens (commented):
Few comments - will look at in more detail later.

eschnett:
This PR was superseded by #1613.

PR #1615, opened 2026-08-14
Support conjugation of FRB gains

ljgray:
Most of the diff here is related to merge conflicts rather than actual changes. This branch was branched off of `frbNetwork4`, but was then rebased against `develop` (with a few cherry-picked commits from other branches removed).

The only actual changes here should be a couple of minor updates to `chime_upgrade_frb.j2`, and options to conjugate and swap real/imaginary ordering in the `processFeedGains` and `loadFeedGains` stages.


rhaas80:
let me rebase frbNetwork4 on develop then to see if I can reduce the number of changed files

rhaas80 (approved):

PR #1616, opened 2026-08-16
EigenN2Iter: mask the feeds the N2 frames flag as bad

jbmertens:
## Summary

`EigenN2Iter` built its solver mask once from `exclude_inputs`, so a feed that went bad during a run kept contributing until a config edit and restart. This PR folds the N2 frames' per-element `flags` into the mask.

- `EigenN2Iter` masks the elements the incoming frame flags as bad (0.0), alongside `exclude_inputs`. The mask is rebuilt only when the flags change. `mask_flagged_inputs` (default **false**) turns this on.
- A frame with no more unmasked elements than `num_ev` is reported failed (to `failed_buf` where configured) instead of solved.
- `N2Accumulate` fills `flags` with 1.0 (good) instead of 0, which under the pipeline-wide convention meant every element bad. The convention is documented on `N2FrameView::flags`.
- `exclude_inputs` and `diagonal_bands_filled` are validated against the frame's element count.
- New metrics: `masked_elements`, `num_mask_updates`, `num_insufficient_elements`.
- `FakeN2` gains `flagged_inputs` / `flag_start_frame`, with five new boost cases. Sphinx: `EigenN2Iter` page added; N2 HDF5 pages no longer list `flags` = 0 as a sentinel.

## Notes

- `eval0` scales with the number of unmasked elements, so it steps when the mask changes. The flags travel on the output frame, so the count is recoverable per frame.
- Masked entries are matrix-completed, as for `exclude_inputs`. Any non-zero flag counts as good.
- The flags fill change reaches every `N2Accumulate` consumer: the `flags` dataset in N2 HDF5 files changes from 0 to 1, and flag-honouring stages downstream of `N2FrameToVisFrame` (e.g. `BadInputFlag`) become active.
- `mask_flagged_inputs` defaults off: every existing N2 producer predates it, so a config that does not opt in still sees flags the pipeline never populates with meaning, and CHIME's `EigenN2Iter` deployments keep their current unmasked behaviour with no config change. Turning it on still needs both ends of a `bufferSend` link on this branch: a consumer reading an older producer's all-zero flags fails every frame.
- `EigenVisIter` (CHIME vis path) is unchanged; a changing mask there needs a datasetState.


jbmertens:
Once we have full pathfinder up and running, we may not want to mask bad feeds as CHIME does. But for now with so many feeds "bad", this might be the only way we get a sensible eigendecomposition.

rhaas80:
`EigenVisIter` not actually used in CHIME since it uses `EigenN2Iter` and `N2FrameToVisFrame` instead. So likely it could be removed from the source code.

lib/stages/EigenN2Iter.hpp
```diff
@@ -64,7 +64,15 @@
  * @conf  block_fill_size  UInt, default 0. Mask out blocks of this size on the diagonal.
  * @conf  exclude_inputs   List of UInts, optional. Inputs to exclude (rows and
  *                         columns to set to zero) in visibilities prior to
- *                         factorization.
+ *                         factorization. These are indices into the incoming
+ *                         frame's elements, which for a subsetted buffer are the
+ *                         subset's own indices, not the full array's.
+ * @conf  mask_flagged_inputs  Bool, default true. Also mask the inputs the
```
andrerenard: Do we want to set this to false for the CHIME system (default false)?    While updating the mask seems like the best way to do this, I'm not sure why CHIME kept a static mask.  Allowing the mask to be dynamic for CHIME might cause some unexpected effects, particularly around rain events.   This might need a wider discussion with the CHIME calibration group.   I'm also not sure what the effects would be on corrcal of changing the mask mid-transit?

andrerenard (approved):
This looks good to me, in the narrow sense that I don't see anything wrong with this code's application of the mask.    There is however a larger question of how we get these flags and _if and when_ we should apply these flags.  I don't recall why CHIME decided not to update this mask, was it just because we didn't implement it, or is there some interaction between changes to the mask and how corr_cal makes use of the resulting eigenvalues?   In addition do we have to worry about situations like a change to the mask in the middle of a calibration source transit?

lib/stages/EigenN2Iter.hpp
jbmertens: Yeah, probably should default false so nothing is affected.

jbmertens:
CHIME doesn't apply an active mask, as I understand, for a couple reasons:
1. Its bad feed flagger gets stuck in a flag-this-feed-as-bad loop, if the feeds are being masked. So enabling this on CHIME will cause some feeds to be permanently flagged. For CHORD, the flagger is being redesigned so we can adjust this behavior.
2. A couple, or even a few, bad feeds doesn't actually change the eigenvalues appreciably. But for CHORD 80% of our feeds are "bad" at the moment.

lib/stages/EigenN2Iter.hpp
rhaas80: I recall being old that there was a feedback loop if the life bad feeds were used in computing the eigenvalues (that are used to compute the gains). Mandana may know .

jbmertens: It is set to false by default, so shouldn't affect behavior.

PR #1619, opened 2026-08-18
chordMetadata: Fix the fixed-length string fields and the serializers

eschnett:
We use fixed-size strings on the `chordMetadata` class. We don't always handle them correctly. Dustin introduced safe wrappers some time ago, but not all code uses them. This PR fixes this, as well as some related edge cases in serialization.

The bulk of this PR is for metadata serialization. I think much of this code – the old-style serialization which doesn't use JSON – is probably outdated and could be removed. But that's a separate decision.

Implemented with the help of AI. Their explanation follows.

The fixed-size `name` and `dim_name` fields are deliberately not NUL-terminated, which every accessor inside the class handled correctly (strnlen bounded by the field size) but callers did not: asdfFileWrite passed `meta->name` straight to fmt, which applies strlen and reads past the field for a maximal-length name. Document the convention on CHORD_META_MAX_DIMNAME and on the fields themselves, use get_name() at that call site, and route the remaining direct writers through set_name()/the new set_dimension_name() so the NUL padding and the truncation warning live in one place.

Serialization fixes:

- serialize() called get_nfreq() unconditionally, which throws when COARSE_FREQ is not set. Per-frequency arrays are now optional, and the ones that are set must agree on their length rather than being silently resized to the length of coarse_freq. Since such an array is either unset or non-empty, a sentinel in its first element is enough to say which; the setters enforce that invariant.
- STREAM_IDS, RFI_FRAME_EXCISION_ENABLED and RFI_FRAME_EXCISION_THRESHOLDS were dropped by both serializers. They are in the byte format now, and from_json copies every non-structural key instead of a hand-maintained list, so new metadata entries cannot go missing again.
- to_json() built vectors from an inverted iterator range for a default-constructed object (dims == -1).
- set_from_bytes()/from_json() validated their input with asserts only, so a release build would write past dim[], dim_name[] etc. from data that bufferRecv accepted off the network. The checks are real now, and dims == CHORD_META_MAX_DIM is no longer rejected as out of range.
- to_json()/from_json() accessed the json metadata without holding the lock.
- deepCopy() adopted the source object's metadata pool.

The byte format grew fields for the entries that used to be dropped, so its size changed; bufferRecv already rejects a metadata size it does not expect, so a version skew fails cleanly rather than silently corrupting frames.

jsonMetadata:

- The `#ifdef DEBUG` around the beamCoord tail fill was never true for a debug build (the build defines DEBUGGING) and true only for translation units that had included kotekanLogging.hpp first, which defines DEBUG as a logging macro. Without the fill, from_json left the unused beams uninitialized and to_json then read indeterminate floats. The fill is unconditional now.
- Removed the unused jsonMetadata::CHORD_META_MAX_FREQ, whose value (4096) disagreed with the global one actually in force (12288).
- MAX_NUM_BEAMS and MAX_NUM_RFI_THRESHOLDS are constants rather than macros.
- Moved the serializers out of the header and range-check timeval on the way in.

Also: atomic_add_lost_timesamples() requires the count to be set, bounds are checked instead of asserted, sample_bytes() and get_dimension_name() reject unset/out-of-range input, the mutex assignment operator returns a reference, locals no longer shadow the `lock` member, and set_from_frame_desc() does copy the array name (the doc comment said otherwise) and reports invalid Symbols instead of passing a null pointer to strncmp.


lib/metadata/chordMetadata.cpp
```diff
@@ -194,161 +248,305 @@ size_t chordMetadata::get_serialized_size() {
...
+// would hide the `fmt` namespace used here and by the logging macros.
+//
+// These two functions validate data that is exchanged with other kotekan
+// instances (see bufferSend/bufferRecv), so they throw instead of calling
+// FATAL_ERROR; the caller decides how bad a malformed frame is. This also matches
+// from_json below.
+
+size_t chordMetadata::set_from_bytes(const char* bytes, size_t length) {
+    if (length < sizeof(chordMetadataFormat))
+        throw std::runtime_error(
```
jbmertens: Consider FATAL_ERROR here. (And other throws below)

lib/metadata/chordMetadata.cpp
```diff
@@ -161,27 +197,45 @@ struct chordMetadataFormat {
...
 
-    // Per-frequency arrays
+    // Per-frequency arrays. Each array is either unset or has `nfreq` entries,
+    // and `nfreq` is at least 1 whenever any of them is set. An array is unset
+    // when its first element holds the invalid value noted below, which is
+    // unambiguous exactly because a set array is never empty.
     int32_t nfreq;
 
-    // frequencies -- integer (0-2047) identifier for FPGA coarse frequencies
+    // frequencies -- integer (0-2047) identifier for FPGA coarse frequencies;
```
andrerenard: Stale comment. 
```suggestion
    // frequencies -- integer (0-CHORD_META_MAX_FREQ) identifier for FPGA coarse frequencies;
```

lib/metadata/chordMetadata.hpp
```diff
@@ -27,19 +28,29 @@
...
+// These names are stored in fixed-size char arrays that are NOT NUL-terminated:
+// a name of exactly CHORD_META_MAX_DIMNAME characters fills its field
+// completely. The fields are NUL-padded, so shorter names do end in a NUL, but
+// no code may rely on that. Read them via get_name() and get_dimension_name(),
+// which bound the length with strnlen; never pass them to strlen, strcpy,
+// printf("%s") or fmt, all of which read until they find a NUL.
 const int CHORD_META_MAX_DIMNAME = 24;
 
+// Maximum number of stream IDs in metadata array
+const int CHORD_META_MAX_STREAM_IDS = 64;
```
andrerenard: What is the idea behind having a `stream_id` in the CHORD metadata?  I do have a concept of stream_ids as it related to the F-engine packet format.  Is that what this is trying to capture, or is this a different stream concept?

lib/metadata/chordMetadata.hpp
```diff
@@ -79,18 +90,26 @@ class chordMetadata : public metadataObject {
...
+            // A mutex is not copied; this object keeps its own.
             return *this;
         }
     } lock;
 
     // TODO: Replace by NDArray
-    char name[CHORD_META_MAX_DIMNAME]; // "E", "J", "I", etc
+    /// The name of the array, e.g. "E", "J", "I". NUL-padded but not
+    /// NUL-terminated when the name uses all CHORD_META_MAX_DIMNAME characters;
+    /// use set_name() and get_name() instead of accessing the field directly.
```
andrerenard: Why not just add one to the sizes so they can always be NULL-terminated?   Or if we aren't supposed to access this directly, should these be private members?

lib/metadata/chordMetadata.hpp
```diff
@@ -79,18 +90,26 @@ class chordMetadata : public metadataObject {
...
+    char name[CHORD_META_MAX_DIMNAME];
```
andrerenard: I realize this isn't part of the PR scope, but `CHORD_META_MAX_DIMNAME` seems like it should be a different constant here (even if it has the same value).

lib/metadata/chordMetadata.hpp
```diff
@@ -352,18 +381,28 @@ class chordMetadata : public metadataObject {
         return metadata.at(jsonMetadata::LOST_TIMESAMPLES).template get<int32_t>();
     }
 
+    /// Adds to the lost time sample count. The count must already be set;
+    /// otherwise there is no way to tell "no samples lost" from "not counted".
     void atomic_add_lost_timesamples(const int32_t lost_samples) {
```
andrerenard, line 390: I need to check if this is still needed... Will remove separately if it isn't.

lib/metadata/jsonMetadata.hpp
```diff
@@ -14,14 +14,15 @@
...
 
 // json based metadata structures to hold "physics" metadata, ie. data that
-// various at most once per frame
+// varies at most once per frame
+
+// These must match chimeMetadata and chordMetadata for now. They are used
+// unqualified all over the code base, so they live in the global namespace
+// (they used to be macros).
+constexpr int MAX_NUM_BEAMS = 20;
+constexpr int MAX_NUM_RFI_THRESHOLDS = 8;
```
andrerenard, line 23: I think both these values apply only to code which is no longer used, but they can stay for now.  I need to do a code clean up soon...

andrerenard (changes requested):

lib/metadata/chordMetadata.hpp
eschnett: That decision (fixed-length non-NUL-terminated strings) was made ages ago, when we were worried about memory bandwidth of copying metadata.

These days we have metadata in JSON, so all these decisions could be updated. This PR, however, only fixes code according to the current standard. Also, much of these metadata now live in buffer objects anyway and could be removed, so I'm not keen on redesigning it.

lib/metadata/chordMetadata.hpp
eschnett: It should. But then, this should probably have been a `std::string` instead...

lib/metadata/chordMetadata.hpp
eschnett: I do not know. I assume that yes, this is about the F-engine's streams.

We inherited much of this from CHIME. The comment is "Stream IDs - the stream identifiers for packets received". I don't think we're either setting or reading it in any code I've touched, except here in the metadata.

andrerenard: Ah, I got confused with another PR, and thought this might be something else.   stream_ids are used in part of the F-engine packet handling logic in the current code, but the value you should 128 for full CHORD compatibility. 
```suggestion
const int CHORD_META_MAX_STREAM_IDS = 128;
```

lib/metadata/chordMetadata.hpp
andrerenard: That's fair, is it worth making it `private` though to catch any current (or future) direct accesses?

eschnett: The helpers were added late. There is existing code that accesses the fields directly. I can make these private, but that might become a larger PR, including possibly regenerating Julia-generated kernels. What do you suggest?

andrerenard: I think it's worth making them private and either switching over the rest of the code to use the accessors.  It's a strange enough field with the lack of NULL termination to possibly cause problems in the future.  Having said that, we could do that as a separate PR to make the changes smaller/easier to review.   I'll leave that upto you, and approve for now.

andrerenard (approved):

PR #1620, opened 2026-08-18
Report log events at run time, not by macro redefinition

eschnett:
The way `FATAL_ERROR` is implemented violates the C++ standard (the ODR rule). The `FATAL_ERROR` macro expands to different functions in different parts of Kotekan, which is not allowed for inline functions.

AI found this, its explanation follows.

## Summary

`kotekanLogging.hpp` defined `KTK_BOOST_ERR`/`KTK_BOOST_WARN` two different ways
depending on whether `BOOST_TEST_MODULE`/`BOOST_TEST_MAIN`/`BOOST_TEST_DYN_LINK` was
defined when the header was included — throwing in a boost test, `(void)0` everywhere
else. `ERROR`, `ERROR_NON_OO`, `WARN`, `WARN_NON_OO`, `CHECK_ERROR` and `CHECK_MEM`
expand to them, and so in turn do `FATAL_ERROR` and `EXIT_ERROR`.

That condition is true in a boost test translation unit and false in every library
translation unit. So **every function defined in a header that logs had two different
bodies**. Inline functions, in-class member definitions and function templates have
external linkage and are emitted as weak symbols, so this was an ODR violation: the
linker keeps one arbitrary copy and silently discards the rest.

This PR replaces the compile-time switch with a runtime hook, so the macros expand to
the same tokens in every translation unit.

## Why this matters

**It's not a corner case.** 22 headers under `lib/` contain ~126 uses of the affected
macros, including `lib/core/Config.hpp`, `lib/core/buffer.hpp`,
`lib/core/configTracker.hpp`, `lib/metadata/chordMetadata.hpp` and
`lib/cuda/NDArrayRingBuffer.hpp`.

**It already misbehaves, in both directions.**

- *Library copy wins in a test binary.* An error the test expected to observe is
  ignored, and `FATAL_ERROR` runs on to `exit_kotekan()`, taking the whole test process
  down with SIGTERM instead of failing one assertion. Four tests carry hand-rolled
  workarounds for exactly this symptom (three copies of an `IgnoreSigterm` fixture, plus
  a `sigaction`-based `SigtermGuard` in `test_hdf5N2Write`).
- *Test copy wins inside library code.* `test_config.cpp:115-117` depends on this today:
  `Config::get<T>` reaches `ERROR_NON_OO` through the expression-parser templates in
  `Config.hpp`, and the test asserts those throw.

**Nothing diagnoses it.** GCC's `-Wodr` covers differing *type* definitions, not
differing function bodies. Which copy wins depends on link order, archive member
selection and LTO (`-DUSE_LTO=ON`), so an unrelated change to a link line can flip it.

**Demonstration.** Two translation units that both emit the same header-inline logging
function, one with `BOOST_TEST_MODULE` defined and one without, compiled and compared:

|        | boost TU | library TU |
| ------ | -------- | ---------- |
| before | 161 instructions | 53 instructions — different mnemonic sequence |
| after  | 155 instructions | 155 instructions — identical mnemonic sequence |

## What changed

**Commit 1 — the fix.** `kotekan::log_event`, `log_event_handler`, an inline
`log_event_hook` and `report_log_event()` in `kotekanLogging.hpp`. The reporting macros
are renamed `KTK_REPORT_ERROR`/`KTK_REPORT_WARNING`, since nothing about them is boost
specific any more. New `tests/boost/kotekanTestLogging.hpp` supplies
`kotekan_logging_fixture`, installed via `BOOST_GLOBAL_FIXTURE` in all 34 boost tests;
it throws `std::runtime_error` on an error and registers a boost warning on a warning —
the behaviour test translation units already had, now applied uniformly instead of
depending on the linker.

Two deliberate details:

- The hook is **header-only** (C++17 inline variable + inline function) rather than
  living in `kotekanLogging.cpp`. An out-of-line definition would make it a
  `kotekan_core` symbol, but several tests link only `kotekan_libs kotekan_utils`. This
  way there is no new link dependency at all.
- The macros check for an installed handler **before** formatting, so production pays one
  relaxed atomic load instead of a string format on every `WARN` and `ERROR`.

**Commit 2 — cleanups.** Retires all four SIGTERM workarounds, now subsumed by the
fixture. Fixes `lib/dpdk/FramePrefetchService.hpp:276`, where a debug block was guarded
by `#ifdef DEBUG` — `DEBUG` is a *logging macro*, not a build flag (the build defines
`DEBUGGING`), so the block compiled only when the logging header happened to be included
first. Adds a `tools/lint.sh` check rejecting `#if`/`#ifdef`/`#elif` on the boost test
macros under `lib/`, so a header cannot go back to changing meaning per translation unit.

## Behaviour change to be aware of

A library-TU `ERROR` used to be a silent no-op inside a test binary; it now reaches the
handler and throws. **Tests that pass today while quietly logging errors from library
code will start failing.** Those failures are genuine findings, but they need triage
rather than being assumed benign — this is the main thing to watch in CI.

Relatedly, `FATAL_ERROR` now throws from its leading `ERROR(...)` before reaching
`exit_kotekan()`, uniformly. That is what a boost TU already did; it is preserved rather
than newly introduced, and it is what makes the `IgnoreSigterm` workarounds removable.
`FatalError` derives from `std::runtime_error`, so existing `BOOST_CHECK_THROW`
assertions match either exception.

## Verification

Developed on macOS, where kotekan does not build (`cpu_set_t` in `lib/core/buffer.hpp`),
so coverage is partial and **CI is the real gate**:

- The before/after machine-code comparison above.
- `test_config` builds and passes — the important one, since its throw assertions run
  through `ERROR_NON_OO` inside a `Config.hpp` template, i.e. the ODR-affected path.
- `test_N2FrameDesc` builds and passes with `IgnoreSigterm` removed, confirming the
  fixture subsumes it for out-of-line `FATAL_ERROR_NON_OO`.
- `test_truncate` builds, links with no extra objects and passes, confirming no new link
  dependency for tests that do not link `kotekan_core`.
- The new lint check flags the pre-fix header and passes on this branch.

The remaining 31 boost tests are unverified locally. Worth a `-DUSE_LTO=ON` run too,
since LTO is what makes the arbitrary-copy selection most volatile.


jbmertens:
Like the idea. Part of the reason for the original behavior was we wanted a difference between the boost tests vs. ordinary operation - will review once those are passing again.

jbmertens:
A Claude review for Claude code 🤖 

-------

## Summary

The analysis is correct and the fix has the correct form. I checked: no `#if` on `BOOST_TEST_*` stays in `lib/` or `tests/`; the lint expression finds `kotekanLogging.hpp:18` on `develop` and stops the CI (`tools/lint.sh:186`); all 34 tests have the fixture; every `BOOST_CHECK_THROW` uses `std::runtime_error` or `std::exception`, so the `FatalError` swap breaks none; all `lib/` targets are static, so the hook has one instance per binary. The CI passes.

## Problems

**1. The handler throws and calls Boost.Test macros from libevent threads.** Boost.Test assertions are not safe for more than one thread.

- `restClient.cpp:509` writes a WARN on the client event thread, and `test_restclient.cpp:192` requests `localhost:1` on purpose. `BOOST_WARN_MESSAGE` then races the main thread. Same at `restClient.cpp:178,217` and `restServer.cpp:450`.
- `restServer.cpp:273,278,302,307` (and `474,481,492`) write `ERROR_NON_OO` on the server thread in `http_general_callback`. The throw goes into the C code of libevent, nothing catches it, and the result is `std::terminate`. Those catch blocks exist to return a 500. The first test that reaches one stops the process instead of failing an assertion.

Fix: keep the thread id in the fixture constructor. Throw and call Boost.Test macros only on that thread. Collect messages from other threads and report them with `BOOST_ERROR` at the end.

The `noexcept` destructors at `restClient.cpp:50` and `RawReader.hpp:400` can do the same, but no test reaches them.

**2. The CMake change opens the same problem again.** `include_directories(.../lib/core)` gives each test the `lib/core` headers without the compile definitions of `kotekan_core`, such as `WITH_SSL` (`lib/core/CMakeLists.txt:82`). Before, a test that included `Config.hpp` also linked `kotekan_core` and got the definition. `Config.hpp:245` puts `#ifdef WITH_SSL` around a member declaration in `class Config` — the same header with two meanings, now on a type. Nothing fails today, but the protection is gone and the lint check cannot find it. Use an INTERFACE target with the same usage requirements as `kotekan_core`.

**3. The code now calculates log arguments always.** `KTK_BOOST_ERR` was `(void)0`; `KTK_REPORT_ERROR` is a call, so the arguments are calculated with or without a handler, and twice when the log level is on. 679 sites. `CHECK_ERROR` also calls `strerror(errno)` twice, with `internal_logging` between, so the second can give a different errno. I found no such call in a per-frame loop, so this is wasted work, not a real-time danger — but the header comment ("production pays one relaxed load") says more than is correct. Fix:

```c
#define KTK_REPORT_ERROR(m, ...)                                                                   \
    (kotekan::log_event_hook.load(std::memory_order_relaxed)                                       \
         ? kotekan::report_log_event(kotekan::log_event::error, __FILE__, __LINE__, fmt(m),        \
                                     ##__VA_ARGS__)                                                \
         : (void)0)
```

**4. The `FakeVisPattern` handlers log and return, and the frame stays empty.** Kotekan stops rather than send data that can be bad. Use `FATAL_ERROR`. The commit keeps this for later, but the PR already changes these lines. Also, the reachable fault was the `{:d}` on a `std::string`, not a null `pool`: the pool lifetimes do not permit a dead pool under a live frame view, so `pool ? … : "unknown"` is extra code that makes the message worse.

**5. `set_error_message()` and `exit_kotekan()` no longer run for a library `FATAL_ERROR` in tests**, because the first `ERROR` throws. `get_error_message()` in `test_logging.hpp:15` is now empty for exactly those errors. Include this in the consolidation the header mentions.

**6. About 50 lines of commit-message history moved into two headers.** The comments on `log_event_handler` and `report_log_event`, and the block in `kotekanTestLogging.hpp`, tell history, not what the code does. Keep the rule that the macros must expand to the same tokens in each translation unit. Remove the rest.

**7. `log_event` repeats `logLevel`.** `kotekan::logLevel` already has `ERROR = 1, WARN = 2` at `kotekanLogging.hpp:232`. Give `logLevel` to the handler instead.

## Minor items

- Three names for one item: file `kotekanTestLogging.hpp`, guard `KOTEKAN_LOGGING_FIXTURE_HPP`, type `kotekan_logging_fixture`. A guard mismatch was the bug commit 4 fixed. Also `log_event` uses underscores next to `logLevel` and `FatalError`.
- The handler builds its location string with four allocations. `FORMAT("{}:{}: {}", …)` does one.
- The lint check also scans `.c` and `.cpp`, where a per-unit branch does no damage. Scan headers only.
- The fixture line splits a comment from the test it describes in `test_NDArrayBufferConfig.cpp`, `test_hdf5N2Write.cpp` and `test_dataset_manager.cpp`.
- The fixture restores the hook but not the previous SIGTERM disposition.
- `#ifdef DEBUG` → `DEBUGGING` (`FramePrefetchService.hpp:276`) is good. The `stream_id_list` loop above it (259–266) is the same wasted work in a release build, and has no guard.

jbmertens (changes requested):
Just explicitly requesting changes -- I'll re-review once the AI's comments are addressed.

eschnett:
@jbmertens I implemented "your" corrections and suggestion. Please re-review.

PR #1623, opened 2026-08-19
ConfigTracker updates & FPGA monitoring

jbmertens:
- The configTracker should support kotekan instances running on other ports.
 - Kotekan should poll the FPGA occasionally to verify consistency (otherwise, new acquisition).
 - Hopefully also fixes some daemon restart issues...

### DIFF
--- a/config/chord_pathfinder.j2
+++ b/config/chord_pathfinder.j2
@@ -607,6 +607,16 @@ host_rfi_sktilde_ringbuffer_1:
 
 # Stages
 
+##  Monitor the F-engine for drift ##
+# The config tracker reads the F-engine controller once at startup and hands
+# that snapshot downstream; this re-reads it so a controller that is
+# reprogrammed or resynced mid-acquisition is caught rather than silently
+# invalidating every config already propagated.
+fpga_monitor:
+    kotekan_stage: FPGAMonitor
+    poll_interval_seconds: 5
+    fetch_timeout_seconds: 5
+
 ##  Packet capture from the F-engine ##
 dpdk:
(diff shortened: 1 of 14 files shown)


config/chord_pathfinder.j2
```diff
@@ -607,6 +607,16 @@ host_rfi_sktilde_ringbuffer_1:
...
 # Stages
 
+##  Monitor the F-engine for drift ##
+# The config tracker reads the F-engine controller once at startup and hands
+# that snapshot downstream; this re-reads it so a controller that is
+# reprogrammed or resynced mid-acquisition is caught rather than silently
+# invalidating every config already propagated.
+fpga_monitor:
+    kotekan_stage: FPGAMonitor
+    poll_interval_seconds: 5
```
andrerenard, line 617: This might be more frequent (and more load) than needed.   A reset in the seq number will cause kotekan to shutdown already.  So I think this only needs to guard against the case that between kotekan startup, and when it starts to get the first packets, the F-engine resets.

andrerenard (approved):
Other than the question about timing, this looks good to me.  No harm in having extra checks against inconsistencies.

config/chord_pathfinder.j2
jbmertens: It's more useful for the recv node, which doesn't run the dpdk checks. This is also a full check of the FPGA config, so should check more than just the sequence number (which was misreporting recently).

jbmertens: But yeah, maybe it can be relaxed as things stabilize...

PR #1624, opened 2026-08-20
Fixes bad feed mask reorder and optionally reorders feeds before sending in RFI broadcast

ljgray:
Confirmed that this produces the expected mapping (compared to manually reordering the bad_inputs updatable_config list in python)

### DIFF
--- a/lib/stages/bufferBadInputs.cpp
+++ b/lib/stages/bufferBadInputs.cpp
@@ -1,7 +1,8 @@
 #include "bufferBadInputs.hpp"
 
-#include "Config.hpp"         // for Config
-#include "N2Util.hpp"         // for frameID
+#include "Config.hpp" // for Config
+#include "N2Util.hpp" // for frameID
+#include "NDArray.hpp"
 #include "StageFactory.hpp"   // for REGISTER_KOTEKAN_STAGE
 #include "Telescope.hpp"      // for Telescope, station_id_t
 #include "buffer.hpp"         // for Buffer
@@ -34,7 +35,7 @@ bufferBadInputs::bufferBadInputs(Config& config_, const std::string& unique_name
     out_buf->register_producer(unique_name);
 
     // Construct the cylinder -> beamformer reorder table.
-    // reorder[beamformer_idx] = cylinder_idx;
+    // want reorder[cylinder_idx] = beamformer_idx
     reorder.resize(num_elements);
 
     // initialize the mask (1 == good)
@@ -42,12 +43,15 @@ bufferBadInputs::bufferBadInputs(Config& config_, const std::string& unique_name
 
     const Telescope& tel = Telescope::instance();
 
-    for (size_t beamformer_idx = 0; beamformer_idx < num_elements; ++beamformer_idx) {
-        station_id_t st_id =
-            tel.element_index_to_station_id(beamformer_idx, ElementOrder::CHIMEBeamformer);
-        reorder.at(beamformer_idx) =
-            tel.station_id_to_element_index(st_id, ElementOrder::CHIMECylinder);
+    // cylinder order is already gives the station id
+    for (station_id_t cyl_idx = 0; cyl_idx < num_elements; ++cyl_idx) {
+        reorder.at(cyl_idx) =
+            tel.station_id_to_element_index(cyl_idx, ElementOrder::CHIMEBeamformer);
     }
+
+    // Set the frame description
+    out_buf->ensure_frame_desc(kotekan::NDArray<kotekan::GetType_t<kotekan::uint8>, 1>::describe(
+        "bad_inputs", {static_cast<ptrdiff_t>(num_elements)}, {"E"}, {1}));
 
     // Listen for bad input list updates
     std::string badInputs = config.get<std::string>(unique_name, "updatable_config/bad_inputs");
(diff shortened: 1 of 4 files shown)


lib/stages/bufferBadInputs.cpp
```diff
@@ -1,7 +1,8 @@
 #include "bufferBadInputs.hpp"
 
-#include "Config.hpp"         // for Config
-#include "N2Util.hpp"         // for frameID
+#include "Config.hpp" // for Config
+#include "N2Util.hpp" // for frameID
```
rhaas80, line 4: does this pass the linter? Different number of " " before the `//`

lib/stages/bufferBadInputs.cpp
```diff
@@ -34,21 +35,23 @@ bufferBadInputs::bufferBadInputs(Config& config_, const std::string& unique_name
...
 
     const Telescope& tel = Telescope::instance();
 
-    for (size_t beamformer_idx = 0; beamformer_idx < num_elements; ++beamformer_idx) {
-        station_id_t st_id =
-            tel.element_index_to_station_id(beamformer_idx, ElementOrder::CHIMEBeamformer);
-        reorder.at(beamformer_idx) =
-            tel.station_id_to_element_index(st_id, ElementOrder::CHIMECylinder);
+    for (size_t cyl_idx = 0; cyl_idx < num_elements; ++cyl_idx) {
+        station_id_t st_id = tel.element_index_to_station_id(cyl_idx, ElementOrder::CHIMECylinder);
```
rhaas80: here (and below): should be declare that the input is a list of channel (station) ids (instead of index into of data sorted in cylinder order) ? That would remove the first call:

```
station_id_t st_id = tel.element_index_to_station_id(cyl_idx, ElementOrder::CHIMECylinder);
```

and make it such that input is always the "name" of the channel and never some index into an array.

lib/stages/rfiBroadcast.cpp
```diff
@@ -1,8 +1,9 @@
 #include "rfiBroadcast.hpp"
 
-#include "Config.hpp"          // for Config
-#include "N2Util.hpp"          // for frameID, modulo
-#include "StageFactory.hpp"    // for REGISTER_KOTEKAN_STAGE
+#include "Config.hpp"       // for Config
+#include "N2Util.hpp"       // for frameID, modulo
+#include "StageFactory.hpp" // for REGISTER_KOTEKAN_STAGE
```
rhaas80, line 5: passes the linter?

lib/stages/rfiBroadcast.cpp
```diff
@@ -54,6 +55,8 @@ rfiBroadcast::rfiBroadcast(Config& config, const std::string& unique_name,
     frames_per_packet = config.get_default<size_t>(unique_name, "frames_per_packet", 1);
     dest_port = config.get<size_t>(unique_name, "destination_port");
     dest_ip = config.get<std::string>(unique_name, "destination_ip");
+    // whether or not a reorder should be applied
+    apply_reorder = config.get_default<bool>(unique_name, "apply_reorder", true);
```
rhaas80: not sure i like that this gives no indication from where to where things are reordered.

lib/stages/rfiBroadcast.cpp
```diff
@@ -100,6 +103,24 @@ rfiBroadcast::rfiBroadcast(Config& config, const std::string& unique_name,
...
+    reorder.resize(num_elements);
+
+    const Telescope& tel = Telescope::instance();
+
+    for (size_t cyl_idx = 0; cyl_idx < num_elements; ++cyl_idx) {
+        if (apply_reorder) {
+            station_id_t st_id =
+                tel.element_index_to_station_id(cyl_idx, ElementOrder::CHIMECylinder);
+            reorder.at(cyl_idx) =
+                tel.station_id_to_element_index(st_id, ElementOrder::CHIMEBeamformer);
```
rhaas80: right now there is no record (unforutnately) in metadata or framedesc about this order. So if one ever (again) passes voltages in cylinder order into the stages then this will not detect this issue.

lib/stages/bufferBadInputs.cpp
ljgray: I ran `clang-format` so I suppose it passes

lib/stages/bufferBadInputs.cpp
ljgray: The issue (I believe) is that this would required changing the order in which the flagging broker sends its updates

ljgray: Which might actually be nice to do down the road

lib/stages/rfiBroadcast.cpp
ljgray: Sure - this option could also just be removed entirely for now

lib/stages/rfiBroadcast.cpp
ljgray: Yeah, that's something that I was wondering about here - it relies entirely on an assumption about the order of the input data

lib/stages/bufferBadInputs.cpp
rhaas80: right now it *is* sending the station (channel) id though, isn't it? Cylinder order index being the same value as station (channel) id. 

Ie I am proposing that right now one can already use the value send by the broker as the station id. That is you can just record the `station id` -> `beamformer index` ordering instead of the `cylinder index` -> `beamformer index` as is done in the pull request (which will actually be the same mapping since `cylinder index` == `station id` [`channel id`]).

rhaas80 (approved):
no actual bugs present

lib/stages/bufferBadInputs.cpp
ljgray: Ah, my bad, I was confused station id with correlator ordering

PR #1626, opened 2026-08-22
Optional fatal on EOP requests outside the table range

jbmertens:
Also guard the "earlier than in table" report in get_EOP_at_UT1() with the same bounds check the instrument-time version uses, so a request for exactly the first table entry is not reported as out of range.


andrerenard (approved):
LGTM!

Issue #184, opened 2018-09-06
Sporadically failing unit tests


nritsche:
See comments for logs of the failing tests.
- [x] failing: `test_vistestpattern.py::test_noise` edit: fixed in #212 
- [ ] failing: `test_accumulate.py::test_gaussian` @jrs65 
- [x] timing out: `test_baseband_readout.py::test_io_errors_and_max_samples` edit: seems to be a frequent victim of concurrent unit testing, fixed in #201 
- [ ] various tests in `test_baseband_readout.py` - they seem to miss early REST commands (introduced by cb13378) @cubranic 
```
test_baseband_readout.py::test_io_errors_and_max_samples Sending interrupt signal to process

sh: line 1: 17366 Terminated              JENKINS_SERVER_COOKIE=$jsc '/var/lib/jenkins/workspace/n_test_receiveflags_speedup-TPWOFC43TVFIPCCKN6PXPGY4FEPTMLPAITPKL6YK7HSVCJ3L6F7Q@2@tmp/durable-70cc0c5a/script.sh' > '/var/lib/jenkins/workspace/n_test_receiveflags_speedup-TPWOFC43TVFIPCCKN6PXPGY4FEPTMLPAITPKL6YK7HSVCJ3L6F7Q@2@tmp/durable-70cc0c5a/jenkins-log.txt' 2>&1

script returned exit code 143
```

nritsche:
Replaced by #673

PR #1631, opened 2026-08-26
test_eigenStages: Pin the Blaze RNG seed so the iterative tests are reproducible

eschnett:
## What

Pin Blaze's global RNG seed in `tests/boost/test_eigenStages.cpp` so the two iterative eigen test cases stop failing at random in CI.

## Why

`eigen_masked_subspace` (`lib/utils/LinearAlgebra.hpp`) initialises its starting subspace with `blaze::rand`:

```cpp
// Initialise (randomly the vector array)
for (unsigned int i = 0; i < n; i++)
    for (unsigned int j = 0; j < k; j++)
        V(i, j) = blaze::rand<MT>();
```

Blaze seeds its global generator from `std::time(0)` unless told otherwise, and kotekan never calls `blaze::setSeed`. So every run started the subspace iteration from a different point and the `erms` values these tests bound moved with it.

There is no warm start to damp this out: `_krylov`/`_subspace` are the algorithm's `p`/`q` parameters, not carried state, so every frame re-randomises from scratch.

## Evidence

Measured on cx67, running the **unmodified** `test_eigenStages` binary 30 times with no recompile between runs:

| | result |
|---|---|
| unmodified `develop` | 28 pass / **2 fail** (~7%) |
| with the seed pinned | **30 pass / 0 fail** |

Both baseline failures were in `eigenVisIter_vis_buffers` or `eigenN2Iter_iterative` — the only two cases that use the randomized iterative solver, and the only two carrying the loose `2e-2` limits that exist because of it. The observed failure is a marginal overshoot, e.g. `0.02127 >= 0.02`.

This is the failure that has been showing up on unrelated PRs and reading as a numerical regression. Related history: #1034, #184.

## Notes for the reviewer

Seeding per pipeline rather than once globally keeps each test independent of how many random numbers the tests before it happened to draw.

Two things this deliberately does **not** fix, recorded in comments rather than papered over:

1. **`eigenN2Iter_concurrent_pipelines` stays nondeterministic.** Blaze's generator is a single shared static with no mutex and no `thread_local` (checked against Blaze's `util/Random.h`), so the two stage threads race on it regardless of the seed. That race is not confined to the test — any two Eigen stages in one process hit it.

2. **The iterative solver never converges in these configurations.** I swept `max_iterations` over 5, 10, 15, 20, 30, 50, 80, 120, 200 and `erms` is negative at *every* setting, i.e. the stage always reports `-eps_eval` rather than a residual. `eps_eval` divides through the near-zero second eigenvalue, so `tol_eval = 1e-6` looks unreachable by construction.

   This matters for how the test reads: `rms_limit` is bounding a convergence metric, not a residual, so a value near the limit is *not* "almost converged". The comment on the check now says so.

   My first draft of this PR raised `max_iterations` and asserted convergence — the sweep showed that would have changed nothing and made the assertion fail 100% of the time, so both were dropped. The seed is the actual fix.

Both of those are worth separate issues for whoever owns the eigen code; I did not want to fold them into a test-flakiness change.

## Testing

- 30x loop of `test_eigenStages` with the seed: 30/30 pass.
- Full boost suite on cx67: 32 passed, 1 failed — the failure is `test_buffer`, which is not on `develop` (untracked local WIP on the machine I built on) and is untouched by this change.
- `clang-format-18` clean.

### DIFF
--- a/tests/boost/test_eigenStages.cpp
+++ b/tests/boost/test_eigenStages.cpp
@@ -22,6 +22,7 @@
 #include <algorithm>
 #include <array>
 #include <atomic>
+#include <blaze/Blaze.h> // for setSeed
 #include <boost/test/included/unit_test.hpp>
 #include <chrono>
 #include <cmath>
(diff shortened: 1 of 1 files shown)


jbmertens:
LGTM.

jbmertens (approved):

eschnett:
@jbmertens Note in particular the comment "The iterative solver never converges in these configurations". I don't know if that was intended. If not, then letting it converge would also be a good idea...

jbmertens:
> @jbmertens Note in particular the comment "The iterative solver never converges in these configurations". I don't know if that was intended. If not, then letting it converge would also be a good idea...

Without knowing the original intent of the test, it asks for 2 eigenvalues of a rank-1 matrix, so the non-convergence isn't surprising. We could switch to asking for convergence of just 1 eigenvalue.

eschnett:
Superseded by https://github.com/kotekan/kotekan/pull/1633.

PR #1632, opened 2026-08-26
eigen_masked_subspace: Draw the starting subspace from a per-thread generator

eschnett:
## The race

`eigen_masked_subspace` (lib/utils/LinearAlgebra.hpp) initialised its starting subspace with `blaze::rand`. That dispatches to `Rand<T>::generate()`, which does `dist( Random<>::rng_ )`, and in `blaze/util/Random.h`:

```cpp
template< typename RNG > RNG Random<RNG>::rng_ ( defaultSeed() );
```

One `static std::mt19937` for the whole process, with neither a mutex nor `thread_local` storage. Every draw advances the twister's internal state array, so two threads drawing at once is a data race on that state, not just an interleaving of draws.

`EigenVisIter` and `EigenN2Iter` each call `eigen_masked_subspace` from their own stage thread, so **any process running two Eigen stages hits this**. The boost test `eigenN2Iter_concurrent_pipelines` runs two `EigenN2Iter` pipelines concurrently and exercises it directly, but the race is in the production path, not a test artifact.

`blaze::rand` is used in exactly one place in kotekan, so this is the only site affected.

## The fix

Draw from a `thread_local` generator instead, so each stage has its own and never touches another's. Blaze's generator is left alone.

The draws are otherwise unchanged — uniform over [0, 1), the range `blaze::rand` uses, for both parts of a complex value. The one difference: the real and imaginary parts are now drawn in a defined order. `blaze::rand<complex<T>>` builds the value as `complex<T>( tmp.generate(), tmp.generate() )`, whose argument evaluation order the compiler chooses.

The per-thread generator is seeded from a fixed constant rather than the clock, so a stage draws the same sequence of starting subspaces on every run and its output is reproducible. Each call still advances the sequence, so successive frames start from different subspaces as before. `eigen_masked_subspace` also gained an optional generator argument for callers that want to control the starting subspace explicitly.

## Tests

`tests/boost/test_linearAlgebra.cpp` is new. It checks that the generator really is per-thread, that a thread which has not touched it starts from the fixed seed, and that eight concurrent decompositions of the same matrix return bit-identical eigenpairs — both through the per-thread generator as the stages use it, and through an explicitly supplied one.

## Verification (cx67)

- Replacing the `thread_local` with a plain `static`, which is what Blaze does, fails four of the six new checks: the threads take each other's numbers and the eigenpairs no longer match. The test catches the bug it exists for.
- With the fix, `test_linearAlgebra` passes 30/30 runs and the full boost suite passes.
- `test_eigenStages` passes 30/30 with no seed pinning.

## Interaction with #eschnett/eigenstages-flaky-seed

That branch pins the Blaze seed inside `test_eigenStages` to stop the ~7% flakiness (2 failures in 30 runs) and records in a `NOTE` that it deliberately does not address this race. With the starting subspace now reproducible on its own, an unseeded `test_eigenStages` passes 30/30 here. If both land, that branch's `seed_blaze()` no longer has any effect on the Eigen stages and the `NOTE` beside it about the shared generator is stale — worth deleting at merge time.

### DIFF
--- a/lib/utils/LinearAlgebra.hpp
+++ b/lib/utils/LinearAlgebra.hpp
@@ -8,6 +8,10 @@
 #include "visUtil.hpp"
 
 #include <blaze/Blaze.h>
+#include <complex>     // for complex
+#include <cstdint>     // for uint32_t
+#include <random>      // for mt19937, uniform_real_distribution
+#include <type_traits> // for is_same_v
 
 // Type defs for simplicity
 // Map complex types to their real equivalent
@@ -220,6 +224,51 @@ struct EigConvergenceStats {
     double rms = 0.0;
 };
 
+/// Seed given to each thread's copy of `eigen_subspace_rng`.
+constexpr uint32_t eigen_subspace_seed = 0x9e3779b9;
+
+/**
+ * @brief The random number generator used to initialise the subspace iteration.
+ *
+ * Every thread gets its own generator, seeded identically. Blaze's `rand` cannot be
+ * used for this: it draws from a single static generator (`Random<RNG>::rng_` in
+ * blaze/util/Random.h) that is neither `thread_local` nor mutex protected, so two
+ * Eigen stages drawing their starting subspaces at the same time would be a data race
+ * on its state.
+ *
+ * The fixed seed also makes the stages reproducible: a given stage draws the same
+ * sequence of starting subspaces on every run.
+ *
+ * @return  The calling thread's generator.
+ **/
+inline std::mt19937& eigen_subspace_rng() {
+    static thread_local std::mt19937 rng(eigen_subspace_seed);
+    return rng;
+}
+
+/**
+ * @brief Draw one random matrix element uniformly from [0, 1).
+ *
+ * That is the range `blaze::rand` uses, for both parts of a complex value.
+ *
+ * @param  rng  The generator to draw from.
+ *
+ * @return      The random element.
+ **/
+template<typename MT>
+MT rand_subspace_element(std::mt19937& rng) {
+    std::uniform_real_distribution<real_t<MT>> dist(0.0, 1.0);
+    if constexpr (std::is_same_v<MT, real_t<MT>>) {
+        return dist(rng);
+    } else {
+        // Draw in a fixed order. `blaze::rand<complex<T>>` passes two draws as
+        // function arguments, whose evaluation order the compiler chooses.
+        const real_t<MT> re = dist(rng);
+        const real_t<MT> im = dist(rng);
+        return MT(re, im);
+    }
+}
+
 /**
  * @brief Find a low rank decomposition of a masked matrix.
  *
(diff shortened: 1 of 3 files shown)


eschnett:
@jbmertens This corrects how random numbers are drawn when using Blaze. It is independent of the other Blaze PR.

lib/utils/LinearAlgebra.hpp
```diff
@@ -220,6 +224,51 @@ struct EigConvergenceStats {
...
+ * Eigen stages drawing their starting subspaces at the same time would be a data race
+ * on its state.
+ *
+ * The fixed seed also makes the stages reproducible: a given stage draws the same
+ * sequence of starting subspaces on every run.
+ *
+ * @return  The calling thread's generator.
+ **/
+inline std::mt19937& eigen_subspace_rng() {
+    static thread_local std::mt19937 rng(eigen_subspace_seed);
```
jbmertens, line 245: So every thread (stage) gets the same seed? Do we want that?

lib/utils/LinearAlgebra.hpp
```diff
@@ -220,6 +224,51 @@ struct EigConvergenceStats {
     double rms = 0.0;
 };
 
+/// Seed given to each thread's copy of `eigen_subspace_rng`.
+constexpr uint32_t eigen_subspace_seed = 0x9e3779b9;
```
jbmertens, line 228: 🥇

lib/utils/LinearAlgebra.hpp
```diff
@@ -245,7 +296,7 @@ std::pair<eig_t<MT>, EigConvergenceStats>
 eigen_masked_subspace(const DynamicHermitian<MT>& A,
                       const DynamicHermitian<float>& W, // Should this be symmetric
                       size_t k, float tol_eval, float tol_evec, size_t maxiter, size_t k_conv = 0,
-                      size_t p = 2, size_t q = 3) {
+                      size_t p = 2, size_t q = 3, std::mt19937& rng = eigen_subspace_rng()) {
```
jbmertens, line 299: Why not seed directly (and configurably) from the stage?

jbmertens (approved):
A few comments - although I don't expect this change will reduce test flakiness (2 eigenvalues of a rank 1 matrix).

But this change fixes a (potential) issue in 2-numa production configs, so CHIME will want to include it @rhaas80 @ljgray .

lib/utils/LinearAlgebra.hpp
rhaas80: every kotena instance also uses the same seed. Apparently harmless :-) Our matrices also add randomness.

eschnett: I don't think that's important. The solution of the eigensolve should be independent of the random initial guess. Having this independent of the thread id is probably good in the long run.

lib/utils/LinearAlgebra.hpp
eschnett: I don't think the seed is that important. I would keep the code simple and not tell the stages about how the RNG that seeds the initial guess is initialized. The stages just want eigenvalues/vectors.

PR #1634, opened 2026-08-28
Loosen lock on copy frames between buffers 

jbmertens:
`Buffer::safe_swap_frame()` would previously (and unnecessarily) lock the entire buffer. Now it takes locks only to read frame pointers. (Any stage that invokes this call must be a sole producer or consumer, so neither frame should be modified by another stage.)

Adds test_buffer_swap, which had no coverage before: a single consumer should get the frames exchanged, a second consumer gets a copy with the source left intact, and copies interleave with both buffers being used from another thread.


andrerenard (approved):
This looks good to me.  I agree we don't need this lock, since the buffer memory access model should guard the actual frames.

PR #1635, opened 2026-08-28
Add source ID (CRS board) remapping to DPDK for CHORD Pathfinder.

andrerenard:
Current Pathfinder data is indexed "incorrectly".  The data is supposed to be indexed as [pol][dish], but currently the dish to CRS board mapping mixes up pol and dish.   This change separates dish and polarization for all systems down stream of the packet capture logic.


jbmertens (approved):
Mostly LGTM - the 
```
crs_board_remap:
    - [0, 2, 4, 6, 8, 10, 12, 14, 1, 3, 5, 7, 9, 11, 13, 15]
```
won't parse correctly though.

The chord_pathfinder_recv.j2 will also need its dish table updated.

jbmertens:
(I note that this will break some downstream stages for chord, but that's fine, will fix shortly.)

jbmertens:
Also another note - reordering the crs board inputs is "free", but a general reorder using `_mm512_permutex2var_epi8` in TransposeBasebandArray is only about 0.3% more expensive.

andrerenard:
> Also another note - reordering the crs board inputs is "free", but a general reorder using `_mm512_permutex2var_epi8` in TransposeBasebandArray is only about 0.3% more expensive.

That's better than I'd have expected, but it's less about the operation cost, and more I'd like to someday remove that transpose stage entirely.  So I don't want to bake us into a situation where we need a full transpose to stay consistent.

PR #1636, opened 2026-08-28
Config: report a malformed value rather than silently defaulting

jbmertens:
get_default() caught every std::runtime_error from get<T>() and returned the default, so a value of the wrong type was indistinguishable from one that was never set. A config could then declare a setting, have it quietly ignored, and run with the default while the rest of the config assumed otherwise -- with no warning anywhere.

### DIFF
--- a/lib/core/Config.hpp
+++ b/lib/core/Config.hpp
@@ -156,8 +156,11 @@ public:
     /**
      * @brief Get a config value or return the default value.
      *
-     * Same as get_int, but if it cannot find the value
-     * (or if it has the wrong type), it returns `default_value`.
+     * A value that is not set anywhere on the path returns @c default_value silently.
+     * A value that is set but cannot be read as @c T also returns @c default_value,
+     * but logs an error, since substituting the default for a malformed entry would
+     * otherwise leave the pipeline running an unintended configuration.
+     *
      * @param base_path     Path to the value in the config.
      * @param name          Name of the value.
      * @param default_value The default value.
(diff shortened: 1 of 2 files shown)


andrerenard (approved):
We should test this on a few "live" configs, but this is a great to have.

jbmertens:
~Switched to FATAL_ERROR as well.~

Along with the config usage reporting (catches stale config params, I'd definitely turn that on in the production CHIME configs), I hope this helps...

jbmertens:
@rhaas80 - just noting it may be useful to include this change in the chime branch; this caught a silent error in the CHORD configs.

lib/core/Config.hpp
```diff
@@ -156,8 +156,11 @@ class Config {
     /**
      * @brief Get a config value or return the default value.
      *
-     * Same as get_int, but if it cannot find the value
-     * (or if it has the wrong type), it returns `default_value`.
+     * A value that is not set anywhere on the path returns @c default_value silently.
+     * A value that is set but cannot be read as @c T also returns @c default_value,
+     * but logs an error, since substituting the default for a malformed entry would
+     * otherwise leave the pipeline running an unintended configuration.
```
rhaas80, line 162: I would lobby to abort in this case.

rhaas80:
oh yikes, this does seem like good error to report. Silently using the default if the given value fails to convert is just wrong.

jbmertens:
I changed this to a fatal error in the chord branch, although that didn't make it in this PR.

PR #1637, opened 2026-09-02
hdf5FileRead: add read_single_file mode to replay single-file dumps

eschnett:
hdf5FileWrite can produce output files with a single data stream, i.e. not producing one dataset per frame.

This PR adds the converse, letting hdf5FileRead read such a data stream.

### DIFF
--- a/lib/stages/hdf5FileRead.cpp
+++ b/lib/stages/hdf5FileRead.cpp
@@ -22,6 +22,8 @@
 #include <highfive/H5DataSpace.hpp>              // for DataSpace, DataSpace::getDimensions
 #include <highfive/H5Exception.hpp>              // for FileException
 #include <highfive/H5File.hpp>                   // for File, File::File, NodeTraits::getDataSet
+#include <highfive/H5Selection.hpp>              // for Selection, SliceTraits::select
+#include <highfive/bits/H5Selection_misc.hpp>    // for Selection::getSpace
 #include <highfive/bits/H5Slice_traits_misc.hpp> // for SliceTraits::read_raw
 #include <iomanip>                               // for operator<<, setfill, setw
 #include <kotekanLogging.hpp>                    // for DEBUG, FATAL_ERROR, ERROR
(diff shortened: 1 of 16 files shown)


rhaas80:
oh, very nice. I will take a look.

lib/stages/hdf5FileReadSingleFile.cpp
```diff
@@ -37,6 +38,46 @@
 using namespace hdf5;
 using namespace HighFive;
 
+/**
+ * @class hdf5FileReadSingleFile
```
jbmertens, line 42: The stage name `hdf5FileReadSingleFile` suggests something more general -- maybe name it hdf5FileReadVoltageFile?

lib/stages/hdf5FileRead.cpp
```diff
@@ -162,7 +226,148 @@ class hdf5FileRead : public kotekan::Stage {
...
     }
 
+    /**
+     * @brief The common part of all file names of this stage:
+     *        "<input_dir>/[<hostname>_][x<rank:04d>_]<file_name>"
+     *
+     * The caller appends the suffix, i.e. ".<frame:08d>.h5" for per-frame files and ".h5"
+     * for a single file.
+     */
+    std::string file_path_prefix() const {
```
jbmertens: This snippet also exists in the hdf5FileWrite stage, so could be abstracted somewhere common to the two files.

lib/stages/hdf5FileRead.cpp
```diff
@@ -328,6 +457,84 @@ class hdf5FileRead : public kotekan::Stage {
...
+    /**
+     * @brief Read all frames from a single file, splitting the dataset along axis 0
+     *
+     * The per-frame axis-0 extent is taken from the buffer's declared frame descriptor
+     * because the file only stores the concatenation of all frames. A missing or
+     * unreadable file is always fatal.
+     */
+    void read_single_file_frames() {
+        auto& read_time_metric = kotekan::prometheus::Metrics::instance().add_gauge(
+            "kotekan_hdf5fileread_read_time_seconds", unique_name);
```
jbmertens: In most stages, I think this is initialized in the constructor, which might be a bit cleaner here.

lib/stages/hdf5FileRead.cpp
```diff
@@ -162,7 +226,148 @@ class hdf5FileRead : public kotekan::Stage {
...
+     *                     mode
+     * @param frame_index  The index of the frame within the dataset. It advances
+     *                     @c fpga_seq_num in single-file mode, where only the value of the
+     *                     first frame is stored on disk; it must be 0 for per-frame files.
+     */
+    void set_metadata(const DataSet& dataset, const std::shared_ptr<chordMetadata>& meta,
+                      const std::vector<std::size_t>& frame_dims,
+                      const std::int64_t frame_index) const {
+        meta->set_name(dataset.getAttribute("name").read<std::string>());
+        meta->type = kotekan::string_to_type(dataset.getAttribute("type").read<std::string>());
```
jbmertens, line 339: This should be validated against the frame data type, right?

lib/stages/hdf5FileRead.cpp
```diff
@@ -162,7 +226,148 @@ class hdf5FileRead : public kotekan::Stage {
...
+     *                     with axis 0 replaced by the buffer's frame extent in single-file
+     *                     mode
+     * @param frame_index  The index of the frame within the dataset. It advances
+     *                     @c fpga_seq_num in single-file mode, where only the value of the
+     *                     first frame is stored on disk; it must be 0 for per-frame files.
+     */
+    void set_metadata(const DataSet& dataset, const std::shared_ptr<chordMetadata>& meta,
+                      const std::vector<std::size_t>& frame_dims,
+                      const std::int64_t frame_index) const {
+        meta->set_name(dataset.getAttribute("name").read<std::string>());
```
jbmertens, line 338: How are errors handled when attributes are missing?

jbmertens (changes requested):
Nice stage to have. Some comments below from Claude 🤖, plus some inline ones from me.

----------

- The reconstructed fpga_seq_num is silently wrong whenever the stream has gaps. hdf5FileRead.cpp:305 computes first + i * extents[0] * tds, which assumes every frame the writer appended was contiguous in time. The writer breaks that assumption itself: write_x_frames/per_y_frames decimation is applied before the single-file append (hdf5FileWrite.cpp:585), so a decimated single file replays with timestamps that are wrong for every frame after the first skipped one. Upstream frame loss does the same.
- Axis 0 is assumed to be the time axis, which goes unchecked: the writer concatenates along axis 0 regardless of what it is, and the reader now bakes the same assumption into the timestamp formula. CHIME-ordered data is frequency-first (that is why hdf5FileReadSingleFile has transpose_frequency_and_time), so such a stream would produce a nonsense file and nonsense sequence numbers with no error.
- The per-frame path still terminates the process on non-file HDF5 errors. The second commit fixes exactly this for hdf5FileReadSingleFile, but hdf5FileRead.cpp:450 still catches only FileException.
- Zero frames is not an error. If the dataset is shorter than one frame, or empty because the writer died before its first flush, hdf5FileRead.cpp:492 yields zero frames, logs at INFO, and the stage returns. The pipeline then idles until an external timeout. A missing per-frame file at index 0 is fatal, and this case should be too.
- do_once misreports. After the single-frame idle at line 527 the stage falls through to line 533 and claims it read all frames.
- The storage type vs. declared type is not validated. If a file's attribute and storage type disagree, the read overruns the frame.
- Silent formula change in hdf5FileReadSingleFile. Line 295 now multiplies by time_downsampling_fpga while line 254 requires the time axis scaling to be 1. When the file is consistent this is a no-op. When it is not, it is unclear which value is right. Either drop the factor or make a non-unit downsampling fatal.
- End-of-input logs at ERROR. The documented normal stop condition for per-frame mode still prints an ERROR line when the next file is missing. That will confuse anyone reading logs after a successful replay and should be WARN or INFO.

lib/stages/hdf5FileReadSingleFile.cpp
```diff
@@ -37,6 +38,46 @@
...
+ * @brief Replay CHIME/F-engine voltage data from a single HDF5 file.
+ *
+ * This is a special-purpose reader for baseband voltages: it reads one dataset holding all
+ * time samples of a recording (rank 4, `[T, F, ...]`, element type
+ * @c int4x2_swapped_withoffset), selects a subset of the frequency channels, and hands the
+ * data out one frame of @c num_times time samples at a time. It can transpose time and
+ * frequency and combine the dish and polarization axes into a single element axis, as the
+ * CHIME F-engine pipeline expects. Anything that does not match these expectations is
```
rhaas80: X-engine

lib/stages/hdf5FileWrite.cpp
```diff
@@ -70,11 +70,15 @@ using namespace HighFive;
  * @conf prefix_host_rank  Bool. Prepend rank to output file names. Default: false.
  * @conf frequency_pool_rank    Int. This stage's rank in the frequency pool.
  * @conf frequency_pool_size    Int. Number of stages in the frequency pool.
- * @conf max_frames  Int. Stop writing after this many frames, Default 0 = unlimited
- *       frames.
+ * @conf max_frames  Int. Default: -1 = unlimited; N >= 0 stops after N frames and
```
rhaas80: change in behaviour, 0 used to be unlimited. Is there an actual usage case to write 0 frames?

lib/stages/hdf5Files.cpp
```diff
@@ -47,7 +47,7 @@ HighFive::DataType chord2hdf5(const kotekan::DataType type) {
             return HighFive::AtomicType<double>();
         default:
             throw std::runtime_error(
-                fmt::format("chord2hdf5 given unknown DataType value: {;d} ", (int64_t)type));
+                fmt::format("chord2hdf5 given unknown DataType value: {:d} ", (int64_t)type));
```
rhaas80, line 54: chord?

lib/testing/testDataGenFloat.cpp
```diff
@@ -78,6 +78,12 @@ testDataGenFloat::testDataGenFloat(Config& config, const std::string& unique_nam
         throw std::invalid_argument("testDataGen: 'array_shape' and 'dim_name' config "
                                     "settings must be the same length!");
     }
+    _dim_scaling = config.get_default<std::vector<std::ptrdiff_t>>(
+        unique_name, "dim_scaling", std::vector<std::ptrdiff_t>(_array_shape.size(), 1));
+    if (_array_shape.size() != _dim_scaling.size()) {
+        throw std::invalid_argument("testDataGenFloat: 'array_shape' and 'dim_scaling' config "
+                                    "settings must be the same length!");
+    }
```
rhaas80, line 86: not really file reader related but sure, that is good to have as well

rhaas80 (approved):
plus all that Jim said

eschnett:
I think I attended to all of the comments except one – I did not rename the "single file reader" stage. We can do that later, it's probably a good idea.

The writer and reader are now restricted to handling data which have time as outermost index. This requires changing input files which output other data to use the other (not-single-file) writer/reader, which is probably a very good idea, but it touches several files.

rhaas80:
I am assuming that time as the outermost index only applies when writing single files? otherwise this would make it impossible to eg write out eg the `host_frb2_weights_buffer` which has `["Fbar", "R", "beamQ", "beamP"]` (and no time dependence).

lib/stages/hdf5FileWrite.cpp
```diff
@@ -123,7 +146,13 @@ class hdf5FileWrite : public kotekan::Stage {
               }),
         buffer(get_buffer("in_buf")) {
 
-        if (max_frames >= 0)
+        if (create_single_file && write_x_frames >= 0 && per_y_frames > 0)
+            FATAL_ERROR("hdf5FileWrite {:s}: write_x_frames/per_y_frames (frame decimation) cannot "
```
jbmertens, line 150: In production I'm not sure we want this behavior - missing frames might just be due to a temporary network issue that resolves? Maybe for @rhaas80 or @andrerenard to comment on.

jbmertens (commented):
One behavior note that people may have an opinion on below, otherwise LGTM.

lib/stages/hdf5FileWrite.cpp
eschnett: If frames can be outright missing then our whole ringbuffer logic needs to be generalized.

jbmertens: The assumption is probably fine for the x-engine engine nodes. But we may want this stage on the receive node, or chime may use it there; upstream x-engine engine nodes might go off randomly. But it can be modified then if need be.

jbmertens (approved):

jbmertens:
There are 2 approvals on this, so I think it can be merged.
The last note about the stage dying on missing frames we can deal with if the issue comes up.

PR #1638, opened 2026-09-03
lib/cuda: build metadata fully, then publish it (fix the publish-then-mutate class)

kvand:
Fixes a metadata modify-in-place hazard that surfaced during GNSS testing, and the other instances of it in `lib/cuda`.

**The pattern.** A stage publishes an object into a ring with `set_metadata()`, reads it back with `get_metadata()` — which returns the LIVE slot-0 object — and then writes derived fields into it. Any consumer polling that ring in the interval sees a half-corrected object. Every instance takes the same fix: build the object, write the derived fields, publish it last.

**What it cost.** `cudaCopyFromRingbuffer::execute` reads `signal_buffer->get_metadata(0)` on every frame and derives

    out_seq = anchor + time_downsampling_fpga * (input_cursor / sample_bytes)

`input_cursor` is the absolute byte count since startup, so a reader catching `time_downsampling_fpga` at 256 instead of 1024 (CHORD's `rfi_downsampling_factor` is 256) does not lose one frame's worth — it quarters the whole accumulated offset, emitting a sequence number `0.75 * uptime` behind. `N2Accumulate` then FATALs on the correlation-vs-RFICounts mismatch, and with `exit_on_worker_failure` the node exits. **Nine CHORD nodes died this way on 2026-09-02/03.**

Once the FATAL was made survivable the signature was measured directly: the gap is `0.75 * node uptime` on three separate nodes, the ratio between successive events on one node is exactly 0.25, and solving for the anchor from each event returns a single per-node constant with zero remainder — the correct anchor with a quartered offset, which is this field and nothing else.

**The sites, after review.** Only ring-backed buffers race: `NDArrayRingBuffer::set_metadata` fills the live slot-0 object, while `NDArrayBuffer::set_metadata` takes a fresh object from the pool on every call (`create_gpu_memory_array_metadata` → `request_metadata_object`) and nothing downstream sees it before `finalize_frame`.

| Site | Buffer | Field written after publish |
|---|---|---|
| `cudaRFISKtilde` | ring | `time_downsampling_fpga` — the measured failure above |
| `cudaRFIS012` | ring | `time_downsampling_fpga`; one hop *upstream*, so the error a reader catches is a factor of `rfi_downsampling_factor` (256), not 4 |
| `cudaRFIS012bar` | ring | `time_downsampling_fpga` |
| `cudaHFB1Accumulate` | ring | `time_downsampling_fpga` |
| `cudaPLMaskExpander` | ring | `time_downsampling_fpga` — **added in review** |
| `cudaPLMaskUpchannelizer` | ring | `time_downsampling_fpga` — **added in review** |
| `cudaFRBBeamReformer` | non-ring | `fpga_seq_num`, written every frame into an object published **once** behind `did_set_metadata` — **fixed in review**, see below |

**Reverted in review:** `cudaCorrelator` and `cudaPL1bitCorrelator`. Both write to `NDArrayBuffer`s, which hand out a fresh object per call, so there is no window there. The comments those changes added described a hazard that does not exist, and are gone with them.

**`cudaFRBBeamReformer`,** which the first round deliberately left alone, is the one non-ring site that does race: `did_set_metadata` meant `set_metadata` ran once, so every later frame's per-frame `fpga_seq_num` write landed in an object that frames already downstream referenced. Calling `set_metadata` per frame gives each frame its own object; the `deepCopy` inside it carries `time_downsampling_fpga`, `coarse_freq`, `freq_upchan_factor` and `freq_upchan_index`, so the flag and the four explicit setters are redundant and removed (net −12 lines).

No behaviour change on the happy path: the published values are identical, only the window in which a concurrent reader can observe an intermediate one goes away.

**Still deliberately out of scope**, as shared infrastructure rather than a caller: `NDArrayRingBuffer::set_metadata` fills slot 0 in place (`allocate_new_metadata_object(0)` is a no-op after frame 0) and writes `name`/`type`/`dims`/`stride` without holding `chordMetadata::lock`. Those values repeat after frame 0 so nothing fails today, and callers patch after the call returns, so the caller fixes stand on their own. Publishing metadata inside `finish_write` alongside the data would subsume them — separate PR.

The consumer-side hardening is #1642.

**Note on CI:** this PR targets `chord`, and `.github/workflows/main.yaml` only triggers on `develop`, so none of the build/test jobs run against it. Verified locally instead: gcc 13 with `-DWERROR=ON`, full CUDA build, and both fleet configs pass `kotekan --check-config`. The review round was re-verified by compiling all nine touched files against the branch headers with the project's own flags; clang-format clean.

### DIFF
--- a/lib/cuda/cudaCorrelator.cpp
+++ b/lib/cuda/cudaCorrelator.cpp
@@ -121,11 +121,8 @@ cudaEvent_t cudaCorrelator::execute(cudaPipelineState&, const std::vector<cudaEv
 
     voltage.check_metadata();
     rfi_RFImask.check_metadata();
-    n2k_correlation.set_metadata(voltage.get_metadata());
-
     const std::shared_ptr<const chordMetadata> voltage_meta = voltage.get_metadata();
     const std::shared_ptr<const chordMetadata> rfi_meta = rfi_RFImask.get_metadata();
-    const std::shared_ptr<chordMetadata> n2k_corr_meta = n2k_correlation.get_metadata();
 
     // Ensure consistency:
     assert(voltage_meta->get_fpga_seq_num()
@@ -133,13 +130,22 @@ cudaEvent_t cudaCorrelator::execute(cudaPipelineState&, const std::vector<cudaEv
            == rfi_meta->get_fpga_seq_num()
                   + rfi_RFImask.get_read_valid().begin() * rfi_meta->get_time_downsampling_fpga());
 
-    // The input ringbuffer metadata do not contain time-dependent metadata,
-    // so we must reconstruct it here. (fpga_seq_num)
-    n2k_corr_meta->set_fpga_seq_num(voltage_meta->get_fpga_seq_num()
-                                    + voltage.get_read_valid().begin()
-                                          * voltage_meta->get_time_downsampling_fpga());
-    n2k_corr_meta->set_time_downsampling_fpga(_sub_integration_ntime
+    // The input ringbuffer metadata do not contain time-dependent metadata, so we must
+    // reconstruct it here (fpga_seq_num). Reconstruct it BEFORE publishing: this used to
+    // publish voltage's metadata and then write the reconstructed fields into the object it
+    // had just handed to the ring, where a consumer reading slot 0 could see the anchor
+    // without the read_valid().begin() term. Unlike the RFImask case that value differs on
+    // every frame, so the window is wider. See cudaRFISKtilde::execute.
+    {
+        auto n2k_corr_meta = std::make_shared<chordMetadata>();
+        n2k_corr_meta->deepCopy(voltage_meta);
+        n2k_corr_meta->set_fpga_seq_num(voltage_meta->get_fpga_seq_num()
+                                        + voltage.get_read_valid().begin()
                                               * voltage_meta->get_time_downsampling_fpga());
+        n2k_corr_meta->set_time_downsampling_fpga(_sub_integration_ntime
+                                                  * voltage_meta->get_time_downsampling_fpga());
+        n2k_correlation.set_metadata(n2k_corr_meta);
+    }
 
     // Set poison for debug checks.
     if (_poison_buffers)
--- a/lib/cuda/cudaCorrelator.cpp
+++ b/lib/cuda/cudaCorrelator.cpp
@@ -121,8 +121,11 @@ cudaEvent_t cudaCorrelator::execute(cudaPipelineState&, const std::vector<cudaEv
 
     voltage.check_metadata();
     rfi_RFImask.check_metadata();
+    n2k_correlation.set_metadata(voltage.get_metadata());
+
     const std::shared_ptr<const chordMetadata> voltage_meta = voltage.get_metadata();
     const std::shared_ptr<const chordMetadata> rfi_meta = rfi_RFImask.get_metadata();
+    const std::shared_ptr<chordMetadata> n2k_corr_meta = n2k_correlation.get_metadata();
 
     // Ensure consistency:
     assert(voltage_meta->get_fpga_seq_num()
@@ -130,22 +133,13 @@ cudaEvent_t cudaCorrelator::execute(cudaPipelineState&, const std::vector<cudaEv
            == rfi_meta->get_fpga_seq_num()
                   + rfi_RFImask.get_read_valid().begin() * rfi_meta->get_time_downsampling_fpga());
 
-    // The input ringbuffer metadata do not contain time-dependent metadata, so we must
-    // reconstruct it here (fpga_seq_num). Reconstruct it BEFORE publishing: this used to
-    // publish voltage's metadata and then write the reconstructed fields into the object it
-    // had just handed to the ring, where a consumer reading slot 0 could see the anchor
-    // without the read_valid().begin() term. Unlike the RFImask case that value differs on
-    // every frame, so the window is wider. See cudaRFISKtilde::execute.
-    {
-        auto n2k_corr_meta = std::make_shared<chordMetadata>();
-        n2k_corr_meta->deepCopy(voltage_meta);
-        n2k_corr_meta->set_fpga_seq_num(voltage_meta->get_fpga_seq_num()
-                                        + voltage.get_read_valid().begin()
+    // The input ringbuffer metadata do not contain time-dependent metadata,
+    // so we must reconstruct it here. (fpga_seq_num)
+    n2k_corr_meta->set_fpga_seq_num(voltage_meta->get_fpga_seq_num()
+                                    + voltage.get_read_valid().begin()
+                                          * voltage_meta->get_time_downsampling_fpga());
+    n2k_corr_meta->set_time_downsampling_fpga(_sub_integration_ntime
                                               * voltage_meta->get_time_downsampling_fpga());
-        n2k_corr_meta->set_time_downsampling_fpga(_sub_integration_ntime
-                                                  * voltage_meta->get_time_downsampling_fpga());
-        n2k_correlation.set_metadata(n2k_corr_meta);
-    }
 
     // Set poison for debug checks.
     if (_poison_buffers)
--- a/lib/cuda/cudaFRBBeamReformer.cpp
+++ b/lib/cuda/cudaFRBBeamReformer.cpp
@@ -192,12 +192,14 @@ cudaEvent_t cudaFRBBeamReformer::execute(cudaPipelineState& /*pipestate*/,
         // Set metadata
         const std::shared_ptr<const chordMetadata> frb1_beams_meta =
             frb1_beams_buffer.get_metadata();
-        frb2_beams_buffer.set_metadata(frb1_beams_meta);
-        const std::shared_ptr<chordMetadata> frb2_beams_meta = frb2_beams_buffer.get_metadata();
+        // Built before publishing, not patched afterwards; see cudaRFISKtilde::execute.
+        auto frb2_beams_meta = std::make_shared<chordMetadata>();
+        frb2_beams_meta->deepCopy(frb1_beams_meta);
         frb2_beams_meta->set_time_downsampling_fpga(frb1_beams_meta->get_time_downsampling_fpga());
         frb2_beams_meta->set_coarse_freq(frb1_beams_meta->get_coarse_freq());
         frb2_beams_meta->set_freq_upchan_factor(frb1_beams_meta->get_freq_upchan_factor());
         frb2_beams_meta->set_freq_upchan_index(frb1_beams_meta->get_freq_upchan_index());
+        frb2_beams_buffer.set_metadata(frb2_beams_meta);
     }
     frb2_beams_buffer.check_metadata();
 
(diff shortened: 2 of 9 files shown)


jbmertens:
@eschnett - I know you had wanted to remove time-dependent metadata from the GPU. Although, more generally we might want to keep that capability. Could you comment on this?

kvand:
On the time-dependent-metadata question: this patch doesn't depend on the answer either way. `time_downsampling_fpga` and its `(128*8)/rfi_downsampling_factor` correction were already here; all that changes is *when* the corrected value becomes visible. If the field is later removed from GPU metadata the whole block goes away with it, and nothing here stands in the way of that.

Two corrections to my original commit message, now fixed on the branch -- both were written against our downstream tree rather than `chord`:

1. It cited the hazard as "documented on `NDArrayRingBuffer::set_metadata`". That comment exists only in our tree, not here.
2. It said `set_metadata` is atomic. On `chord` it is not: it calls `allocate_new_metadata_object(0)`, takes `get_metadata()`, and `deepCopy`s into that live slot-0 object -- and since `GenericBuffer::allocate_new_metadata_object` only allocates when the slot is empty, after frame 0 it is a no-op, so every frame fills the published object in place.

The fix does still close the reported window here, but by a narrower mechanism than I claimed: `time_downsampling_fpga` lives behind `chordMetadata::lock`, so collapsing two locked writes into one leaves no interval in which a reader can observe the uncorrected value.

While rewording it I went looking for the same pattern elsewhere and found six more instances, so I have extended this PR to cover them (see the updated description). The one worth calling out is `cudaRFIS012.cpp:204`, which does publish-then-mutate on this very field one hop upstream of the RFImask fix -- a reader landing in that window sees the voltage stream's downsampling rather than the S012 value, so the error there is a factor of `rfi_downsampling_factor` (256 on CHORD) rather than the factor of 4 that took nine nodes down. `cudaCorrelator` and `cudaPL1bitCorrelator` do it to `fpga_seq_num`, which is reconstructed per frame and so has a wider window still.

Two related things I left alone deliberately, since they are yours to judge rather than mine to transform: `cudaFRBBeamReformer` writing `fpga_seq_num` into the published object on every frame, and `set_metadata` itself filling slot 0 in place (and writing name/type/dims/stride without holding the lock), which is the root of the whole class.

There is also a one-line hardening at the other end: `cudaCopyFromRingbuffer::execute` already takes a locked deep-copy snapshot and then goes back to the live object for every value it uses. Reading its own snapshot makes that consumer immune to this class in any producer, including ones added later. I have it ready as a separate PR if you would like it -- it seemed cleaner not to mix a consumer-side change into this one.

jbmertens:
🤖 feedback below.

1. The PR missed two ring-buffer sites used by CHORD pathfinder: `lib/cuda/cudaPLMaskExpander.cpp:171-177` and `lib/cuda/cudaPLMaskUpchannelizer.cpp:228-236` both publish then modify `time_downsampling_fpga`.
2. The three non-ring sites do not race. `NDArrayBuffer::set_metadata` allocates a new object per call, and no other thread sees it before `finalize_frame`. The comments in `cudaCorrelator` and `cudaPL1bitCorrelator` describe a window that does not exist.
3. The comments keep a verbose record of change and debugging history. Keep the incident account in the PR description.
4. The unchanged FRB site does race. The per-frame write at `lib/cuda/cudaFRBBeamReformer.cpp:219-221` modifies an object that frames on the downstream host already reference. Each frame's metadata should be rebuilt as the other non-ring stages do; the values do not change. `deepCopy` makes the `did_set_metadata` flag and the four explicit setters redundant.
5. The reader-side snapshot helps less than described. It prevents torn reads across fields, but a copy taken between publish and patch still captures the stale value.
6. Fixing the root cause does not replace the caller changes. `NDArrayRingBuffer::set_metadata` writes `name`, `type`, `dims` and `stride` without the lock. Those values repeat after frame 0, so nothing fails today. Callers patch after the call returns, so the caller fixes stand. Only publishing metadata inside `finish_write`, with the data, would subsume them. Separate PR.

jbmertens:
The CHIME branch may also want these fixes. @rhaas80

rhaas80:
I begin to see why some projects ban AI pull requests.

Can someone explain that actual race conditions succinctly? If the statement above is that we set up metadata for frame "N" (if indeed this is frames and not a ringbuffer which has only ever one copy of metadata that never changes) and the fear is that someone may access this metadata before we are done modifying it, does this take into account that no one may access metadata from frame "N" untill they have received it as a full frame (ie after we have called `mark_frame_full`?).

eschnett:
Time-dependendent metadata: They do not make sense for ringbuffers which are, logically speaking, a single infinitely-large buffer. They have a single metadata instance that exists for the whole duration the ringbuffer exists.

For non-ringbuffer frames it's logically consistent to attach time-dependent (or rather frame-dependent) metadata. There is, of course, the question of what one should do when a stage combines frames, and we need to at least document what should be done for all the time-dependent metadata that we introduce. We also need to make sure that we store all these metadata somewhere, i.e. that we don't simply pass on e.g. beam buffers which depend on a bad-feed mask without stating explicitly which version of the time-varying bad-feed mask was used for that beam buffer. Or which versions, if the beams are integrated in time.

If possible I prefer an approach where we keep time-dependent data in their own buffers. I think that, in may cases, people attach time-dependent metadata to a frame simply because this is "easier" than creating a new frame. Creating a frame takes at least 100 lines of code (if you include the yaml), whereas a `setMetadata` with a string key is very easy to do. The problems only occur in the long term, when it turns out that people haven't carefully thought through the consequences (what I describe above) and then need to pile on work-arounds in downstream code.

eschnett:
The problem is that AI is very good at talking to AI, and not so good at talking to humans.

eschnett:
I have the feeling that these changes are papering over a real problem that needs to be corrected elsewhere. If I am correct, this PR implies:
- a ringbuffer reader can start accessing ringbuffer metadata before it has acquired actual data
- this can happen quite early, before the ringbuffer writer has written any data, and before it has written any metadata

If that is correct, then we need to sort this out generally. A ringbuffer reader should not be able to access a ringbuffer's metadata before the writer has written them. It's not that reading metadata "being currently written" is the problem – it's reading metadata "before they have been written" is also a problem.

We need to check whether metadata are read/written in constructors (that would probably be bad) or only after some ringbuffer data have been acquired (then this PR is not necessary).

The changes to `FRBBeamReformer` seem correct. That code assumes it's writing metadata into a ringbuffer, but it's not.

eschnett:
We do not need to redesign the way we write metadata. The `did_set_metadata` flags are necessary (and need to be added where missing), but otherwise our current synchronization is good enough. The writer sets the metadata on the first iteration, and the reader will not (try to) access metadata until the writer has finished.

Thus, overall:
- Keep the changes to `FRBBeamReformer`
- Everywhere (where not already present) we need to introduce a `did_set_metadata` flag, ensuring that metadata are written only once, in the beginning.

Without such a flag @kvand is correct: The metadata can be modified by a writer (in later iterations) while a reader is looking at them. Since the writer would overwrite data with the same data such errors are very difficult to detect at run time. (It's still undefined behaviour and needs to be fixed.)

kvand:
Yeah, sorry about the first long-ish reply. Too much work going on and trying to take shortcuts that just made for more work!

To @rhaas80 's original question: yes, the first write is safe, as you say. The race is the re-write, where someone can get a half-overwritten set of metadata.

I think Erik's got the right of it here. There is a race on later re-use, but the did_set_metadata on the ring writers sounds like a cleaner & more universal solution. Let me rework, and keep the FRBBeamReformer change.

jbmertens:
Yeah, AIs do really tend to spit out word salads.

> The `did_set_metadata` flags are necessary (and need to be added where missing)

This is not documented or enforced anywhere, so I'm not sure how anyone should know about this usage pattern. Claude seems to think there are a number of other hidden and/or conflicting usage patterns, and other stages beyond this PR will bug out eventually too. For example:

> A ring buffer keeps one metadata object for its whole lifetime, so its fpga_seq_num is a constant: the seq number of the first sample written to the ring. Anything that pulls a frame out of a ring has to compute that frame's seq number itself:
>
> `seq = ring_meta->get_fpga_seq_num() + ring.get_read_valid().begin() * ring_meta->get_time_downsampling_fpga()`
>
> cudaCorrelator, cudaPL1bitCorrelator and cudaCopyFromRingbuffer do this. cudaBasebandBeamformer_*, cudaFRBBeamReformer and cudaPLMaskAccumulator drop the first term and use the cursor alone. The two agree only when the ring's fpga_seq_num is 0 — true in the file-driven test configs, false on site, where it is the FPGA seq number of the first frame kotekan ingested. Nothing notices until a downstream stage compares seq numbers across inputs; N2Accumulate does, and FATALs.

eschnett:
Here is my AI's comment:

(Context: I asked AI to read through the PR comments, and it agreed with my conclusion. I then asked it to create a new PR that applies to `develop`, which is now #1643. I also asked it to draft a reply – you can see for yourself whether this reply is easier to understand than the descriptions above.)

To answer @rhaas80's question succinctly: for `NDArrayBuffer` outputs the "nobody reads frame N before it is full" argument holds, because `NDArrayBuffer::set_metadata` takes a fresh metadata object from the pool on every call. For ring buffers it does not hold: a ring buffer has a *single* metadata object for its whole lifetime, `NDArrayRingBuffer::set_metadata` fills that object in place, and these stages called it on every frame. Consumers in other cudaProcess stages (`run_recv_rfi_RFImask`, `run_n2k`, `run_pl_corr`) read that same object while the producer rewrites it. That is the race. The `time_downsampling_fpga` patch after the rewrite is merely what made it visible; rewriting the unlocked `name`/`type`/`dims`/`stride` fields with identical values every frame is a data race just the same.

So the fix is not to publish more carefully on every frame, but to stop rewriting: set ring buffer metadata once, on the first frame, before the first `finish_write`. After that nothing writes, and the ring buffer mutex orders that write before any read. The generated kernels and `cudaCopyToRingbuffer` already do exactly this (`if (instance_num == 0 && !did_set_metadata)`); the hand-written RFI and PL-mask stages had `// TODO: Set these metadata only once` instead. The `instance_num == 0` part matters: `gpuProcess` creates `buffer_depth` instances of each command, and a per-instance flag alone rewrites the metadata `buffer_depth` times at startup.

#1643 does this against `develop` for all eight ring-producing stages, including `cudaRFIS012tilde`, `cudaRFISKbar` and the `rfi_SKtilde` output that this PR leaves as per-frame rewrites, and keeps the `cudaFRBBeamReformer` change from here, which is correct for the mirror-image reason (non-ring output, frame-dependent `fpga_seq_num`). It builds with `WERROR=ON` and passes the nine `gpu_batch` configs that exercise these stages, including `verify_cuda_rfi_sktilde` with the writer and reader in separate cudaProcess stages at `buffer_depth: 4`.

I suggest we take #1643, merge `develop` into `chord`, and add the same guard around `bf_mask_applied->set_metadata` in `chord`'s `cudaRFISKtilde`, which does not exist on `develop`. #1642 can stay; it is harmless.

eschnett:
@kvand Sorry, I started working on a solution before I saw your reply. See #1643.

kvand:
No worries, I hadn't got to it yet -- thanks for taking a stab at it! Will read over when I get a moment!

eschnett:
https://github.com/kotekan/kotekan/pull/1643 has been merged instead of this one.

PR #1639, opened 2026-09-03
most recent set of updates

rhaas80:
the usual mixed bag of stuff.

* code to preserve the ordering of frequencies in UpchannelizationSchedule (mostly just swapping the order of two nested loops)
* code to re-order visibility matrix entries, needed to preserve the ordering expected by the CHIME n2 backends
* a minor fix in chudaBeamReformer to report the correct FPGA sequence number (not starting at 0 at kotekan startup basically)

### DIFF
--- a/lib/stages/N2Accumulate.cpp
+++ b/lib/stages/N2Accumulate.cpp
@@ -33,6 +33,7 @@
 #include <math.h>     // for floor
 #include <memory>     // for shared_ptr, __shared_ptr_access, dynamic_pointer_cast
 #include <ostream>    // for ostream, basic_ostream
+#include <utility>    // for swap
 #ifdef WITH_OMP
 #include <omp.h>
 #endif
@@ -77,7 +78,17 @@ N2Accumulate::N2Accumulate(Config& config, const std::string& unique_name,
     _variance_mode(config.get<N2VarianceMode>(unique_name, "variance_mode")),
     _debug_accum_mode(config.get_default<bool>(unique_name, "debug_accum_mode", false)),
     _profile_info(config.get_default<bool>(unique_name, "profile_info", false)),
-    _tel(Telescope::instance()),
+    _tel(Telescope::instance()), _input_order(config.get_default<ElementOrder>(
+                                     unique_name, "input_order", _tel.fiducial_element_order())),
+    _output_order(config.get_default<ElementOrder>(unique_name, "output_order",
+                                                   _tel.fiducial_element_order())),
+    _reorder([&]() {
+        std::vector<int> reorder(_num_elements);
+        for (int idx_in = 0; idx_in < _num_elements; ++idx_in)
+            reorder.at(idx_in) = _tel.station_id_to_element_index(
+                _tel.element_index_to_station_id(idx_in, _input_order), _output_order);
+        return reorder;
+    }()),
     _feed_positions_m(_tel.get_feed_positions_m(_num_elements, _tel.fiducial_element_order())),
     n_valid_gauge(Metrics::instance().add_gauge("kotekan_N2accumulate_frac_valid_fpga_ticks",
                                                 unique_name, {"freq_id"})),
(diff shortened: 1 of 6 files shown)


lib/stages/N2Accumulate.cpp
```diff
@@ -77,7 +78,17 @@ N2Accumulate::N2Accumulate(Config& config, const std::string& unique_name,
...
+                                     unique_name, "input_order", _tel.fiducial_element_order())),
+    _output_order(config.get_default<ElementOrder>(unique_name, "output_order",
+                                                   _tel.fiducial_element_order())),
+    _reorder([&]() {
+        std::vector<int> reorder(_num_elements);
+        for (int idx_in = 0; idx_in < _num_elements; ++idx_in)
+            reorder.at(idx_in) = _tel.station_id_to_element_index(
+                _tel.element_index_to_station_id(idx_in, _input_order), _output_order);
+        return reorder;
+    }()),
```
jbmertens, line 91: May want to validate this is a permutation.

lib/stages/N2Accumulate.cpp
```diff
@@ -1054,11 +1073,13 @@ bool N2Accumulate::output_and_reset(frameID& in_frame_id, frameID& in_rfiframema
...
 
-                            int64_t n2_idx = N2::cmap(j, i, _num_elements);
+                            int64_t n2_idx = N2::cmap(j_out, i_out, _num_elements);
 
                             // Populate the visibility matrix, remember the upper-tri element
                             // is the conjugate of the lower-tri element.
-                            N2::cfloat v{(float)_vis[2 * idx], (float)_vis[2 * idx + 1]};
+                            N2::cfloat v{
+                                (float)_vis[2 * idx],
+                                (float)(swapped ? +_vis[2 * idx + 1] : -_vis[2 * idx + 1])};
```
jbmertens: Swapped is false by default, so this will change behavior for CHORD in a way we don't want.

jbmertens (commented):
Does CHIME plan to use fringestopping? May be worth an error if fringestopping is turned on for an "incorrect" ordering.

lib/stages/N2Accumulate.cpp
rhaas80: uhm, hmm, it was not intended to. Maybe I got the logic wrong after all? Though I thought I had tested this. 

if there's no swapping then neither instrument is supposed to have an (extra) complex conjugate.

lib/stages/N2Accumulate.cpp
rhaas80: Good idea. That would best go into Telescope which has the mappings, not the "clients". i'll make a (different) pull request.

rhaas80:
> Does CHIME plan to use fringestopping? May be worth an error if fringestopping is turned on for an "incorrect" ordering.

eventually I guess. My understanding is that "beamformer" order is the order also used by CHORD? So right now fringestopping should work. however what does not work is the blanking out of bands in the matrix (since what is a band changes).

lib/stages/N2Accumulate.cpp
jbmertens: Or maybe it now works in a way we _do_ want :) I'll have to check.

rhaas80: hmm, that's it most likley. I guess when rebasing the `conj` below crept back in.

rhaas80: > Or maybe it now works in a way we _do_ want :) I'll have to check.

oh I strongly doubt that. Most likely I just messed up the check or commit.

jbmertens (approved):
LGTM. It may also be worth changing fringestopping to use the stage ordering just in case... but not needed for this PR.

lib/stages/N2Accumulate.cpp
rhaas80: I had tested, but what I had tested was that the new code with reordering when given beamformer ordered voltages and without reordering when given cylinder ordered voltages would give the same answer. I had not tested that either one agreed with the result before the code change (and feeding cylinder ordered data). 

Done now, indeed had to reverse the test on `swapped`.

PR #1640, opened 2026-09-03
GNSS changes: kv/chord #1

kvand:
First stage of a major GNSS-tree merge: some functionality that touches core functions.

(The full GNSS tree is split up into 9 PRs, representing quite a bit of divergence, and worth staging into the main chord trunk.)
**Update 2026-09-04.** Three commits added.

- `utils: modulo<T> reduces on write` -- `modulo<T>` (the type behind `frameID`) kept its
  counter unreduced and computed `norm()` as `_i % _n` with `_n` unsigned, so after 2^32
  increments the wrapped int made that conversion jump by `(2^32 mod n)`: on a 24-frame buffer
  the sequence runs ..., 14, 15, 0, 1, ... and skips eight slots. Any `num_frames` that does
  not divide 2^32 is exposed; powers of two are accidentally safe. For a frameID used as a
  Buffer cursor the second such skip is a deadlock -- producer waiting on a slot the consumers
  have walked past, both idle, no FATAL and no log line. Six CHORD nodes wedged this way on
  2026-09-04, each about 15 h after its own start, with the producing stage's passed+dropped
  frame counters at exactly 2^33. `tests/boost/test_modulo.cpp` pins the semantics for both
  copies of the class.
- `utils: modulo<T> -- reduce the delta, not just the result` -- follow-up from an adversarial
  review of the above: the decrement fix held only for a signed T (an unsigned one wrapped to
  the type maximum first), and `+=` still truncated an argument too large for T. Both mutation
  paths now reduce the delta in int64 before combining.
- `Review fixes: the validation modes must report, not abort` -- `--dry-run` and
  `--check-config` aborted (SIGABRT) on a config with no root `log_level`, a non-string
  `kotekan_stage`, or a wrong-typed `metadata_pool`; the verdict went through the config's own
  log level, so `log_level: off` suppressed it; the missing-config guard printed to syslog; the
  exit status was -1 (255). Also: the blosc2 link line asserted `--as-needed` rather than
  restoring it, which leaked onto every following library and cancelled the `--no-as-needed`
  that `CMAKE_LINK_WHAT_YOU_USE` puts at the front -- now `--push-state/--pop-state`, linking
  the `find_library` path rather than `-lblosc2`. Plus `usleep` -> `sleep_for` in rawFileRead
  (useconds_t is 32-bit), a missing `<stdint.h>`, and `@conf`/`@metric` docs for the keys and
  counter this branch adds.

Note for reviewers: this PR targets `chord`, and `.github/workflows/main.yaml` only triggers on
`develop`, so none of the build/test jobs run against it. Everything above was verified locally
(gcc 13, `-DWERROR=ON`, boost tests) rather than by CI.

**Update 2026-09-07.** Review pass (2 commits):

- `Review: keep comments to what/why, move the history out` — per review, comments across
  the PR now state the invariant and the reason for it; the incident history lives here:
  - *modulo*: six CHORD nodes wedged on 2026-09-04, ~15 h after each start, with a stage
    counter at exactly 2^33 — the unreduced int wrapped and `% 24` skipped 8 slots of the
    shared bf-mask buffer.
  - *blosc2 link line*: two `libasdf-cxx.a` installs of the same declared version (8.0.0),
    one referencing 2 blosc1 symbols, the other 2 blosc1 + 7 blosc2 symbols; the `.pc` file
    says `-lblosc` for both. Neither can drop blosc1 (libblosc2 2.13 does not export
    `blosc_compress_ctx`).
  - *valve counters*: on 2026-07-27 a per-frame WARN put 4642 lines in one soak log while
    the actual loss (dropped frames became zero-filled sample_seq gaps) was invisible
    without a denominator.
  - *dry-run producer check*: the first live CHORD GNSS pipeline stalled silently because a
    pruned stage was the only producer of a mask the transposes consume.
- `Valve: rate-limit the drop WARN by time; drop the unreachable stage-duplicate check` —
  WARN on the first drop then at most once per 60 s (a valve feeding a newest-is-best
  consumer drops most of its input by design; at 1e5 frames/s one line per 100 drops is
  GB/h). `--check-config` keyed stages by full config path, which a JSON tree cannot
  duplicate, so that branch never fired; buffers and pools keep the check. `--dry-run`
  help names the no-producer check.

Built with the branch's staged build (gcc 13), `test_modulo` 5/5 pass, `--check-config`
PASSED on `config/chime_science_run_recv.yaml`.

### DIFF
--- a/kotekan/CMakeLists.txt
+++ b/kotekan/CMakeLists.txt
@@ -26,6 +26,22 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
                 -Wl,--no-whole-archive)
     if(ASDF_CXX_FOUND)
         target_link_libraries(kotekan PRIVATE blosc)
+        # asdf-cxx's pkg-config declares -lblosc (c-blosc 1.x), but SOME INSTALLS of the same
+        # version are compiled against c-blosc2 and reference blosc2_schunk_* out of ndarray.cxx.
+        # Measured 2026-08-05: cx19's libasdf-cxx.a (2026-01-20) needs 2 blosc1 symbols and no
+        # blosc2; cf06's (2026-01-22, same declared version 8.0.0) needs 7 blosc2 symbols. Same
+        # kotekan flags, same CMake cache, different library -- so this cannot be detected from
+        # the .pc file, only from the archive.
+        #
+        # Link blosc2 too where it exists. Raw flags rather than the resolved path because CMake
+        # hoists imported library paths ahead of -lasdf-cxx, where --as-needed then discards the
+        # .so as unused; and --as-needed is restored immediately so this does not leak onto
+        # everything that follows.
+        find_library(BLOSC2_LIB blosc2)
+        if(BLOSC2_LIB)
+            target_link_libraries(kotekan PRIVATE "-Wl,--no-as-needed" "-lblosc2"
+                                                  "-Wl,--as-needed")
+        endif()
     endif()
 elseif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
     target_link_libraries(kotekan PRIVATE -Wl,-all_load kotekan_libs)
--- a/kotekan/CMakeLists.txt
+++ b/kotekan/CMakeLists.txt
@@ -30,17 +30,16 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
         # version are compiled against c-blosc2 and reference blosc2_schunk_* out of ndarray.cxx.
         # Measured 2026-08-05: cx19's libasdf-cxx.a (2026-01-20) needs 2 blosc1 symbols and no
         # blosc2; cf06's (2026-01-22, same declared version 8.0.0) needs 7 blosc2 symbols. Same
-        # kotekan flags, same CMake cache, different library -- so this cannot be detected from
-        # the .pc file, only from the archive.
+        # kotekan flags, same CMake cache, different library -- so this cannot be detected from the
+        # .pc file, only from the archive.
         #
         # Link blosc2 too where it exists. Raw flags rather than the resolved path because CMake
-        # hoists imported library paths ahead of -lasdf-cxx, where --as-needed then discards the
-        # .so as unused; and --as-needed is restored immediately so this does not leak onto
-        # everything that follows.
+        # hoists imported library paths ahead of -lasdf-cxx, where --as-needed then discards the .so
+        # as unused; and --as-needed is restored immediately so this does not leak onto everything
+        # that follows.
         find_library(BLOSC2_LIB blosc2)
         if(BLOSC2_LIB)
-            target_link_libraries(kotekan PRIVATE "-Wl,--no-as-needed" "-lblosc2"
-                                                  "-Wl,--as-needed")
+            target_link_libraries(kotekan PRIVATE "-Wl,--no-as-needed" "-lblosc2" "-Wl,--as-needed")
         endif()
     endif()
 elseif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
--- a/kotekan/CMakeLists.txt
+++ b/kotekan/CMakeLists.txt
@@ -33,13 +33,18 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
         # kotekan flags, same CMake cache, different library -- so this cannot be detected from the
         # .pc file, only from the archive.
         #
-        # Link blosc2 too where it exists. Raw flags rather than the resolved path because CMake
-        # hoists imported library paths ahead of -lasdf-cxx, where --as-needed then discards the .so
-        # as unused; and --as-needed is restored immediately so this does not leak onto everything
-        # that follows.
+        # Link blosc2 too where it exists. --no-as-needed because the reference is from asdf-cxx's
+        # archive, which the linker has not seen yet at this point in the line, so the .so would
+        # otherwise be dropped as unused. push-state/pop-state rather than a bare --as-needed
+        # afterwards: --as-needed is positional and sticky, so asserting it here would silently
+        # apply to every library that follows (and would cancel the --no-as-needed that
+        # CMAKE_LINK_WHAT_YOU_USE puts at the front of the line). pop-state restores whatever the
+        # toolchain default was. Link the resolved path, not -lblosc2: find_library searches CMake's
+        # prefixes, which the linker's own -l search does not.
         find_library(BLOSC2_LIB blosc2)
         if(BLOSC2_LIB)
-            target_link_libraries(kotekan PRIVATE "-Wl,--no-as-needed" "-lblosc2" "-Wl,--as-needed")
+            target_link_libraries(kotekan PRIVATE "-Wl,--push-state,--no-as-needed" ${BLOSC2_LIB}
+                                                  "-Wl,--pop-state")
         endif()
     endif()
 elseif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
--- a/kotekan/CMakeLists.txt
+++ b/kotekan/CMakeLists.txt
@@ -26,21 +26,12 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
                 -Wl,--no-whole-archive)
     if(ASDF_CXX_FOUND)
         target_link_libraries(kotekan PRIVATE blosc)
-        # asdf-cxx's pkg-config declares -lblosc (c-blosc 1.x), but SOME INSTALLS of the same
-        # version are compiled against c-blosc2 and reference blosc2_schunk_* out of ndarray.cxx.
-        # Measured 2026-08-05: cx19's libasdf-cxx.a (2026-01-20) needs 2 blosc1 symbols and no
-        # blosc2; cf06's (2026-01-22, same declared version 8.0.0) needs 7 blosc2 symbols. Same
-        # kotekan flags, same CMake cache, different library -- so this cannot be detected from the
-        # .pc file, only from the archive.
-        #
-        # Link blosc2 too where it exists. --no-as-needed because the reference is from asdf-cxx's
-        # archive, which the linker has not seen yet at this point in the line, so the .so would
-        # otherwise be dropped as unused. push-state/pop-state rather than a bare --as-needed
-        # afterwards: --as-needed is positional and sticky, so asserting it here would silently
-        # apply to every library that follows (and would cancel the --no-as-needed that
-        # CMAKE_LINK_WHAT_YOU_USE puts at the front of the line). pop-state restores whatever the
-        # toolchain default was. Link the resolved path, not -lblosc2: find_library searches CMake's
-        # prefixes, which the linker's own -l search does not.
+        # asdf-cxx's pkg-config declares -lblosc (c-blosc 1.x), but an install built against
+        # c-blosc2 references blosc2_* from its static archive as well, and the .pc file does not
+        # say which. Link blosc2 too where it exists: --no-as-needed because the reference comes
+        # from an archive the linker has not seen yet at this point in the line (the .so would be
+        # dropped as unused); push/pop-state so the flag does not stick to the libraries that
+        # follow; the resolved path rather than -lblosc2 so CMake's prefixes are searched.
         find_library(BLOSC2_LIB blosc2)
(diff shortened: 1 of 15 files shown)


lib/stages/bufferRecv.cpp
```diff
@@ -301,10 +301,23 @@ void bufferRecv::main_thread() {
     listener = socket(AF_INET, SOCK_STREAM, 0);
     evutil_make_socket_nonblocking(listener);
 
-    // Bind even when connections from a previous run are still in TIME_WAIT.
-    // Without this, a restart within the TIME_WAIT window (a couple of minutes)
-    // fails with EADDRINUSE on a port nothing is listening on any more, which
-    // is exactly the window a supervisor restarts kotekan in.
+    // SO_REUSEADDR, UNCONDITIONALLY (2026-08-14). This used to be gated on `!drop_frames`,
```
jbmertens: The change comment here seems relevant to a specific workflow rather than general kotekan use.

kotekan/kotekan.cpp
```diff
@@ -437,6 +455,182 @@ void start_new_kotekan_mode(Config& config, bool dump_config) {
...
+ * opened, and the REST port is never bound. Catches the errors that are cheap
+ * to catch statically -- unknown stage types, duplicate names, and dangling
+ * metadata-pool references. Everything else (buffer wiring, per-stage required
+ * config keys) is left to `--dry-run`, which finds them authoritatively by
+ * actually building the pipeline.
+ *
+ * @param config The parsed config to check.
+ * @return The number of problems found; 0 means the config passed.
+ */
+int validate_config_static(Config& config) {
```
jbmertens, line 517: This (and dry runs) perform some yaml validation we may want to include as part of CI. Worth considering which configs should be part of that.

lib/stages/rawFileWrite.cpp
```diff
@@ -45,6 +45,16 @@ rawFileWrite::rawFileWrite(Config& config, const std::string& unique_name,
     _num_frames_per_file = config.get_default<uint32_t>(unique_name, "num_frames_per_file", 1);
     _prefix_hostname = config.get_default<bool>(unique_name, "prefix_hostname", true);
     _exit_after_n_files = config.get_default<uint32_t>(unique_name, "exit_after_n_files", 0);
+    // OPT-IN escape from the NDArray refusal below. The refusal exists because the frame
+    // DESCRIPTOR is set dynamically and is not written to the file, so a reader cannot
+    // recover the shape from the file alone -- a real hazard for anything self-describing.
+    // It is NOT a hazard when the shape is fixed and known out of band, which is exactly the
```
jbmertens, line 51: Arguably better is to use a file format with a header (hdf5 read/write). But this is fine for raw read/writes.

kotekan/CMakeLists.txt
```diff
@@ -26,6 +26,22 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
                 -Wl,--no-whole-archive)
     if(ASDF_CXX_FOUND)
         target_link_libraries(kotekan PRIVATE blosc)
+        # asdf-cxx's pkg-config declares -lblosc (c-blosc 1.x), but SOME INSTALLS of the same
+        # version are compiled against c-blosc2 and reference blosc2_schunk_* out of ndarray.cxx.
+        # Measured 2026-08-05: cx19's libasdf-cxx.a (2026-01-20) needs 2 blosc1 symbols and no
+        # blosc2; cf06's (2026-01-22, same declared version 8.0.0) needs 7 blosc2 symbols. Same
```
jbmertens, line 32: Could we just drop blosc1 support?

tests/boost/test_modulo.cpp
```diff
@@ -0,0 +1,107 @@
...
+// by (2^32 mod n) -- on a 24-frame buffer the sequence went 15 -> 0 and skipped
+// eight slots. Two such skips left the shared bf-mask buffer with a producer
+// waiting on a full slot and its consumers waiting on an empty one, wedging the
+// node ~15 h after start. These tests pin the reduce-on-write semantics for both
+// copies of the class.
+
+template<typename M>
+static void check_sequence_survives_2_32(unsigned n) {
+    // Nine chunks of 1e9 pass through both 2^31 and 2^32 -- the old
+    // implementation overflows its int at the third chunk and never recovers.
```
jbmertens, line 21: A change note is good, but there are a lot of comments referencing old code or debugging steps. Might be worth having a CHANGELOG or even AGENTS.md file where that sort of thing can go instead.

jbmertens (changes requested):
Mostly minor comments. The CI changes I can do later.

lib/stages/bufferRecv.cpp
kvand: Agreed — trimmed to the general statement in 799c35705: bind over TIME_WAIT so a
restart inside that window does not fail with EADDRINUSE on a dead port, and why that
is safe (SO_REUSEADDR on a listener does not let two live listeners share a port; that
would need SO_REUSEPORT). The supervisor-restart narrative is gone.

kotekan/kotekan.cpp
kvand: Agreed, and `--check-config` is built for exactly that: constructs nothing, no devices,
no hugepages, exits non-zero on failure, so it can run in the existing build job on any
runner. Suggested set: every `config/*.yaml` that is a full pipeline (the CHIME and
CHORD recv/gpu configs), plus `config/tests/*.yaml`. `--dry-run` needs the node's
DPDK/CUDA and is better left to a deploy-time check than CI. Happy to add the CI step
in a follow-up PR (this one targets `chord`, which `main.yaml` does not trigger on) —
or as a commit here if you would rather have it now.

lib/stages/rawFileWrite.cpp
kvand: Agreed for anything self-describing. The raw path is kept for the deterministic replay
case, where the shape is fixed in both configs and the value of the format is that the
bytes on disk are the bytes in the frame. Comment shortened to say that and no more.

kotekan/CMakeLists.txt
kvand: Not yet, unfortunately: both installed `libasdf-cxx.a` archives we build against
reference blosc1's `blosc_compress_ctx`/`blosc_decompress_ctx`, which `libblosc2.so.2`
(2.13) does not export — only `libblosc.so.1` does. One of the two installs additionally
pulls seven `blosc2_schunk_*` symbols, which is why blosc2 is linked conditionally on
top rather than instead. Once the asdf-cxx installs are uniform and built against blosc2
only, the `blosc` line can go and the `find_library(BLOSC2_LIB)` block becomes the only
one. Comment shortened to the general statement (799c35705).

tests/boost/test_modulo.cpp
kvand: Agreed. 799c35705 goes through every file in the PR: comments now state the invariant and
why (e.g. why reduce-on-write is not equivalent to reducing on read), and the dated
incident narrative — which node, which soak, which install — is moved into the commit
messages and the PR description. I have not added a CHANGELOG/AGENTS.md in this PR since
it would be a repo-level convention worth agreeing separately; glad to start one if you
want it.

kotekan/kotekan.cpp
```diff
@@ -437,6 +457,180 @@ void start_new_kotekan_mode(Config& config, bool dump_config) {
...
+    for (const auto& buffer : buffers) {
+        const std::string pool =
+            config.get_default<std::string>(buffer.path, "metadata_pool", "none");
+        if (pool != "none" && pool_names.count(pool) == 0)
+            report(fmt::format(fmt("buffer '{:s}' requests metadata pool '{:s}', which is not "
+                                   "defined"),
+                               buffer.name, pool));
+    }
+
+    INFO_NON_OO("config check: {:d} stages, {:d} buffers, {:d} metadata pools", stages.size(),
```
jbmertens, line 587: The static check accepts a config without a root `log_level`, while `--dry-run` and the runtime reject the same config. Verified on a copy of `run_n2accum.yaml` with the key removed: `--check-config` PASSED, `--dry-run` FAILED. Add an existence check for the root key inside the existing try block, so a CI run of the static check catches the case.

kotekan/kotekan.cpp
```diff
@@ -218,6 +225,19 @@ void print_help() {
 
     printf("    --print-json (-p)              Prints the json version of the config.\n");
     printf("    --print-yaml (-y)              Prints the yaml version of the config.\n\n");
+    printf("Validation modes (require --config; neither binds the REST port nor runs\n");
+    printf("the pipeline, and both exit non-zero if the config does not pass):\n");
+    printf("    --check-config                 Static checks only: unknown stage types,\n");
```
jbmertens, line 230: The sphinx user guide documents the direct-invocation flags in the "Direct invocation" paragraph of `docs/sphinx/user/user_introduction.rst` (`-b`, `-s`, `-e`). Add one sentence there covering the two validation modes.

lib/stages/valve.cpp
```diff
@@ -65,8 +74,20 @@ void Valve::main_thread() {
...
         } else {
-            WARN("Output buffer full. Dropping incoming frame {:d}.", frame_id_in);
+            // Rate-limited BY TIME, carrying the running total. A valve in front of a
+            // newest-is-best consumer drops most of its input by design -- at 1e5 frames/s
+            // even one line per hundred drops is gigabytes of log per hour -- while a valve
+            // that should never drop wants its first loss reported at once.
+            ++n_dropped;
+            auto now = std::chrono::steady_clock::now();
+            if (n_dropped == 1 || now - last_warn >= std::chrono::seconds(60)) {
+                WARN("Output buffer full, dropping frames: {:d} lost so far (downstream "
```
jbmertens, line 85: Valve is a general-purpose stage. Remove the clause "each loss is a gap the consumer must zero-fill", which describes one downstream consumer, and report the dropped count alone.

jbmertens (commented):
Re-reviewed at 3801468f8. The 09-07 commits resolve the earlier threads. Built in a CUDA Debug tree: `test_modulo` passes 5/5, and `--check-config` / `--dry-run` were exercised on `config/ci-tests/cpu_batch/run_n2accum.yaml` plus deliberately broken copies (unknown stage type, non-string `kotekan_stage`, dangling metadata pool, duplicate buffer name, `log_level: OFF`, consumer-only buffer). Three remaining items inline.

jbmertens (approved):
Few more minor comments -- nothing worth blocking the merge though.

kotekan/kotekan.cpp
kvand: Fixed in dc103e85b, and reproduced first: on `config/ci-tests/cpu_batch/run_n2accum.yaml`
with the key removed, `--check-config` did pass while `--dry-run` failed.

Rather than a standalone existence check I split the string→level mapping out of
`update_log_levels()` as `parse_log_level()`, and the static check now runs the config
through that same function. A second copy of the accepted-value list would have been free to
drift from the runtime's, which is this same bug in miniature. It also means an unusable
*value* is caught, not just a missing key:

- unmodified: PASSED, exit 0
- `log_level` removed: FAILED — "The config option: log_level is required, but was not found in the path: /"
- `log_level: LOUD`: FAILED — with the runtime's own list of valid values

The first two both passed before.

kotekan/kotekan.cpp
kvand: Added in dc103e85b, in the "Direct invocation" paragraph: what each mode does, that both exit
non-zero, and the operational difference that decides which to reach for — `--check-config`
constructs nothing, opens no device and binds no port, so it is safe on a node with a live
pipeline, while `--dry-run` builds the pipeline and needs the node's DPDK/CUDA and hugepages
to itself. With a two-line example block showing which one takes `sudo`.

lib/stages/valve.cpp
kvand: Removed in dc103e85b — the WARN now reports the dropped count and nothing about what a
particular consumer does with the gap. clang-format in the same commit also moves the
`<chrono>` include this PR added into the system-include block, where the rest live.

kotekan/kotekan.cpp
kvand: Leaving the CI step to you, as offered — thanks. For whenever you pick it up: `--check-config`
is now strict enough to be worth running (see the `log_level` thread above; it previously
passed configs the runtime rejects), it constructs nothing and binds nothing so it needs no
special runner, and it exits non-zero on failure.

PR #1643, opened 2026-09-09
lib/cuda: set ring buffer metadata once, on the first frame

eschnett:
## Summary

Ring buffer metadata are written **once**, on the first frame, instead of on every frame. This is the `develop` counterpart to the discussion on #1638: same race, different remedy.

A ring buffer has a single metadata object for its whole lifetime, and `NDArrayRingBuffer::set_metadata` fills that live slot-0 object in place. The RFI and PL-mask stages called it on every frame (and patched `time_downsampling_fpga` afterwards), so a consumer in another cudaProcess could observe the object mid-rewrite: `cudaCopyFromRingbuffer` read the RFImask's downsampling factor as 256 instead of 1024 and emitted sequence numbers `0.75 * uptime` behind, which took down nine CHORD nodes (#1638). Even with unchanged values, the per-frame rewrite is a data race on the unlocked `name`/`type`/`dims`/`stride` fields.

### Ring stages: write once

```cpp
// Set the ring buffer metadata once; see `NDArrayRingBuffer::set_metadata`
if (instance_num == 0 && !did_set_metadata) {
    did_set_metadata = true;
    rfi_RFImask.set_metadata(rfi_S012.get_metadata());
    // Correct RFImask metadata
    ...
}
```

After the first `finish_write` no writer touches the object, and the `RingBuffer` mutex (`finish_write` / `wait_and_claim_readable`) orders that write before any consumer's read, so no further synchronization is needed. The guard is `instance_num == 0 && !did_set_metadata` because `gpuProcess` creates `buffer_depth` instances of every command: frame 0 is always handled by instance 0, and a per-instance flag alone would rewrite the metadata `buffer_depth` times at startup. This is the convention the generated FRB kernels and `cudaCopyToRingbuffer` already follow, and it resolves the `// TODO: Set these metadata only once` comments.

Hand-written stages: `cudaRFIS012`, `cudaRFIS012bar`, `cudaRFIS012tilde`, `cudaRFISKtilde`, `cudaRFISKbar`, `cudaHFB1Accumulate`, `cudaPLMaskExpander`, `cudaPLMaskUpchannelizer`.

Generated: `cudaTranspose2048_chime` had the same per-frame rewrite of its ring output's metadata with no flag at all; the guard is added to `julia/kernels/xpose2048_template.cxx` and, identically, to the checked-in generated file. `cudaUpchannelizer_*` guarded with a function-local `static bool did_set_metadata`, i.e. one flag per *class* per process: in a multi-GPU process (one kotekan process per node) the upchannelizer stages of all GPUs but the first would skip setting their ring's metadata. It is now a member, as in the FRB beamformers (`julia/kernels/upchan_template.cxx` + 27 generated files). The other generated families were audited: the FRB beamformers already write once with a member flag; the baseband beamformer's output is not a ring buffer and is correctly set per frame.

### `cudaFRBBeamReformer`: the opposite fix

Its output is **not** a ring buffer, and `NDArrayBuffer::set_metadata` takes a fresh pool object per call, so the metadata is now set every frame. Previously `did_set_metadata` made `set_metadata` run once, and the per-frame `fpga_seq_num` write then modified an object that frames already downstream still referenced. The `deepCopy` inside `set_metadata` carries the fields the four explicit setters copied, so those and the flag are removed.

### `NDArrayRingBuffer::set_metadata`

Gains a doc comment stating the contract (single object for the ring's lifetime, call once before the first `finish_write`, guard by instance 0 + flag).

### Second commit: per-class static state (multi-GPU single-process runs)

A multi-GPU node runs all its GPU pipelines in one kotekan process, so a class has one stage object per GPU. All function-local `static` state was audited:

- **PTX build.** Every generated kernel built its PTX under a per-class `static std::once_flag`. `cudaDeviceInterface::build_ptx` loads the module into *that device's* `runtime_kernels` and is fatal if the kernel already exists there, so the flag kept the `buffer_depth` instances on one device from building twice -- but it also kept every GPU after the first from building at all, leaving those stages without a kernel. Now: build iff this device does not have the kernel yet (`runtime_kernels` is public). That also covers two stages of the same class on one GPU, which the hand-written `if (instance_num == 0)` idiom in `cudaUpchannelize.cpp` would not. Stages are constructed sequentially, so no lock. Six templates + all 59 generated files.
- **Test data generators.** `testDataGen`, `testDataGenFloat`, `testDataGenQuad` kept a `static struct timeval` (and `testDataGenFloat` its random-number engine) shared across instances; now per stage.
- Everything else found (`lib/core` singletons, registries, `fftwPlannerLock`, `gdalFileRead`'s `once_flag`, `frb_I_lock`) is intentionally process-wide.

## Verification

On cx67 (A40), Release `-O3` with `WERROR=ON` and asserts enabled:

- clang-format clean.
- Full build, no warnings, after each commit.
- All 19 `config/ci-tests/gpu_batch` configs pass after each commit, including `verify_cuda_rfi_sktilde` (writer and RFImask reader in separate cudaProcess stages with `buffer_depth: 4`, i.e. the multi-instance, cross-thread path this change is about) and `test_xpose2048.j2` for the transpose.
- `cudaHFB1Accumulate`, `cudaPLMaskUpchannelizer` and `cudaFRBBeamReformer` have no CI config and are compile-verified only.

## Notes

- `chord`'s `cudaRFISKtilde` has an additional ring output (`bf_mask_applied`) whose `set_metadata` belongs inside the same guard when this is merged there.

### DIFF
--- a/julia/kernels/bb_template.cxx
+++ b/julia/kernels/bb_template.cxx
@@ -271,15 +271,16 @@ cuda{{{kernel_name}}}::cuda{{{kernel_name}}}(Config& config,
 
     set_command_type(gpuCommandType::KERNEL);
 
-    // Build the PTX only once
-    static std::once_flag build_ptx_flag;
-    std::call_once(build_ptx_flag, [&]() {
+    // Build the PTX once per device: the kernels live in this device's `runtime_kernels`, shared
+    // by the `buffer_depth` instances of this command (building twice is fatal), while a stage on
+    // another GPU has its own device. (A static flag would be shared by the stages of all GPUs.)
+    if (!device.runtime_kernels.count("{{{kernel_name}}}_" + std::string(kernel_symbol))) {
         const std::vector<std::string> opts = {
             "--gpu-name={{{cuda_arch}}}",
             "--verbose",
         };
         device.build_ptx("lib/cuda/generated/{{{kernel_name}}}.ptx", {kernel_symbol}, opts, "{{{kernel_name}}}_");
-    });
+    }
 }
 
 cuda{{{kernel_name}}}::~cuda{{{kernel_name}}}() {}
--- a/julia/kernels/xpose2048_template.cxx
+++ b/julia/kernels/xpose2048_template.cxx
@@ -169,6 +169,9 @@ private:
         {{/isscalar}}
     {{/kernel_arguments}}
 
+    // Set once, on the first frame; see `NDArrayRingBuffer::set_metadata`
+    bool did_set_metadata;
+
     // To avoid trailing comma below
     int dummy;
 };
@@ -237,6 +240,7 @@ cuda{{{kernel_name}}}::cuda{{{kernel_name}}}(Config& config,
         {{/isscalar}}
     {{/kernel_arguments}}
 
+    did_set_metadata(false),
     dummy()                      // avoid trailing comma
 {
     // Register host memory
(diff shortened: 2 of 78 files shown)


kvand:
This looks fine to me -- I'd vote for adopting this and dropping #1638. We should probably merge / cherry-pick it over to the CHORD branch as well once in.

julia/kernels/bb_template.cxx
```diff
@@ -271,15 +271,16 @@ cuda{{{kernel_name}}}::cuda{{{kernel_name}}}(Config& config,
 
     set_command_type(gpuCommandType::KERNEL);
 
-    // Build the PTX only once
-    static std::once_flag build_ptx_flag;
-    std::call_once(build_ptx_flag, [&]() {
+    // Build the PTX once per device: the kernels live in this device's `runtime_kernels`, shared
+    // by the `buffer_depth` instances of this command (building twice is fatal), while a stage on
+    // another GPU has its own device. (A static flag would be shared by the stages of all GPUs.)
+    if (!device.runtime_kernels.count("{{{kernel_name}}}_" + std::string(kernel_symbol))) {
```
rhaas80, line 277: you are checking for a value of `0`, not `false`, aren't you? In that case I tend to prefer the explicit `a == 0` over `!a` to make it clear that the actual value matters not truth.

julia/kernels/xpose2048_template.cxx
```diff
@@ -326,40 +331,45 @@ cudaEvent_t cuda{{{kernel_name}}}::execute(cudaPipelineState& /*pipestate*/, con
...
-                                        Ein_buffer.get_buffer_name(), d, ndarray.dimname(d), metadata->get_dimension_name(d));
-                        // The ring buffer direction is special
-                        if (d > 0)
-                            assert(metadata->dim[d] == int(ndarray.extent(d)));
-                        assert(metadata->dim_scaling[d] == ndarray.dimscaling(d));
-                        if (!(metadata->stride[d] == ndarray.stride(d)))
-                            FATAL_ERROR("buffer name: {:s}, dimension: {:d}: metadata stride: {:d}, ndarray stride: {:d}",
-                                        Ein_buffer.get_buffer_name(), d, metadata->stride[d], ndarray.stride(d));
+    // Since we use a ring buffer we need to set the metadata only once
+    if (instance_num == 0 && !did_set_metadata) {
```
rhaas80, line 335: since we are talking race conditions, is it guaranteed that `instance_num == 1` cannot start running before `instance_num == 0` has finished? Otherwise this is a race.

rhaas80 (approved):
Nothing actually wrong, but a stylistic complaint.

PR #1644, opened 2026-09-09
bufferBadInputs: start the mask stream at the voltage stream's first seq num

eschnett:
> **Stacked on #1645.** This branch is rebased onto the CI fix so that its checks can run (the docker workflow's `changed-files` step failed on the original `opened` event, see #1645); the first commit in the diff belongs to #1645 and disappears once that merges.

## Summary

`bufferBadInputs` produced bad feed mask frames whose FPGA sequence numbers counted from zero (`frame_index * bf_mask_lifetime_in_samples`), i.e. it assumed the voltage stream starts at sequence number zero. It does not: the DPDK input starts at the FPGA counter. The consumers of the bad feed mask ring buffer locate mask sample `k` at `k * bf_mask_lifetime_in_samples` FPGA samples after the voltage ring buffer's logical beginning, which is the sequence number of the first voltage frame — so that is where the mask stream has to start as well.

This follows the existing precedent in `setBBBeams`: an optional `in_clock_buf` is registered as a consumer, its first frame's `fpga_seq_num` is read, the frame is released and the stage unregisters (so the producer never waits for it again), and the mask stream starts at that sequence number. Without a clock buffer the stream still starts at zero, so existing configs are unaffected.

Configured in:
- `config/chime_upgrade_frb.j2`, `config/chime_upgrade_n2.j2`: `in_clock_buf: host_voltage_buffer_0`
- `config/ci-tests/gpu_batch/verify_cuda_rfi_sktilde.yaml`: both `bufferBadInputs` stages clock off `host_rfi_s012_buffer`, and the S012 generator gets `first_frame_index: 3` so that this CI config exercises a **nonzero** origin (all CI generators started at 0 before, which is why this class of bug was never caught).

Related: #1643 (ring metadata write-once) and a follow-up fixing the same start-at-zero assumption in `cudaBasebandBeamformer`, `cudaFRBBeamReformer` and `cudaPLMaskAccumulator`.

## Verification

On cx67 (A40), Release `-O3`, `WERROR=ON`, asserts enabled: clang-format clean, `j2lint` clean on the edited configs, full build without warnings, all 19 `gpu_batch` configs pass. With `first_frame_index: 3`, `verify_cuda_rfi_sktilde` logs `Bad feed mask FPGA sequence numbers start at 24576, from the first frame of host_rfi_s012_buffer` for both stages and both checkers pass. The CHIME upgrade configs cannot run on cx67 (no FPGA controller); they are render-checked only.

### DIFF
--- a/.github/workflows/build_push_docker.yaml
+++ b/.github/workflows/build_push_docker.yaml
@@ -56,6 +56,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
+        with:
+          # `changed-files` with `since_last_remote_commit` needs the history: on a newly
+          # opened PR it falls back to `HEAD^`, which a depth-1 checkout does not have.
+          fetch-depth: 0
 
       # Has the Dockerfile ever changed vs base branch? (for tag selection)
       - name: Check Dockerfile diff vs base
--- a/config/chime_upgrade_frb.j2
+++ b/config/chime_upgrade_frb.j2
@@ -551,6 +551,8 @@ set_bf_mask:
     updatable_config:
         bad_inputs: /updatable_config/bad_inputs
     out_buf: host_bf_mask_buffer
+    # The mask stream starts at the sequence number of the first voltage frame
+    in_clock_buf: host_voltage_buffer_0
 
 # Send the bad feed mask to its GPU ringbuffer. One mask element covers
 # bf_mask_lifetime_in_samples input samples, so this runs far less often than the other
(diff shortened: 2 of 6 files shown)


jbmertens (approved):
Could you update the chord_pathfinder.j2 config as well?

jbmertens:
Just a note - this does not work as a single stage supplying the mask for both numas, since they have different voltage input buffers, so different offsets. So 2-numa configs will need 2 copies of this stage (see #1655).

jbmertens:
Proceeding with merging this so it's easier to base on top of it.

PR #1646, opened 2026-09-09
Config: make a malformed get_default() value fatal

jbmertens:
`Config::get_default()` returned the default for a value that is set but cannot be read as the requested type, and logged an error. Substituting the default for a malformed entry leaves the pipeline running an unintended configuration, which is worse than stopping, so the case is now `FATAL_ERROR_NON_OO`. An absent value still returns the default silently.

`test_config` covers both cases: an absent value yields the default; a value of the wrong type throws (inside a boost test the error logged by the fatal macro throws before `exit_kotekan` runs).

Carried on `chord` since August; extracted from the chord/develop diff.

### DIFF
--- a/lib/core/Config.hpp
+++ b/lib/core/Config.hpp
@@ -6,7 +6,7 @@
 #ifndef CONFIG_HPP
 #define CONFIG_HPP
 
-#include "kotekanLogging.hpp" // for ERROR_NON_OO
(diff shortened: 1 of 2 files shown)


andrerenard (approved):
This seems reasonable to me.  Even if we break some configs, better we learn about those issues than allow them to keep running with the error.

PR #1647, opened 2026-09-09
cudaCopyFromRingbuffer: read the metadata snapshot, not the live ring object

jbmertens:
Upstreams #1642 from `chord` (author kvand; reviewed and merged there on 2026-09-09). The commit is cherry-picked unchanged.

`execute()` already deep-copies ring slot 0 into `out_meta` under `chordMetadata::lock`, then read every value it used (`sample_bytes()`, `fpga_seq_num`, `time_downsampling_fpga`) from the live ring object, which the producer may still be writing. The output sequence number is `anchor + time_downsampling_fpga * (input_cursor / sample_bytes)` with `input_cursor` the absolute byte count since startup, so one torn read of `time_downsampling_fpga` skews the whole accumulated offset, and `N2Accumulate` then fatals on the correlation/RFI-counts mismatch. Nine CHORD nodes exited that way on 2026-09-02/03.

Read the values from the snapshot instead. No behaviour change when no producer write is in flight: the snapshot is a byte copy of the object the old code read.

### DIFF
--- a/lib/cuda/cudaCopyFromRingbuffer.cpp
+++ b/lib/cuda/cudaCopyFromRingbuffer.cpp
@@ -6,7 +6,7 @@
 #include "cudaUtils.hpp"      // for CHECK_CUDA_ERROR
 #include "cuda_runtime_api.h" // for cudaHostGetFlags, cudaMemcpyAsync, cudaHostRegister, cudaH...
 #include "gpuCommand.hpp"     // for gpuCommandType
-#include "kotekanLogging.hpp" // for DEBUG
+#include "kotekanLogging.hpp" // for DEBUG, FATAL_ERROR
 
 #include "fmt.hpp" // for compile_string_to_view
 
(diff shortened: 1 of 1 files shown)


jbmertens:
This was merged into `chord`, but not yet `develop`.

lib/cuda/cudaCopyFromRingbuffer.cpp
```diff
@@ -114,21 +114,29 @@ cudaEvent_t cudaCopyFromRingbuffer::execute(cudaPipelineState& pipestate,
...
+    // the whole uptime.
+    assert(input_cursor % out_meta->sample_bytes() == 0);
     if (initial_fpga_seq_num == -1) { // first time
         if (instance_num == 0) {      // we handle frame 0 of the buffer depth
             assert(input_cursor == 0);
-            initial_fpga_seq_num = in_meta->get_fpga_seq_num();
+            initial_fpga_seq_num = out_meta->get_fpga_seq_num();
         } else { // handle one of the later frames, frame 0 handler has set metadata
-            initial_fpga_seq_num = in_meta->get_fpga_seq_num();
+            initial_fpga_seq_num = out_meta->get_fpga_seq_num();
```
eschnett: This `if` statement does the same thing in the `then` and the `else` branch. Keep the `input_cursor == 0` check in the `if` statement, turn it into a real (non-assert) check that aborts on error, and move the assignment out of the `if` statement.

lib/cuda/cudaCopyFromRingbuffer.cpp
```diff
@@ -114,21 +114,29 @@ cudaEvent_t cudaCopyFromRingbuffer::execute(cudaPipelineState& pipestate,
...
         }
     } else { // not first time
-        assert(in_meta->get_fpga_seq_num() == initial_fpga_seq_num);
+        assert(out_meta->get_fpga_seq_num() == initial_fpga_seq_num);
     }
-    out_meta->set_fpga_seq_num(in_meta->get_fpga_seq_num()
-                               + in_meta->get_time_downsampling_fpga()
-                                     * (input_cursor / in_meta->sample_bytes()));
-    assert(input_cursor % in_meta->sample_bytes() == 0);
+    out_meta->set_fpga_seq_num(out_meta->get_fpga_seq_num()
```
eschnett: We could use `initial_fpga_seq_num` here.

eschnett (approved):
I'd prefer my first suggestion to be applied, but this change is correct either way.

jbmertens:
> I'd prefer my first suggestion to be applied, but this change is correct either way.

Both suggestions should be implemented.

PR #1648, opened 2026-09-09
testDataGen: per-instance timestamp; gains stages: require_frame_desc

jbmertens:
Two independent one-line fixes carried on `chord`.

- `testDataGen`: the `timeval` holding the frame timestamp was a function-local static, so every generator instance shared one struct and concurrent instances tore each other's `gettimeofday()` results (`chime_upgrade_n2_fakedata` runs 20). Each stage thread now uses a local.
- `loadFeedGains`, `processFRBFeedGains`: call `require_frame_desc()` instead of `ensure_frame_desc()` on the gains buffers, as `processFeedGains` already does. These stages consume declared buffers, so a missing descriptor is a config error rather than something the stage should originate.

### DIFF
--- a/lib/stages/loadFeedGains.cpp
+++ b/lib/stages/loadFeedGains.cpp
@@ -86,7 +86,7 @@ loadFeedGains::loadFeedGains(Config& config, const std::string& unique_name,
                         num_local_freq * num_elements * 2 * sizeof(float), buf->frame_size);
         }
         // Set frame description
-        buf->ensure_frame_desc(kotekan::NDArray<kotekan::GetType_t<kotekan::float32>, 3>::describe(
+        buf->require_frame_desc(kotekan::NDArray<kotekan::GetType_t<kotekan::float32>, 3>::describe(
             "gain",
             {static_cast<ptrdiff_t>(num_local_freq), static_cast<ptrdiff_t>(num_elements),
              static_cast<ptrdiff_t>(2)},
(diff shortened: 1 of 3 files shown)


andrerenard (approved):
LGTM

rhaas80:
These are both producers of the buffers. So they should set the metadata or ensure it has the value they will set it to. They should not require it to be already set, shouldn't they?

lib/stages/loadFeedGains.cpp
```diff
@@ -86,7 +86,7 @@ loadFeedGains::loadFeedGains(Config& config, const std::string& unique_name,
                         num_local_freq * num_elements * 2 * sizeof(float), buf->frame_size);
         }
         // Set frame description
-        buf->ensure_frame_desc(kotekan::NDArray<kotekan::GetType_t<kotekan::float32>, 3>::describe(
+        buf->require_frame_desc(kotekan::NDArray<kotekan::GetType_t<kotekan::float32>, 3>::describe(
```
rhaas80, line 89: Note that in line 77 above it registers itself as a producer. So it cannot require. It must ensure.

lib/stages/processFRBFeedGains.cpp
```diff
@@ -41,7 +41,7 @@ void processFRBFeedGains::copy_upchannelize_f(const float* src_f, float16_t* dst
 }
 
 void processFRBFeedGains::set_frame_desc(Buffer* buf) {
-    buf->ensure_frame_desc(kotekan::NDArray<kotekan::GetType_t<kotekan::float16>, 5>::describe(
+    buf->require_frame_desc(kotekan::NDArray<kotekan::GetType_t<kotekan::float16>, 5>::describe(
```
rhaas80, line 44: `processFRBFeedGains` is a sub-class of `processFeedGains` which calls `set_frame_desc` on `out_buf` which is has registered itself a producer of. So it must ensure and cannot require.

PR #1649, opened 2026-09-09
Add bufferDedup: forward frames only when their contents change

jbmertens:
A generic pass-through stage that compares each input frame against the last frame it forwarded and drops repeats. The first frame after startup is always forwarded, and metadata passes through, so a forwarded frame keeps the FPGA sequence number of the input that carried the change. An optional `resend_after_frames` re-emits unchanged contents periodically, so a downstream consumer that can drop frames (e.g. a `bufferSend` link that was down) recovers the current state.

On `chord` the stage deduplicates the applied bad-feed mask echoed back from the GPU before the `bufferSend` link to the receive node, so only mask changes cross the network. The stage itself has no CHORD-specific dependency.

Covered by `tests/boost/test_bufferDedup`. Carried on `chord`; extracted from the chord/develop diff.

### DIFF
--- a/lib/stages/CMakeLists.txt
+++ b/lib/stages/CMakeLists.txt
@@ -68,6 +68,7 @@ add_library(
     pulsarNetworkProcess.cpp
     bufferMerge.cpp
     bufferCopy.cpp
+    bufferDedup.cpp
     bufferSwitch.cpp
     bufferDelay.cpp
     bufferQuiet.cpp
--- a/lib/stages/bufferDedup.cpp
+++ b/lib/stages/bufferDedup.cpp
@@ -1,0 +1,67 @@
+#include "bufferDedup.hpp"
+
+#include "StageFactory.hpp"   // for REGISTER_KOTEKAN_STAGE
+#include "kotekanLogging.hpp" // for FATAL_ERROR, DEBUG
+#include "visUtil.hpp"        // for frameID, modulo
+
+#include "fmt.hpp" // for compile_string_to_view
+
+#include <cstring>    // for memcmp, memcpy
+#include <functional> // for bind, function
+
+using kotekan::bufferContainer;
+using kotekan::Config;
+using kotekan::Stage;
+
+REGISTER_KOTEKAN_STAGE(bufferDedup);
+
+bufferDedup::bufferDedup(Config& config, const std::string& unique_name,
+                         bufferContainer& buffer_container) :
+    Stage(config, unique_name, buffer_container, std::bind(&bufferDedup::main_thread, this)),
+    in_buf(get_buffer("in_buf")), out_buf(get_buffer("out_buf")),
+    resend_after_frames(config.get_default<int>(unique_name, "resend_after_frames", 0)) {
+
+    in_buf->register_consumer(unique_name);
+    out_buf->register_producer(unique_name);
+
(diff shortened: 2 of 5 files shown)


andrerenard (changes requested):
I'm a bit concerned about the `resend_after_frames` logic.  If we have downstream stages that can drop things, we introduce the possibility of desynchronized bad-feed masks with the data to which they were applied?   I'm generally not concerned about the exact time that the mask is applied, but I do want us to accurately record _what was done_ to each frame of data.   Perhaps I'm missing something here, but it seems like we should be following a config tracker style tag and retrieve model, not a best effort model?

eschnett:
I don't understand the "echoed back from the GPU" part. This bit just copies the actual bad feed mask into a copy. Why is this useful – why not just use the original, actual bad feed mask?

If all you want is a bad feed mask at a shorter cadence then we can just change the respective yaml parameter. (Or copy the bad feed mask on the CPU. Or reuse the bad feed mask on the CPU multiple times. It doesn't matter – the bad feed masks are small.) But adding extra logic to copy bad feed masks to an RFI stage seems like a wrong approach.

jbmertens:
The "simple" bookkeeping I had in mind was to just pass the bad feed mask used on the gpu back to the cpu, similar to the RFI mask. I.e, the mask should be treated as data, not metadata. De-duplication was just a way to compress and pipeline that, although it isn't much data in the first place.

Yes, we could re-process the output from bufferBadInputs instead by mirroring exactly what was done on the gpu, in a new stage or something. Which is fine, equivalent, and avoids copying the mask off GPU. But at the expense of something "specialized"; the added complexity didn't seem worth saving such a little bit of bandwidth.

eschnett:
The difference between RFI and bad feed mask is that the bad feed mask is generated on the CPU, so it's already available there.

I'd be happy to write the code. What exactly do you need from the bad feed mask that the already-on-CPU bad feed mask doesn't do? Is there a particular stage where it needs to be plumbed in?

jbmertens:
> What exactly do you need from the bad feed mask that the already-on-CPU bad feed mask doesn't do? Is there a particular stage where it needs to be plumbed in?

The synchronization information. The bad feed mask doesn't get a useful label in bufferBadInputs, and I wasn't going to change that stage since I don't know what CHIME does with it. Plus, the label could be different for the different NUMAs, and the details may change if we change the cadence the mask is sent to the gpu or how operations are performed on the gpu. So a new stage would be needed that mirrors what each GPU is doing, and has to be kept updated if that changes. The management overhead on that doesn't seem worth it to me,

jbmertens:
The downstream stages that need this information are N2Accumulate, which ANDs the masks together, and the hdf5N2 file writer, which writes the record to disk.

eschnett:
The GPU states do not do anything with the bad feed mask, they only access it read-only. The only "writing" happens in the copy-cpu-to-gpu stage.

Let's talk in person in 9 minutes in the CHIME meeting.

eschnett:
@jbmertens needs the initial FPGA sequence number of the bad feed mask to be correct, not just set to zero.

eschnett:
The synchronization information is introduced by PR [#1644](https://github.com/kotekan/kotekan/pull/1644).

jbmertens:
I'm going to close this one and rework #1655 to use #1644.

PR #1650, opened 2026-09-09
RfiSKMetrics: export per-feed spectral kurtosis from SKbar

jbmertens:
New CPU stage consuming the host copy of `cudaRFISKbar`'s single-feed SK output (`[T, F, {SK, b, sigma}, P, D]`; a negative sigma marks an invalid cell). Each element's SK is averaged over its valid cells per frame and folded into an exponential moving average (`ema_frames`, default 256).

The averages are served on a JSON GET endpoint `<unique_name>/sk` for machine consumers (the bffs feed flagger polls it) and as Prometheus gauges `kotekan_rfi_sk_per_feed{element=}` and `kotekan_rfi_sk_per_feed_valid_frac` for dashboards. Elements use the flat `[P][D]` element index, the same indexing as `/updatable_config/bad_inputs`. The single-feed SK is computed for every feed regardless of the bad-feed mask, so flagged feeds keep being measured and can heal.

Sphinx page added under processes. The CHORD pathfinder config block that wires the stage follows with the config PRs.

### DIFF
--- a/docs/sphinx/user/processes/RfiSKMetrics.rst
+++ b/docs/sphinx/user/processes/RfiSKMetrics.rst
@@ -1,0 +1,5 @@
+************
+RfiSKMetrics
+************
+
+.. doxygenclass:: RfiSKMetrics
--- a/lib/stages/CMakeLists.txt
+++ b/lib/stages/CMakeLists.txt
@@ -120,6 +120,7 @@ add_library(
     N2FrameToVisFrame.cpp
     RfiMaskSum.cpp
     RfiFrameMask.cpp
+    RfiSKMetrics.cpp
     CountLostPLSamplesScalar.cpp
     PLMaskExpandedToCompact.cpp
     # Julia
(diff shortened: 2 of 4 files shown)


eschnett (dismissed):
There are no checks ensuring the `in_buf` has `float` type elements (you can check the metadata for this), and that the buffer has enough elements to not segfault when its elements are accessed. This should probably be metadata check. I don't think metadata can be accessed in the constructor, but the check is cheap either way.

Ideally the check would ensure that the memory layout matches what you expect, but at the very least the datatype and the number of elements need to be correct.

jbmertens:
There should now be a check for the buffer layout and type.

andrerenard (approved):
This looks useful to have!   No real comments on the code.  But I did have a larger question about the "RFI broker" equivalent for CHORD.  Are you thinking of using this type of structure for the broker to compute the cross-frequency per element SK for feed flagging, or follow the CHIME push model?

jbmertens:
This follows a model where the broker pulls, so this stage takea the place of RFIBroadcast stage.

This simplifies the design of the bad feed flagger from a full standalone app to a simple service script integrated with choco, and shouldn't really have much behavior difference. I was aiming for something easier to maintain.

PR #1651, opened 2026-09-09
dpdk: validate CRS packets in the distributors before ring indexing

jbmertens:
`crs16BoardDistributor` and `crs1BoardDistributor` computed `worker_ring_ids[stream_id / (128 / size)]` straight from packet bytes. The capture ports run in promiscuous mode, so stray traffic (LLDP, ARP, IPv6 ND) also reaches the distributor, and its bytes at the stream_id offset are arbitrary: any value >= 128 indexed past the ring-id vector and handed `rte_ring_enqueue` a wild `rte_ring*`. Observed on the CHORD pathfinder as an immediate lcore SEGV in `handle_packet`, or as heap corruption surfacing later in unrelated threads. The capture workers already validate `pkt_len` and the 0xCF cookie, but they run after the distributor.

Drop anything with the wrong `pkt_len`, a bad IP checksum or an out-of-range stream_id before touching the rings; count each class plus ring-full drops in per-port Prometheus gauges (`kotekan_dpdk_distributor_*_total`); warn once per class. The expected size comes from the `packet_size` config key, which every config using these handlers already sets.

Squashes two `chord` commits: the validation, and Andre Renard's review follow-ups (co-authored). Running on the pathfinder since July.

### DIFF
--- a/lib/dpdk/crs16BoardDistributor.hpp
+++ b/lib/dpdk/crs16BoardDistributor.hpp
@@ -27,8 +27,15 @@ public:
     /// Processes the incoming packets
     int handle_packet(struct rte_mbuf* mbuf) override;
 
-    /// Update stats, not used by this handler yet.
-    virtual void update_stats() override {};
+    /// Export the distributed/dropped packet counters.
+    virtual void update_stats() override {
+        std::vector<std::string> port_label = {std::to_string(port)};
+        packets_total_metric.labels(port_label).set(total_packets);
+        bad_checksum_packets_total_metric.labels(port_label).set(bad_checksum_packets);
+        non_crs_packets_total_metric.labels(port_label).set(non_crs_packets);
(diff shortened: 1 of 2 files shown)


andrerenard (changes requested):
Ah this is going to cause me some fun merge conflicts.  You can merge this in now so things work, but I have another PR coming that is going to conflict with this...  One small change that might be worth adding, I've made invalid stream_ids fatal.  It would be really strange to have a packet with a valid checksum and CRS cookie that also had a bad stream_id unless the F-engine or X-engine has a configuration error.   I've also removed the metric for that.

jbmertens:
If there is no diff, hopefully that won't be too bad 😬 

Do you want to approve these changes, and issue another PR for fixes, or add changes  on top of this PR?

andrerenard (approved):
I'll approve, my PR has a bunch of other changes.  So I'll just resolve the conflicts.

PR #1652, opened 2026-09-09
Check the SK kernel against analytically known values in CI

jbmertens:
`verify_cuda_rfi_sktilde` compares the GPU kernel to the CPU oracle, which cannot catch a bug shared by both implementations. `verify_cuda_rfi_sktilde_analytic` feeds one constant S012 stream (via `testRFIS012Gen`'s per-element value arrays) to three SK kernels differing only in their bad-feed mask and asserts values known on paper. The two feed groups share (S0, S1), so the Nita-Gary bias and sigma (table interpolations with no closed form) are identical for every feed, and the bias cancels exactly in the difference between the unmasked and half-masked paths: 257/510 in the feed-averaged SK, zero in the bias component. The third path masks feeds past the validity gate's threshold and must hold the exact `[0, 0, -1]` dummy values. All thresholds are dyadic, so no comparison sits on a float rounding boundary.

Adds `testDataCheckExpected`, a generic checker of float frames against expected constants cycled per component, optionally on the difference of two buffers, with a skip list for components that have no analytic value (here sigma, which shrinks with the feed count by a table value).

Runs with the rest of `config/ci-tests/gpu_batch`. Verified standalone: passes on an A40 pair (2 devices), 2s runtime.

### DIFF
--- a/config/ci-tests/gpu_batch/verify_cuda_rfi_sktilde_analytic.j2
+++ b/config/ci-tests/gpu_batch/verify_cuda_rfi_sktilde_analytic.j2
@@ -1,0 +1,259 @@
+{#-
+  verify_cuda_rfi_sktilde_analytic.j2
+
+  Checks cudaRFISKtilde against analytically known values, complementing the
+  GPU-vs-CPU-oracle comparison in verify_cuda_rfi_sktilde.yaml (which cannot catch a
+  bug shared by both implementations).
+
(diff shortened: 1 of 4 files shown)


eschnett (approved):
A test can't hurt. I think this only tests with a single magic number. It would be nice to add a comment to that magic number (`check_mask_difff/expected`) describing how it was conjured up.

PR #1653, opened 2026-09-09
TransposeBasebandArray: write the scalar path element-major

jbmertens:
The scalar fallback in `TransposeBasebandArray` looped `element_long` outside `time_short`, so it wrote `element_short` bytes and then stepped by `out_time_stride` (48 KiB) to the next time sample, coming back for the next 8 bytes only after touching all 16 output rows. No output cache line was ever completed in one visit.

Swapping the loops fills each time sample's whole row of elements in order, two complete cache lines at a time, keeping the same 8-byte moves since an `element_long`'s lanes are contiguous at both ends. Measured on cx69 with 3 MiB frames and the scalar path forced: 2.40 to 0.58 ms/frame, 4.1x. Adds `disable_avx512` so the scalar path can be exercised on a machine that has AVX512, and a boost test covering both paths, including that unwritten source elements arrive as `0x88` and a lost block is filled with it. No CHORD node runs the scalar path; this only matters where AVX512 is unavailable.

### DIFF
--- a/lib/stages/TransposeBasebandArray.cpp
+++ b/lib/stages/TransposeBasebandArray.cpp
@@ -142,16 +142,23 @@ STAGE_CONSTRUCTOR(TransposeBasebandArray) {
             NUM_ELEMENTS / 8));
     }
 
-    // AVX512 fast path is always enabled with these constants
+    // Escape hatch to run the scalar path on a machine that supports AVX512, so that both
+    // paths can be exercised and compared without rebuilding for a different -march.
+    const bool disable_avx512 = config.get_default<bool>(unique_name, "disable_avx512", false);
+
     use_avx512_fast_path = false;
+    if (disable_avx512) {
+        INFO("TransposeBasebandArray: AVX512 fast path disabled by config, using scalar path");
+    } else {
 #ifdef __AVX512F__
-    // With our constants: time_short=16, element_short=8, num_elements=128, element_long=16
-    // All requirements are met
-    use_avx512_fast_path = true;
-    INFO("TransposeBasebandArray: AVX512 fast path enabled");
+        // With our constants: time_short=16, element_short=8, num_elements=128, element_long=16
+        // All requirements are met
+        use_avx512_fast_path = true;
+        INFO("TransposeBasebandArray: AVX512 fast path enabled");
 #else
-    INFO("TransposeBasebandArray: AVX512 not available, using scalar path");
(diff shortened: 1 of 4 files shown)


andrerenard:
I don't expect the scalar fallback to ever get used, but I guess no harm making it faster. 

The `FATAL_ERROR` vis `throw` I'm not sure is the right approach here.   The original idea was to use a throw in the constructor, since that can interrupt the stage construction process.   Doing a `FATAL_ERROR` allows things to progress into the controlled shutdown system - but that shouldn't be needed at the stage constructor step (nothing is running yet).  Although perhaps we should add a little unwinding logic (by catching the throw) in kotekan mode.   But we should be able to exit a bit more efficiently at the construction step of kotekan mode.

jbmertens:
Yeah, this was noticed when playing with reordering in this stage, which is now irrelevant I guess. The error handling is mostly for consistent behavior. 

Both changes are somewhat vestigial. They can be reverted, or not.

andrerenard:
I'm tempted to revert at least the throw vs. ERROR parts of this.  I think throwing in the constructors is actually the better pattern to follow.  Otherwise we are unnecessarily completing more startup and shutdown steps then are needed at this point in the execution.  If we have issues with that error gets returned to the logs/terminal, we should address them with modifications to kotekanStage by catching the exceptions and dealing with them in the framework startup logic.

lib/stages/TransposeBasebandArray.cpp
eschnett, line 34: CHORD final vs. CHORD pathfinder?

This constant can probably remain a run-time parameter without decreasing efficiency.

eschnett: This isn't done on the scalar path?

eschnett (approved):

eschnett:
This only affects the scalar fallback, so it's not efficient anyway.

jbmertens:
Dropped the FATAL_ERROR commit; the constructor throws `std::runtime_error` on bad config again, as before. The PR is now the scalar-path change alone. Rebuilt and ran `test_TransposeBasebandArray` at the new tip.

lib/stages/TransposeBasebandArray.cpp
jbmertens: I believe @andrerenard's eventual plan is to not use this stage.

jbmertens: Added.

jbmertens:
PR updated so the fatals were reverted to throws.

PR #1654, opened 2026-09-09
testDataGenFewHot: test data generator to probe input reordering

jbmertens:
`testDataGenFewHot`: a test data generator that sets a few "hot" elements in an otherwise-zero baseband frame, used by `chime_upgrade_frb.j2` to probe input reordering through the pipeline. Four commits by Roland Haas, kept separate: the initial stage, a data-shape hack to produce `upchan_U16`-like output, a switch to `ensure_frame_desc`, and an adjustment to `dimscalings`.

Carried on `chord`; extracted from the chord/develop diff. Reformatted to the current clang-format style, which the original commits predate.

### DIFF
--- a/lib/stages/CMakeLists.txt
+++ b/lib/stages/CMakeLists.txt
@@ -104,6 +104,7 @@ add_library(
     FPGAMonitor.cpp
     givenDataGen.cpp
     testDataGen.cpp
+    testDataGenFewHot.cpp
     DPDKShuffleSimulate.cpp
     BadInputFlag.cpp
     hexDump.cpp
--- a/lib/stages/testDataGenFewHot.cpp
+++ b/lib/stages/testDataGenFewHot.cpp
@@ -98,10 +98,10 @@ void testDataGenFewHot::main_thread() {
 
         chordmeta->set_frame_counter(seq_num);
 
-        buf->allocate_ndarray_frame_desc<kotekan::GetType<kotekan::int4x2_swapped_withoffset>::type,
-                                         4>("Ebar", {1024,256,2, NUM_ELEMNS/2}, {"Tbar", "Fbar", "P", "D"});
+        buf->ensure_frame_desc(kotekan::GenericNDArray::describe(kotekan::int4x2_swapped_withoffset,
+                                         "Ebar", std::vector<ptrdiff_t>{1024,256,2, NUM_ELEMNS/2}, std::vector<kotekan::Symbol>{"Tbar", "Fbar", "P", "D"}));
         /* test that things are consistent */
-        chordmeta->check_frame_desc(buf->get_ndarray_frame_desc());
+        chordmeta->check_frame_desc(buf->get_frame_desc<kotekan::GenericNDArray>());
 
         std::memset(frame, 0x88 /* 0 volts */, 1024*256*NUM_ELEMNS);
 
--- a/lib/stages/testDataGenFewHot.cpp
+++ b/lib/stages/testDataGenFewHot.cpp
@@ -20,13 +20,13 @@ using kotekan::Stage;
 REGISTER_KOTEKAN_STAGE(testDataGenFewHot);
 
 #define NUM_ELEMNS 2048
-#define NUM_FREQ 1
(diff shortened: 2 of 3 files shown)


lib/stages/testDataGenFewHot.cpp
```diff
@@ -0,0 +1,122 @@
...
+        chordmeta->set_array_dimension(0, 1024, "Tbar", 16);
+        chordmeta->set_array_dimension(1, 256, "Fbar", 1);
+        chordmeta->set_array_dimension(2, 2, "P", 1);
+        chordmeta->set_array_dimension(3, NUM_ELEMNS / 2, "D", 1);
+        chordmeta->set_strides_simple();
+        chordmeta->type = kotekan::int4x2_swapped_withoffset;
+        std::vector<int> coarse_freq(NUM_FREQ);
+        std::vector<int> freq_upchan_factor(coarse_freq.size());
+        std::vector<int> freq_upchan_index(coarse_freq.size());
+        for (int f = 0; f < 256; f++) {
```
andrerenard: There are a lot of locations where a hard-coded value, e.g. 256 is used instead of a constant, e.g. `NUM_FREQ`.

lib/stages/testDataGenFewHot.hpp
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @par Buffers
+ * @buffer out_buf Buffer to fill
+ *         @buffer_format any format
+ *         @buffer_metadata chordMetadata
+ *
+ * @conf  type                  String. "single", "double"
+ * @conf  freq_id               Int. Frequency to produce data for.
+ * @conf  dish                  Int. Dish to set.
+ * @conf  samples_per_data_set  Int. How often to produce data.
```
andrerenard:
```suggestion
 * @conf  samples_per_data_set  Int. How many time samples per frame.
```

lib/stages/testDataGenFewHot.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+#include <memory>   // for shared_ptr
+#include <stddef.h> // for ptrdiff_t
+#include <stdint.h> // for uint32_t
+#include <string>   // for string, basic_string
+#include <vector>   // for vector
+
+/**
+ * @class testDataGenFewHot
+ * @brief Generate test data as a standin for DPDK.
```
andrerenard: Some description of what the data looks like?

lib/stages/testDataGenFewHot.hpp
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @par Buffers
+ * @buffer out_buf Buffer to fill
+ *         @buffer_format any format
+ *         @buffer_metadata chordMetadata
+ *
+ * @conf  type                  String. "single", "double"
+ * @conf  freq_id               Int. Frequency to produce data for.
+ * @conf  dish                  Int. Dish to set.
```
andrerenard: "dish" is used here, but it's treated as an "element" ordering internally - and they mean different things.

lib/stages/CMakeLists.txt
```diff
@@ -104,6 +104,7 @@ add_library(
     FPGAMonitor.cpp
     givenDataGen.cpp
     testDataGen.cpp
+    testDataGenFewHot.cpp
```
andrerenard, line 107: I realize this is following a pattern, but these should ideally move to `lib/testing`.  Especially now that I've factored out the strange boost conflict.

andrerenard (changes requested):
Just some minor pedantic comments.

lib/stages/CMakeLists.txt
jbmertens: The test data generation stages actually have been used in production for CHORD, standing in for missing brokers. Any objection to moving these in a future PR?

andrerenard: Yeah we can move this to a later PR.

andrerenard (approved):
This is looking much better with the layout inferred for the buffer.

PR #1655, opened 2026-09-09
Bad-feed mask plumbing and pathfinder configs

jbmertens:
Based on develop, which now includes #1644 (bufferBadInputs clocked off the voltage stream), #1657, #1658 and #1650. Replaces the earlier version of this PR, which echoed the applied mask off the GPU and deduplicated it (#1649, closed).

## Summary

The bad feed mask that gates the GPU RFI statistics is recorded with the data it was applied to, using the mask stream itself rather than a copy from the GPU.

- **N2Accumulate** takes an optional `in_bad_feed_mask_buf`: the bufferBadInputs stream the GPU gates on, consumed 1:1 with the correlation frames and checked against them by FPGA sequence number. #1644 clocks that stream off the voltage buffer's first frame and the GPU indexes the mask by ring position, so mask frame k is exactly the element applied to correlation frame k. The mask is ANDed over each accumulation bin into the per-element `flags`, mapped through the input->output reorder; N2TimeDownsample folds the flags over its window.
- **bufferBadInputs** masks elements the telescope places outside the main array (CHORD's Fake dishes and RFI antennas) on top of the posted list, through the Telescope grid lookup, and stamps the clock buffer's coarse frequencies on every frame so streams from several instances can be told apart downstream. `in_clock_buf` stays optional; configs without it are unchanged.
- **cudaRFIS012tilde, cudaRFISKtilde, cudaRFISKbar** stop when the mask ring's origin seq differs from the data ring's: a mask stream clocked to another GPU's voltage stream would gate the wrong samples. Same pattern as cudaCorrelator's voltage/RFI-mask consistency assert, but live in Release.
- **cudaPLMaskAccumulator** names the kernel's 128 samples per mask element once, uses it for the pl_counts seq offset and downsampling, and stops if the mask's `time_downsampling_fpga` disagrees with it, following up on #1657.
- **EigenN2Iter** accepts compact DishInputs frames: the full-triangle check now counts products (n(n+1)/2 over the frame's own elements) instead of requiring the FullUpperTri layout name. The pathfinder configs run eigencalc on DishInputs buffers, which develop's check rejected at the first frame. eigenN2Iter_dish_inputs covers it.
- **hdf5N2Write** takes an optional `in_bad_feed_mask_buf` and writes a `/bad_feed_mask` group per file: one row per mask frame on the streams' common FPGA grid over the file's tick span, one column per stream (an X-engine half, identified by the coarse frequencies it applied the mask to, which bufferBadInputs stamps on every frame), -1 where a stream's frame did not arrive, and the streams' frequency lists (`stream_freq_id`). Every frequency with data in a file must belong to one of its streams; masks and data come from the same X-engine process, so anything else is a miswired mask send and the writer stops. A dedicated thread takes the frames in, so the buffer drains however long a file write takes; per-stream frame, missing-frame and late-frame counters are exported (a frame arriving after its row was written stays recorded as -1). Documented in n2_vis_hdf5.rst; `/flags` stays the per-integration record.
- **chord_pathfinder configs**: one bufferBadInputs per NUMA half, each clocked to its half's voltage buffer (the two DPDK captures pick their start seqs independently, so the halves can start frames apart) and feeding only that half's GPU copy, N2Accumulate and send. Each half's stream goes whole to the recv node (port 11029) through a private bufferCopy'd buffer, so bufferSend's congestion heuristic sees only its own backlog rather than the frames N2Accumulate holds; the recv node's hdf5N2Write folds it into `/flag_updates`. The configs also carry the pathfinder's deployed state (live dish table, first-stage excision, DPDK remap, DishInputs layout, RfiSKMetrics, frame descriptors on the sends). A mask stream must not cross NUMA halves: the GPU and N2Accumulate checks catch a cross-wired stream whenever the halves started on different frames; the config comments cover the remaining case. The sends' reconnect interval is now on the key bufferSend reads (`reconnect_time`; `retry_time` was never read), the recv writers no longer log at DEBUG (an INFO line per received frame), and dead keys (`num_gpus`, writer `file_name`/`prefix_hostname`/`zip_compression`/`num_ev`/`n2_layout`, unused jinja variables) are gone.

## Verification

Test build (gcc, WERROR, CUDA, DPDK, boost tests) clean. test_hdf5N2Write (12 cases, including the new test_writer_bad_feed_mask), test_eigenStages (8 cases, including the new eigenN2Iter_dish_inputs), test_N2Subset, test_N2FrameDesc and test_CHORDTelescope pass; tests/test_n2_accumulate.py 42/42; run_n2accum.yaml and run_n2accum_chime.yaml pass via run_tests.sh with the mask stream clocked off the correlation buffer. chord_pathfinder_recv.j2 constructs its full pipeline (it fails only at the site output mount). chord_pathfinder.j2 constructs every CPU-side stage with the DPDK and GPU stages stripped from the rendered config, and a static pass over all 34 GPU command blocks finds every config key develop's commands require; the devbox has no hugepages and its NVIDIA driver is currently unloadable, so the GPU origin check, the PLMaskAccumulator change and the gpu_batch configs need the GPU CI run.

### DIFF
--- a/config/chord_pathfinder.j2
+++ b/config/chord_pathfinder.j2
@@ -4,9 +4,6 @@
 -#}
 
 
-# num_output_frames: fully accumulated N2 frames per frequency that should reach hdf5N2Write
-{% set num_output_frames = 100 %}
-
 # num_local_freq: number of frequency IDs to simulate on this GPU;
 #    must be divisible by 4 because pl_mask shapes use num_local_freq/4.
 {% set num_local_freq = 384 %}
@@ -16,8 +13,6 @@
 {% set sampling_rate_MHz = 3.2e3 %}
 {% set fft_length = 16384 %}
 {% set nyquist_zone = 1 %}
-{% set df_MHz = (sampling_rate_MHz / fft_length) %}
-{% set freq0_MHz = 0 %}
 
 # Science band narrowed to just the simulated freq_ids above
 {% set science_min_freq_MHz = 300 %}
@@ -40,17 +35,8 @@
 # of sub-integrations produced per frame. Choose any positive divisor of samples_per_data_set (does not need to be 1024).
 {% set sub_integration_ntime = 8192 %}
 
-# n2k sub-integrations contained in one generated frame (derived)
-# (= samples_per_data_set / sub_integration_ntime).
-{% set integrations_per_frame = samples_per_data_set // sub_integration_ntime %}
-
-# n2k sub-integrations to combine into each output N2 frame by N2Accumulate.
-# Choose a multiple of integrations_per_frame so n2k_frames_per_output is an integer and generators run long enough.
-# 238 ~= 10 seconds. 
-{% set num_subintegrations_per_bin = 238 %} 
-
-# correlation frames needed to form one output N2 frame
-{% set n2k_frames_per_output = num_subintegrations_per_bin // integrations_per_frame %}
+# n2k sub-integrations to combine into each output N2 frame by N2Accumulate. 238 ~= 10 seconds.
+{% set num_subintegrations_per_bin = 238 %}
 
 # RFI Detection Parameters
 {% set rfi_downsampling_factor = 256 %}
@@ -83,7 +69,6 @@ config_tracker:
 
 numa_node: 0
 
-num_gpus: 1
 buffer_depth: 4
 num_gpu_frames: 4
 host_side_buffer_depth: buffer_depth * 6
@@ -1610,7 +1595,7 @@ buffer_send_bf_mask:
     buf: host_bf_mask_send_buffer
     server_ip: 10.222.0.51
     server_port: 11029
-    retry_time: 10.0
+    reconnect_time: 10
     drop_frames: true
 
 buffer_send_bf_mask_1:
@@ -1619,7 +1604,7 @@ buffer_send_bf_mask_1:
     buf: host_bf_mask_send_buffer_1
     server_ip: 10.222.0.51
     server_port: 11029
-    retry_time: 10.0
+    reconnect_time: 10
     drop_frames: true
 
 # Buffer sends. use_frame_desc sends the frame descriptor once per connection, so the receiver
@@ -1631,7 +1616,7 @@ buffer_send_n2_subset:
     buf: n2_eigen_buffer
     server_ip: 10.222.0.51
     server_port: 11025
-    retry_time: 10.0
+    reconnect_time: 10
     drop_frames: true
 
 buffer_send_n2_full:
@@ -1640,7 +1625,7 @@ buffer_send_n2_full:
     buf: n2_buffer
     server_ip: 10.222.0.51
     server_port: 11027
-    retry_time: 10.0
+    reconnect_time: 10
     drop_frames: true
 
 buffer_send_n2_subset_1:
@@ -1649,7 +1634,7 @@ buffer_send_n2_subset_1:
     buf: n2_eigen_buffer_1
     server_ip: 10.222.0.51
     server_port: 11025
-    retry_time: 10.0
+    reconnect_time: 10
     drop_frames: true
 
 buffer_send_n2_full_1:
@@ -1658,5 +1643,5 @@ buffer_send_n2_full_1:
     buf: n2_buffer_1
     server_ip: 10.222.0.51
     server_port: 11027
-    retry_time: 10.0
+    reconnect_time: 10
     drop_frames: true
--- a/config/chord_pathfinder.j2
+++ b/config/chord_pathfinder.j2
@@ -57,9 +57,6 @@
 {% set rfi_second_downsampling_factor = 32 %}
 {% set rfi_num_times = samples_per_data_set // rfi_downsampling_factor %}  {# = 32 #}
 {% set rfi_num_times_bar = rfi_num_times //  rfi_second_downsampling_factor %}  {# = 1 #}
-# Number of live/connected feeds (A01,A06,A07 x 2 pol = 6). Used to scale the feed-averaged
-# SK validity threshold so the gate can pass with only these feeds present (see below).
-{% set rfi_num_good_feeds = 6 %}
 
 ---
 type: config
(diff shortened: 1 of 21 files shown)


eschnett:
You introduce a new term "frame order" to describe the memory or disk layout of dishes/polarizations. This is confusing. We already have quite a few names, including "correlator order", "beamformer order", and "cylinder order" (for CHIME), and introducing a new "frame order" is only useful if this is indeed a new order. We decided to use beamformer order in both CHIME and CHORD everywhere in the pipeline (except possibly in the very first stage that sees the network input and needs to convert to beamformer order). Thus I think your "frame order" is actually "beamformer order", and I would use that term. If not, please add a respective enum element and description to the telescope object definition.

(Internally we unfortunately have separate "chime beamformer order" and "chord beamformer order", but that's an implementation detail because every run will only ever use one of these, depending on the telescope type. There is also a function which returns the standard (fiducial) order for each telescope, for the telescope functions which require an order as input.)



Applying bad feed masks: Using a queue and then the node's wall clock time isn't useful. The problem is that, downstream, we want to be able to combine frames, and they are combined by sequence number, and combining only works if they use the same bad feed mask. Thus queuing/scheduling needs to be based on the FPGA sequence number, not on wall time.

Also, where is the respective code? I see the text in the PR description, but I don't see a queue in `bufferBadInputs`.



"reject a mask that postdates the data it is folded into": How could that happen? The scenario would be a metadata description that has a particular FPGA sequence number, and also has a mask that has a later FPGA sequence number. That should never happen, unless a previous stage mixes up things and applies the wrong mask to its data. It might be that just the description of the mechanism is wrong, or the justification that AI made up; otherwise, the described mechanism (both an "update id" and an "effective FPGA sequence number") seems right. I assume that the update id is a new metadata key, and the effective FPGA sequence number is just the regular sequence number of the bad feed mask frame?



"cudaRFISKtilde optionally echoes the mask it applied back to the host": This is a GPU stage, why would it copy to the host? We have special stages for copying data between host and GPU. We're not doing this for anything else (weights, phases, packet loss mask), so it seems weird to distrust the pipeline only for the bad feed mask. I think an explicit metadata check in this stage is enough. This mechanism should be removed.



DishInputs: This changes the layout of the file and is not backward compatible, is it? I can't judge whether this is a good idea, so I assume it is. How has this been tested? Testing file formats is difficult. You could produce a file from known input and use a short Python script to see whether the data is what you expect. I see that the Boost test checks the index maps, but it doesn't test the actual data.

eschnett (dismissed):
I left comments in the PR.

jbmertens:
Thanks! This PR is probably worth discussing in person. Briefly -

> You introduce a new term "frame order" to describe the memory or disk layout of dishes/polarizations. This is confusing. We already have quite a few names, including "correlator order", "beamformer order", and "cylinder order" (for CHIME), and introducing a new "frame order" is only useful if this is indeed a new order. 

When we subset further down the N2 pipeline (i.e. stack baselines and/or mask), we need both information about the original ordering and how the subset was taken. In that code, something like (beamformer order + dish inputs subset) is needed to fully specifies which element(s) live(s) where in the visibility matrix.

Since more than one element can be combined when stacking, this isn't an element ordering any more. "Visibility ordering" is perhaps a better term.

> Applying bad feed masks: Using a queue and then the node's wall clock time isn't useful. The problem is that, downstream, we want to be able to combine frames, and they are combined by sequence number, and combining only works if they use the same bad feed mask. Thus queuing/scheduling needs to be based on the FPGA sequence number, not on wall time.
> 
> Also, where is the respective code? I see the text in the PR description, but I don't see a queue in `bufferBadInputs`.
>
> "reject a mask that postdates the data it is folded into": How could that happen? The scenario would be a metadata description that has a particular FPGA sequence number, and also has a mask that has a later FPGA sequence number. That should never happen, unless a previous stage mixes up things and applies the wrong mask to its data. It might be that just the description of the mechanism is wrong, or the justification that AI made up; otherwise, the described mechanism (both an "update id" and an "effective FPGA sequence number") seems right. I assume that the update id is a new metadata key, and the effective FPGA sequence number is just the regular sequence number of the bad feed mask frame?

Discussion on #1649 for this. The relevant changes are mostly config changes, plus the stage in that PR. Many of the code changes were already merged into develop for CHIME, and now the CHORD pathfinder config just uses them. But, I think the idea needs discussion in general, and will be relevant for how we handle calibration as well.


> "cudaRFISKtilde optionally echoes the mask it applied back to the host": This is a GPU stage, why would it copy to the host? We have special stages for copying data between host and GPU. We're not doing this for anything else (weights, phases, packet loss mask), so it seems weird to distrust the pipeline only for the bad feed mask. I think an explicit metadata check in this stage is enough. This mechanism should be removed.

The bad feed flagger uses this information to check if a feed looks like it is bad. (This is something CHIME already does, that chord pathfinder is catching up on.) We do trust the pipeline, this isn't added GPU code, just an added config snippet. The changes just exist to serve up the per-feed SK data so it can be pulled, rather than being pushed as it was for CHIME.

> DishInputs: This changes the layout of the file and is not backward compatible, is it? I can't judge whether this is a good idea, so I assume it is. How has this been tested? Testing file formats is difficult. You could produce a file from known input and use a short Python script to see whether the data is what you expect. I see that the Boost test checks the index maps, but it doesn't test the actual data.

The subset stage itself is tested, and the file writer is tested, but yes the two aren't tested together and should be.

config/ci-tests/cpu_batch/run_n2accum.yaml
```diff
@@ -74,6 +78,16 @@ earth_rotation_data:
     # yamlfix:on
 
 
+bad_inputs:
+    kotekan_update_endpoint: json
+    # yamlfix:off
+    bad_inputs: [3, 17, 40]
```
andrerenard, line 84: We should likely avoid the use of the term "inputs", since we have either `[pol][dish]` or `[element]` indexing.   So "inputs" seems possibly ambiguous?

docs/sphinx/user/file_formats/n2_vis_hdf5.rst
```diff
@@ -652,6 +654,52 @@ Completeness tracking and configuration snapshots
...
+     - uint64
+     - Absolute FPGA sequence number of the first sample each row was applied to.
+   * - ``stream_freq_id``
+     - (streams, freqs)
+     - int32
+     - The coarse frequency ids each stream's masks were applied to, -1 padded.
+       A file frequency's stream is the row that contains it.
+
+The group attribute ``uncovered_freq_id`` (int32 array) lists the file frequencies
+that hold data but belong to no stream; it is absent when every frequency is covered.
```
andrerenard: This doesn't seem like it should be possible?

lib/stages/bufferBadInputs.cpp
```diff
@@ -91,6 +93,35 @@ bufferBadInputs::bufferBadInputs(Config& config_, const std::string& unique_name
         }
     }
 
+    // Baseline mask from the telescope's dish table: elements whose dish is not a real
+    // array dish (Fake or an RFI antenna) are never valid inputs and stay masked
+    // independent of the posted bad-inputs list.
+    //
+    // Every dish missing from `dish_inputs` is reported as Fake, so a telescope configured
```
andrerenard: I wonder if we should use the term "missing" instead of "fake".   Fake seems to suggest we've generated some fake data for that location, when in fact there is just nothing there (all zeros or similar).

lib/stages/hdf5N2Write.cpp
```diff
@@ -993,6 +993,98 @@ std::optional<std::string> N2FileData::_get_final_filename() {
...
+}
+
+void N2FileData::create_bf_mask(const BfMaskStreams& streams, std::size_t num_pol,
+                                std::size_t num_dish) {
+    const std::size_t num_stream = streams.size();
+    std::size_t num_freq = 0;
+    for (const auto& stream : streams)
+        num_freq = std::max(num_freq, stream.second.size());
+
+    h5_file->createGroup("/bf_mask");
```
andrerenard: Should this be `bad_feed_mask`?   I read it as beamform mask, which I guess context should tell me otherwise, but someone just reading this file might not know what this means.

config/chord_pathfinder.j2
```diff
@@ -254,36 +252,73 @@ n2_pool:
...
+# get their own metadata pool: one object each, hundreds alive at a time.
+bf_mask_buffer_depth: 2 * host_side_buffer_depth
+bf_mask_send_buffer_depth: 512
+bf_mask_pool:
+    kotekan_metadata_pool: chordMetadata
+    num_metadata_objects: 2 * (bf_mask_buffer_depth + bf_mask_send_buffer_depth) + 4 * buffer_depth
+
+# The bad feed mask is valid for this many input samples: one frame (~42 ms), so each
+# correlation frame is gated by exactly one mask element, which N2Accumulate requires.
+bf_mask_lifetime_in_samples: samples_per_data_set
```
andrerenard, line 267: I think this could be longer, although ok to leave for now.  But possibly worth checking it still works with some larger multiple of samples_per_data_set?

andrerenard (approved):
Looks good over all.   Not sure if any of these changes are actually needed, but there is some terminology inconsistency (feeds vs. inputs vs. elements.)   Elements is kind of an internal indexing term, so not something that should be exposed at the file level.  However we should decide on feeds vs inputs...  mixing those is a bit confusing (although I realize the code does this everywhere...)

config/ci-tests/cpu_batch/run_n2accum.yaml
jbmertens: Yes, will be another PR. We should probably decide between "inputs" and "elements", and go with one everywhere...

lib/stages/bufferBadInputs.cpp
jbmertens: Agreed, will be another PR.

lib/stages/hdf5N2Write.cpp
jbmertens: Yes, the abbreviation is explicitly written out in a number of places now. And probably should be everywhere, but will be another PR.

config/chord_pathfinder.j2
jbmertens: Will leave for now, but agreed it's worth checking if there is any performance impact.

docs/sphinx/user/file_formats/n2_vis_hdf5.rst
jbmertens: The mask itself carries information about what frequencies it should have been applied to. This can happen if the visibility data makes it to the file writer, but mask data doesn't, e.g. due to a misconfiguration. I'm making it a fatal error if this happens instead.

PR #1657, opened 2026-09-10
Derive fpga_seq_num from the ring buffer's origin, not from zero

eschnett:
A ring buffer's `fpga_seq_num` is the sequence number of its logical beginning, and a ring position counts samples from there. Two stages that produce non-ring outputs from a ring position set the output's `fpga_seq_num` to the position alone, i.e. assumed the stream starts at sequence number zero, which it does not (the DPDK input starts at the FPGA counter):

- cudaBasebandBeamformer (julia/kernels/bb_template.cxx and the five generated cudaBasebandBeamformer_*.cpp, edited identically): `T_min`
- cudaPLMaskAccumulator: `begin * 128`

Add the ring's `fpga_seq_num`, the way cudaCorrelator, cudaPL1bitCorrelator and cudaCopyFromRingbuffer already do, and the way cudaFRBBeamReformer was fixed by @rhaas80 in 396b48b0f.  cudaPLMaskAccumulator keeps its literal 128: it is the ring's time-axis scaling, hardcoded in the kernel indexing as well.

### DIFF
--- a/lib/cuda/cudaPLMaskAccumulator.cpp
+++ b/lib/cuda/cudaPLMaskAccumulator.cpp
@@ -183,8 +183,12 @@ cudaEvent_t cudaPLMaskAccumulator::execute(cudaPipelineState& /*pipestate*/,
     pl_mask.check_metadata();
     pl_counts.set_metadata(pl_mask.get_metadata());
 
+    const auto& pl_mask_meta = pl_mask.get_metadata();
     const auto& pl_counts_meta = pl_counts.get_metadata();
-    pl_counts_meta->set_fpga_seq_num(pl_mask.get_read_valid().begin() * 128);
+    // The ring buffer's `fpga_seq_num` is the sequence number of its logical beginning, not
+    // zero; each pl_mask element along the time axis spans 128 FPGA samples.
+    pl_counts_meta->set_fpga_seq_num(pl_mask_meta->get_fpga_seq_num()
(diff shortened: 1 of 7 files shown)


rhaas80 (approved):

lib/cuda/cudaPLMaskAccumulator.cpp
```diff
@@ -183,8 +183,12 @@ cudaEvent_t cudaPLMaskAccumulator::execute(cudaPipelineState& /*pipestate*/,
     pl_mask.check_metadata();
     pl_counts.set_metadata(pl_mask.get_metadata());
 
+    const auto& pl_mask_meta = pl_mask.get_metadata();
     const auto& pl_counts_meta = pl_counts.get_metadata();
-    pl_counts_meta->set_fpga_seq_num(pl_mask.get_read_valid().begin() * 128);
+    // The ring buffer's `fpga_seq_num` is the sequence number of its logical beginning, not
+    // zero; each pl_mask element along the time axis spans 128 FPGA samples.
+    pl_counts_meta->set_fpga_seq_num(pl_mask_meta->get_fpga_seq_num()
```
jbmertens, line 190: One small suggestion: scale the offset by time_downsampling_fpga, not by 128.

jbmertens (commented):
One suggestion below, not needed before merging.

jbmertens:
Proceeding with merging this so it's easier to base on top of it.

PR #1658, opened 2026-09-10
N2Layout::DishInputs compact layout and per-element index maps

jbmertens:
Adds an `N2Layout::DishInputs` option. This layout runs on the CHORD pathfinder, where the visibility subset written to disk holds only the connected dishes.

## `N2Layout::DishInputs`

- A compact N2 layout: the dense triangle over the telescope's connected elements, those whose dish type is not `Fake` (RFI antennas included), in the fiducial element order. Requires a `CHORDTelescope`.
- `EigenN2Iter`'s full-triangle check accepts these frames.
- `CHORDTelescope::get_connected_elements(ElementOrder)` is the one new telescope method, providing the indices of the connected elements.

## `hdf5N2Write` per-element index maps

- The `/index_map` input tables (`dish_idx`, `pol`, `grid_x_idx`, `grid_y_idx`, `feed_pos_disp_m`, `coelev_disp_deg`, `type`, `label`) now hold one row per visibility element, matching how `/index_map/prod` and the per-(f, e, t) datasets already index elements. Each row therefore maps an element to a dish+polarization in the file's `input_order`.
- Before this the tables were per dish (`num_dishes` rows, some with axis label `dish`) while `prod` counted elements.


andrerenard (approved):
LGTM.   Although I need to go take a look at my baseband subsetting logic for the `/index_map` conventions.  I think we have some duplication in the hdf5N2Write, and my changes in hdf5FileWrite.   I'll see if there is anything we can abstract there.  Once this is merged in I'll try and address those issues.

PR #1659, opened 2026-09-10
Remove EigenVis* stages, and various minor reversions and corrections.

jbmertens:
Relevant to the most recent PRs.
Also removes the un-runnable chime production configs, which will still exist in history and in the hotfix branch.


lib/stages/N2FrameToVisFrame.cpp
```diff
@@ -170,7 +170,7 @@ void n2FrameToVisFrame::main_thread() {
             const state_id_t visState_id = dm.create_state<gatingState>(*visSpec).first;
             ds_id_vis = dm.add_dataset(visState_id, base_dataset_id);
 
-            // eigenVis creates a new state by adding the number of eigenVectors
+            // Eigen decomposition adds a state with the number of eigenvectors
```
rhaas80: please keep "eigenVis creates a new ..." since I specifically want to refer back to production CHIME that I need to mimic at that point in the code.

lib/stages/N2FrameToVisFrame.cpp
```diff
@@ -114,9 +114,9 @@ n2FrameToVisFrame::~n2FrameToVisFrame() {}
 
 void n2FrameToVisFrame::main_thread() {
 
-    // base dataset_id of visAcummulate (before gating)
+    // base dataset_id of visAccumulate (before gating)
     dset_id_t base_dataset_id;
-    // added state of eigenVis
+    // eigenvalue state
```
rhaas80: please keep as this documents where I copied this from and why we mimic things like so

julia/docs/chime_science_run_gpu.yaml
```diff
@@ -479,8 +479,6 @@ vis_buffers:
   num_frames: buffer_depth
   visbuf_10s_merge:
     kotekan_buffer: vis
-  visbuf_eig_10s_merge:
-    kotekan_buffer: vis
```
rhaas80, line 483: I'd assume this file can go away as well. All Julia stuff is CHORD / upgrade related. No Julia in production CHIME.

python/kotekan/runner.py
```diff
@@ -1738,14 +1738,6 @@ def has_hdf5():
     return len(missing) == 0
 
 
-def has_eigenvis():
```
rhaas80, line 1741: this still runs? No "unknown function `has_eigenvis`"?

One could also use:

```
def has_eigenvis():
    """ This stage no longer exists """
    return False
```

rhaas80 (approved):

python/kotekan/runner.py
jbmertens: If tests pass, then it should be ok, right? 😬 
~Will return False~

jbmertens: This _should_ be safe to remove, since nobody calls it I believe the only place this might be called would be code using the hotfix branch anyways.

PR #1660, opened 2026-09-11
round of updates, mostly minor fixes but also data layout changes for CHIME quantized beamformer data

rhaas80:
*  5f134a6f4 - hdf5FileWrite: check frame size against array description
* 56f2256cc -  testQuantizeKernel8Chime: sort sub-frequencies in descending order
* 78873d741 - cudaQuantize8Chime: update metadata with revesed upchan freq idx
* 15ea14d4c - cudaQuantizeKernel8Chime: store sub-frequencies in descending order
* a38452aa0 - fix(bufferBadInputs): fix inversed cyl -> beamformer order mapping
* 170bf28a3 - fix(rfiBroadcast): ensure that output packets can be transposed to cylinder order
* fdfbbb4ea - feat(processFeedGains): optionally take the complex conjugate of gains while upchannelizing
* 53d0bcd79 - ICETelescope: check that user specified input reorder is a permutation

the changes in 15ea14d4c, 78873d741, 56f2256cc could be done in a number of places, but reversing the order last, when quantizing to CHIME's FRB beamformer format means all previous stages are the same for CHIME / CHORD / others (but CHIME and CHORD still label their frequency channels in reverse order).

### DIFF
--- a/lib/stages/bufferBadInputs.cpp
+++ b/lib/stages/bufferBadInputs.cpp
@@ -78,9 +78,12 @@ bufferBadInputs::bufferBadInputs(Config& config_, const std::string& unique_name
         for (size_t idx = 0; idx < num_elements; ++idx)
             reorder.at(idx) = idx;
     } else {
-        for (size_t output_idx = 0; output_idx < num_elements; ++output_idx) {
-            station_id_t st_id = tel.element_index_to_station_id(output_idx, output_order);
-            reorder.at(tel.station_id_to_element_index(st_id, input_order)) = output_idx;
+        // Note: this is overkill. The input received by this stage's REST
+        // endpoint takes channel ids (which happen to by element indices in
+        // cylinder order in CHIME).
+        for (station_id_t station_id = 0; station_id < num_elements; ++station_id) {
+            const int output_idx = tel.station_id_to_element_index(station_id, output_order);
+            reorder.at(tel.station_id_to_element_index(station_id, input_order)) = output_idx;
         }
     }
 
--- a/lib/cuda/cudaQuantize8Chime.cpp
+++ b/lib/cuda/cudaQuantize8Chime.cpp
@@ -13,7 +13,7 @@
 #include "gpuCommand.hpp"               // for gpuCommandType
 #include "metadata.hpp"                 // for metadataObject
 
-#include <algorithm>          // for max
+#include <algorithm>          // for max, reverse
 #include <array>              // for array
 #include <cassert>            // for assert
 #include <cmath>              //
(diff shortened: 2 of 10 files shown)


lib/stages/bufferBadInputs.cpp
```diff
@@ -78,9 +78,12 @@ bufferBadInputs::bufferBadInputs(Config& config_, const std::string& unique_name
         for (size_t idx = 0; idx < num_elements; ++idx)
             reorder.at(idx) = idx;
     } else {
-        for (size_t output_idx = 0; output_idx < num_elements; ++output_idx) {
-            station_id_t st_id = tel.element_index_to_station_id(output_idx, output_order);
-            reorder.at(tel.station_id_to_element_index(st_id, input_order)) = output_idx;
+        // Note: this is overkill. The input received by this stage's REST
+        // endpoint takes channel ids (which happen to by element indices in
+        // cylinder order in CHIME).
+        for (station_id_t station_id = 0; station_id < num_elements; ++station_id) {
```
jbmertens, line 84: This was useful for CHORDEarly ordering, but we aren't using that any more. I suppose it's still nice to be explicit about the ordering of the incoming feed mask, and ordering of the mask in the output buffer.

jbmertens (commented):
LGTM.

Regarding upchannelization order, I'll rely on @eschnett or @andrerenard to say if this has any unexpected side effects.

rhaas80:
> Regarding upchannelization order, I'll rely on @eschnett or @andrerenard to say if this has any unexpected side effects.

It should not. I produces the ordering produced by production CHIME right now. It's hard to put in a CI test for this (or at least cumbersome), but here's some documentation for manual testing: https://github.com/chime-experiment/RTPipeline/issues/136

> Using the F-Engine simulator and setting a frequency in MHz of (for example):

```
fi = 13
f = 978
adc_frequency = 1.6e+9     # [Hz]
pfb_nsamples = 4096
Delta_f = adc_frequency / pfb_nsamples
f0 = 800e6 - (800e6-400e6)/1024 * f - Delta_f/2 + Delta_f/16/2 + Delta_f/16*fi
```

> I set up a monochromatic source in the sky (at zenith) with frequency f0.
> 
> I then run this data through most of the beamformer pipeline (namely upchan, FRB1, FRB2, but not the quantizer FRB3) and plot the resulting array. This shows that for sub-frequencies fi=0..15 the beam shows up (only) in the expected bin (with some aliasing of much lower amplitude given that f=978 is close to 400Mhz already).

ljgray (approved):

PR #1663, opened 2026-09-11
Shutdown and GPU memory safety fixes from chord

jbmertens:
Four small shutdown and safety fixes from the chord branch (July and August) that no other PR carried, plus one lock-discipline nit.

## Summary

- **gpuDeviceInterface**: stop with FATAL_ERROR when a second user of a named GPU memory block asks for a different length, or asks for a single frame of a block another user allocated as an array. Both were `ERROR` plus an `assert`, so a Release build silently aliased frame 0.
- **basebandApiManager**: hold the readout registry lock across iteration instead of per lookup, so a readout registering or unregistering while a request walks the registry cannot invalidate the iterator.
- **basebandReadout**: on shutdown, hand a frame range that is already locked to `extract_data`, which releases the locks on every path, and only unlock the frames the loop did not reach; previously a shutdown mid-dump double-unlocked some frames and leaked the rest.
- **BasebandWriter**: the file-closing thread waited on a condition variable with no predicate, so a shutdown notification that arrived before the wait began was lost and the thread slept out a full sweep cadence. Wait on a `closing_stop` flag.
- **Buffer::json_description** reads `peek_hold_enabled` under the buffer lock like the other fields it snapshots.

### DIFF
--- a/lib/core/basebandApiManager.cpp
+++ b/lib/core/basebandApiManager.cpp
@@ -51,15 +51,17 @@ basebandApiManager::basebandReadoutRegistry::operator[](const uint32_t& key) { /
     return readout_map[key];
 }
 
+std::unique_lock<std::mutex> basebandApiManager::basebandReadoutRegistry::lock() {
+    return std::unique_lock<std::mutex>(map_lock);
+}
+
 basebandApiManager::basebandReadoutRegistry::iterator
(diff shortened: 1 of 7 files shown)


andrerenard (approved):
These all look like good changes to have.  Particularly the GPU memory errors, and basebandApiManager locks.

PR #1664, opened 2026-09-11
Upstream chord #1592 and #1640: REST shutdown quiesce, config validation modes, frameID fix

jbmertens:
Upstreams two of kvand's chord PRs whose content never reached develop: #1592 and the develop-relevant part of #1640. Commits keep kvand's authorship; the #1640 squash is reworded to say what it contains, since the GNSS parts of that PR are already on develop.

## Summary

- **restServer::stop_processing()** (#1592): before pipeline teardown, stop dispatching to endpoint callbacks and wait out any handler already running on the server thread; further requests get 503. Closes a use-after-free where a SIGINT/SIGTERM teardown on the main thread destructed a stage while the server thread was still inside one of its handlers. Called from kotekan.cpp's shutdown path. restServer::stop() also no longer joins a server thread that was never started.
- **`kotekan --check-config` and `--dry-run`** (#1640): validation modes that exit non-zero on a bad config without binding the REST port or running the pipeline. `--check-config` is static (unknown stage types, duplicate names, dangling metadata pools) and constructs nothing, so it is safe next to a live pipeline; `--dry-run` builds the whole pipeline and tears it down without starting it. log_level parsing is shared with the runtime so the check accepts exactly what the runtime accepts. Documented in user_introduction.rst.
- **modulo<T> / frameID reduce-on-write** (#1640), in both N2Util.hpp and visUtil.hpp: the value is kept in [0, n) after every mutation. Counting unreduced and taking `% n` on read is discontinuous once a signed counter overflows (2^32 mod n is nonzero unless n divides 2^32), and a decrement below zero now lands on n-1 for unsigned T as well. test_modulo pins both copies through 2^32 increments.
- **rawFileRead `frame_period_us`** replay pacing, **rawFileWrite `allow_ndarray`**, and **valve** `kotekan_valve_passed_frames_total` plus a time-rate-limited drop warning (#1640).
- kotekan/CMakeLists.txt links blosc2 where asdf-cxx was built against c-blosc2 (#1640).

### DIFF
--- a/kotekan/kotekan.cpp
+++ b/kotekan/kotekan.cpp
@@ -707,6 +707,15 @@ int main(int argc, char** argv) {
 
         if (sig_value == SIGINT || sig_value == SIGTERM) {
             INFO_NON_OO("Got SIGINT or SIGTERM, shutting down kotekan...");
+            // Quiesce the REST server before destructing the stages: this
+            // teardown runs on the main thread while the server thread may
+            // still be handling a request that captured a stage's `this`
+            // (e.g. a monitor polling a stage's get_status endpoint), which
+            // would use-after-free once the stage is deleted below. Must
+            // happen before kotekan_state_lock is taken: the /start, /stop
+            // and /status handlers block on that lock, so draining while
+            // holding it would deadlock against any such handler in flight.
+            rest_server.stop_processing();
             std::lock_guard<std::mutex> lock(kotekan_state_lock);
             if (kotekan_mode != nullptr) {
                 INFO_NON_OO("Attempting to stop and join kotekan_stages...");
--- a/kotekan/kotekan.cpp
+++ b/kotekan/kotekan.cpp
@@ -27,6 +27,7 @@
 #include <locale>      // for locale
 #include <map>         // for map, operator!=, _Rb_tree_iterator
 #include <mutex>       // for mutex, lock_guard
+#include <set>         // for set
 #include <stdexcept>   // for runtime_error, out_of_range
 #include <stdio.h>     // for printf, fprintf, fclose, stderr, fdopen, feof, fgets
 #include <stdlib.h>    // for exit, free, WEXITSTATUS, WIFEXITED
@@ -195,6 +196,12 @@ else:
     sys.stdout.write(json.dumps(config_yaml))
 )";
 
+// Long-only option values, past the range of any short option character.
+enum : int {
+    OPT_CHECK = 1000,
+    OPT_DRY_RUN,
+};
+
 kotekanMode* kotekan_mode = nullptr;
 bool running = false;
 std::mutex kotekan_state_lock;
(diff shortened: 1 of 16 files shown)


kotekan/kotekan.cpp
```diff
@@ -218,6 +225,19 @@ void print_help() {
...
+    printf("                                   duplicate names, dangling metadata pools.\n");
+    printf("                                   Constructs nothing -- safe to run next to a\n");
+    printf("                                   live pipeline.\n");
+    printf("    --dry-run                      Builds the full pipeline (every constructor,\n");
+    printf("                                   every config key), fails on any buffer with\n");
+    printf("                                   consumers but no producer, then tears down\n");
+    printf("                                   without starting. Authoritative, but stage\n");
+    printf("                                   constructors open devices (DPDK/CUDA) and\n");
+    printf("                                   buffers claim hugepages -- only run this where\n");
+    printf("                                   the node is free.\n\n");
```
andrerenard, line 240: This is a great feature to have!    My only comment is we are starting to make kotekan.cpp quite long, maybe not in this PR, but we might want to consider breaking it up, just to make it easier to see what's going on here (and pushing out things these test functions and some of the python code to other files).

lib/stages/rawFileRead.hpp
```diff
@@ -26,7 +27,25 @@
...
+ *                          needs prefix_hostname: true here to be found.
+ * @conf   frame_period_us  Uint64. Sleep this long after publishing each frame, so a
+ *                          capture replays at roughly the rate it was acquired at.
+ *                          Set it to one frame's worth of acquisition time: e.g. a frame
+ *                          of 199680 samples taken at 20 MS/s is 9984 us.
+ *                          Default 0 = unthrottled: frames are read as fast as the
+ *                          downstream pipeline drains them, which is the behaviour you
+ *                          want unless something in the pipeline is tied to the wall
+ *                          clock. Note this is a floor on the frame period, not a rate
+ *                          lock -- the read and per-file open overhead add to it, so a
```
andrerenard, line 44: In theory this could be somewhat corrected by measuring time elapsed between frames and accounting for that in the sleep.  And seems a bit more like the intended behaviour?  Assuming the read is faster than the frame time, then you would actually recover the ~true frame rate - which could be more useful for benchmarks.

lib/utils/visUtil.hpp
```diff
@@ -820,16 +820,17 @@ class modulo {
     /// Assignment of a number into the modular number.
     modulo<T>& operator=(const T& i) {
         _i = i;
+        reduce();
```
andrerenard, line 823: Are we sure there are no cases that the underlying "unreduced" value was used?  I seem to recall some code that might have done something with it... but I could be wrong.  

Did we run into situations where this overflows?

Separately if we are keeping the version of this in N2Util and visUtil in sync, should we just remove the duplication (at least for this class)?

andrerenard (approved):
Largely I think this could be merged as is, but do consider the comments inline.   In particular double check the reduce logic in the `modulo` doesn't have unforeseen consequences.

kotekan/kotekan.cpp
jbmertens: Agreed. I also want to add CI coverage for more yamls using this.

lib/stages/rawFileRead.hpp
jbmertens: @kvand - I'm going to merge this PR in without changes, but see Andre's comment above in case you'd like to follow up.

lib/utils/visUtil.hpp
jbmertens: I remember a comment somewhere with a gnss pipeline change where this happened. @kvand will have to comment though.

I'll keep the visUtil one separate for now to avoid changing behavior for CHIME, but in principle they shouldn't be duplicated...

jbmertens: Looks like commit 189884644 is where the overflow happened. The underlying unreduced value `_i` is private, so shouldn't be seen anywhere else.

PR #1665, opened 2026-09-11
Recalculate the baseband phase matrix periodically

eschnett:
In anticipation of tracking baseband beams:

Recalculate the baseband beam phase matrix periodically (e.g. every second) instead of only once at startup.

This does not yet introduce a mechanism to actually update the beam pointings dynamically.

### DIFF
--- a/julia/kernels/bb_template.cxx
+++ b/julia/kernels/bb_template.cxx
@@ -39,6 +39,16 @@ std::array<T, D> reverse(const std::array<T, D>& values) {
         result[d] = values[D - 1 - d];
     return result;
 }
+
+// Override the leading (slowest) dimension's scaling with a run-time value. Used for inputs
+// that are valid for a configurable number of FPGA samples, such as the phase matrix.
+template<std::size_t D>
+std::array<std::ptrdiff_t, D> with_leading_dimscaling(std::array<std::ptrdiff_t, D> dimscalings,
+                                                      const std::ptrdiff_t dimscaling) {
+    static_assert(D > 0);
+    dimscalings[0] = dimscaling;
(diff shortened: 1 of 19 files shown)


julia/kernels/bb_template.cxx
```diff
@@ -372,6 +467,16 @@ cudaEvent_t cuda{{{kernel_name}}}::execute(cudaPipelineState& /*pipestate*/, con
         J_meta->set_fpga_seq_num(E_meta->get_fpga_seq_num()
                                  + T_min * E_meta->get_time_downsampling_fpga());
         assert(J_meta->get_time_downsampling_fpga() == 1);
+
+        // Element `k` of a slowly varying input covers the samples `k * lifetime` onwards,
+        // counted from the voltage ring buffer's logical beginning -- so the two streams have
+        // to start at the same sequence number.
+        {{#kernel_arguments}}
+            {{#haslifetime}}
+                assert({{{name}}}_buffer.get_metadata()->get_fpga_seq_num()
```
jbmertens, line 476: This check (and other asserts I see) will be absent in release builds. Is it worth promoting this (or any others) to actual fatal errors?

jbmertens (commented):
I don't follow all the details of what is going on here, so can't give this a proper review quickly. Will leave that to @andrerenard or someone else, but we could go through this in person if needed.

andrerenard (approved):
The logic here all looks good to me.

julia/kernels/bb_template.cxx
eschnett: I think it is. I will do a separate pass over all `assert` statements to ensure this happens everywhere.

PR #1666, opened 2026-09-11
Clarify meaning of time_downsampling_fpga, and apply it everywhere

eschnett:
`time_downsampling_fpga` had become ambiguous for buffers whose time direction is split across two axes: does it describe the slow outer axis or the fast inner one? The majority of the code means the outer axis, i.e. `dim_scaling[0]`. This meaning is now added to its definition, and the two beamformers which assumed a different meaning are updated.

### DIFF
--- a/config/chime_upgrade_frb.j2
+++ b/config/chime_upgrade_frb.j2
@@ -1886,7 +1886,7 @@ write_frb:
     base_dir: frb_beams
     in_buf: host_frb3_beams_buffer_post
     file_name: host_frb3_beams_buffer
-    create_single_file: false  # split time axis: dim_scaling[0] != time_downsampling_fpga
+    create_single_file: false  # dim 0 has extent 1; single file needs a gap-free stream
     input_order: CHIMEBeamformer
 
 # enable to dump sk statistics data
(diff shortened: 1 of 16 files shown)


rhaas80:
:-) I just ran into this as well. Naturally I feel that the "actual" definition is not to be just an alias for `dimscaling[0]` but instead that scaling of the actual time, ie the smallest `dimscaling` of any time axis. But this is tricky to test in say the writer right now since it has to then identify which axes are parts of "time".

If we do enforce `time_downsampling_fpga` to be the same as `dimscaling[0]` then I'd rather just remove it altogether (it is not inherited from CHIME production, so I have no emotional attachment to it).

eschnett:
Removing sounds good to me. That's more work than this PR because we'll have to go through all places in Kotekan that use it.

jbmertens:
Checking they agree seems fine with me, but I'll leave it to @rhaas80 to approve.

jbmertens (commented):
Checking they agree seems fine with me for now, but I'll leave it to @rhaas80 to approve.

rhaas80:
> Checking they agree seems fine with me for now, but I'll leave it to @rhaas80 to approve.

Me? Well if its going to go away soon anyway (preferably before I actually have to care when I try and incorporate all these changes into the CHIME upgrade branch).

rhaas80 (approved):

andrerenard (approved):
I think this is a reasonable way to define `time_downsampling_fpga`.

PR #1668, opened 2026-09-11
inventBFMask: do not require a CHORDTelescope

eschnett:
The stage unconditionally cast the telescope to CHORDTelescope on every frame, even though only the "auto" mask mode needs a telescope at all. Since the CHIME configs recently switched to CHIMETelescope, this threw std::bad_cast and aborted the process, breaking the `chime.j2` F-Engine simulator tests.


jbmertens:
LGTM.

jbmertens (approved):

PR #1669, opened 2026-09-12
N2Accumulate: fix weights for masked data

djgormley:
DTV frame masking leaves gaps in the even/odd pairs used by `N2Accumulate`. We need the output weights to reflect the samples that survive that masking.

The variance normalization currently counts every pair, including rejected pairs and pairs with an empty frame. Count only accepted pairs with samples in both frames. Keep valid samples from an incomplete pair in the visibility mean, and return zero weight when no variance estimate is available.

This preserves the existing inverse-estimated-variance convention. Weight estimates from very few usable pairs can still be noisy and biased. The doxygen block on `N2VarianceMode` is updated to match, which means dropping the old claim that EvenOddPosDef gives an unbiased variance estimate, since masking on the data breaks the assumptions behind it.

Tested with a CPU Test build and `WERROR=ON`. Tests cover masked pairs, empty frames, unequal counts between frames, and pairs spanning input frames.


lib/stages/N2Accumulate.cpp
```diff
@@ -1114,15 +1123,16 @@ bool N2Accumulate::output_and_reset(frameID& in_frame_id, frameID& in_rfiframema
                     loop_over_block([&](int64_t idx, [[maybe_unused]] N2::cfloat v) {
                         float weight = 0.0f;
 
-                        int64_t num_var_samp = _vis_samples_in_out_frame / 2;
-                        int64_t norm = ns * num_var_samp;
+                        // Normalize by pairs that contributed to the variance estimate.
+                        int64_t num_var_samp = _n_usable_variance_pairs.at(f);
+                        double norm = static_cast<double>(ns) * num_var_samp;
```
jbmertens: Small optimization: this doesn't need to be computed for every block, so can live outside of the `loop_over_block` lambda.

jbmertens (changes requested):
Thanks, good find on the issue.

Besides the comment below, the only thing I notice is the type conversions which I don't think are needed and introduce some churn. Could you revert that? Then happy to merge this in.

lib/stages/N2Accumulate.cpp
djgormley: Good call, moved it above the lambda so it is computed once per frequency (fd7588c).

djgormley:
Thanks. Reverted the widening in fd7588c: the counters are int32 again and the variance product is the original float expression. That also dropped the 65536-sample test case, which only existed to exercise the widening; the int32 product overflow it hit is there on develop already and is not part of this fix. The pair count now lives outside the block loop. Rebuilt with WERROR=ON and the two N2Accumulate test modules pass (52 cases).

jbmertens (approved):
LGTM.

Just noting on the `isfinite` calls, visAccumulate (old accumulation code) used to require a minimum number of good samples, rather than infinite variance. We may want to reintroduce that at some point, but TBD based on live runs I guess.

djgormley:
Thanks for the approval. CI is green on fd7588c; I do not have write access here, so could you merge when convenient? Squash is fine.

jbmertens:
@rhaas80 noting this for CHIME upgrade use. This should scale the variances as per the old estimator. Relevant in practice only for large numbers of missing samples.

rhaas80:
by the old? Meaning that the current code in `develop` (or really in dd645ce91) does not?

jbmertens:
The original visAccumulate code (hotfix?) computes weights ~ N_good/\sigma^2
The develop code before this PR computed weights ~ N_all/\sigma^2
The develop code after this PR computes weights ~ N_good/\sigma^2  again.
(edited for simplicity)

jbmertens:
(edited above.)
Arguably, we might actually want N_all / sigma^2, just depends what downstream expects...

rhaas80:
doesn't CHIME use `(_variance_mode == N2VarianceMode::CHIMEv1)` and that did not change?

jbmertens:
Ah, I don't know. If so, yes, that shouldn't have changed.

rhaas80:
> Ah, I don't know. If so, yes, that shouldn't have changed.

Had to check myself and (drumroll):

```
config/chime_upgrade_frb.j2:    variance_mode: CHIMEv1
```

so whatever this change changes, it doesn't affect what CHIME is using. For better or worse :-)

PR #1670, opened 2026-09-12
rawFileRead: read the size header for each frame

djgormley:
Our offline DTV tests replay recorded frames through the pipeline, so the reader needs to handle complete multi-frame recordings.

`rawFileWrite` writes a metadata-size header before every frame, but `rawFileRead` reads only the first one. For files with multiple frames, subsequent headers are read as part of the metadata or payload.

Read each frame's header and check for incomplete records and mismatched metadata. Use 64-bit file sizes to avoid truncating sizes above 4 GiB.

Trailing bytes that don't make up a whole frame are warned about and skipped rather than treated as an error, so a capture that was cut short still replays the frames it did finish. There's a second reason not to be strict here: `N2Buffer.calculate_layout` pads the frame to `maxalign` while `n2frame_layout_t::total_size()` doesn't, so the two sizes only agree when `num_elements` divides by 4. A python-written N2 dump at 1, 2 or 3 elements is a few bytes longer than the reader's record size, and the field offsets are identical, so those files read fine if the tail is ignored.

Also updates `docs/sphinx/user/file_formats/raw_frames.rst`, which described the single-header behaviour this fixes.

Tested with a CPU Test build and `WERROR=ON`: 11 raw replay tests pass. They cover standard and scalar N2 buffers, malformed records, tolerated trailing bytes, and file sizes above 4 GiB.


jbmertens (commented):
The change is fine with me. I've requested a review from @kvand who I think has been modifying this stage, just so he is aware, and in case he has any feedback. I'll leave final approval to him.

Although in general, I would suggest using the hdf5 read/writers over the raw read/writer if possible, since those stages output more structured data that can be validated.

djgormley:
Thanks. Agreed on the HDF5 writers for anything downstream; the raw path is what the capture writes on the node, and the offline DTV replay needs the frames back byte for byte, metadata included, which is why this reader had to cope with multi-frame files. While waiting on the second look I dropped one guard that could never trip (c2c213b), so the diff is a little smaller; the 11 replay tests still pass.

djgormley:
One more pass over this: back to plain fseek/ftell/rewind with a 64-bit size (the seek and rewind guards could not fire on a file that had just been opened), a short first header is now fatal like the other framing errors instead of ending the reader quietly, the N2-only metadata check is gone since the consumers own that, and the test no longer needs to be serial. 10 replay tests pass.

kvand:
Hey, sorry to be slow on this. I don't think this touches my earlier fiddling with raw read / write objects, I'm broadly ok with this, and I generally agree with Jim's hdf5 suggestion if the data is conducive to it.

I did throw it at Claude for interest, and it suggested a few things, nothing that looks too major to me... there are some esoteric edge cases, but I don't think it's worth belabouring the issue on those. No blocking objections from me, go for it!

kvand (approved):
No blockers to me, go ahead.

PR #1671, opened 2026-09-12
cudaPL1bitCorrelator: validate count geometry at startup

djgormley:
`cudaPL1bitCorrelator` fills the `N2` counts buffer with the nominal integration length when the count kernel doesn't implement the array geometry, and logs about it once.  Two checks at the top of the constructor, same shape as the ones in `cudaPLMaskAccumulator`: `num_polarizations` and `num_dishes` have to be positive, and `num_dishes` has to divide by 8. Those are buffer geometry preconditions, invalid rather than just unimplemented. `div_noremainder` only asserts them, so a debug build already aborts in there. The value is in `Release`, which today truncates silently instead of stopping.


lib/cuda/cudaPL1bitCorrelator.cpp
```diff
@@ -121,6 +119,7 @@ cudaPL1bitCorrelator::cudaPL1bitCorrelator(kotekan::Config& config, const std::s
     num_frequencies(config.get<int>(unique_name, "num_frequencies")),
     num_polarizations(config.get<int>(unique_name, "num_polarizations")),
     num_dishes(config.get<int>(unique_name, "num_dishes")),
+    num_station_groups(kotekan::n2_count_station_groups(num_polarizations, num_dishes)),
```
eschnett: Moving the almost trivial calculation (dividing by 8) and the error check into a new file makes the code more difficult to understand. Having the check in the constructor is a good idea, but the check should remain in the same file, and the simple calculation ("num_groups = num_dishes / 8") should remain visible here.

The concept of a "station group" isn't used elsewhere in the code, so this is best just a local variable, calculated right where it is used.

docs/n2-count-geometry.md
```diff
@@ -0,0 +1,20 @@
+# Packet/RFI count geometry
+
+`cudaPL1bitCorrelator` requires positive polarization and dish counts, with
+`num_dishes` divisible by eight. The count kernel supports
```
eschnett: This is just the current state of the kernel, this is not carved in stone. As needed, the kernel will become more flexible. We have other telescopes (CHIME, Hirax, CHARTS) which have other dish counts and which need to remain supported.

As written, this reads as if this 16-or-128 was a design decision. It is not. It is just a current limitation. It's dangerous to document current limitations too well, because others (especially AI agents) tend to latch onto this and are then afraid to fix the code.

lib/cuda/cudaPL1bitCorrelator.cpp
```diff
@@ -91,6 +91,7 @@ class cudaPL1bitCorrelator : public cudaCommand {
     const int num_frequencies;
     const int num_polarizations;
     const int num_dishes;
+    const int num_station_groups;
```
eschnett: I would treat this as a local variable (it's called `Sds` below), not as a class-wide constant. It's really just `dishes / 8`, so introducing a new name makes things look more complicated than they are.

lib/cuda/cudaPL1bitCorrelator.cpp
```diff
@@ -262,31 +257,17 @@ cudaEvent_t cudaPL1bitCorrelator::execute(cudaPipelineState& /*pipestate*/,
...
-        if (!warned_about_unsupported_Sds) {
-            // These cases are not yet implemented in n2k. Pretend that there is no packet loss.
-            ERROR("The 1-bit correlator calculating the n2k counts is not yet implemented for {:d} "
-                  "dishes. Pretending there was no packet loss. The n2k counts will be wrong if "
-                  "there "
-                  "was packet loss.",
-                  num_dishes);
-            warned_about_unsupported_Sds = true;
-        }
-        cudaMemsetInt(n2k_counts_memory, Nds, n2k_counts.get_ndarray().size());
```
eschnett, line 288: This code is there for a reason: Other telescopes need to be supported, and you are now making this an error. Please add the fallback path back, convert the error into a warning.

(Of course, generalizing the kernel to support other dish counts would be even better, but I realize that this is a lot to ask.)

lib/utils/n2CountGeometry.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include <string>
+
+namespace kotekan {
+
+/// Return P*(D/8), or throw if the count kernel does not support it.
+inline int n2_count_station_groups(int64_t num_polarizations, int64_t num_dishes) {
+    if (num_polarizations <= 0 || num_dishes <= 0)
+        throw std::invalid_argument("N2 count geometry requires positive polarizations and dishes");
+    if (num_dishes % 8 != 0)
+        throw std::invalid_argument("N2 count geometry requires dishes divisible by eight");
```
eschnett: I like this explicit error.

Checking for negative dish counts is really quite thorough, but why not.

Comparing to `int64_t::max` below is really quite unnecessary. I would assume this error is impossible to trigger.

lib/utils/n2CountGeometry.hpp
```diff
@@ -0,0 +1,29 @@
...
+    const int64_t groups_per_polarization = num_dishes / 8;
+    if (groups_per_polarization > std::numeric_limits<int64_t>::max() / num_polarizations)
+        throw std::invalid_argument("N2 count station-group geometry overflow");
+    const int64_t groups = num_polarizations * groups_per_polarization;
+    if (groups != 16 && groups != 128)
+        throw std::invalid_argument("N2 count kernel only supports Sds=16 or Sds=128; got Sds="
```
eschnett: Can you write "currently supports" here instead?

tests/test_n2_count_geometry.py
```diff
@@ -0,0 +1,99 @@
...
+import shutil
+import subprocess
+from pathlib import Path
+
+import pytest
+
+
+@pytest.fixture(scope="module")
+def geometry_guard(tmp_path_factory):
+    compiler = shutil.which("c++") or shutil.which("g++") or shutil.which("clang++")
```
eschnett: I'm not sure that compiling C++ code from scratch is a good idea for our test cases. This ignores all kinds of compiler settings. We have both Python tests (which run a pre-built Kotekan with various input files, that's what probably works best here) or Boost tests (which test from a C++ level). We don't usually test C++ classes just on their own. This doesn't work well because we usually need quite specific build options because Kotekan is very hardware dependent.

eschnett (changes requested):

djgormley:
@eschnett -- Thank you for all of the feedback.  I have been _**very**_ hesitent and nervous to work on code that changes functionality out side of my own components, so I tried to be very thorough -- Maybe too thorough?  As long as the concept of the change is agreeable I am more than happy to implement it to your specifications.

lib/cuda/cudaPL1bitCorrelator.cpp
djgormley: Done. Header's gone and `Sds` is back to `num_dishes / 8 * num_polarizations` as a local in `execute()`. All that's left in the constructor is the two `FATAL_ERROR` checks.

docs/n2-count-geometry.md
djgormley: Yeah, fair point. Deleted it.

lib/cuda/cudaPL1bitCorrelator.cpp
djgormley: Removed, it's a local again.

lib/cuda/cudaPL1bitCorrelator.cpp
djgormley: Put it back. Only change is `ERROR` to `WARN`, rest of the branch is the same. CHIME and HIRAX behave like they did before.

lib/utils/n2CountGeometry.hpp
djgormley: Gone with the header. Kept the positive check, it's a` FATAL_ERROR` in the constructor now.

lib/utils/n2CountGeometry.hpp
djgormley: That string only lived in the header so it went with it. What's left on this path is your original "not yet implemented for N dishes", which already reads as a current thing, so I left it alone. Can reword it if you'd still like a change.

tests/test_n2_count_geometry.py
djgormley: Agreed, deleted it. Didn't add a replacement. The checks are inside the CUDA command now so there's no host-side entry point a Boost test can reach, and the pytest jobs all build with `USE_CUDA=OFF` so the class isn't linked into those binaries. `gpu_batch` is the only harness that could get to it. Could add a yaml there if you want one, otherwise this matches the checks in `cudaPLMaskAccumulator`, which don't have a test either.

djgormley:
Removed the header, the doc and the test, put the fallback back with `WARN` instead of `ERROR`, and `Sds` is a local again. Down to +13/-6 in `cudaPL1bitCorrelator.cpp`.

I kept the two checks you liked, at the top of the constructor: positive `num_polarizations`/`num_dishes`, and `num_dishes` divisible by 8. Used `FATAL_ERROR` to match what `cudaPLMaskAccumulator` and `cudaRFIS012` do (different checks, same shape). One caveat-- the initializer list hits `div_noremainder` first, so in a debug build you get its assert instead of my message. Only `Release` actually reaches it. Still better than `Release` quietly truncating, but it isn't a check that fires everywhere.

Also worth mentioning on the downgrade: `WARN` needs log level > 1, `ERROR` only needed > 0. So anything running at `log_level`: error won't see this at all now. Fine by me either way, just letting you know.

djgormley:
Opened #1672 for the counts wrap so it doesn't clutter this PR.

djgormley:
The five points from the review are in 5f8c19b (warn instead of fail, the check kept next to its use, no new files) and the branch is up to date with develop; ready for another look when you have a moment.

djgormley:
@eschnett Hey Erik.  I applied your changes.  Thank you for your feedback.  Can you approve this?

djgormley:
After Erik's feedback this is just adds two additional error checks to the constructor.

lib/cuda/cudaPL1bitCorrelator.cpp
```diff
@@ -278,11 +285,11 @@ cudaEvent_t cudaPL1bitCorrelator::execute(cudaPipelineState& /*pipestate*/,
     } else {
         if (!warned_about_unsupported_Sds) {
             // These cases are not yet implemented in n2k. Pretend that there is no packet loss.
-            ERROR("The 1-bit correlator calculating the n2k counts is not yet implemented for {:d} "
-                  "dishes. Pretending there was no packet loss. The n2k counts will be wrong if "
-                  "there "
-                  "was packet loss.",
-                  num_dishes);
+            WARN("The 1-bit correlator calculating the n2k counts is not yet implemented for {:d} "
```
eschnett: Why are you downgrading this from an error to a warning? Since this is functionality that is really not supported, an error (which is not fatal in Kotekan) seems like the right choice.

I also recall that you moved this check into the constructor at some point. This was the most valuable part of your change, and it avoids the `warned_about_unsupported_Sds` variable as well.

eschnett (changes requested):

lib/cuda/cudaPL1bitCorrelator.cpp
djgormley: Put it back to ERROR and moved the check into the constructor, so it fires once at startup and the flag in execute() is gone. The fallback branch itself is unchanged.

djgormley:
Hey @eschnett.  Is this okay now or should I just close this PR?

eschnett:
@djgormley I need to let the CI workflow run manually when you make a change, so my responses are delayed.

eschnett (approved):

PR #1673, opened 2026-09-14
N2Accumulate: check the inputs before accumulating

djgormley:
The DTV mask #1677 rewrites the RFI mask stream this stage consumes, and I found `N2Accumulate` takes its five inputs on trust, no check that they cover the same frequencies, share a period, or land on the frame grid. That is also why two CI configs had been running wrong unnoticed. This PR just makes the stage verify its inputs before using them, so the mask fold can be trusted.


djgormley:
Rebased onto develop now that #1669 is in, so this shows its one commit (a5309bf). Same content, with the counters kept at int32 as in #1669 and the 65536-sample test case gone for the same reason; the period-4 case stays. Rebuilt with WERROR=ON, 73 N2Accumulate test cases pass.

jbmertens:
Could you provide some context for these changes? A voltage sample and FPGA tick do refer to the same thing for both CHORD and CHIME; I'm not sure when these would be different.

djgormley:
Yeah, you're right, they're the same thing. The ratio was coming from the time_downsampling_fpga metadata, which the stage never checked against sub_integration_ntime, and I generalised where I should have just checked. Dropped it: the stage now requires the metadata period to equal the configured sub-integration and stays in samples throughout. The renames and the period-4 test went with it.

What's left is the useful bit: the five streams are checked for the same frequencies and period, consecutive frames on the frame grid, counts equal across products and in range, and consistent totals. That's what caught the two CI configs (freq_ids only on the correlation generator, count_max above the sub-integration so packet loss went negative), fixed here. Title and description updated, diff is +243/-7 now.

djgormley:
Tidied after another look: counts that differ across products now warn once and carry on with the first entry (a routine per-link packet loss must not stop a running pipeline), the totals check moved out of the OpenMP region, the frequency-order state is a local, the duplicate include and the negative-sequence guard are gone, and the class doc is three lines. Both n2accum configs run clean, 65 test cases pass.

djgormley:
Yeah, you're right, they're the same thing. The ratio was coming from the `time_downsampling_fpga metadata`, which the stage never checked `against sub_integration_ntime`, and I generalized where I should have just checked. I dropped it: the stage now requires the metadata period to equal the configured sub-integration and stays in samples throughout.

Context I should have been more clear about in the description: the DTV mask #1677 rewrites the RFI mask stream this stage consumes, and I found `N2Accumulate` takes its five inputs on trust, no check that they cover the same frequencies, share a period, or land on the frame grid. That is also why two CI configs had been running wrong unnoticed. This PR just makes the stage verify its inputs before using them, so the mask fold can be trusted.

lib/stages/N2Accumulate.cpp
```diff
@@ -419,6 +423,114 @@ void N2Accumulate::main_thread() {
...
+                                t, f);
+                }
+                int64_t block_idx = 0;
+                for (int64_t ihi = 0; ihi < _n2k_counts_lin_blocks; ++ihi) {
+                    for (int64_t jhi = 0; jhi <= ihi; ++jhi, ++block_idx) {
+                        for (int64_t ilo = 0; ilo < _n2k_counts_blocksize; ++ilo) {
+                            for (int64_t jlo = 0; jlo < _n2k_counts_blocksize; ++jlo) {
+                                if (ihi == jhi && jlo > ilo)
+                                    continue;
+                                const int64_t idx =
```
jbmertens: The nested loops here could be compacted to a single loop over idx; no harm in a little extra checking for the mirrored triangle blocks.

lib/stages/N2Accumulate.cpp
```diff
@@ -962,6 +1074,22 @@ bool N2Accumulate::output_and_reset(frameID& in_frame_id, frameID& in_rfiframema
...
+        const uint64_t span = static_cast<uint64_t>(ticks_in_accum);
+        // Nominal valid counts (the correlator's unsupported-geometry fallback) plus real packet
+        // loss add up to more than the span; say so once and carry on rather than stop a running
+        // pipeline.
+        static bool warned_totals = false;
+        if ((static_cast<uint64_t>(_n_valid_fpga_samples_in_vis.at(f)) + _n_pl_samples_in_vis.at(f)
+                 > span
+             || _n_rfi_samples_in_vis.at(f) > span)
+            && !warned_totals) {
+            WARN("N2Accumulate valid, packet-loss and RFI counts at frequency {:d} exceed the "
```
jbmertens: This WARN might happen often; can we reduce it to INFO or DEBUG level?

jbmertens (changes requested):
The added checks are good to have. Couple small comments, otherwise LGTM.

lib/stages/N2Accumulate.cpp
djgormley: As a professor from my undergrad used to say "it looks like a damn christmas tree" lol.  Yeah, I'll take care of it.

lib/stages/N2Accumulate.cpp
djgormley: Dropped it to DEBUG.

lib/stages/N2Accumulate.cpp
djgormley: Done, single loop over the products now. Checked first that the mirrored entries are always written: n2k writes full 8x8 tiles, the fallback memsets the whole array, and the poison check would catch anything left over. The one test that relied on them being skipped is now a rejection case.

lib/stages/N2Accumulate.cpp
```diff
@@ -1002,8 +1113,11 @@ bool N2Accumulate::output_and_reset(frameID& in_frame_id, frameID& in_rfiframema
         meta->frame_length_fpga_ticks = ticks_in_accum;
         meta->n_valid_fpga_ticks = _n_valid_fpga_samples_in_vis.at(f);
         meta->n_rfi_fpga_ticks = _n_rfi_samples_in_vis.at(f);
-        meta->n_rfi_only_fpga_ticks =
-            ticks_in_accum - _n_valid_fpga_samples_in_vis.at(f) - _n_pl_samples_in_vis.at(f);
+        // Signed so nominal counts plus packet loss cannot wrap to 2^64 (#1672).
```
jbmertens: Let's avoid referencing PRs in the comments.

lib/stages/N2Accumulate.cpp
```diff
@@ -962,6 +1060,19 @@ bool N2Accumulate::output_and_reset(frameID& in_frame_id, frameID& in_rfiframema
 #endif
 
     // Wait for a block of frames to be available.  Grab them and get them metadata.
+    for (int64_t f = 0; f < _num_freq_per_n2k_frame; ++f) {
+        const uint64_t span = static_cast<uint64_t>(ticks_in_accum);
+        // Nominal valid counts (the correlator's unsupported-geometry fallback) plus real packet
+        // loss add up to more than the span; note it and carry on rather than stop a running
+        // pipeline.
+        if (static_cast<uint64_t>(_n_valid_fpga_samples_in_vis.at(f)) + _n_pl_samples_in_vis.at(f)
```
jbmertens: If these are checked earlier on, I don't think these checks are needed at output frame time.

lib/stages/N2Accumulate.cpp
```diff
@@ -962,6 +1060,19 @@ bool N2Accumulate::output_and_reset(frameID& in_frame_id, frameID& in_rfiframema
...
+                > span
+            || _n_rfi_samples_in_vis.at(f) > span) {
```
jbmertens: Reading this, this particular check should be fatal too.

lib/stages/N2Accumulate.cpp
```diff
@@ -419,6 +423,100 @@ void N2Accumulate::main_thread() {
...
+                const int64_t scalar_offset = t * _num_freq_per_n2k_frame + f;
+                const int32_t rfi_count = rficounts[scalar_offset];
+                const int32_t pl_count = plcounts[scalar_offset];
+                if (rfi_count < 0 || rfi_count > _n_fpga_samples_per_n2k_correlation || pl_count < 0
+                    || pl_count > _n_fpga_samples_per_n2k_correlation) {
+                    FATAL_ERROR("N2Accumulate diagnostic count out of range at subintegration {}, "
+                                "frequency {}",
+                                t, f);
+                }
+                for (int64_t i = 0; i < _n2k_counts_num_products; ++i) {
```
jbmertens: So this loop isn't needed, strictly speaking. There are some annoying mismatches between all of these buffer shapes; only the i=0th element is actually used by this stage, and changing that would take some GPU pipeline reworking.

The only checks we need here should amount to (according to Claude and I agree):
```
const int64_t count = counts_mat[t * counts_stride_t + f * counts_stride_f];
const int64_t rfi_count = rficounts[t * _num_freq_per_n2k_frame + f];
const int64_t pl_count = plcounts[t * _num_freq_per_n2k_frame + f];
if (count < 0 || pl_count < 0 || count + pl_count > _n_fpga_samples_per_n2k_correlation
    || rfi_count < 0 || rfi_count > _n_fpga_samples_per_n2k_correlation)
    FATAL_ERROR(...);
```
This tidies up the logic a log, and then the checks in output_and_reset should not be needed.

jbmertens (changes requested):
Few more suggestions...

lib/stages/N2Accumulate.cpp
djgormley: Done.

lib/stages/N2Accumulate.cpp
djgormley: Done.

lib/stages/N2Accumulate.cpp
djgormley: Done.

lib/stages/N2Accumulate.cpp
djgormley: Done.

jbmertens (approved):
LGTM.

PR #1675, opened 2026-09-15
Kv/gpu stream locking

kvand:
Two changes to the CUDA pipeline to handle more complex pipelines: Lock CUDA command queuing per stream, not per device; offset the used queues across multiple pipelines.

cudaDeviceInterface::gpu_command_mutex was one device-wide lock held across each pipeline's whole command queuing, including blocking driver calls, so every cudaProcess on a GPU serialised against every other one however little they shared. On CHORD GNSS (ten cudaProcess stages per GPU) one pipeline stuck inside that lock stalled the entire GPU half. This replaces it with one recursive_mutex per CUDA stream: a pipeline takes the locks for exactly the streams its commands enqueue onto, so pipelines with disjoint stream sets no longer exclude each other.

cuda_stream_base (new, default 0). The default assignment COPY_IN/COPY_OUT/KERNEL → 0/1/2 is shared by every pipeline on a device, so all their kernels land on one in-order stream. Setting cuda_stream_base: b on a cudaProcess shifts its default triple to b+0/1/2; bases three apart give disjoint streams, hence independent queues and no shared lock. Base 0 reproduces the existing assignment exactly. cudaSyncInput/cudaSyncOutput derive their source streams from the same rule, so a shifted pipeline waits on its own copy-in stream rather than another pipeline's.

Tests. tests/boost/test_cuda_stream_assignment (CPU-only; ten cases) covers the role→stream map, the ascending lock order — driven with host mutexes from three concurrent threads, with a negative control that deadlocks when the sort is removed — and the stage bound. config/ci-tests/gpu_batch/test_xpose2048_stream_base.j2 runs the transpose config at base 3 (streams {3, 5} instead of {0, 2}) through compare_data; full gpu_batch passes 21/21 on an L40S.


lib/cuda/cudaCommand.hpp
```diff
@@ -55,9 +55,22 @@ extern std::shared_ptr<cudaCommandState> no_cuda_command_state;
...
+ *                    owning cudaProcess; config lookup walks up, so its commands inherit it.
+ *                    Two pipelines on one GPU get disjoint CUDA streams when their bases are
+ *                    at least 3 apart -- the triples themselves must not overlap, so bases 0
+ *                    and 3 are disjoint while 0 and 1 share two streams. Disjoint streams let
+ *                    their kernels overlap (a stream is an in-order queue, so pipelines
+ *                    sharing stream 2 serialize) and mean they take no queuing mutex in
+ *                    common. This is not validated across stages: overlapping bases are legal
+ *                    and simply share what they overlap on. The owning cudaProcess must
+ *                    declare num_cuda_streams greater than base+2. An explicit @c cuda_stream
+ *                    is absolute and ignores this.
```
andrerenard: Allowing the base to have values other than a multiple of 3 seems strange to me...  We'd get into a lot of trouble easily if say someone makes a pipeline with base = 1 and then we got kernels and copies mixed up on the same streams.

jbmertens (changes requested):
Could you update the chord_pathfinder.j2 config as well, to work with these changes?

Couple (unnecessary, but fine) suggestions from Claude,
- cudaSyncOutput.cpp:23 — Debug log prints device.get_num_streams() (device-wide) but the loop below is bounded by the stage's own num_cuda_streams — misleading when a device is shared. Fix: log the stage-local count.
- cuda_stream_base/num_cuda_streams read independently from config in 3 places each (cudaCommand.cpp, cudaSyncInput.cpp, cudaSyncOutput.cpp, cudaSyncStream.cpp, cudaProcess.cpp) — risk of drift. Fix: compute once in cudaProcess, pass down via constructors.

Otherwise LGTM.

andrerenard (changes requested):
It took me a while, but I'm reasonably confident this won't break anything. 

However, I'm also wondering if the need for this change is the result of a violation of the expected gpuCommand/Process pattern, and not an actual problem.

Generally any blocking calls should happen on the `wait_for_precondition()` and not on the execute.   What blocking calls are happening in the execute that results in this stalling? 

I don't see any harm in making this change, since it will still reduce lock contention in systems with more than 3 streams, but blocking calls in execute still run the risk of stalls in those other pipelines. 

I also don't really understand why we wouldn't use `cuda_stream_base*3` in all the stream math to insure we'd always have non overlapping pipelines in the cases were we run additional pipelines along side the main one.

kvand:
> Could you update the chord_pathfinder.j2 config as well, to work with these changes?

It should work already. All of the gnss configs are derived from that file, so in some sense I've been running it on-sky for a month or so with this in place.

>     * cudaSyncOutput.cpp:23 — Debug log prints device.get_num_streams() (device-wide) but the loop below is bounded by the stage's own num_cuda_streams — misleading when a device is shared. Fix: log the stage-local count.

Yup, agreed. Will get that done shortly.

>     * cuda_stream_base/num_cuda_streams read independently from config in 3 places each (cudaCommand.cpp, cudaSyncInput.cpp, cudaSyncOutput.cpp, cudaSyncStream.cpp, cudaProcess.cpp) — risk of drift. Fix: compute once in cudaProcess, pass down via constructors.

I agree with this one in principle, but it changes 87 files by Claude's count -- every CUDA command changes its signature, and needs this added. Totally on board, but should go in a subsequent PR?

jbmertens:
> > Could you update the chord_pathfinder.j2 config as well, to work with these changes?

> It should work already. All of the gnss configs are derived from that file, so in some sense I've been running it on-sky for a month or so with this in place.

We don't want to add `cuda_stream_base` or other settings? This PR would be an appropriate place to do that, rather than later.

> I agree with this one in principle, but it changes 87 files by Claude's count -- every CUDA command changes its signature, and needs this added. Totally on board, but should go in a subsequent PR?

Yes, separate PR sounds good.

kvand:
> However, I'm also wondering if the need for this change is the result of a violation of the expected gpuCommand/Process pattern, and not an actual problem.
>
> Generally any blocking calls should happen on the `wait_for_precondition()` and not on the execute. What blocking calls are happening in the execute that results in this stalling?

I don't think it's driven by a violation of the command/process pattern. The (repeated) wedge that motivated it was found in backtraces to be inside cuEventRecord, after the wait_for_precondition() was satisfied. Quoting Claude's summary:
"What blocks is the CUDA driver itself: every command's execute() ends in record_end_event, which calls cudaEventCreate and cudaEventRecord unconditionally... the cx19 backtrace caught [these], with five correlator threads parked on one futex in queue_commands and the sixth holding it inside cuEventRecord. So the stall is GPU back-pressure surfacing in an enqueue call..."
That also makes nailing down the origin extra-hard, since it lives inside cuda things...

It seems likely some underlying race between the 6 threads above that's at the actual root, the claim here is that this removes the amplifier: one chain deadlocking doesn't freeze up a node, and since spreading things I've never managed to re-excite it.


> I don't see any harm in making this change, since it will still reduce lock contention in systems with more than 3 streams, but blocking calls in execute still run the risk of stalls in those other pipelines.

Ah, here's a proper rabbit-hole to fall down!

I feel like one mutex per cudaProcess makes more immediate sense than one globally -- but digging in on it, I'm not even sure (and neither is Claude) why the global mutex was there. This PR splits it up and that empricially reduces (to undetectable) a deadlock under the single mutex, but it's not obvious why the single one needs to be there in the first place. The cuda runtime and drive apis are documented thread-safe, and ordering is guaranteed by the stream... I had Claude dig into it, and as best it can tell, the global mutex is only precautionary, there's some chance the better fix here is to purge it and the per-process mutex.

Claude dug up exactly one piece of unguarded device state, `cudaDeviceInterface::runtime_kernels` -- a checked accessor may be a reasonable thing to fold into this PR to make it properly bulletproof.

> I also don't really understand why we wouldn't use `cuda_stream_base*3` in all the stream math to insure we'd always have non overlapping pipelines in the cases were we run additional pipelines along side the main one.

What's the actual problem with copies and kernels sharing streams? We're not tying them with events, so they'd only be delayed by the microseconds of time it takes to queue something? I don't see the issue with using queues in a mixed way... It's easy enough to shift by 3, but I'm not sure what that buys here.

andrerenard:
> I don't think it's driven by a violation of the command/process pattern. The (repeated) wedge that motivated it was found in backtraces to be inside cuEventRecord, after the wait_for_precondition() was satisfied. Quoting Claude's summary: "What blocks is the CUDA driver itself: every command's execute() ends in record_end_event, which calls cudaEventCreate and cudaEventRecord unconditionally... the cx19 backtrace caught [these], with five correlator threads parked on one futex in queue_commands and the sixth holding it inside cuEventRecord. So the stall is GPU back-pressure surfacing in an enqueue call..." That also makes nailing down the origin extra-hard, since it lives inside cuda things...
> 
> It seems likely some underlying race between the 6 threads above that's at the actual root, the claim here is that this removes the amplifier: one chain deadlocking doesn't freeze up a node, and since spreading things I've never managed to re-excite it.

Oh that's stranger than I was expecting - and somewhat unsatisfying.  Although yes this would reduce the chance of that kind of stall - assuming (as seems to be the case) the driver isn't sharing a lock between streams. 

> Ah, here's a proper rabbit-hole to fall down!
> 
> I feel like one mutex per cudaProcess makes more immediate sense than one globally -- but digging in on it, I'm not even sure (and neither is Claude) why the global mutex was there. This PR splits it up and that empricially reduces (to undetectable) a deadlock under the single mutex, but it's not obvious why the single one needs to be there in the first place. The cuda runtime and drive apis are documented thread-safe, and ordering is guaranteed by the stream... I had Claude dig into it, and as best it can tell, the global mutex is only precautionary, there's some chance the better fix here is to purge it and the per-process mutex.
> 
> Claude dug up exactly one piece of unguarded device state, `cudaDeviceInterface::runtime_kernels` -- a checked accessor may be a reasonable thing to fold into this PR to make it properly bulletproof.

The reason for this follows an older (but still valid) design pattern.   In previous code we'd linked together similar kernels with single blocks of GPU memory without ring buffers.  This has a lot of advantages in terms of lower CPU overhead, and lower end-to-end latency - since you don't have to bounce back to the CPU every time.  However with the ring buffers, it's natural to require that round trip, and makes the metadata tracking easier, so that's mostly what's been done since they were developed. 

But this is where the lock becomes important.  If two gpuProcesses invoke a chain of kernels which happen to be the same set of commands, and use the same name for the intermediary memory, then not locking the quote can create a nasty race condition.  This type of pattern isn't hard to imagine, since it could be just applying the same process to two streams of data.

The other case this could happen is if we have multiple kernels queued by the same execute, then the same problem with intermediary memory pops up. 

One way to fix this, would be at the memory model layer.  We append a unique identifier to all get_gpu_memory calls so each gpuProcess has its own blocks of intermediary memory.  Or we can disallow that pattern, and require rings between every call, but I'm somewhat hesitant to do that, since this model has a lot of overhead, and it's not entirely clear to me yet if that's acceptable in all cases. 

> > I also don't really understand why we wouldn't use `cuda_stream_base*3` in all the stream math to insure we'd always have non overlapping pipelines in the cases were we run additional pipelines along side the main one.
> 
> What's the actual problem with copies and kernels sharing streams? We're not tying them with events, so they'd only be delayed by the microseconds of time it takes to queue something? I don't see the issue with using queues in a mixed way... It's easy enough to shift by 3, but I'm not sure what that buys here.

Copies cannot happen in parallel, there is only one "real" D2H and H2D queue.   So if a large copy is happening in one of those queues, and a copy lands on a kernel stream, it will block that stream until the other copy clears.  Since we _do_ have very large copy operations, this was actually a common source of pipeline stalls in the early CHORD benchmark tests. 

Generally there is very little value in having multiple copy streams, expect possibly for the random cuda API stalls you've experienced. 

As a result, I don't really want to make it super easy for people to mix copies and kernels in the same stream.  e.g. adding a small gpuProcess with a base offset of 1, and then copy landing on stream 1, could stall the main set of kernels on the base offset.

andrerenard (changes requested):
Just to say it here.   I was wrong before about this being safe.   There is the possibility of creating a race condition in a multi-kernel cudaProcess with `get_gpu_memory()` allocated intermediate memory and different streams (and thus locks).   Reducing the locking granularity requires adjusting the memory allocator.  I _think_ it should be sufficient to make the memory unique to the gpuProcess instead of the gpuDevice.  In which case, I also think the locking can likely be removed entirely (assuming cuda API thread safety). 

However also as discussed, using multiple streams to get around a cuda API stalls doesn't feel like the best solution.  Although maybe it's our only option?

andrerenard:
However, that will have knock on effects to how the ring buffers share memory between gpuProcesses, so this isn't an easy thing to fix... and may require more thought.

kvand:
There's been some back-and-forth on this, I wanted to post a couple of updates since the thread here has gone a few days out-of-date. @andrerenard is away the rest of this week and said he'd dig into this when back, so just to summarize as I understand for when we pick it up:
* The current global mutex + single-ordered-queue model enforces a kernel queuing order, which makese most things safe, but limits what we can push on the GPU. Anything with less-than-full occupancy leaves resources on the table.
* We can switch to an unordered queue, or add support for multiple queues, or insist every kernel spans the device or eats the efficiency loss. Either queue change generates a bunch of race conditions; the kernel one I really don't want to force.
* This PR is to allow multiple queues, it allows different queues for different cudaProcess'es, and gives each one its own mutex to keep things from racing during queuing.
* That opens the door to memory races. The rings are mostly fine with this, each one guards its producers and consumers -- unless there are identically named memory pools backing them, in which case things race and clobber. None of our configs current trip things, but I've added a startup guard, fatals out if your config tries to do this.
* I've added a framejoin process to synchronize the output from multiple queues, that's widely used in the current gnss pipe.

PR #1676, opened 2026-09-15
N2: add per-product sample counts

djgormley:
Packet loss can leave different inputs with different numbers of accepted samples in the same 10 s accumulation. Our DTV masking replay needs the right normalization when frame masking and per-input packet loss occur together. This adds a valid-sample count per product and uses it for the mean and weight.

The counts are stored in N2 frames, Python N2 buffers, and CHORD_0.1 HDF5 files. Scalar mode remains the default. Fringestopping uses the configured input order, and changes to the dataset or RFI policy produce a warning, as in scalar mode.

Tested with a CPU Test build and WERROR=ON: the N2FrameDesc and hdf5N2Write Boost tests, plus all 89 scalar accumulation, masked normalization, per-product support, and rotation tests.


djgormley:
Rebased on the reduced #1673 (the tick-versus-sample split is gone there, so the period-4 test cases are gone here too and the per-product tick count is the sample count). Boost N2FrameDesc and hdf5N2Write pass, 82 pytest cases pass, WERROR=ON.

jbmertens:
I could use some context for these changes too. Maybe worth in-person discussion before proceeding. My impression is there is a lot of conditional logic that could be condensed by just passing along a number or two in the N2 frames.

djgormley:
Fair. Looking at it again, DTV masking doesn't actually need this: the mask is per frequency and hits every input the same way, so counts stay equal across products. This branch really implements the old packet_loss_is_scalar: false path, which is a bigger design question than a DTV fix. Closing it for now and we can talk it through in person; you may well be right that a couple of numbers per frame would do. The DTV mask stage doesn't depend on this one, so I'll open that next.

PR #1678, opened 2026-09-16
Convert GPU kernel `assert` statements into proper run-time errors

eschnett:
Convert C++ `assert` statements in the auto-generated CPU kernels into proper error messages. Handle the NDArray helper classes as well. No asserts are on a hot path.


jbmertens (approved):
LGTM.

PR #1679, opened 2026-09-16
UpchannelizationSchedule: turn multiton into a parse function

eschnett:
The current upchannelization schedule is cached in a static map. This does not work when different consumers (different GPUs) have different lists of coarse frequencies. This PR turns the schedule into a function that takes the coarse frequencies as input and does not cache anything.

This could supersede most of https://github.com/kotekan/kotekan/pull/1599.


lib/testing/inventVoltage.cpp
```diff
@@ -47,22 +47,12 @@ class inventVoltage : public kotekan::Stage {
...
     const bool reuse_frames = config.get_default<bool>(unique_name, "reuse_frames", false);
 
-    // Kotekan can take a long time to start up because e.g. beamforming matrices need to be
-    // calculated. We want to delay starting the benchmark timings until this startup has finished.
-    // The "startup buffers" are those that we wait for before starting to push data through the
-    // pipeline as fast as we can. We register ourselves as a consumer of these buffers and then
-    // wait for their first frame in our main thread.
-    const std::vector<std::string> startup_buffer_names =
-        config.get<std::vector<std::string>>(unique_name, "startup_buffer_names");
-
```
rhaas80, line 57: these changes seem unrelated

config/fengine/include/hfb_chime.j2
```diff
@@ -193,7 +193,10 @@ run_calc_hfb2_weights:
     frb2_num_frequencies: upchan_hfb_upchan_U128_max_num_output_channels
     frb2_beam_positions: host_hfb2_beam_positions_buffer
     frb2_weights: host_hfb2_weights_buffer
-    num_threads: 32
+    metadata_source:
+{% for id in range(16) %}   # num_frequencies = 16
+        - host_voltage_buffer_{{ id }}
+{% endfor %}    num_threads: 32
```
rhaas80: unintentional line join (`num_threads`)

rhaas80 (changes requested):
minor change requested

lib/stages/UpchannelizationSchedule.hpp
```diff
@@ -32,10 +34,25 @@ class UpchannelizationSchedule : kotekan::kotekanLogging {
...
+    // This is a plain value type, not a singleton. Parsing a schedule
+    // is cheap -- a handful of configuration lookups and some small
+    // maps -- so every stage that needs a schedule constructs its
+    // own.
+    //
+    // There deliberately is no process-wide instance. A single
+    // kotekan process can drive several GPUs, and each GPU handles a
+    // different set of coarse frequency channels, so there is no such
+    // thing as "the" schedule of a process.
+    //
```
andrerenard: This seems like an AI internal comment, and not something useful to keep in the code.

lib/stages/UpchannelizationSchedule.hpp
andrerenard, line 15: This class should likely live in `lib/utils`, not `lib/stages` now that I think of it.

lib/testing/inventVoltage.cpp
eschnett: They are not! Before removing this code `inventVoltage` would wait until the FRB2 weights have been calculated. These weights depend on the frequencies, which are only known from the voltage metadata, which are set by `inventVoltage`! I needed to break this circular dependency. And since the weight calculation is now 1000x times faster than it used to be, we don't need to wait any more.

andrerenard (changes requested):
Looks good, just some small comments.

lib/stages/UpchannelizationSchedule.hpp
eschnett: Moved.

lib/stages/UpchannelizationSchedule.hpp
eschnett: Removed.

config/fengine/include/hfb_chime.j2
eschnett: Un-joined.

andrerenard (approved):

andrerenard:
@rhaas80 if you are happy with this can you approve?

rhaas80 (approved):

PR #1680, opened 2026-09-16
chordMetadata: Make the fixed-length string fields private

eschnett:
The "string" fields in `chordMetadata` are char arrays which are not NUL terminated. This is confusing for callers. Make these fields private and update all users to go through the accessors.


julia/kernels/xpose_template.cxx
```diff
@@ -194,9 +194,8 @@ cudaEvent_t cuda{{{kernel_name}}}::execute(cudaPipelineState& /*pipestate*/, con
                 assert({{{name}}}_meta->type == {{{name}}}_type);
                 assert({{{name}}}_meta->dims == {{{name}}}_rank);
                 for (std::size_t dim = 0; dim < {{{name}}}_rank; ++dim) {
-                    assert(std::strncmp({{{name}}}_meta->dim_name[dim],
-                                        {{{name}}}_labels[{{{name}}}_rank - 1 - dim],
-                                        sizeof {{{name}}}_meta->dim_name[dim]) == 0);
+                    assert({{{name}}}_meta->get_dimension_name(dim)
```
jbmertens: Noting along with the PR I just reviewed, assert() won't be in release builds.

lib/metadata/chordMetadata.hpp
```diff
@@ -101,19 +101,10 @@ class chordMetadata : public metadataObject {
     } lock;
 
     // TODO: Replace by NDArray
```
andrerenard, line 103: Should this todo also move, or be removed?

andrerenard (approved):
Looks good, thanks for making this change!

jbmertens:
Asked in slack as well, but is there a reason for keeping the dim names in the chordMetadata at all? Or should they be (only) part of the frame descriptions?

eschnett:
I'm not sure we're using the frame descriptors everywhere already. So this PR just fixes a small inconsistency, reviewing all the code and moving everything to properly setting and reading frame descriptors is a larger project. It's the future, but I don't know whether we're there yet.

lib/metadata/chordMetadata.hpp
eschnett: This is the to-do item telling us to use the frame descriptors instead. These descriptors use the `NDArray` class internally.

PR #1681, opened 2026-09-16
N2Metadata: carry dataset_id through the raw format and JSON

djgormley:
`serialize` and `set_from_bytes` copy the N2 metadata fields by hand and skip `dataset_id`; the JSON does the same. So an N2 frame written with `rawFileWrite` and read back with `rawFileRead` arrives with a null dataset id, although N2Accumulate set one and the writers read it from the frame view. Five lines to carry it through, and one Boost case that round-trips a frame through both paths and checks that JSON without the field still decodes to null.

Tested with the CPU Test build and `WERROR=ON`: test_N2FrameDesc passes.


jbmertens:
Just requested a review/approval from @rhaas80 or @ljgray. I don't _think_ this affects the CHIME pipeline, since N2Metadata is not serialized in it? But want to check. Otherwise LGTM.

rhaas80:
I don't think this is required (for CHIME at least). The dataset id is created from scratch in `N2FrameToVisFrame` based on what CHIME's `visAccumulate` (and the following stages) would have done. The only change in dataset ID that is not fixed (as in never changes) is (https://github.com/kotekan/kotekan/blob/abd4d980c0c0e87c3bd831652fa6834472220a19/lib/stages/N2FrameToVisFrame.cpp#L153) to record if any gating (used in connection with the 26m telescope) happens. Since no such gating is anticipated for the near future, `visAccumulate` (and by extension `N2Accumulate`) have no need to create a dataset id. 

Unless I am missing something, this change is thus not needed (and would simplify your life I assume if there was no need to carry around not-needed values).

djgormley:
Makes sense, nothing of mine depends on it. I only submitted it because the field is set by N2Accumulate and read by the writers but dropped by serialize, so there is a gap. If nobody sets it in practice there's no point carrying it.  Just something I noticed while working on my project that was easy enough to fix with a few simple lines.

Feel free to close if you think it's not worth it.

jbmertens (approved):
Iirc the metadata field was originally included so the python shared memory interface would continue to work. But yeah I'd like to just remove the field entirely from N2 buffers... eventually. This corrects behavior for now, though glad it wouldn't be an issue in practice anyways.

PR #1683, opened 2026-09-17
Rename dish type Fake to Missing

jbmertens:
"Fake" described the placeholder dishInfo entries the telescope uses for unpopulated array slots, which reads as if the dish is simulated rather than simply absent. Rename DishType::Fake to DishType::Missing (JSON value, default dishInfo label, and all doc/comment references) to make the naming match the concept.


lib/utils/CHORDTelescope.hpp
```diff
@@ -27,7 +27,7 @@ using dish_index_t =
  * @brief Enum for denoting the type of dish input into Kotekan
  */
 enum class DishType : int32_t {
-    Fake = -1,     // Not a real dish
+    Missing = -1,  // An unpopulated dish slot
```
andrerenard: dish input/element? 
```suggestion
    Missing = -1,  // An unpopulated dish input
```

andrerenard (approved):

PR #1684, opened 2026-09-17
upchan: Repair the in-file self-test

eschnett:
This PR adds self-tests to the Julia code that generates the upchannelization kernels. This tests these kernels in a simple setup in pure Julia, without Kotekan. I wrote these tests when writing the kernels; apparently they were disabled at some point, and this PR re-enables them and corrects bit-rot. The kernels themselves do not change, so this should be an uncontroversial change. AI description of the PR follows below.

The upchannelizer's in-file self-test (`main(; run_selftest=true)` in `julia/kernels/upchan.jl`) has been broken for some time. It is not normally run — the call at the bottom of the file is commented out — so the rot went unnoticed. The kernel is fine for most upchannelization factors; the CPU-side reference model was wrong.

## The kernel's actual convention

Measured by sweeping a tone across all fine frequencies on an A40:

    Ē[u, t̄] = 1/U Σ_s W(s) exp(+2πi (u - (U-1)/2) s/U) E[U t̄ + s]

so a tone `E[t] = amp exp(2πi ν t)` lands in fine frequency `u = (U-1)/2 - ν U`, with `Ē[u, t̄] ∝ exp(2πi t̄ ν U)`.

## Reference model fixes

- **Frequency sign.** `test_freq = bin - (U-1)/2 + delta` put the tone into fine frequency `U-1-bin`, not `bin` — the reference assumed the conjugate of the kernel's transform. Now `test_freq = (U-1)/2 - (bin + delta)`.
- **Time origin.** The `t̄ - (M-1) + M/2` offset is spurious; there is none. For `M = 4` it contributed a constant factor of −1, which is why every expected value came out sign-flipped.
- **Attenuation.** The hard-coded `attenuation_factors` table was wrong, not merely imprecise: it claims 0.0374 at `delta = 0.5` where the window response is 0.5008, and 1.00007 at `delta = 0` where it is 1.0127. Replaced by the closed form evaluated from the same `Wkernel` the kernel uses, which also supplies the phase the response acquires for `delta ≠ 0`.
- **Precision.** Input and output phases were reduced mod 2 in Float32, losing precision for large `t`; now done in Float64.
- `amp` drops 7.5 → 6.0 so the expected output no longer clips against the 4-bit range.

## `Fmax` overrun (separate bug)

`main` set `Fmax = F` although `Ē` only holds `F̄ = F_per_U[U] * U` fine frequencies. With the production setups — e.g. `pathfinder` at `U = 16`, where `F = 384` but `F_per_U[16] = 64` — the kernel wrote past the end of `Ē`, and the `mod(addr + offset, length)` postprocess wrapped those writes around, overwriting earlier output. `Fmax` is now clamped to `F_per_U[U]`.

## Running it

Adds a `selftest` setup with `F <= F_per_U[U]`, drivers per upchannelization factor, and `julia/bin/upchan_selftest.sh`. Result on an A40:

```
selftest_U2: PASS
selftest_U4: FAIL (known kernel bug)
selftest_U8: PASS
selftest_U16: PASS
selftest_U32: PASS
selftest_U64: PASS
selftest_U128: FAIL (known kernel bug)
```

The passing cases have a maximum error of 0.076. The output is quantized to 4 bits, so anything below 0.5 is exact agreement.

The self-test call at the bottom of `upchan.jl` stays commented out so that kernel regeneration is not slowed down; the comment points at the script instead.

## Two kernel bugs the repaired test exposes

`upchan_selftest.sh` lists U=4 and U=128 as known failures and fails if either starts passing. Both are special cases of the FFT decomposition, and both are kernel bugs rather than reference bugs.

- **U=4** (`n == 0`): a *constant* input — exactly representable, so quantization cannot be blamed — produces a time-dependent output with power in three of the four fine frequencies, instead of a constant, equal split across the middle two. Every other U passes this same test. Reproduce with `main(; run_selftest=true, bin=1, delta=0.5f0)`, which makes `test_freq == 0`.
- **U=128** (the extra `Γ⁴` stage, `k = 7 > 6`): the tone is rotated between fine frequencies `u` and `u + 64` by an angle proportional to `u mod 64` — entirely in `u + 64` for `u = 0`, split evenly at `u = 31`, entirely in `u` at `u = 63`. That is the signature of a wrong twiddle factor in the final butterfly.

Both are fixed in #1686, which is stacked on this PR.

## Unrelated observation

`lib/testing/gpuSimulateCudaUpchannelize.cpp:213` uses `exp(-2πi (u-(U-1)/2) s/U)` — the conjugate of the real kernel — so it disagrees for *all* U. Nothing in CI compares the two, so neither validates the other. Left alone here.


jbmertens:
Do you want these to run as part of CI, or just locally?

jbmertens (approved):
Suggestions from Claude below. Nothing blocking imo.

-------------
1. julia/bin/upchan_selftest.sh:39: grep for ^Found [0-9]* errors before classifying, and report ERROR with tail -20 and status=1 when it is absent. A crash before "Checking results" currently counts as a known failure and exits 0.
2. julia/bin/upchan_selftest.sh:35: replace wait || true with wait. Bare wait always returns 0.
3. julia/kernels/setup_selftest.jl:31,34: drop Fbar_out and Tds_U1. upchan.jl never reads them.
4. julia/kernels/upchan.jl:2093: add the main(; run_selftest=true, bin=1, delta=0.5f0) reproduction recipe to this comment. It only lives in the PR description now.
5. julia/kernels/upchan.jl:1640: state whether the +2πi sign is what eqn. (83)/(84) specify or the measured kernel behaviour. The CPU simulator uses the conjugate, and the test now pins one of them.

eschnett:
These should be run when the kernels are regenerated. It's not necessary to run them in CI because we don't generate the kernels automatically either.

PR #1685, opened 2026-09-17
Add AGENTS.md file

jbmertens:
And a few README.md notes for humans.

Adds an AGENTS.md file with guidelines for higher quality code editing and reviewing, and a CLAUDE.md file that points to it.


eschnett:
I have asked Claude Fable to comment on this. The response is quite good, and I'll past it verbatim below. (Note: I already have a CLAUDE.md locally, auto-generated by Claude after analyzing the repo. Claude is referring to this "local" file below). The largest suggestions seems to be:
- don't mix your local setup with global information
- it's a good idea to point to other files with more specific information, e.g. coding style
- some things (e.g. threading model etc.) are apparently not described anywhere else, and thus should be listed here; otherwise, agents will either have to re-discover that (expensively) or won't know about it.



Verdict first: the PR's `CLAUDE.md` is just a one-line `@AGENTS.md` import, so the real content to review is AGENTS.md. The import is the right mechanism for Claude Code and needs no changes. AGENTS.md itself is useful but mixes repo facts with one developer's machine setup, contains four claims that don't hold on develop, and lacks the things your local file records that the docs don't.

## Claims in AGENTS.md that are wrong or stale

- **`--check-config` "needs 15435f812 or later".** It was added in 2acd5ab7b, PR #1664, which is already an ancestor of that commit and on develop. The caveat is moot for a file living on develop. Drop it.
- **"Shell scripts in tests/ci-scripts/push_pull run as part of tests."** No workflow on develop or chord references that directory, and `git log -S` shows none ever did. They are manual scripts, or the sentence describes a CI that no longer exists.
- **"chord ... should be kept equal to develop."** It isn't, and the workflow doesn't aim for that. The chord tip is a merge of develop into chord plus PR branches merged early:

| only in chord (by patch-id) | only in develop | content diff |
|---|---|---|
| 265 commits | 2 commits | 54 files |

  The sentence should describe the actual policy (PRs land on chord first, develop is merged back) or be removed.
- **"Never `-n auto`."** CI itself runs pytest with `-n auto` in [main.yaml:90](.github/workflows/main.yaml:90). Rephrase as "in a container on a large host use `-n 4`".
- **"Python tooling exists in /opt/kotekan_env."** The repo only probes that path when present, in [lint.sh:16](tools/lint.sh:16) and [make_version.cmake.in:115](lib/version/make_version.cmake.in:115). Say "on hosts that have it".

## Content that belongs in a personal file, not the repo

The Common Commands block reads like one shell history. `-j 40`, podman with "docker daemon is not available", the `localhost/u2404-cpu` image names (not in the repo; CI pulls from ghcr.io), the `<scratch>/wt-<x>` worktree scheme, and the force-push recipe all describe James's hosts and habits. Suggest they move to that developer's `CLAUDE.local.md`, or get a "CHORD dev hosts" qualifier. Same for `build-2404`: correct for the GitHub runners, but `/code/kotekan` is only the lint and IWYU mount point.

## Policy statements the maintainers should decide, not the file

- **"Development of new pytests is discouraged."** You wrote the HDF5 round-trip test as a pytest recently. If that's not team consensus, this line will make agents refuse to add pytests.
- **The "lazy ladder" and "avoid lambdas" advice** is generic agent guidance, and partly contentious. The repo already has `docs/sphinx/dev/dev_style_guide_code.rst`. AGENTS.md should point there instead of restating style, and it currently never mentions `docs/sphinx/dev/` at all.
- **"Individual commits within a PR do not all need to pass tests"** is consistent with the squash-merge history, so it's fine as written.

## What's missing, and what happens to your local file

Everything AGENTS.md says about building and testing overlaps your local "Building and testing" section, with small disagreements (boost timeout 30 vs 180, lint dry-run vs `-i`). What it has nothing on is your Threading model, Locking conventions, and Error handling sections: REST callbacks on one libevent thread, teardown order, destroyed-at-exit singletons, shared `cudaCommandState`, the buffer mutex rule, the datasetManager lock set, and never calling `FATAL_ERROR` from a destructor. Those are exactly the facts an agent can't derive cheaply and the sphinx docs don't record. I'd offer them as a section for AGENTS.md, as a commit on jbm/agents or a follow-up PR.

Practical point on the collision: once this merges, `git pull` on develop will refuse to proceed because your untracked CLAUDE.md would be overwritten. Rename it to `CLAUDE.local.md` beforehand. Claude Code loads that file alongside CLAUDE.md, so you get AGENTS.md via the import plus your notes, and add it to `.git/info/exclude` so it never shows in status. After that, trim the local file to what is yours alone: the macOS/cx67 build notes, the `initalize_stages` spelling, and whatever threading content isn't upstreamed.

The README layout and glossary additions are new information, not a duplicate of `overview.rst`, and look correct on a spot check. One nit: `python/kotekan/` also holds config rendering and the test runner, not only buffer readers.

andrerenard:
Overall this looks great. I agree with claude that we need a kotekan life-cycle, thread and memory model, and general code flow overview.  I will attempt to write one.

jbmertens:
Thanks - regarding each of these,

 - The local setup is shared for anyone developing on ctc or CHORD site nodes, which I think is most of us. So the commands I have included I think are appropriate, but we could add a note that these only apply on those nodes.
 - The lint scripts cover coding style. Beyond this, I don't have anything in mind.
 - If something (like the threading model) isn't described anywhere, we should also add a description for humans! The AGENTS file can refer to that. This also applies to e.g. design documents on GPU kernels which I still do not have access to despite repeated requests.

andrerenard:
Yeah that's a good point about also needing that threading model description for humans...   It should be here: https://kotekan.readthedocs.io/latest/overview_theory_of_operation.html but isn't...  I will just deal with that today.

jbmertens:
Added a few more changes. I hope this works as an initial document -- we can revise as we go along.

As Claude noted, using `CLAUDE.local.md` or `.claude/CLAUDE.md` is best for personal preferences.

PR #1686, opened 2026-09-18
upchan: Fix the U=4 and U=128 FFT bugs

eschnett:
The upchannelization self-test (#1684) found errors in the upchannelization kernels. This corrects them. AI text follows.

The repaired self-test from #1684 (now merged) is what found these. Rebased onto `develop`; this PR is just the two commits below.

Both bugs are in special cases of the FFT decomposition.

## U=4 — a typo in `Γ¹`

`n == 0` for U=4, so the length-4 FFT is done entirely by `Γ¹`. The high bit of `timehi` has value `2^(m-1)`, which is 2 for `m = 2`:

```julia
timehi0 = 2i32 * 0i32 + 1i32 * thread1
timehi1 = 1i32 * 1i32 + 1i32 * thread1   # should be 2i32 * 1i32
```

The two halves of the `Float16x2` therefore carried `timehi` in `{thread1, 1 + thread1}` rather than `{thread1, 2 + thread1}`. The `U == 2` (`1i32 * 1i32`, `m = 1`) and `U >= 8` (`4i32 * 1i32`, `m = 3`) branches are right, as is the corresponding `timelo` code in `Γ³` for every U. The build-time bound check carried the same typo, and `2 < 2^m = 4` holds, so it never fired.

## U=128 — two things in the `Γ⁴` stage

1. Step 6.5 read the un-twiddled `WWW_t0`/`WWW_t1` instead of `ZZZ_t0`/`ZZZ_t1`, so the `Γ⁴` phases computed in step 6.4 were dead code and eqn. (61) was skipped entirely. That produced the rotation between fine frequencies `u` and `u + 64` by an angle proportional to `u mod 64`.
2. With that fixed the tone became a single clean peak, but in the wrong half: the two outputs of the final length-2 FFT are swapped. Eqn. (62) puts the sum in `uhi = 0` and the difference in `uhi = 1`; here it is the other way round, so either the two halves of `Time(:time, 1, 2)` or those of `Freq(:freq, U/2, 2)` are labelled opposite to the note. I originally attributed this to the frequency axis being conjugated, but #1691 showed that conjugation came from the broken `chimify`, not from the kernel — the swap is needed either way, and the self-test pins the orientation.

The `# TODO: shouldn't this be eqn. (78)?` above `Γ⁴` was a red herring: the `Γ⁴` *values* match eqn. (61) correctly, they just weren't used.

## Verification

On an A40, `./julia/bin/upchan_selftest.sh` (added in #1684, with its `known_failures` list now removed):

```
selftest_U2: PASS
selftest_U4: PASS
selftest_U8: PASS
selftest_U16: PASS
selftest_U32: PASS
selftest_U64: PASS
selftest_U128: PASS
```

Maximum error 0.076 throughout. The output is quantized to 4 bits, so anything below 0.5 is exact agreement.

This is on top of #1691, so each run covers both test cases — the tone and the constant — with a distinct gain per fine frequency and a tone in every coarse frequency. Before the fix, U=4 put power into three of its four fine frequencies and U=128 landed a constant in `u = 0` and `u = 63` instead of the middle two.

Earlier, before #1691 corrected the encoder, I also swept the injected tone across **every** fine frequency for U ≤ 32 and across 12 (U=64) and 19 (U=128) representative ones — 93 in total, no failures. Those runs went through the old `chimify`, so read them as evidence that the fix holds across the whole frequency axis rather than as a check of the convention.

## Regenerated kernels

Second commit: mechanical re-run of `./julia/bin/upchan.sh` on cx67 (A40, Julia 1.12.7) from `~/src/kotekan`.

The semantic change is confined to the U=4 and U=128 kernels, one line in each generator output:

```
U=4    timehi1 = (1i32) * (1i32) + ...  ->  (2i32) * (1i32) + ...
U=128  YYY_u* = WWW_t0 +/- WWW_t1       ->  ZZZ_t0 -/+ ZZZ_t1
```

All 34 `.yaml` files — the kernel ABI — are byte-identical, and so are all `.cpp` wrappers, since neither the argument list nor the template changed. The remaining churn in the `.jl`, `.ptx` and `.sass` of the other upchannelization factors is the usual toolchain drift: a shifted `upchan.jl` line number in a provenance comment, plus LLVM register renumbering. Seven setups came back byte-identical. Nothing outside the upchannelizer is touched, and the regenerated PTX assembles cleanly under `ptxas -arch=sm_86`.

Note that CI does not exercise these kernels — `config/ci-tests/gpu_batch/test_pl_mask_upchannelizer.yaml` drives the hand-written `cudaPLMaskUpchannelizer.cu`, not the Julia-generated `cudaUpchannelize` — so the A40 self-test above is the only coverage.


julia/kernels/upchan.jl
```diff
@@ -1402,8 +1402,13 @@ function upchan!(emitter)
...
                         # Step 6.5 (equivalent to 6.3, but without mma): Length 2 FFT: Y = exp(...) Z
-                        apply!(emitter, :YYY_u0, [:WWW_t0, :WWW_t1], (WWW_t0, WWW_t1) -> :($WWW_t0 + $WWW_t1))
-                        apply!(emitter, :YYY_u1, [:WWW_t0, :WWW_t1], (WWW_t0, WWW_t1) -> :($WWW_t0 - $WWW_t1))
+                        # The kernel's fine frequency axis runs opposite to the `u` of
+                        # eqn. (62): a tone at `test_freq` lands in `(U-1)/2 - test_freq`,
+                        # not in `test_freq + (U-1)/2`. That reflection inverts the `uhi`
+                        # bit, so the sum belongs to `uhi = 1` and the difference to
+                        # `uhi = 0`.
+                        apply!(emitter, :YYY_u0, [:ZZZ_t0, :ZZZ_t1], (ZZZ_t0, ZZZ_t1) -> :($ZZZ_t0 - $ZZZ_t1))
+                        apply!(emitter, :YYY_u1, [:ZZZ_t0, :ZZZ_t1], (ZZZ_t0, ZZZ_t1) -> :($ZZZ_t0 + $ZZZ_t1))
```
rhaas80, line 1410: the comment in line 1404 mentions Y=...Z (which is what the proposed code does). The previous code used "W". Is this part of the bugfix (ie those are not just temporaries)?

julia/kernels/upchan.jl
```diff
@@ -667,7 +667,7 @@ function upchan!(emitter)
                         elseif U == 4
                             quote
                                 timehi0 = 2i32 * 0i32 + 1i32 * thread1
-                                timehi1 = 1i32 * 1i32 + 1i32 * thread1
+                                timehi1 = 2i32 * 1i32 + 1i32 * thread1
```
rhaas80, line 667: I'll have to take that AI's word for it.

rhaas80 (commented):
I am happy to try and review, but all I can do is verify that the code matches the comments and the summary. This is too deep in the guts for me to provide a review on the math implemented (if needed, it will take me reading and at least partially digesting the upchannelization LaTeX file).

julia/kernels/upchan.jl
eschnett: Yes, that is correct. This was a typo of mine. This combines bit patterns. It could also have been written as `... << 1` instead of `2i32 * ...`. Both lines should use the same bit shifts, hence `2i32 * ...` both times.

julia/kernels/upchan.jl
eschnett: Yes.

eschnett:
After re-enabling self tests in `upchan.jl` (which should never have been disabled), these kernels fail. With these changes they pass again.

PR #1688, opened 2026-09-18
Add the 2-NUMA configuration for CHIME

andrerenard:
- Adds a 2 NUMA domain (two CPUs, two GPU, and two NICs) configuration for CHIME. 
- Removes the requirement to have a frequency list in the [N2FrameToVisFrame] stage, by adding a "metadata buffer"
- Removes the FRB send stages API call for changing the offset (which I believe is never used in production).


andrerenard:
@rhaas80 Still in draft while I double/triple check the semantic correctness.  However I'm added you as a reviewer now for a stylistic review.   Do you like this layout, etc?   I've tried to avoid the usual multi-NUMA tendency of just duplicating everything by using aggressive jinja2 templating, but that adds its own nastiness when it comes to ease of readability.

andrerenard:
I've also not included some of the simulated data testing logic, since I don't see much value in using both CPUs and GPUs to run to those kinds of tests.   However let me know if you'd prefer I add that in.

PR #1691, opened 2026-09-18
Replace `chimify` with `CUDASIMDTypes.swap_offset`

eschnett:
For context: the Julia GPU kernel generators contain self-tests, and these self-tests need to convert from and to CHIME's 4+4-bit offset-encoded "swapped" representation. This corrects errors in the tests (not the kernels -- the kernels remain unchanged!), and the self-tests are now running, and they show all is fine (except for the already-open https://github.com/kotekan/kotekan/pull/1686).

`chimify` must swap the nibbles within each byte *and* add 8. Four of the five generators only added the offset; only `frb.jl` was correct. `CUDASIMDTypes.swap_offset` does exactly this and is an involution, so it replaces both `chimify` and `unchimify`. Host side only — `julia/output/` is unchanged.

The missing swap conjugates the whole transfer function, so fixing it broke upchan's self-test: the `exp(+2πi …)` convention documented in #1684 had been *measured through* the broken encoder. The reference model is corrected back to `exp(-2πi …)`; the kernel is not touched. `gpuSimulateCudaUpchannelize.cpp:221` already uses that sign.

Self-test hardening, since reference-model rot has now gone unnoticed twice:

- `bb.jl`/`frb.jl` counted mismatches and exited 0; they now `error()`. `frb.jl`'s `*** FOUND ERROR ***` banner was unreachable.
- `bb.jl` indexed `J_wanted` with stride `T` instead of `Tout` and died with a `BoundsError` before comparing anything.
- `frb.jl`'s commented-out coverage assert could never hold; replaced by "the kernel wrote nothing outside its range".
- `upchan_selftest.sh` scored a crash as a known failure and exited 0.
- upchan: per-fine-frequency gains, a tone in every coarse frequency, and a constant-input case.

Verified on an A40: upchan U2/U8/U16/U32/U64 pass both cases, U4/U128 fail both as before; a U=8 bin sweep 0…7 passes; frb, bb and chimefrb pass.


julia/bin/upchan_selftest.sh
```diff
@@ -21,22 +22,41 @@ setups='
 
 # Setups that are known to fail, with the kernel bug they expose. Both are
 # special cases of the FFT decomposition (`n == 0` for U=4, the extra `Γ⁴`
-# stage for U=128); all other upchannelization factors pass.
+# stage for U=128); all other upchannelization factors pass. Both test cases
+# fail for both, which is worth recording: the tone and the constant see the
+# same bug from two independent directions.
```
rhaas80, line 27: this is a temporary issue only? Given that there is  another pull request with a bug fix.

julia/kernels/upchan.jl
```diff
@@ -1634,57 +1640,107 @@ function main(;
...
+            E1 = Evalue(freq, time)
         else
             E1 = 0.0f0 + 0im
         end
         E1 = clamp(round(Int, E1), -7, +7)
         E_memory[Eidx + 1] = Int4x2(E1.re, E1.im)
     end
 
-    # map!(i -> zero(Int4x2), Ẽ_wanted, Ẽ_wanted)
+    # map!(i -> zero(Int4x2), Ẽ_wanted, Ẽ_wanted)
```
rhaas80, line 1733: I don't see a difference between before and after. (whenever Ẽ or Ē are  involved).

rhaas80 (approved):
I guess? At least I am convinced the actual CUDA code was doing the chimify correctly since it would have shown up as an error in the FRB1 beamformer (for which a Python code exists).

julia/bin/upchan_selftest.sh
eschnett: Yes, exactly.

julia/kernels/upchan.jl
eschnett: It's a unicode thing. `Ẽ` (E-tilde) has two different encodings, and apparently the encoding changed. It's debatable whether Github's diff should flag that, it should probably normalize the unicode sequence first.

PR #1692, opened 2026-09-18
Add self-tests for the transpose kernels, and an umbrella script

eschnett:
Follow-up to #1691, which fixed `chimify` and hardened the self-tests that already existed. This adds the ones that were missing.

**`xpose.jl` / `xpose2048.jl` had no self-test** — `main` had no `run_selftest` keyword, so the commented-out call at the bottom of each would have raised a `MethodError`. A transpose is a permutation, so it can be checked exactly: random input, permutation applied on the CPU, compared with no tolerance and full coverage. `xpose2048` also gets a random `scatter_indices` permutation; the identity one it ran with would hide any mistake in applying the scatter at all. 537M and 34M samples, 0 errors.

**`xpose.jl` did not run at all.** It carries its own setup block instead of including a `setup_*.jl`, so it never picked up the constants #1608 moved there, and `./julia/bin/xpose.sh` dies with `UndefVarError: ptx_compat`. There are no `xpose_*` artifacts under `julia/output/` or `lib/cuda/generated/`, so it has been dead a while. Fixed by adding the three constants. Nothing is regenerated here.

**New upchan cases:** an impulse probe (the kernel is linear, so impulses recover the transfer matrix directly — this is what would have caught the U=4 and U=128 bugs without any tone-convention reasoning), and the time window wrapped around the end of the ring buffer. `Tmin`/`T̄min` were asserted to be zero, so the `mod(addr + offset, length)` wrap had never run with a nonzero offset — and that wrap is what turns an out-of-range write into silent corruption. The reference is now ring-aware, and anything outside `T̄min:T̄max` must still hold the fill value.

**The error measure was wrong**, though never wrong enough to fail: `Ē` is quantized per component, but the check compared the complex magnitude, bounded by `0.5·√2 = 0.707` against a tolerance of 0.8. The impulse probe sat right at that bound. Measuring per component restores the intended 0.5 bound.

**`julia/bin/selftests.sh`** runs every generator's self-test and exits nonzero on failure. It works in a scratch copy, because running a generator in-tree rewrites `julia/output/` without the clang-format/JuliaFormatter passes that `julia/bin/*.sh` apply.

Verified on an A40: all seven generators PASS; upchan clean in all four cases for U in {2,8,16,32,64}, fails all four for U=4/U=128, unchanged.


julia/kernels/xpose.jl
```diff
@@ -482,6 +504,69 @@ function main(; compile_only::Bool=false, output_kernel::Bool=false)
...
 
+    if run_selftest
+        println("Allocating input data...")
+        Ein_memory = Array{Int4x8}(undef, idiv(D, 4) * P * F * T)
+        Eout_memory = Array{Int4x8}(undef, idiv(D, 4) * P * F * T)
+        info_memory = Array{Int32}(undef, num_threads * num_warps * num_blocks)
+
+        println("Setting up input data...")
+        # The kernel is a pure permutation, so a random input can be checked
+        # exactly: complete coverage, no reference arithmetic, no tolerance.
```
rhaas80, line 515: not quite true. since there's more elements (2048) than distinct values to fill the array with (15**2 = 225).

julia/kernels/xpose2048.jl
```diff
@@ -32,6 +33,17 @@ end
...
 
+# `-8` (nibble `0b1000`) is the NaN sentinel for `Int4x8`, so random test data
+# must avoid it. Map that one nibble value to `0` and leave the other 15 alone.
+# This is done on whole words because the arrays are large.
+function kill_nan_nibbles(r::UInt32)
+    lo = r & 0x77777777             # the low three bits of each nibble
+    lo |= lo >> 0x1
+    lo |= lo >> 0x2                 # bit 0 of each nibble: are they nonzero?
+    nonzero = (lo & 0x11111111) << 0x3
+    return r & ~(r & 0x88888888 & ~nonzero)
```
rhaas80, line 44: seems overly optimized for a test (I'd much prefer the explicit test for the NaN value [a hex nibble, no need to show bits I think])

julia/kernels/xpose2048.jl
```diff
@@ -413,9 +425,21 @@ function main(; compile_only::Bool=false, output_kernel::Bool=false)
     info_memory = Array{Int32}(undef, num_threads * num_warps * num_blocks)
 
     println("Setting up input data...")
-    map!(i -> Int4x8(-4, -3, -2, -1, 0, 1, 2, 3), Ein_memory, Ein_memory)
-    for i in (D * P - 1):-1:0
-        scatter_indices_memory[i + 1] = i
+    if run_selftest
+        # The kernel is a pure permutation, so a random input can be checked
+        # exactly: complete coverage, no reference arithmetic, no tolerance.
```
rhaas80, line 430: see above

rhaas80 (changes requested):
some comments seem misleading. Contains non-transpose changes in the upchan.jl files.

julia/kernels/xpose.jl
eschnett: This is not the CHIME transpose kernel, this is a currently-unused (and probably-will-never-be-used) kernel that could replace part of the packet acquisition code that Andre wrote. (At the time we didn't know whether the packet layout would allow efficient reordering to fill CHORD's input buffers.)

julia/kernels/xpose2048.jl
eschnett: Yes, this code is cleverly written. Claude is showing off its programming skills here. But the alternative is 8 comparisons, to check whether any of the values is -8.

julia/kernels/xpose.jl
eschnett: Comment updated.

julia/kernels/xpose2048.jl
eschnett: Code simplified.

julia/kernels/xpose2048.jl
eschnett: Comment corrected as well.

rhaas80: well *does* it achieve complete coverage? That would require N**M tests if there's M array entries each one of which can take N values (namely one has to try each and every value of a M positions N values-per-digit number).

julia/kernels/upchan.jl
```diff
@@ -1623,18 +1626,43 @@ function main(;
     !silent && println("Setting up input data...")
 
     # Give every fine frequency its own gain. With a constant gain a transposed
-    # or otherwise misindexed gain array would be invisible. The gains stay
-    # `≤ 1` so that the expected output remains inside the 4-bit output range,
-    # and are exact in `Float16` so that they contribute no rounding error.
-    gain(fbar) = Float16(1 - (fbar % 4) / 8)
+    # or otherwise misindexed gain array would be invisible. The gains are exact
+    # in `Float16` so that they contribute no rounding error of their own, and
```
rhaas80, line 1630: "are correct so contribute no rounding error of their own" seems in correct. 

If say the gain is `1/2` (exactly representable) and the values is `1+eps` (eps being the regular floating point epsilon such that `1+eps != 1.`) then this does give rounding error (in that  bit of information got lost).

julia/kernels/upchan.jl
```diff
@@ -1623,18 +1626,43 @@ function main(;
...
         G_memory[fbar + 1] = gain(fbar)
     end
 
-    @show Tmin = Int32(0)
-    @show Tmax = Int32(idiv(T, 4))
-    @show T̄min = Int32(0)
-    @show T̄max = Int32(idiv(Tmax, U) - (M - 1))
+    # `E` and `Ē` are ring buffers, of `T` and `T̄ring = T/U` time samples. The
+    # kernel processes input times `Tmin:Tmax` and writes output times
+    # `T̄min:T̄max`, both modulo the respective buffer length -- `Tmax` may exceed
```
rhaas80, line 1642: I would not use the same `Tmin:Tmax` for input and output if the input is `T` and the output is `T/U`.

julia/kernels/upchan.jl
```diff
@@ -1623,18 +1626,43 @@ function main(;
...
+    # `T`, which is what makes the window wrap. With `twrap` the window is
```
rhaas80, line 1643: for what it's worth: `Tmin` may exceed `T` (if using just the regular indices rather than the one ` % T`).

julia/kernels/upchan.jl
```diff
@@ -1685,7 +1713,11 @@ function main(;
         # fine frequency channels are offset by half a channel from the DFT
         # bins.
         test_freq = freq -> (testbin(freq) + delta) - (U - 1) / 2.0f0
-        Evalue = (freq, time) -> amp * cispi(Float32(mod(2 * time * Float64(test_freq(freq)) / U, 2.0)))
+        Evalue =
+            (dish, polr, freq, time) -> begin
+                dish == 0 && polr == 0 || return 0.0f0 + 0.0f0im
```
rhaas80, line 1718: a rather strange way of expressing

```
if (dish == 0 && polar == 0) {
  // single tone
  return amp * cispi(Float32(mod(2 * time * Float64(test_freq(freq)) / U, 2.0)))
} else {
  // not used but must return a value
  return 0.0f0 + 0.0f0im
}
```

I can see that this way may be more efficient, but this is a test code and should be understandable, not efficient I think.

julia/kernels/upchan.jl
```diff
@@ -1697,12 +1729,12 @@ function main(;
         # predicted to be zero, which holds only for `delta = 0`; a nonzero
         # `delta` leaks into the neighbouring bins.
         Ēvalue =
-            (freq, u, tbar) -> begin
-                u == testbin(freq) || return 0.0f0 + 0.0f0im
+            (dish, polr, freq, u, tbar) -> begin
+                dish == 0 && polr == 0 && u == testbin(freq) || return 0.0f0 + 0.0f0im
```
rhaas80, line 1733: same as above, admittedly the "unintuitive" construct (which is subjectived admittedly) has been present before.

julia/kernels/upchan.jl
```diff
@@ -1713,30 +1745,63 @@ function main(;
         # kernel bug, where a constant input produces a time dependent output
         # with power in three of the four fine frequencies.
         cvalue = 6.0f0 - 3.0f0im
-        Evalue = (freq, time) -> cvalue
-        Ēvalue = (freq, u, tbar) -> cvalue * response(u, 0.0f0)
+        Evalue = (dish, polr, freq, time) -> (dish == 0 && polr == 0 ? cvalue : 0.0f0 + 0.0f0im)
```
rhaas80, line 1748: same as above

rhaas80 (approved):
so many AI generated comments. The language leaves to be desired as far as conciseness if concerned. They read like one is being lectured.

Tests look ok though.

julia/kernels/upchan.jl
eschnett: This is a statement about the test, not about the kernel. The input values to the test are always `-7 ... +7` and never `1 + eps`.

julia/kernels/upchan.jl
eschnett: That's `T` for the input and `T-tilde` for the output.

julia/kernels/upchan.jl
eschnett: Yes, and in general it will. `T` is the buffer size, and `Tmin` the starting time which grows monotonically. `Tmin > T` will happen after 4 iterations or so.

julia/kernels/upchan.jl
eschnett: This is not about efficiency, that's a standard idiom in Julia. The `||` and `&&` operators short-cut, and that's just a concise way of exiting a function early.

julia/kernels/upchan.jl
eschnett: Yes, standard Julia...

julia/kernels/xpose2048.jl
eschnett: I think you replied to the old version of the code. The new version has a different comment.

julia/kernels/upchan.jl
rhaas80: oh, right. There's an extra bit on top of `T`. I need to pay closer attention.

julia/kernels/upchan.jl
rhaas80: Interesting. I understand the code (it's a standard `bash` idiom as well). I am not familiar with Julia to recognize it's standard patterns, so am judging this based on C/C++ patterns,where this would be odd since the condition checked is exactly the one in which case the executed code (the return statement) does not execute).  So it's a `unless(foo)` sort of thing, which C/C++ lacks. Now `perl` on the other hand... 
:-P

Anyhow, if this is standard Julia code (and given that this is Julia code): sure, fine with me. Learned something new today.

PR #1693, opened 2026-09-21
Add CPU upchannelization stage for testing

eschnett:
Add a new CPU upchannelization stage as a comparison test to the existing efficient GPU state. The new stage is simple and uses float32 arithmetic so that we can detect float16-caused problems as well.

This is a drop-in implementation, i.e. it expects its inputs and outputs on the GPU and will copy the data to/from CPU. (That's fine since efficiency is not important for testing.)

This PR also adds the respective helper functions to `NDArrayRingBuffer` to copy data from a ring buffer to a `std::vector` on the host and back.


config/ci-tests/gpu_batch/verify_cuda_upchan.j2
```diff
@@ -0,0 +1,257 @@
...
+          out_buf: host_{{ chain }}_voltage_buffer
+{% endfor %}
+
+# The kernel accumulates in float16 and the reference in float32, so the two round
+# differently: a sample within half a quantization step of a boundary can legitimately land
+# on either side. Measured on an A40: 45360 of 16777216 bytes differ (0.270%), and *every*
+# one differs by exactly +-1 in a 4-bit component -- not one component anywhere is off by
+# more than 1. The four +-1 cases split 24.7/24.8/25.2/25.3%, and 26 bytes differ in both
+# components, which is what independent rounding predicts. The threshold below is 1%, ~3.7x
+# the observed rate.
```
rhaas80: given that this sounds like AI text: are you (the human controlling the AI tool) willing to put your name behind this statement (and the numbers quoted)?

lib/cuda/gpuSimulateCudaUpchannelizer.cpp
```diff
@@ -0,0 +1,483 @@
...
+        FATAL_ERROR("num_taps must be positive, not {:d}", _num_taps);
+    // The ring bookkeeping subtracts a whole number of output samples as overlap.
+    if (_granularity % _upchannelization_factor != 0)
+        FATAL_ERROR("granularity_number_of_timesamples ({:d}) must be a multiple of the "
+                    "upchannelization factor ({:d})",
+                    _granularity, _upchannelization_factor);
+    if (!(0 <= _Fmin && _Fmin <= _Fmax && _Fmax <= _num_frequencies))
+        FATAL_ERROR("Invalid frequency span [{:d},{:d}): there are {:d} input frequencies", _Fmin,
+                    _Fmax, _num_frequencies);
+    // Each input channel becomes U output channels, and they have to fit.
```
rhaas80, line 219: fit exactly? Ie test for (non-)equality?

lib/cuda/gpuSimulateCudaUpchannelizer.cpp
```diff
@@ -0,0 +1,483 @@
...
+
+    G_buffer.check_metadata();
+    E_buffer.check_metadata();
+    Ebar_buffer.set_metadata(E_buffer.get_metadata());
+
+    const auto E_meta = E_buffer.get_metadata();
+    auto Ebar_meta = Ebar_buffer.get_metadata();
+
+    const auto E_nfreq = E_meta->get_nfreq();
+    if (!(0 <= _Fmin && _Fmin <= _Fmax && _Fmax <= E_nfreq))
```
rhaas80, line 299: `_Fmax <= E_nfreq` implicitly assumes (`_Fmin == 0`) or does it not?

lib/cuda/gpuSimulateCudaUpchannelizer.cpp
```diff
@@ -0,0 +1,483 @@
...
+    // Mismatched gains would scale each frequency by another frequency's gain.
+    const auto G_meta = G_buffer.get_metadata();
+    const auto G_nfreq = G_meta->get_nfreq();
+    if (G_nfreq != Ebar_nfreq)
+        FATAL_ERROR("Gain buffer G holds {:d} frequencies, but this kernel produces {:d}", G_nfreq,
+                    Ebar_nfreq);
+    const auto G_coarse_freq = G_meta->get_coarse_freq();
+    for (int freq = 0; freq < Ebar_nfreq; ++freq)
+        if (Ebar_coarse_freq.at(freq) != G_coarse_freq.at(freq))
+            FATAL_ERROR("Gain buffer G is for coarse frequency {:d} at index {:d}, but this "
```
rhaas80, line 332: probably ok to abort on first error, though often in these cases it is nice to print all bad array elements first, then abort. But really, here this should not matter.

lib/cuda/gpuSimulateCudaUpchannelizer.cpp
```diff
@@ -0,0 +1,483 @@
...
+                        "kernel produces coarse frequency {:d} there",
+                        G_coarse_freq.at(freq), freq, Ebar_coarse_freq.at(freq));
+
+    // The buffers must be large enough for the frequencies we are about to read and write.
+    if (E_meta->dim[1] != E_nfreq)
+        FATAL_ERROR("Input buffer E reports {:d} frequencies, but its frequency dimension has "
+                    "extent {:d}",
+                    E_nfreq, E_meta->dim[1]);
+    if (Ebar_meta->dim[1] < Ebar_nfreq)
```
rhaas80, line 346: check for (non-)equality instead?

lib/cuda/upchannelizeReference.cpp
```diff
@@ -0,0 +1,122 @@
...
+    const int U = upchannelization_factor;
+    const int P = num_polarizations;
+    const int D = num_dishes;
+    assert(M > 0);
+    assert(U > 0);
+    assert(P > 0);
+    assert(D > 0);
+    assert(num_times_out >= 0);
+
+    // Precompute the phase factor exp(-2 pi i (u - (U-1)/2) s / U). It depends only on
```
rhaas80, line 67: this combines the "Extra phas" and the regular phase. Might be worth a comment since this is not the exactly the expression in (83).

lib/cuda/upchannelizeReference.hpp
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * @f[
+ *   \bar{E}[\bar\tau, f, u, \pi, d] = \mathrm{Quantize}\left(
+ *       G[u + U f] \sum_{s=0}^{MU-1} W(s)\,
+ *       e^{i\pi s (U-1)/U} e^{-2\pi i u s/U}\, E[U\bar\tau + s, f, \pi, d] \right)
+ * @f]
+ *
+ * The two phase factors collapse into the single factor
+ * @f$ \exp(-2\pi i (u - (U-1)/2) s / U) @f$, which is what the code below evaluates.
+ * There is no FFT, no bit reversal and no twiddle table: the transform is a literal
```
rhaas80: without the FFT I'd be surprised to find a bit reversal or twiddle factors.

lib/cuda/upchannelizeReference.hpp
```diff
@@ -0,0 +1,123 @@
...
+namespace kotekan {
+
+/// Number of PFB taps used by the CHORD upchannelizer (`julia/kernels/upchan.jl:92`).
+constexpr int upchan_default_num_taps = 4;
+
+/**
+ * @brief The sinc-Hanning PFB window @f$W(s)@f$, eqn. (11) of the notes with @f$N \to U@f$.
+ *
+ * @f[
+ *   s' = \frac{2s - (MU-1)}{2MU}, \qquad
```
rhaas80: I'd cancel `2` in numerator and denominator to exactly match (11).

thus uses `U` where (11) uses `N` in the denominator but then the `(MU-1)` does not match the expected `MN` term in the numerator.

This does not obviously look like (11) .

lib/cuda/upchannelizeReference.hpp
```diff
@@ -0,0 +1,123 @@
...
+////////////////////////////////////////////////////////////////////////////////
+// Sample encodings
+//
+// CHIME/CHORD pack a complex sample with the REAL part in the HIGH bits and the
+// IMAGINARY part in the LOW bits; this is what "swapped" names in the kotekan type
+// `int4x2_swapped_withoffset`. See `external/n2k/include/n2k/Correlator.hpp`
+// (`real_part_in_low_bits = false`, `offset_encoded = true`).
+//
+// Decoding the other way round yields the complex conjugate of eqn. (83) -- consistently
+// on input and output, so it is invisible unless compared against the GPU.
```
rhaas80, line 94: so?

lib/cuda/upchannelizeReference.hpp
```diff
@@ -0,0 +1,123 @@
...
+inline std::complex<float> upchan_decode_int4(const std::uint8_t byte) {
+    const int imag = int(byte & 0x0f) - 8;
+    const int real = int(byte >> 4) - 8;
+    return std::complex<float>(float(real), float(imag));
+}
+
+/// Encode one `int4x2_swapped_withoffset` byte, clamping to +-7 and rounding half to even.
+std::uint8_t upchan_encode_int4(std::complex<float> value);
+
+/// Decode one `cint8` sample: plain two's complement, real component first.
```
rhaas80: The comment is in conflict with what the code does. The code only "casts" a `std::complex<int8_t>` to a `std::complex<float>` without any assumptions which component is first or assumption on the number encoding.

lib/testing/testDataCheck.hpp
```diff
@@ -165,6 +166,12 @@ testDataCheck<A_Type>::testDataCheck(kotekan::Config& config, const std::string&
 
     num_frames_to_test = config.get_default<int32_t>(unique_name, "num_frames_to_test", 0);
     max_num_errors_logged = config.get_default<int32_t>(unique_name, "max_num_errors_logged", 100);
+    // How many differing elements still count as a pass. The default, 0, demands exact
+    // equality. Raise it when comparing a float16 GPU kernel against a float32 CPU reference:
```
rhaas80, line 170: doesn't the GPU also handle (say) denormals differently than the IEEE conforming CPU code does?

tests/boost/CMakeLists.txt
```diff
@@ -338,3 +338,11 @@ else()
 endif()
 target_include_directories(test_TransposeBasebandArray PRIVATE ${KOTEKAN_SOURCE_DIR}/lib/stages
                                                                ${KOTEKAN_SOURCE_DIR}/lib/utils)
+
+# test_upchannelizeReference - the serial float32 upchannelization reference.
+# The reference lives in lib/cuda, next to its only caller, but has no CUDA dependency of its
+# own. Compile it directly rather than linking kotekan_cuda, so these tests still build and run
+# on a machine without CUDA.
```
rhaas80: the *tests* run? What do they compare against  then (if there's no runnable CUDA implementation)?

rhaas80: hmm, so there's another even more obvious implemenation of (83) in the Boost tests?

tests/boost/test_upchannelizeReference.cpp
```diff
@@ -0,0 +1,316 @@
...
+// centring; centring on MU/2 instead would break it.
+BOOST_AUTO_TEST_CASE(window_is_symmetric) {
+    for (const int U : all_U) {
+        const std::vector<float> window = upchan_window(M, U);
+        for (int s = 0; s < M * U; ++s)
+            BOOST_CHECK_SMALL(window.at(s) - window.at(M * U - 1 - s), 1.0e-6f);
+    }
+}
+
+// The window has sinc nulls: it changes sign. An unnormalized sinc(x) = sin(x)/x over this
```
rhaas80: I don't know what "sinc nulls" are. Is this a common term?

tests/boost/test_upchannelizeReference.cpp
```diff
@@ -0,0 +1,316 @@
...
+// The transform
+////////////////////////////////////////////////////////////////////////////////
+
+// A constant (DC) input must land in fine channels U/2 - 1 and U/2, split equally, and be
+// time-independent. This needs no convention reasoning and has no quantization noise; it is
+// the cheapest structural check on the transform, and the one that caught two real FFT bugs
+// in `upchan.jl` (U = 4 and U = 128).
+//
+// DC sits exactly half a bin from each of the two centre channels, so each picks up the
+// window's half-bin response, 0.5008. The remaining channels are not exactly zero -- they see
```
rhaas80, line 128: just to be sure `0.5008` is expected even for infinite accuracy math and to just a `float16` (or `float32`) artifact?

tests/boost/test_upchannelizeReference.cpp
```diff
@@ -0,0 +1,316 @@
...
+        }
+    }
+}
+
+// A tone at normalized frequency (bin - (U-1)/2)/U must land in fine bin `bin`, and the
+// output must advance in phase as exp(2 pi i tbar nu U) with no time-origin offset.
+//
+// The response is not a delta: a 4-tap PFB leaks a fixed fraction into its neighbours. Those
+// fractions are a property of the window alone, so they are the same for every U, and pinning
+// them is a much sharper test than merely requiring "not much" leakage.
```
rhaas80: "pinning"?

tests/boost/test_upchannelizeReference.cpp
```diff
@@ -0,0 +1,316 @@
...
+//
+// The fine-frequency axis is cyclic, so bin 1 is a *neighbour* of bin U-1. Where +k and -k
+// land on the same channel (k == U/2) the two contributions cancel instead of adding, so the
+// fixed fractions only apply for k < U/2.
+BOOST_AUTO_TEST_CASE(single_tone_lands_in_its_bin) {
+    // On-bin response, and the leakage at cyclic distance 1 and 2, as fractions of the peak.
+    constexpr double on_bin = 1.0127;
+    constexpr double leak[3] = {1.0, 6.0414e-3, 1.8051e-4};
```
rhaas80, line 177: citeable source for the numbers?

tests/boost/test_upchannelizeReference.cpp
```diff
@@ -0,0 +1,316 @@
...
+    const int U = 8;
+    const auto plain = run(U, 1, [](int) { return std::complex<float>(1.0f, 0.0f); }, 1.0f);
+    const auto scaled = run(U, 1, [](int) { return std::complex<float>(1.0f, 0.0f); }, 3.0f);
+    for (int u = 0; u < U; ++u)
+        BOOST_CHECK_SMALL(std::abs(scaled.at(0).at(u) - 3.0f * plain.at(0).at(u)), 1.0e-5f);
+}
+
+// The output at tbar reads inputs [U*tbar, U*tbar + M*U), and nothing outside that range.
+// An impulse in the part of tbar=1's window that tbar=0's window does not cover must show up
+// in tbar=1 and be *exactly* absent from tbar=0.
```
rhaas80: "*exactly* absent" sounds strange. What is "not exactly absent"? It contributes bit? Thus is not absent.

tests/boost/test_upchannelizeReference.cpp
```diff
@@ -0,0 +1,316 @@
...
+BOOST_AUTO_TEST_CASE(output_reads_its_own_input_window) {
+    const int U = 8;
+    const int num_times_out = 2;
+    // tbar=0 covers [0, M*U); tbar=1 covers [U, U + M*U). This sits in the second but not
+    // the first, and far enough from the window edge to carry a non-negligible weight.
```
rhaas80: if its far from the window, why does it carry *non-*negligible weight?

tests/boost/test_upchannelizeReference.cpp
```diff
@@ -0,0 +1,316 @@
...
+// Clamped to +-7, not [-8, +7]: -8 is the poison / missing-data marker.
+BOOST_AUTO_TEST_CASE(int4_saturates_symmetrically) {
+    BOOST_CHECK_EQUAL(upchan_decode_int4(upchan_encode_int4({100.0f, -100.0f})),
+                      std::complex<float>(7.0f, -7.0f));
+    BOOST_CHECK_EQUAL(upchan_decode_int4(upchan_encode_int4({-8.0f, -7.6f})),
+                      std::complex<float>(-7.0f, -7.0f));
+}
+
+// Ties round to even, matching the `cvt.rni` in the generated PTX. `floor(x + 0.5)` -- what
+// the older CPU simulation uses -- would give 1, 2, 3, 4 here instead.
```
rhaas80: what does it mean by "older CPU simulation"? Given that (was there) so far there has not been any CPU simulation (for the CHORD like beamformer or PFB based upchannelization)?

rhaas80 (changes requested):
I cannot match all the claimed expressions with the cited formulae. For a test case, I would consider this a required change.

lib/cuda/gpuSimulateCudaUpchannelizer.cpp
eschnett: Yes. This is true here because the data are copied from their ring buffer into a "packed" buffer on the host, and in this buffer `Fmin = 0` and `Fmax = E_nfreq`. The check here is looser than what actually happens. I read this check as "this needs to hold for the code below to be correct", not "this is what will hold in the actual test".

tests/boost/test_upchannelizeReference.cpp
eschnett: "pinning" as in "requiring this exact number", as opposed to "below a threshold".

It seems I am getting used to AI speak! I don't notice certain patterns any more.

tests/boost/test_upchannelizeReference.cpp
eschnett: Measured today! So this is correct until something changes.

Given the quantized nature and floating-point arithmetic I doubt one can derive these numbers easily from scratch.

lib/cuda/gpuSimulateCudaUpchannelizer.cpp
rhaas80: it's not testing `0 == _Fmin` though, only `0 <= _Fmin`.

tests/boost/test_upchannelizeReference.cpp
eschnett: There is a previous CPU stage (`gpuSimulateCudaUpchannelize`) that simulates the upchannelizer. It has bugs and doesn't use ring buffers. I am not deleting that stage here, but I think it's unused.

lib/cuda/gpuSimulateCudaUpchannelizer.cpp
eschnett: The code is correct even if they do not fit exactly. This checks whether the output array is large enough, the output array could be padded.

lib/cuda/gpuSimulateCudaUpchannelizer.cpp
eschnett: Done.

lib/cuda/upchannelizeReference.cpp
eschnett: Added.

lib/cuda/upchannelizeReference.hpp
eschnett: Removed that part of the comment.

lib/cuda/upchannelizeReference.hpp
eschnett: It isn't exactly (11).

Footnote (4) in the notes say "I may have some off-by-one errors in Eq. (10), e.g. I’m not sure whether the MN denominator should be (MN−1). I didn’t bother tracking these down since they don’t affect any of the results in this note significantly.". The exact equations with explanation ("symmetric about the centre") are given in the beginning of the file `upchannelizeReference.hpp`.

eschnett: I canceled the `2`.

lib/cuda/upchannelizeReference.hpp
eschnett: Well, we do compare against the GPU, so we would catch this problem.

I don't think there is good way to check that we have the right convention in this file (not real/imag swapped), except (a) being careful when writing the decoding/encoding, and (b) comparing against another code to see whether the encoding matches.

lib/cuda/upchannelizeReference.hpp
eschnett: I made the comment more explicit.

lib/cuda/upchannelizeReference.hpp
rhaas80: ok.

lib/testing/testDataCheck.hpp
eschnett: I think this is only the case when denormals are generated. When numvers are already denormals, they are handled correctly (I think), but just slower. So in this function that shouldn't matter.

tests/boost/CMakeLists.txt
eschnett: You can run this test without CUDA, e.g. to check whether a certain input produces a certain output. That's mostly useful to test this test, but then, AI loves tests.

tests/boost/test_upchannelizeReference.cpp
rhaas80: "private communication" :-)

tests/boost/test_upchannelizeReference.cpp
eschnett: [`sinc`](https://en.wikipedia.org/wiki/Sinc_function) is a function, it's either `(sin x) / x` or `(sin pi x) / x`. It has zeros. I should call these zeros instead.

tests/boost/test_upchannelizeReference.cpp
eschnett: `0.5008` is a measured value (see a few lines further up).

tests/boost/test_upchannelizeReference.cpp
eschnett: I'm saying "completely absent" instead.

tests/boost/test_upchannelizeReference.cpp
rhaas80: right, so would it be 1/2 and 0 for perfect arithmetic (I honestly have no idea, `0.5008` looks temptingly close to 1/2 )? I'd guess that the comment would refer to that or explicitly say that it reports on what was found before by running the tested code (and thus may be subject to bugs in said code instead of bugs in the math). The comments says "are" so yes that would imply "this was observed" compared to "should be" as derived without arithmetic error.

tests/boost/test_upchannelizeReference.cpp
eschnett: I think this is supposed to mean "if there was an error, and the weight was non-zero, then it would be non-negligible". That isn't a strong argument and I removed that part of the sentence.

eschnett:
@rhaas80 Thanks for your review. I think I either answered your comments or updated the code. Can you have another look?

lib/cuda/gpuSimulateCudaUpchannelizer.cpp
rhaas80: In this case: what is put into the "unused" frequency locations in the production run? Poisson? Left uninitialized?

config/ci-tests/gpu_batch/verify_cuda_upchan.j2
```diff
@@ -253,15 +253,12 @@ run_recv_{{ chain }}_voltage:
 # differently: a sample within half a quantization step of a boundary can legitimately land
 # on either side. Measured on an A40: 45360 of 16777216 bytes differ (0.270%), and *every*
 # one differs by exactly +-1 in a 4-bit component -- not one component anywhere is off by
-# more than 1. The four +-1 cases split 24.7/24.8/25.2/25.3%, and 26 bytes differ in both
-# components, which is what independent rounding predicts. The threshold below is 1%, ~3.7x
-# the observed rate.
+# more than 1. 26 bytes differ in both components, which is what independent rounding
```
rhaas80: bytes? Values! (they may be "bytes" but we don't compare bytes, we compare values)

rhaas80 (approved):
test itself is ok. Some comments are still strange to me.

Should serve to catch at least some regressions.

Thank you!

lib/cuda/gpuSimulateCudaUpchannelizer.cpp
eschnett: Left uninitialized. This is meant for several stages writing into the same buffer.

PR #1695, opened 2026-09-21
Fix the dish label ordering

andrerenard:
Dishes within a CRS board were labeled in the config top to bottom, instead of bottom to top.  No cable changes were done, this is a configuration change to fix the labels.

Older data before this version will have wrong labels, but the order of the data in the files in unchanged.


eschnett (approved):
There's not much to review here... One really would have to trace the cables to check whether this is correct.

I hope we have code in Kotekan that checks that the dish labels are unique etc.

This pattern looks regular. A Jinja loop would simplify the code.

andrerenard:
This has been separately reviewed by Ian and Dallas as having the prefered labeling, so I'm going to merge it in.

